PR #29, opened 2015-09-18
Separate JobClient into SyncJobClient and AsyncJobClient decorator

azakkerman:
* Separate JobClient into SyncJobClient and AsyncJobClient
* Cleanup the job client code dealing with closing Request and Response objects by using Closeable and try-with-resources construct
* Fix and update the unit test


azakkerman:
I mean I can just restore the JobClient with only the try-with-resources fixes in this review, since you are opposed to splitting it into base and decorator

dgrnbrg:
I think it would be good to have the components that are necessary to free automatically freed.

One question I have is what testing you'll do for this? When wenbo wrote the code, he ran a benchmark and watched to see if connections weren't getting freed. My main worry w/ testing a new resource release mechanism is that the resource gets leaked for some obscure reason.

dgrnbrg:
Are you still interested in pursuing this? If not, I'll close the issue.

dgrnbrg:
I'm going to close this for now; please reopen if you'd like to continue to pursue this approach.

PR #38, opened 2015-09-22
Why is this test flaky?

leifwalsh:
https://travis-ci.org/twosigma/Cook/jobs/81488613

It's those circuit breaker tests!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
     - MESOS_NATIVE_JAVA_LIBRARY=/usr/lib/libmesos.so
     - CLJ_HTTP_ASYNC_POOL_TEST_DURATION_MULTIPLIER=5
   matrix:
-    - TEST_DIR=scheduler TEST_CMD='lein test :all'
-    - TEST_DIR=jobclient TEST_CMD='mvn test'
+    - TEST_DIR=scheduler DEPS_CMD='lein with-profiles +test deps' TEST_CMD='lein test :all'
+    - TEST_DIR=jobclient DEPS_CMD='mvn dependency:resolve' TEST_CMD='mvn test'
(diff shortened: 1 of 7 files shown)


leifwalsh:
Ok, hopefully that's all of them, and all the math still works out, and that factor is high enough.

leifwalsh:
Not good enough yet it seems.

dgrnbrg:
Maybe only certain durations should get increased, like wait times to stable states?

Is it possible to script the interaction, so that there's an order to the steps, and a timeout? You could make a map from step-name to `promise`s, and then have a state machine `deliver` those promises to have thread-safe, rudimentary orchestration by means of blocking calls to `deref`. Or maybe that wouldn't help much; I haven't read the tests too closely.

leifwalsh:
There is already sequencing, though done differently than you described.

If this doesn't fix it I think we should disable these tests until I get back and have a chance to merge the other fixes that didn't make it into the open-source snapshot.  The tests try to test general-ish behavior without being too specific to the underlying implementation, but in order to be fast they have to be kind of specifically tuned.  This is annoying because the state is kind of hard to make general statements about; this may be a design decision we should revisit.

leifwalsh:
@dgrnbrg I have a green build (for circuit breaker stuff) in https://travis-ci.org/twosigma/Cook/builds/81792099, now trying to improve travis logs a bit.  A metatransaction test are now failing, probably need to use ex-data instead of assuming it's got .getCause() but that's just a guess.

leifwalsh:
@dgrnbrg tests now look good, you may want to peek at the build I linked above though.  Travis logs are nicer too (maven ones aren't perfect, but aren't worse).  Please merge if you like it or complain if you don't.

dgrnbrg:
awesome!

PR #45, opened 2015-09-22
Add JMX and Graphite metrics. Fixes #1 and Fixes #8

### LINKED ISSUE #1: Add graphite metrics reporting

### LINKED ISSUE #8: Add support for riemann reporter with metrics v3
On the earliest commits, it seems to have metrics v3 and v2 coexisting. I don't understand why that would work, but it should be possible to build this correctly using compatible versions.

dgrnbrg:


dgrnbrg:
@icexelloss I'd love your feedback here, too.

PR #47, opened 2015-09-23
Elide some job params from db for better compression. Fixes #25

### LINKED ISSUE #25: Remove the default settings of properties in rest api implementation
Things like `max-runtime` and `priority` are being set and written to the DB, rather than picking up defaults in the scheduler. This makes the schema and DB use ugly and inefficient; these should be cleaned up.

dgrnbrg:
This should begin to reduce the amount of DB storage we use for default params, which could lead to better scan performance for some workloads.

Leif, as resident DB expert, what do you think about this style?

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -95,15 +95,20 @@
                                   :environment/name k
                                   :environment/value v}]))
                             env)
+                ;; These are optionally set datoms w/ default values
+                maybe-datoms (concat
+                               (when (and name (not= name "cookjob"))
+                                 [[:db/add id :job/name name]])
+                               (when (and priority (not= util/default-job-priority priority))
+                                 [[:db/add id :job/priority priority]])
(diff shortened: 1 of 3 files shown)


leifwalsh:
I don't know how much this will speed up datomic but from the db side I think it's fine.

In terms of implementation, it might be better to somehow encode this as part of the schema, or put the defaults in the database somehow so that they're not just defined in the code only, which is fragile.  What if we wrote a "default job" to the database and just always merge whatever we read for a specific job over top of the default job?

Also, maybe factor this out into a default-handling db accessor library?

dgrnbrg:
I see how this can be fragile, but it's not clear to me that the default belongs in the DB rather than in the code.

leifwalsh:
The default is itself data, so belongs in the db, IMO. We can set it in the same place in the code where we define the schema. Or maybe a config file eventually?




Just a suggestion though. If you agree about the fragility of this then I think I've done my job here, I trust you to handle it. 



-- 
Cheers,
Leif

PR #44, opened 2015-09-24
Infinite loop during lost job races

dgrnbrg:
It seems that it's possible for reconciliation to cause Cook to keep recieving updates about an instance, and those updates will keep failing, which will kill due to consuming too many resources.

We must investigate under what circumstances we repeatedly try to transact an instance fail update that is repeatedly bailed out of by the transaction function _and_ triggers reconciliation or Mesos status updates (or maybe those are in response to a repeatedly generated `killTask`?).

```
2015-09-22 19:53:29,693 WARN  cook.mesos.scheduler [clojure-agent-send-off-pool-5] - Attempting to kill task 19b52a71-b837-4758-ac3d-41dae28e4050 as instance 17592186045465 with :job.state/completed and :instance.status/unknown should've been put down already
2015-09-22 19:53:29,706 WARN  cook.datomic [async-dispatch-23] - Retrying txn ([:instance/update-state 17592186045465 :instance.status/failed] [:db/add 17592186045465 :instance/end-time #inst "2015-09-22T19:49:23.169-00:00"]) attempt 10 of 10
2015-09-22 19:53:29,708 WARN  cook.datomic [async-dispatch-39] - Retrying txn ([:instance/update-state 17592186045465 :instance.status/failed] [:db/add 17592186045465 :instance/end-time #inst "2015-09-22T19:53:29.258-00:00"]) attempt 2 of 10
2015-09-22 19:53:29,708 WARN  cook.datomic [async-dispatch-41] - Retrying txn ([:instance/update-state 17592186045465 :instance.status/failed] [:db/add 17592186045465 :instance/end-time #inst "2015-09-22T19:53:29.222-00:00"]) attempt 2 of 10
2015-09-22 19:53:29,710 INFO  cook.mesos.scheduler [clojure-agent-send-off-pool-11] - Mesos status is:  {:task-id 19b52a71-b837-4758-ac3d-41dae28e4050, :state :task-lost, :message Reconciliation: Task is unknown, :timestamp 1.442951609695113E9, :source :source-master, :reason :reason-reconciliation}
2015-09-22 19:53:29,706 WARN  cook.datomic [async-dispatch-3] - Retrying txn ([:instance/update-state 17592186045465 :instance.status/failed] [:db/add 17592186045465 :instance/end-time #inst "2015-09-22T19:50:19.917-00:00"]) attempt 10 of 10
2015-09-22 19:53:29,714 WARN  cook.datomic [async-dispatch-43] - Retrying txn ([:instance/update-state 17592186045465 :instance.status/failed] [:db/add 17592186045465 :instance/end-time #inst "2015-09-22T19:53:29.714-00:00"]) attempt 1 of 10
```

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -209,10 +209,11 @@
                (handle-throughput-metrics complete-throughput-metrics
                                           job-resources
                                           instance-runtime))
-             (when (or (nil? instance)
+             ;; This code kills any task that "shouldn't" be running
+             (when (or (nil? instance) ; We could know nothing about the task, meaning a DB error happened and it's a waste to finish
                        (not= task-state :task-lost) ; killing an unknown task causes a TASK_LOST message. Break the cycle!
-                       (= prior-job-state :job.state/completed)
-                       (= prior-instance-status :instance.status/failed))
+                       (= prior-job-state :job.state/completed) ; The task is attached to a failed job, possibly due to instances running on multiple hosts
+                       (= prior-instance-status :instance.status/failed)) ; The kill-task message could've been glitched on the network
                (log/warn "Attempting to kill task" task-id
                          "as instance" instance "with" prior-job-state "and" prior-instance-status
                          "should've been put down already")
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -210,10 +210,14 @@
                                           job-resources
                                           instance-runtime))
              ;; This code kills any task that "shouldn't" be running
-             (when (or (nil? instance) ; We could know nothing about the task, meaning a DB error happened and it's a waste to finish
-                       (not= task-state :task-lost) ; killing an unknown task causes a TASK_LOST message. Break the cycle!
-                       (= prior-job-state :job.state/completed) ; The task is attached to a failed job, possibly due to instances running on multiple hosts
-                       (= prior-instance-status :instance.status/failed)) ; The kill-task message could've been glitched on the network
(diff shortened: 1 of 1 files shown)


dgrnbrg:
Li, what do you think the risk of this change is?

icexelloss:
I think it more clean to say kill a task if:

(1) mesos status = Running and (2) datomic status = Failed

dgrnbrg:
This code is specifically meant to kill tasks that are not expected to avoid weird cluster states, such as:

- Launched, but somehow not written to DB (the `nil?` check)
- Was supposed to be completed already, but was actually still running (a slave got disconnected but then reconnected when running mesos masters in non-strict mode) `(= prior-state :completed)`
- We thought the instance was dead, but our kill message was lost due to a network anomaly `(= instance-prior-state :failed)`

This new check specifically fixes when killTasks are sent to IDs that don't even exist, which would otherwise trigger a loop.

Given that, what do you think now?

icexelloss:
If I understand correctly, the code block is used to kill a task that shouldn't be running. So I think it's safe to assume we only want do execute this block when mesos report a task is running (and we think it shouldn't). That's why I argue it is cleaner to check if the new mesos task status is TASK_RUNNING, than to do the special case for TASK_LOST

dgrnbrg:
Hey Li,
I updated the comments to make it clearer what each case is catching.

Are you suggesting that the condition should be `(and running? (or ...reasons-to-kill))`?

icexelloss:
Yes, exactly.

dgrnbrg:
I see what you're saying. I made sure it kills if the task is in any of the several "active" task states.

icexelloss:
Ship it :+1:

Issue #50, opened 2015-09-28
Provide ability to update Priority of a Waiting job.


tnn1t1s:
Users would like programmatic ability to sort the queue. More generally, can we provide the ability to update any mutable property of a job via an UPDATE method.

dgrnbrg:
What properties of the job should be mutable? Currently, none of them are definitely safe to mutate--instead, the better type of design for this sort of system would be to provide a clone + cancel API, which would create new jobs based on an existing job's template, with some properties changed.

This will keep the data model mostly immutable, which will make it easier to continue to avoid consistency bugs.

tnn1t1s:
IMO priority is mutable. after discussion over here, i don't think other properties should be mutable e.g name. 

>>  the better type of design for this sort of system would be to provide a clone + cancel API,

this would be result in a different user experience than updating the queue in place. in the use case I have in mind, which I can explain 1/1, this experience delta would probably be significant.

dgrnbrg:
Could you write it up here? It's much better to have clearly written design discussions.

dgrnbrg:
Check out PR #52 and these docs: https://github.com/twosigma/Cook/blob/clone-api/scheduler/docs/scheduler-rest-api.asc#updatemodifyclone-a-job

This is what I have in mind, since it gives you almost the same user experience (you can update priorities of running jobs, restart jobs w/ more or fewer CPUs or memory, and change the user a job is running as).

Does this seem to solve your problems?

tnn1t1s:
I don't understand what the impact of changing a running job would be. It's already been scheduled. For queued jobs, I think it's important to retain the submission time and any other properties of the job that are considered in the scheduling order. In the case of priority, cancel/resubmit is not the same as update in place.

leifwalsh:
I thought the point of datomic was that updates preserve history for you so you can ignore this problem in the application. 



-- 
Cheers,
Leif

dgrnbrg:
[...]

PR #52, opened 2015-09-28
draft of clone api

dgrnbrg:
We'd like to have a way to update a job's properties, as asked for in #50. However, to maintain immutability throughout the system, we'll add this "clone" API, to enable creating new jobs based on existing ones, and optionally simultaneously killing the existing one.


dgrnbrg:
Moved to fork

PR #64, opened 2015-10-07
Add timeout when query slave state

icexelloss:


icexelloss:
This is to port patch from @WenboZhao . Without this, rest call to cook scheduler can hang.

PR #65, opened 2015-10-07
Fix 'multiple instance' race condition in cook scheduler

icexelloss:


dgrnbrg:
Why do you keep the first occurrence of each job in the sorted order?

icexelloss:
Which line are you referring to? To explain this, because the way we define "waiting task" (job/state = waiting) and "running job" (has a task with task/status = unknown or running), a newly launched job (job/state = waiting and task/status = unknown) can be considered both waiting and running.

Multiple instances happens when ranking function runs right after an instances launched and before the job/state gets changed to running.

Another possible fix is that in the filter function of line 667, only taking task ents that is in pending-task-ents. 

In the long run we should better define this to make them disjoint.

dgrnbrg:
I'm not sure that we can make waiting and running disjoint, b/c there's always an "unknown" transition state, when we're not sure that the task launch message has been delivered.

My question is whether the order of the job entity matters, and whether we should take the earlier or later occurrence of the job (since that might affect the priority). Which job is the "right one"?

icexelloss:
Good point. I pushed another change. I think this is cleaner now.

dgrnbrg:
Nice!

PR #68, opened 2015-10-08
Cleanup comments and add a standalone mesos example

tnn1t1s:
I tried to standardize on a comment format 
     - comment starts above relevant line
     - no more than 80 chars on line
     - two semicolons at start, and one to follow.

### DIFF
--- a/scheduler/dev-config.edn
+++ b/scheduler/dev-config.edn
@@ -1,17 +1,27 @@
 {:port 12321
- ;; We'll set the user to vagrant, since that's the default for many Vagrant-based Mesos setups
+ ;; We'll set the user to vagrant, since that's the default for many 
+ ;  Vagrant-based Mesos setups
  :authorization {:one-user "vagrant"}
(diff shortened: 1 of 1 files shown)


leifwalsh:
This is a bit of a weird style for lisp, where did you see it? 




Emac a is quite helpful with a style like this, obviously: http://www.gnu.org/software/emacs/manual/html_node/elisp/Comment-Tips.html




-- 
Cheers,
Leif

leifwalsh:
Emacs. Come on, ios. 



-- 
Cheers,
Leif

dgrnbrg:
The clojure style is 2 semicolons if the comment is on its own line, one
comment if it's sharing a line with code, and 3 semicolons if it's
toplevel. We use that everywhere; I think it's an emacs thing.

leifwalsh:
@dgrnbrg see the link I pasted



-- 
Cheers,
Leif

PR #69, opened 2015-10-09
Partial Solution: Accept host constraints in the cook json api

shamsimam:


scheduler/docs/scheduler-rest-api.asc
```diff
@@ -75,6 +75,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`mem` | double | MB of requested memory.
 |`uris` | list of URI objects | A list of URIs that will be fetched into the container before launch.
 |`env` | JSON map | A map of environment variables to be provided to the job.
+|`host_constraints` | JSON map | A map of host constraints to be provided to the job.
```
dgrnbrg: Once we cement the semantics, we'll need to make this more descriptive, and possibly include it in its own section. Also, we should indicate somewhere that it's optional.

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -108,14 +108,17 @@
...
   (let [conn (restore-fresh-database! "datomic:mem://mesos-api-test")
-        job (fn [cpus mem] {"uuid" (str (java.util.UUID/randomUUID))
-                            "command" "hello world"
-                            "name" "my-cool-job"
-                            "priority" 66
-                            "max_retries" 100
-                            "max_runtime" 1000000
-                            "cpus" cpus
-                            "mem" mem})
+        job-with-constraints (fn [cpus mem host-constraints]
```
dgrnbrg: you can do:

```clojure
(fn inner-name
  ([x y] (inner-name x y 0))
  ([x y z] (+ x y z)))
```

dgrnbrg:
Soon this will be something we can extend once Fenzo is merged to work end-to-end.

dposada:
@shamsimam Is this still active or can we close?

shamsimam:
I think this PR is now outdated and not required since we have Fenzo integration already working.

Closing.

Issue #13, opened 2015-09-14
Propagate the reason field for failed tasks


dgrnbrg:
The reason we show in json will be the enum value, but all lowercase, changing underscores to spaces, and dropping the leading `reason `.

Discussed with @icexelloss.

dgrnbrg:
@icexelloss what is the name of the field in the instance's json? Should it be `"reason"`, which is only set when it failed? Or should it have a different name, or have a default value? Are there other cases we're missing?

icexelloss:
@dgrnbrg, can you look at the initial design docs for cook error code?

https://github.com/icexelloss/Cook/blob/cook-error-code/scheduler/docs/cook-error-code

dgrnbrg:
This looks like a good design.

icexelloss:
Great!

I will start working on this.

icexelloss:
#104

PR #72, opened 2015-10-21
Include whether an instance was preempted in rest API

dgrnbrg:


icexelloss:
I prefer to do this as part of #13 , what do you think?

dgrnbrg:
I agree that it'd be good as part of #13

PR #84, opened 2015-11-18
Add Fenzo for scheduling

dgrnbrg:
This adds Fenzo as the task placer. To take advantage of Fenzo, we want to provide it with many tasks to place every cycle. In order to guarantee that big tasks get scheduled and Fenzo sees a lot of tasks, we exponentially reduce the # of tasks that Fenzo sees every scheduling cycle as long as it doesn't schedule the head of the queue. Once the head is scheduled, we reset the # of jobs that Fenzo gets to consider.

When Fenzo starts a task out of order, it is flagged as "backfilled". A backfilled task is bumped to the highest priority for the preemption system. Once a backfilled task moves to the head of the queue, however, we upgrade it to be non-backfilled. The backfilled tasks don't actually make a job register as "running", since they could be killed at any time.

### DIFF
--- a/scheduler/dev-config.edn
+++ b/scheduler/dev-config.edn
@@ -21,6 +21,4 @@
        :levels {"datomic.db" :warn
                 "datomic.peer" :warn
                 "datomic.kv-cluster" :warn
-                "cook.mesos.scheduler" :debug
-                "com.netflix.fenzo.TaskScheduler" :debug
                 :default :info}}}
--- a/scheduler/dev-config.edn
+++ b/scheduler/dev-config.edn
@@ -21,4 +21,6 @@
        :levels {"datomic.db" :warn
                 "datomic.peer" :warn
                 "datomic.kv-cluster" :warn
+                "cook.mesos.scheduler" :debug
+                "com.netflix.fenzo.TaskScheduler" :debug
                 :default :info}}}
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -31,10 +31,10 @@
                  [amalloy/ring-buffer "1.1"]
                  [lonocloud/synthread "1.0.4"]
                  [org.clojure/tools.namespace "0.2.4"]
-                 [org.clojure/core.cache "0.6.3"]
-                 [org.clojure/core.memoize "0.5.6"]
+                 [org.clojure/core.cache "0.6.4"]
+                 [org.clojure/core.memoize "0.5.8"]
                  [clj-time "0.9.0"]
-                 [org.clojure/core.async "0.1.346.0-17112a-alpha"]
+                 [org.clojure/core.async "0.2.374"]
                  [prismatic/schema "0.2.1"
                   :exclusions [potemkin]]
                  [clojure-miniprofiler "0.4.0"]
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -42,6 +42,7 @@
                  [org.clojure/data.priority-map "0.0.5"]
                  [swiss-arrows "1.0.0"]
                  [riddley "0.1.10"]
+                 [com.netflix.fenzo/fenzo-core "0.7.10"]
 
                  ;;Logging
                  [org.clojure/tools.logging "0.2.6"]
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -42,7 +42,10 @@
                  [org.clojure/data.priority-map "0.0.5"]
                  [swiss-arrows "1.0.0"]
                  [riddley "0.1.10"]
-                 [com.netflix.fenzo/fenzo-core "0.7.10"]
+                 [com.netflix.fenzo/fenzo-core "0.8.2"
+                  :exclusions [org.apache.mesos/mesos
+                               org.slf4j/slf4j-api
+                               org.slf4j/slf4j-simple]]
 
                  ;;Logging
                  [org.clojure/tools.logging "0.2.6"]
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -16,7 +16,7 @@
 (defproject cook "1.0.1-SNAPSHOT"
   :description "This launches jobs on a Mesos cluster with fair sharing and preemption"
   :license {:name "Apache License, Version 2.0"}
-  :dependencies [[org.clojure/clojure "1.6.0"]
+  :dependencies [[org.clojure/clojure "1.7.0"]
 
                  ;;Data marshalling
                  [org.clojure/data.codec "0.1.0"]
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -81,7 +81,8 @@
         datomic-report-chan (async/chan (async/sliding-buffer 4096))
         mesos-pending-jobs-atom (atom [])
         mesos-heartbeat-chan (async/chan (async/buffer 4096))
-        {:keys [scheduler view-incubating-offers view-mature-offers]}
+        current-driver (atom nil)
+        {:keys [scheduler view-incubating-offers]}
         (sched/create-datomic-scheduler
                    mesos-datomic-conn
                    (fn set-or-create-framework-id [framework-id]
@@ -89,13 +90,13 @@
                       curator-framework
                       zk-framework-id
                       (.getBytes framework-id "UTF-8")))
+                   current-driver
                    mesos-pending-jobs-atom
                    mesos-heartbeat-chan
                    offer-incubate-time-ms
                    task-constraints)
         framework-id (when-let [bytes (curator/get-or-nil curator-framework zk-framework-id)]
                        (String. bytes))
-        current-driver (atom nil)
         leader-selector (LeaderSelector.
                           curator-framework
                           zk-prefix
@@ -136,8 +137,7 @@
                                                                                                          :driver driver
                                                                                                          :mesos-master mesos-master
                                                                                                          :pending-jobs-atom mesos-pending-jobs-atom
-                                                                                                         :view-incubating-offers view-incubating-offers
-                                                                                                         :view-mature-offers view-mature-offers}))
+                                                                                                         :view-incubating-offers view-incubating-offers}))
                                     (counters/inc! mesos-leader)
                                     (async/tap mesos-datomic-mult datomic-report-chan)
                                     (let [kill-monitor (cook.mesos.scheduler/monitor-tx-report-queue datomic-report-chan mesos-datomic-conn current-driver)]
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -252,6 +252,7 @@
                     end (:instance/end-time instance)
                     base {:task_id (:instance/task-id instance)
                           :hostname hostname
+                          :backfilled (:instance/backfilled? instance false)
                           :slave_id (:instance/slave-id instance)
                           :executor_id (:instance/executor-id instance)
                           :status (name (:instance/status instance))}
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -55,7 +55,7 @@
    :max-retries (s/both s/Int (s/pred pos? 'pos?))
    :max-runtime (s/both s/Int (s/pred pos? 'pos?))
    (s/optional-key :uris) [Uri]
-   (s/optional-key :ports) [(s/pred zero? 'zero)] ;;TODO add to docs the limited uri/port support
+   (s/optional-key :ports) (s/pred pos? 'pos?)
    (s/optional-key :env) {NonEmptyString s/Str}
    :cpus PosDouble
    :mem PosDouble
@@ -67,10 +67,8 @@
   [conn jobs :- [Job]]
   (doseq [{:keys [uuid command max-retries max-runtime priority cpus mem user name ports uris env]} jobs
           :let [id (d/tempid :db.part/user)
-                ports (mapv (fn [port]
-                              ;;TODO this schema might not work b/c all ports are zero
-                              [:db/add id :job/port port])
-                            ports)
+                ports (when (and ports (not (zero? ports)))
+                        [[:db/add id :job/port ports]])
                 uris (mapcat (fn [{:keys [value executable? cache? extract?]}]
                                (let [uri-id (d/tempid :db.part/user)
                                      optional-params {:resource.uri/executable? executable?
@@ -144,7 +142,7 @@
                   :priority (or priority util/default-job-priority)
                   :max-retries max_retries
                   :max-runtime (or max_runtime Long/MAX_VALUE)
-                  :ports (or ports [])
+                  :ports (or ports 0)
                   :cpus (double cpus)
                   :mem (double mem)}
                  (when uris
@@ -239,7 +237,7 @@
        :status (name (:job/state job))
        :uris (:uris resources)
        :env (util/job-ent->env job)
-       ;;TODO include ports
+       :ports (:job/port job 0)
        :instances
        (map (fn [instance]
               (let [hostname (:instance/hostname instance)
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -252,7 +252,9 @@
                     end (:instance/end-time instance)
                     base {:task_id (:instance/task-id instance)
                           :hostname hostname
+                          ;;TODO validate that these show up in API
                           :backfilled (:instance/backfilled? instance false)
+                          :preempted (:instance/preempted? instance false)
                           :slave_id (:instance/slave-id instance)
                           :executor_id (:instance/executor-id instance)
                           :status (name (:instance/status instance))}
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -55,7 +55,7 @@
    :max-retries (s/both s/Int (s/pred pos? 'pos?))
    :max-runtime (s/both s/Int (s/pred pos? 'pos?))
    (s/optional-key :uris) [Uri]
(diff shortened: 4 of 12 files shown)


scheduler/dev-config.edn
```diff
@@ -21,4 +21,6 @@
        :levels {"datomic.db" :warn
                 "datomic.peer" :warn
                 "datomic.kv-cluster" :warn
+                "cook.mesos.scheduler" :debug
```
dgrnbrg: these levels should be removed before merge

scheduler/src/cook/mesos.clj
```diff
@@ -81,7 +81,8 @@
         datomic-report-chan (async/chan (async/sliding-buffer 4096))
         mesos-pending-jobs-atom (atom [])
         mesos-heartbeat-chan (async/chan (async/buffer 4096))
-        {:keys [scheduler view-incubating-offers view-mature-offers]}
+        current-driver (atom nil)
```
dgrnbrg: Exposing this data here feels hacky...perhaps this file should be refactored?

scheduler/src/cook/mesos/api.clj
```diff
@@ -252,6 +252,7 @@
                     end (:instance/end-time instance)
                     base {:task_id (:instance/task-id instance)
                           :hostname hostname
+                          :backfilled (:instance/backfilled? instance false)
```
dgrnbrg: This isn't exposed in the Java API

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -218,10 +218,22 @@
...
+    (log/fatal "Implementation detail failed; needed priority, got" (class task->scored-task))
+    (throw (ex-info "Implementation detail failed" {})))
   (let [{pending-job-mem :mem pending-job-cpus :cpus} (util/job-ent->resources pending-job-ent)
         pending-job-dru (compute-pending-job-dru state pending-job-ent)
+
         ;; This will preserve the ordering of task->scored-task
         host->scored-tasks (->> task->scored-task
+                                ;;TODO maybe we can just change every backfilled task here to have the worst possible DRU
+                                ;;this should bias us towards always killing those tasks first...
+                                (reduce-kv (fn [m k {:keys [task] :as scored-task}]
```
dgrnbrg: Will this actually work?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -354,14 +366,31 @@
...
+        requests (mapv (fn [job]
+                         (let [job-id (:db/id job)]
+                           (->TaskRequestAdapter job (job->task-info db fid (:db/id job)))))
+                       considerable)
+        result (.scheduleOnce fenzo requests leases)
+        assignments (.. result getResultMap values)]
+    (log/debug "Found this assigment:" result)
+    (when (log/enabled? :debug)
+      (log/debug "Task placement failure information follows:")
+      (doseq [failures (.. result getFailures values)
```
dgrnbrg: Some of this debugging logic should maybe be removed or cleaned up...but how?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -407,104 +416,239 @@
...
+  "Converts the DB function :job/allowed-to-start? into a predicate"
+  [db job]
+  (try
+    (d/invoke db :job/allowed-to-start? db (:db/id job))
+    true
+    (catch clojure.lang.ExceptionInfo e
+      false)))
+
+;;TODO LOL
+#_(let [db (db (d/connect "datomic:mem://cook-jobs"))]
```
dgrnbrg: This helper code should go away

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -407,104 +416,239 @@
...
-                driver
-                (:ids offer)
-                (mapv #(dissoc % :job-id :num-ports) task-infos))))))
+              (doseq [{:keys [tasks leases]} matches
+                      :let [offers (mapv :offer leases)
+                            task-infos (mapv (fn [^TaskAssignmentResult task]
+                                               (:task-info (.getRequest task)))
+                                             tasks)
+                            slave-id (:slave-id (first offers))]]
+                ;TODO if the ports are in the offer, then Fenzo is identifying the ports with a list .getAssignedPorts on the TaskAssignmentResult
```
dgrnbrg: We're still not handling ports at the very end, but the entire complex bits of the code path are handled. We should finish this before merge.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -536,6 +682,13 @@
                       :instance.status/running :task-running}]
     (when (seq running-tasks)
       (log/info "Preparing to reconcile" (count running-tasks) "tasks")
+      (doseq [[task-id] running-tasks
+              :let [task-ent (d/entity db task-id)
+                    hostname (:instance/hostname task-ent)
+                    job (:job/_instance task-ent)]]
+        (.. fenzo
+            (getTaskAssigner)
```
dgrnbrg: This isn't thread safe--should be done on a channel w/ the `offer-handler`'s loop or else we'll see IllegalStateExceptions.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -759,180 +915,34 @@
...
-      (catch Exception e
-        (log/error e "Error in start-offer-matcher!")))))
-
-(meters/defmeter [cook-mesos scheduler offer-back-pressure])
 (meters/defmeter [cook-mesos scheduler mesos-error])
 
+(defn make-fenzo-scheduler
+  [driver offer-incubate-time-ms]
+  (.. (com.netflix.fenzo.TaskScheduler$Builder.)
+      (disableShortfallEvaluation) ;; We're not using the autoscaling features
```
dgrnbrg: Maybe we should enable a bin packer?

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: @icexelloss Could you take a look and confirm this code does what I'm asserting?

scheduler/src/cook/mesos/api.clj
dgrnbrg: fixed

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -218,10 +218,22 @@
   [{:keys [task->scored-task host->spare-resources] :as state}
    {:keys [min-dru-diff safe-dru-threshold] :as params}
    pending-job-ent]
+  ;;We need to rely on this being a priority map TODO check that this is true
```
icexelloss: This is true. It has to be a priority map

scheduler/src/cook/mesos/rebalancer.clj
icexelloss: Can we change the per user sorting function to reflect backfilled?

dgrnbrg: Where is that defined? Which one? I had some trouble figuring out which sorting functions were used for which code.

icexelloss: https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/util.clj#L110

icexelloss: Feel free to put it other file. Util might not be the best place :(

dgrnbrg: We can't make the change there, since that code is used in many places, and we only want to penalize backfilled jobs during preemption, not during normal scheduling.

icexelloss: I prefer to have other two orders (preemption order and scheduling order) defined in two separate functions and revisit the use of the old sorting function to be replaced by one of the two and double check the assumption and invariant. We can sit together and do this if you want.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -354,14 +367,31 @@
...
+  (let [t (System/currentTimeMillis)
+        leases (mapv #(->VirtualMachineLeaseAdapter % t) offers)
+        requests (mapv (fn [job]
+                         (let [job-id (:db/id job)]
+                           (->TaskRequestAdapter job (job->task-info db fid (:db/id job)))))
+                       considerable)
+        result (.scheduleOnce fenzo requests leases)
+        failure-results (.. result getFailures values)
+        assignments (.. result getResultMap values)]
+    (log/info "Found this assigment:" result)
```
dgrnbrg: We should also look at and log these: http://netflix.github.io/Fenzo/fenzo-core/com/netflix/fenzo/TaskScheduler.html#scheduleOnce-java.util.List-java.util.List-

scheduler/project.clj
```diff
@@ -31,10 +31,10 @@
                  [amalloy/ring-buffer "1.1"]
                  [lonocloud/synthread "1.0.4"]
                  [org.clojure/tools.namespace "0.2.4"]
-                 [org.clojure/core.cache "0.6.3"]
-                 [org.clojure/core.memoize "0.5.6"]
+                 [org.clojure/core.cache "0.6.4"]
+                 [org.clojure/core.memoize "0.5.8"]
                  [clj-time "0.9.0"]
-                 [org.clojure/core.async "0.1.346.0-17112a-alpha"]
+                 [org.clojure/core.async "0.2.374"]
```
icexelloss: Some of the newer core.async seems to require clojure 1.7, is this upgrade safe?

scheduler/src/cook/mesos.clj
icexelloss: What is current-driver used for?

scheduler/src/cook/mesos/util.clj
```diff
@@ -108,15 +108,22 @@
...
+   If consider-backfilled-jobs? is true, then we treat jobs which were backfilled as being the lowest priority"
+  [consider-backfilled-jobs?]
+  (letfn [(task->feature-vector
+            [task]
             ;; Last two elements are aribitary tie breakers.
             ;; Use :db/id because they guarantee uniqueness for different entities
             ;; (:db/id task) is not sufficient because synthetic task entities don't have :db/id
             ;; This assumes there are at most one synthetic task for a job, otherwise uniqueness invariant will break
-            [(- (:job/priority (:job/_instance task) default-job-priority))
+            [(if (and consider-backfilled-jobs? (:instance/backfilled? task))
```
icexelloss: I forgot the conclusion of our discussion last but I feel backfilled should be just the first element of the vector instead of wiping out the original priority information.

scheduler/src/cook/mesos/schema.clj
```diff
@@ -221,6 +227,12 @@
     :db/index true
     :db.alter/_attribute :db.part/db}])
 
+(def migration-port-is-a-count
```
icexelloss: Can you explain what is this?

scheduler/src/cook/mesos/scheduler.clj
icexelloss: Why do we need to call TaskAssigner here? What's the invariant that we are trying to keep?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -407,107 +417,232 @@
...
+(defn make-offer-handler
+  [conn driver-atom fenzo fid-atom pending-jobs-atom]
+  (let [offers-chan (async/chan (async/buffer 5))
+        resources-atom (atom (view-incubating-offers fenzo))
+        timer-chan (chime-ch (periodic/periodic-seq (time/now) (time/seconds 1))
+                             {:ch (async/chan (async/sliding-buffer 1))})
+        max-considerable 1000]
+    (async/thread
+      (loop [num-considerable max-considerable]
+        ;;TODO make this cancelable
```
icexelloss: This comment is confusing...Can you elaborate in the comment?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -407,107 +417,232 @@
...
+            (reset! front-of-job-queue-mem-atom
+                    (or (:mem first-considerable-resources) 0))
+            (reset! front-of-job-queue-cpus-atom
+                    (or (:cpus first-considerable-resources) 0))
+            (cond
+              (empty? matches) (if (seq considerable)
+                                 nil ;; This case means we failed to match anything. We need to start head of line blocking now
+                                 true) ;; This case means there was nothing to match--don't penalize the scheduler for the next burst of submitted jobs
+              (not (compare-and-set! pending-jobs scheduler-contents new-scheduler-contents))
+              (do
```
icexelloss: Is it still safe to do so? Is the call into fenzo idempotent?

icexelloss: If I remember correctly, this code deals with multiple thread running handle-resource-offers! How does multiple thread  with a single fenzo instance work? Do we still to need to handle multiple thread case or it will be single threaded?

icexelloss:
In general, it would be great if the effect of backfilled-instances are well documented in a single place to be better maintainable. Now it's kind of all over the place, things like:
(1) Backfilled instances are of the lowerest priority?
(2) Having backfilled intance doesn't change the job state to be running?
...

scheduler/project.clj
dgrnbrg: Good point, moved to 1.7

scheduler/src/cook/mesos/util.clj
dgrnbrg: :+1:

scheduler/src/cook/mesos/schema.clj
dgrnbrg: In an earlier commit, the port schema in the DB was meant to be a set of the ports allocated. Now, it's the # of ports allocated.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: Eventually, if we were to clean up the scheduler's loops to make it more repl-development friendly, this is a loop that should be terminate-able when shutting down the system. Today, there are many things that don't have any way to be stopped, but I've been trying to tag them so that they're easy to find in the future, should we decide to make this type of refactoring.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: This is single-threaded--it's called by the `make-offer-handler` loop only. This race only fails if the prioritizer raced, in which case we recycle the offers and try again.

dgrnbrg:
Where do you think that backfilled instances could be documented?

dgrnbrg:
@icexelloss @wyegelwel additional comments?

scheduler/src/cook/mesos/scheduler.clj
icexelloss: Ok

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -407,107 +417,232 @@
...
-          (reset! front-of-job-queue-cpus-atom
-                  (or (:cpus first-considerable-resources) 0))
-          (cond
-            (empty? matches) (decline-offers driver (:ids offer))
-            (not (compare-and-set! pending-jobs scheduler-contents new-scheduler-contents))
-            (do (log/info "Pending job atom contention encountered")
+      offer-size-mem
+      (get-in offer [:resources :mem] 0)))
+  (log/debug "invoked handle-resource-offers!")
+  (let [offer-stash (atom nil)] ;; This is a way to ensure we never lose offers fenzo assigned if an errors occures in the middle of processing
```
icexelloss: Maybe put a comment:

If we fail to launch some of the tasks, we will put all the offers - used and unused - onto offer-stash. Those used offers will be feed back to frenzo again for scheduling, which will probably cause task lost next time we use it.

icexelloss:
Put docs about backfilled instance and the state machine of jobs

scheduler/src/cook/mesos/api.clj
```diff
@@ -55,7 +55,7 @@
    :max-retries (s/both s/Int (s/pred pos? 'pos?))
    :max-runtime (s/both s/Int (s/pred pos? 'pos?))
    (s/optional-key :uris) [Uri]
-   (s/optional-key :ports) [(s/pred zero? 'zero)] ;;TODO add to docs the limited uri/port support
+   (s/optional-key :ports) (s/pred #(not (neg? %)) 'nonnegative?)
```
wyegelwel: Is ports allowed to be a list?

scheduler/src/cook/mesos/api.clj
```diff
@@ -67,10 +67,8 @@
   [conn jobs :- [Job]]
   (doseq [{:keys [uuid command max-retries max-runtime priority cpus mem user name ports uris env]} jobs
           :let [id (d/tempid :db.part/user)
-                ports (mapv (fn [port]
-                              ;;TODO this schema might not work b/c all ports are zero
-                              [:db/add id :job/port port])
-                            ports)
+                ports (when (and ports (not (zero? ports)))
```
wyegelwel: Why do we discount the port if it is zero? Simply because we default to it later? Why does everything get a port?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -342,7 +343,6 @@
                                     (fn [now]
                                       (let [host->combined-offers
```
wyegelwel: Which offers is this seeing? Pre or post fenzo? If pre, change the name since we are no longer combining offers here.

scheduler/src/cook/mesos/api.clj
wyegelwel: I see it is the number of ports.

wyegelwel:
I'm closing this in favor of #145 which is just about ready to merge.

PR #85, opened 2015-11-19
Add mesos master hosts configuration

icexelloss:


icexelloss:
This allows us to use mesos http api

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -359,7 +358,7 @@
                                                       (map (fn [[k v]]
                                                              [(keyword (name (keyword k))) v]))
                                                       (into {}))
-                                              utilization (get-mesos-utilization mesos-master)
+                                              utilization (get-mesos-utilization mesos-master-hosts)
```
dgrnbrg: What is the purpose of this? Is it just so that the rebalancer only runs when the utilization is above some minimum threshold?

scheduler/src/cook/components.clj
```diff
@@ -291,6 +292,13 @@
                                   offer-incubate-ms)
      :mesos-master (fnk [[:config [:mesos master]]]
                         master)
+
+     :mesos-master-hosts (fnk [[:config [:mesos master {master-hosts nil}]]]
```
dgrnbrg: If this is supposed to be configurable, it should be exposed in the end-user config.

dgrnbrg:
If the goal of this is to be getting global cluster stats + utilization, that's something that Fenzo already dynamically calculates w/o the additional dependency on reading this data over HTTP--Fenzo infers and remembers the size of the cluster as observed. We should probably just port this to use that collector, to reduce the config complexity overall.

scheduler/src/cook/components.clj
icexelloss: This is exposed in end-user config. User can specify master-hosts in the config.

scheduler/src/cook/mesos/rebalancer.clj
icexelloss: Yes

icexelloss:
Using Fenzo for this sounds like a good idea. Until that is done, this change allows people to run mesos master the zookeeper on different hosts. In the future, this can also be used for mesos http api.

scheduler/src/cook/components.clj
dgrnbrg: Can you add this to the docs?

icexelloss: Updated

scheduler/docs/configuration.asc
```diff
@@ -73,6 +73,10 @@ We'll look at the configurable options in turn:
   This option sets the Mesos master connection string.
   For example, if you are running Mesos with a Zookeeper node on the local machine (a common development setup), you'd use the connection string `zk://localhost:2181/mesos`.
 
+`:master-hosts`::
```
dgrnbrg: Can you give an example of how to format this string here?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -376,19 +375,6 @@
        (shutdown-rebalancer))))
 
 (comment
-  (let [mesos-master "zk://sf-test-zk1.pit.twosigma.com:2181,sf-test-zk2.pit.twosigma.com:2181,sf-test-zk3.pit.twosigma.com:2181/mesos"
```
dgrnbrg: Can you leave this here? I think it might be the only place that the rebalancer reconfiguration is demonstrated...at least until we make it easier to configure.

icexelloss: I left another example. It's at the end of file.

scheduler/docs/configuration.asc
icexelloss: Done

dgrnbrg: This is a list parameter?

scheduler/src/cook/components.clj
```diff
@@ -291,6 +292,13 @@
...
+                              (if master-hosts
```
dgrnbrg: This should be something like

```
(if master-hosts
  (cond
    (string? master-hosts)
    (throw (ex-info ":master-hosts should be a list (e.g. ['host1.example.com', ...])" {}))
    (not (every? string? master-hosts))
    (throw "master hosts should be a sequence of strings of the hostnames")
    ...
```

icexelloss: Hmm...Wonder if we can do schema checks instead of checking this manually.

dgrnbrg: The important thing with the diagnostic messages on the main input is to
not use Schema, but instead try to provide actionable instructions to the
user.

scheduler/docs/configuration.asc
icexelloss: Done

icexelloss:
@dgrnbrg, anything more needs to be done?

scheduler/src/cook/components.clj
```diff
@@ -291,6 +292,14 @@
...
      :mesos-master (fnk [[:config [:mesos master]]]
                         master)
+     :mesos-master-hosts (fnk [[:config [:mesos master {master-hosts nil}]]]
+                              (if master-hosts
+                                (if (and (sequential? master-hosts) (every? string? master-hosts))
+                                  master-hosts
+                                  (throw (ex-info ":mesos-master should be a list of hostnames (e.g. [\"host1.example.com\", ...])" {})))
+                                (->> master
+                                     (re-seq #"[/|,]?([^/,:]+):\d+")
+                                     (map second))))
```
dgrnbrg: Can you make this a `mapv` or follow it with a `doall` so that regex parse failures trigger sooner? Also, validate that the result of this passes `(every? string? %)`

dgrnbrg:
@icexelloss ping

icexelloss:
@dgrnbrg , I updated

dgrnbrg:
Could you add the `(every? string?)` validation as well?

dgrnbrg:
Can you squash the commits? This is good to merge.

PR #86, opened 2015-11-20
Catch exception when updating offer metrics

icexelloss:


dgrnbrg:
I don't think that this is correct approach. Instead, we should have a default value for the get-in. Since now, we just sometimes silently hide useful metrics.

PR #89, opened 2015-11-20
Revert "Catch exception when updating offer metrics"

dgrnbrg:
Reverts twosigma/Cook#86

Instead, this should provide defaults for the offer cpus and mem.


dgrnbrg:
@icexelloss can you acknowledge this?

icexelloss:
We no longer need default here because of  https://github.com/twosigma/Cook/pull/87

This is merely defensive. Without this, an exception in publishing the metrics could crash the entire offer buffer thread. I don't think it's necessary to revent the change.

PR #101, opened 2016-03-11
simulator:  Introduce simulation tests.

mforsyth:


mforsyth:
@wkf Removing the "into []" causes the lazy sequence not to get realized, so I left it as is... other than that, resolved all of your "round 1" comments and re-pushed.

PR #102, opened 2016-03-11
Implement offerRescinded callback

icexelloss:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -925,9 +934,12 @@
         (log/debug "Preparing to pull offer (offer-matcher)")
         (let [offer (async/<!! matcher-chan)
               fid (:fid offer)
+              offer-ids (:ids offer)
               driver (:driver offer)]
           (log/debug "(offer-matcher) Offer taken " offer)
-          (handle-resource-offer! conn driver fid pending-jobs-atom offer))
+          (if-let [invalid-offer-id (some #(cache/lookup @rescinded-offer-id-cache %) offer-ids)]
+            (log/info "Received rescinded offer" invalid-offer-id "in offer matcher")
```
wyegelwel: Add a meter metric for this so we can track the rate that this happens

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -101,6 +102,14 @@
              "UTF-8")
      executor-key executor-value}))
 
+(defn rescind-offer!
+  [rescinded-offer-id-cache offer-id]
+  (log/info "Rescinding offer" offer-id)
+  (swap! rescinded-offer-id-cache (fn [c]
```
wyegelwel: Add a meter here so we know how often offers are rescinded

scheduler/src/cook/mesos/scheduler.clj
icexelloss: Isn't this the same as the rescinding rate?

wyegelwel: I'm not sure =)

wyegelwel: I'm not sure. At the very least, the diff in time will give us information.

icexelloss:
Due to resource constraints, I am deferring metrics to a later mr

PR #110, opened 2016-03-29
simulator:  Introduce simulation tests.

mforsyth:


simulator/config/two-sigma-settings.edn
```diff
@@ -0,0 +1,65 @@
...
+               :job-memory {:mean 40960
+                            :std-dev 10240
+                            :floor 1024
+                            :ceiling 512000}
+               :job-cpu {:mean 3.0
+                         :std-dev 1.0
+                         :floor 1.0
+                         :ceiling 8.0}}
+
+              {:description "Fcat users"
```
wyegelwel: Can you change the names to reflect what the jobs are, essentially:

short running low resource intensive for this group, long running high memory for another...

simulator/resources/job_schedule.edn
```diff
@@ -0,0 +1,103 @@
+{:model
+ [[{:db/id #db/id[:db.part/db]
+    :db/ident :model.type/cook
+    :db/doc "Cook scheduling jobs for a Mesos cluster and reporting on results."}
+   {:db/id #db/id[:db.part/db]
+    :db/ident :model/num-researchers
+    :db/valueType :db.type/long
+    :db/doc "Number of researchers who schedule jobs"
```
wyegelwel: Can we change from researchers -> users

simulator/src/dev/cook/sim/repl.clj
```diff
@@ -0,0 +1,208 @@
...
+  (-> (settings) :cook-db-uri datomic.api/delete-database))
+
+(defn list-scheds
+  []
+  (schedule/list-job-schedules (sim-db)))
+
+(defn list-sims
+  [sched-id]
+  (report/list-sims (sim-db) (cook-db) sched-id
+                    #inst "2016-06-21T23:50:03.892-00:00"))
```
wyegelwel: Any reason why this is hardcoded to this time?

simulator/config/two-sigma-settings.edn
mforsyth: Well - this file is actually supposed to proportionally mimic Two Sigma's specific use case.  I wonder, is it even a good idea to have such a file in this repo at all?

I think it will be difficult to concisely describe the profiles we are talking about in a way that would be suitable for a label on a graph... for example, "Spark users" are unusual in these sims because the job-duration is always exactly 7200 seconds.  "Long running, consistent runtime, high memory, consistent CPU" wouldn't be good label on a graph...  this string is really supposed to describe WHY the collection of attributes is the way it is, rather than just summarize the attributes.

The problem would only get worse if there were a lot more profiles, or the behavior of each profile was modelled in a more sophisticated way.

PR #117, opened 2016-04-06
Add metrics-clojure-ring, move us back to mainline metrics-clojure

sophaskins:
It appears that in purging the codebase of voom (5c44eff1a52339176f1b6692639effd4970b8a6f) we lost the automatic Ring metrics.

This adds them back, and additionally puts us back on the upstream published version of metrics-clojure, now that it's been published.

### DIFF
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -32,7 +32,8 @@
             [compojure.core :refer (GET POST routes context)]
             [compojure.route :as route]
             [plumbing.graph :as graph]
-            [cook.curator :as curator])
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/components.clj
```diff
@@ -158,7 +159,8 @@
                                                        wrap-stacktrace-web
                                                        wrap-no-cache
                                                        wrap-params
-                                                       health-check-middleware)
+                                                       health-check-middleware
+                                                       instrument)
```
icexelloss: What is the effect of "instrument"?

PR #122, opened 2016-04-14
Add timers for job submission

sophaskins:
Timers around the two places that are possible entry points for a job

### DIFF
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -20,7 +20,8 @@
             [metatransaction.utils :as dutils]
             [clojure.tools.logging :as log]
             [clj-mesos.scheduler]
-            [metrics.counters :as counters]
+            [metrics.counters :as counters
+            [metrics.timers :as timers]
             [cook.datomic :refer (transact-with-retries)]
             [cook.mesos.scheduler :as sched]
             [cook.mesos.heartbeat]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos.clj
```diff
@@ -61,7 +63,8 @@
                      base-wait 500 ; millis
                      opts {:retry-schedule (cook.util/rand-exponential-seq retries base-wait)}]]
          (if (and (<= memory 200000) (<= ncpus 32))
-           (dutils/transact-with-retries!! conn opts (concat [txn] additional-txns-per-job))
+           (timers/time! submit-to-mesos-duration
```
wyegelwel: We should also have a counter of number of jobs submitted, and perhaps a timer that is normalized for the number of jobs run (submit time / job ). One would hope submit time / job == `submit-to-mesos-duration`. Even if the submit time is linear, we can start to reason about how long a submission of n jobs *should* take.

scheduler/src/cook/mesos/api.clj
```diff
@@ -115,11 +117,11 @@
                                     {:resource/type :resource.type/mem
                                      :resource/amount mem}]}]]
     ;; TODO batch these transactions to improve performance
-    @(d/transact conn (-> ports
-                          (into uris)
-                          (into env)
-                          (into maybe-datoms)
-                          (conj txn))))
+    (timers/time! @(d/transact conn (-> ports
```
wyegelwel: I think you are missing the `submit-jobs-duration` var after timers/time!

Same comment as in submit-to-mesos about counter on jobs and timer on the `doseq`.

Issue #123, opened 2016-04-15
Cook API attempts to log Kerberos user even if configured auth isn't Kerberos


mforsyth:
This causes 500 errors on API requests when not using Kerberos. 

This was introduced here:
bb2b63228566400c423925e4ac30f838ebddf052

Should leave the same access logging behavior in,  just do it only when it's actually configured to use Kerberos.

wyegelwel:
[Ring_ Stacktrace.pdf](https://github.com/twosigma/Cook/files/225053/Ring_.Stacktrace.pdf)

Attached the error message when this occurs. It appears that this stackoverflow discusses the issue: http://stackoverflow.com/questions/15029148/tomcat-fails-to-authenticate-using-spnego-cannot-locate-default-realm

Perhaps the best thing to do is change to use a different auth method since it is only used as a shim  to get a user associated with the request.

wyegelwel:
This is no longer an issue following the merging of PR #125

PR #125, opened 2016-04-19
Create shim principal for access logs

wyegelwel:
Swap out the kerberos principal which was causing errors for non-kerberized auth environments (issue #123) and instead just pass in a shim principal.


wyegelwel:
@dgrnbrg @icexelloss Any comments?

PR #126, opened 2016-05-10
Add support for job labels

rustyconover:
Add support for job labels.

### DIFF
--- a/jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -93,6 +93,7 @@ final public class Job {
         private List<FetchableURI> _uris = new ArrayList<>();
         private Map<String,String> _env = new HashMap<>();
         private JSONObject _container;
+        private Map<String,String> _labels = new HashMap<>();
 
         /**
          * Prior to {@code build()}, command, memory and cpus for a job must be provided.<br>
@@ -118,7 +119,7 @@ final public class Job {
                 _status = Status.INITIALIZED;
             if (_name == null)
                 _name = "cookjob";
-            return new Job(_uuid, _name, _command, _memory, _cpus, _retries, _status, _priority, _instances, _env, _uris, _container);
+            return new Job(_uuid, _name, _command, _memory, _cpus, _retries, _status, _priority, _instances, _env, _uris, _container, _labels);
         }
 
         /**
@@ -135,6 +136,7 @@ final public class Job {
             setEnv(job.getEnv());
             setUris(job.getUris());
             setContainer(job.getContainer());
+            setLabels(job.getLabels());
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -213,6 +215,44 @@ public Builder setEnv(Map<String,String> enviroment) {
...
+        /**
+         * Adds a collection of label to the job.
+         *
+         * This resets the labels of the job; it will remove labels that were previously set.
+         *
+         * @param labels specifies the labels to set for this job.
+         * @return this builder
+         */
+        public Builder setLabels(Map<String,String> labels) {
+            _labels = ImmutableMap.copyOf(labels);
```
leifwalsh: we won't be able to do setLabels() followed by addLabel(), I assume that's ok

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -371,6 +413,7 @@ private Job(UUID uuid, String name, String command, Double memory, Double cpus,
         } else {
             _container = null;
         }
+        _labels = ImmutableMap.copyOf(labels);
```
leifwalsh: similarly, we won't be able to initialize with a label map and then add additional labels

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
rustyconover: This is the same restrictions as the environment, so I believe we're okay.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
rustyconover: Same reply as above.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -507,6 +558,7 @@ public static JSONObject jsonizeJob(Job job)
         object.put("max_retries", job.getRetries());
         object.put("status", job.getStatus());
         object.put("env", env);
+        object.put("labels", env);
```
wyegelwel: Looks like we are setting the `"labels"` to `env`

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -617,6 +669,16 @@ public static JSONObject jsonizeJob(Collection<Job> jobs)
                 }
                 jobBuilder.setEnv(envMap);
             }
+            if (json.has("labels")) {
+                JSONObject labelsJson = json.getJSONObject("labels");
+                Map<String, String> labelsMap = new HashMap<>();
+                if (labelsJson.length() > 0) {
```
wyegelwel: Is this check necessary?

rustyconover: I think so, I'm just following how the environment is handled.

PR #127, opened 2016-05-13
labels for raw scheduler

tnn1t1s:

PR #128, opened 2016-05-16
Labels

mforsyth:
This pull request is based on https://github.com/twosigma/Cook/pull/127.  

This one includes a fix (correcting an attribute name in Datomic) and also introduces a bit of unit test coverage to make sure the labels are really populated.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -57,6 +57,7 @@
    (s/optional-key :uris) [Uri]
    (s/optional-key :ports) [(s/pred zero? 'zero)] ;;TODO add to docs the limited uri/port support
    (s/optional-key :env) {NonEmptyString s/Str}
+   (s/optional-key :labels) {NonEmptyString s/Str}
    :cpus PosDouble
    :mem PosDouble
    ;; Make sure the user name is valid. It must begin with a lower case character, end with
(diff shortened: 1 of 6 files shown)


tnn1t1s:
Can you also validate  https://github.com/twosigma/Cook/pull/126

This will allow us to set labels in the CoaseGrainedScheduler.

mforsyth:
OK - I have actually never used JobClient yet so depending on what's involved it may take me some time to get that set up.  Will get to it this afternoon.

scheduler/src/cook/mesos/api.clj
```diff
@@ -95,6 +96,13 @@
                                   :environment/name k
                                   :environment/value v}]))
                             env)
+                labels (mapcat (fn [[k v]]
+                              (let [label-var-id (d/tempid :db.part/user)]
```
wyegelwel: Fix indent

wyegelwel:
This all looks good. Did you have a chance to test it against a mesos cluster to ensure that it is set correctly on mesos's side?

mforsyth:
@wyegelwel Ah, I didn't even realize that this was intended to make labels attached to the actual Mesos tasks.  I just tried it and no, the tasks that come from the jobs don't have any labels.

Just to be clear:   We want the labels attached to the Cook jobs to also be attached to all Mesos tasks that are instances of the job?

rustyconover:
@mforsyth Attaching the labels to all of the generated Mesos tasks from a Cook job is the intention.

mforsyth:
OK, the problem was with clj-mesos.  This PR is now waiting on 

https://github.com/twosigma/clj-mesos/pull/1

... which adds Labels support.  Once that PR is merged, in this branch I'll need to change Cook scheduler's project.clj to use Two Sigma's fork of clj-mesos in order to get access to the fix.

leifwalsh:
@mforsyth I merged twosigma/clj-mesos#1, you can go ahead and update project.clj and run any last tests you want to.  Let me know when you're ready to have this merged.

mforsyth:
OK, I added the new dep for clj-mesos  that supports Labels.  I just tested it locally end-to-end one more time, and it works (labels populated on Mesos tasks).

PR #130, opened 2016-05-27
Using trusty mesos package

pradeepchhetri:


dposada:
Is this still active or can it be closed?

dposada:
Closing since this seems to be inactive

PR #132, opened 2016-05-28
Make sort for rank jobs lazy

wyegelwel:
When there are many jobs in the queue, the ranking function can take a long time because it must sort all jobs. This PR makes the final sort across all jobs lazy so that the affect of many jobs in the queue is greatly reduced.

### DIFF
--- a/scheduler/src/cook/mesos/dru.clj
+++ b/scheduler/src/cook/mesos/dru.clj
@@ -64,19 +64,38 @@
                           task-resources)]
     scored-tasks))
 
-(defn scored-task->priority
-  "The higher the dru of task, the lower the priority"
-  [scored-task]
-  (- (:dru scored-task)))
+(defn sorted-merge
+  "Accepts a seq-able datastructure `colls` where each item is seq-able.
+   Each seq-able in `colls` is assumed to be sorted based on `key-fn`
+   Returns a lazy seq containing the sorted items of `colls`
+
+   Initial code from: http://blog.malcolmsparks.com/?p=42"
+  ([key-fn ^java.util.Comparator comp-fn colls]
+   (letfn [(next-item [[_ colls]]
+             (if (nil? colls)
+               [:end nil] ; Allow nil items in colls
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/mesos/dru.clj
```diff
@@ -64,19 +64,38 @@
...
+                     (sort-by (comp key-fn first) comp-fn colls)]
+                 [yield (if remaining (cons remaining other-colls) other-colls)])))]
+     (->> colls
+       (vector :begin) ; next-item input is [item colls], need initial item
+       (iterate next-item)
+       (drop 1) ; Don't care about :begin
+       (map first)
+       (take-while (partial not= :end)))))
+  ([key-fn colls]
+   (sorted-merge key-fn compare colls))
```
icexelloss: Is passing "compare" preferable here? It does make it more complicated to read

scheduler/src/cook/mesos/dru.clj
```diff
@@ -64,19 +64,38 @@
...
+   Returns a lazy seq containing the sorted items of `colls`
+
+   Initial code from: http://blog.malcolmsparks.com/?p=42"
+  ([key-fn ^java.util.Comparator comp-fn colls]
+   (letfn [(next-item [[_ colls]]
+             (if (nil? colls)
+               [:end nil] ; Allow nil items in colls
+               (let [[[yield & remaining] & other-colls]
+                     (sort-by (comp key-fn first) comp-fn colls)]
+                 [yield (if remaining (cons remaining other-colls) other-colls)])))]
```
icexelloss: it's a bit strange to use cons instead of conj here (since we don't care where it's added?), but I guess it doesn't matter

icexelloss:
Looks good to me. There are minor comments but I think you can merge it as-is.

wyegelwel:
I'm going to add tests as well as another performance improvement.

Would you like me to split out into two PR or use this one?

icexelloss:
We can make it separate

PR #133, opened 2016-06-03
Use unfiltered db when possible for query improvement

wyegelwel:
This commit has the `sort-jobs-by-dru` use an unfiltered db when possible. Throughout most (all) of the code, we use a filtered datomic database which will remove datoms which have a metatransaction status of uncommitted. We do this so that jobs will only be considered valid once they have all been submitted so that we could avoid a case where we try to schedule a partial submission. Unfortunately, using a filtered database is not free, and in some cases can be costly. In the case of ranking jobs, it is very costly. The reason for the cost is that the filter must check every datom to see if it passes. Worse, if the datom doesn't have a metatransaction linked to it, it has to do a more expensive search to confirm that. If we can avoid using the filtered db where possible, both the initial poll for data gets fast as well as looking up additional attributes on the entity. 

As an aside, we should revisit the idea of metatransaction because it comes with such hefty performance consequences.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -687,12 +687,17 @@
                (zipmap (util/get-all-resource-types db) (repeat 0.0)))})))
 
 (defn sort-jobs-by-dru
-  "Return a lazy list of job entities ordered by dru"
-  [db]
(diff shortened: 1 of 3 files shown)


icexelloss:
This looks good to me

PR #91, opened 2015-12-02
First draft of container support.

sdegler:
Schema (validator),  Schema(datomic), and input code needed to add container support to cook.
Still has spurious pprints in, these need to be removed prior to commit.
Clojure style and idiom comments more than welcome.

PR #135, opened 2016-06-14
Add docker support 

wyegelwel:
This code was derived from sdegler's pull request #91.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -24,15 +24,43 @@
             [clojure.tools.logging :as log]
             [clojure.string :as str]
             [clojure.core.cache :as cache]
+            [clojure.walk :as walk]
             [clj-http.client :as http]
             [ring.middleware.json]
+            [clojure.data.json :as json]
             [liberator.core :refer [resource defresource]]
+            [cheshire.core :as cheshire]
             [cook.mesos.util :as util]
             [clj-time.core :as t])
   (:import java.util.UUID))
 
 (def PosDouble
   (s/both double (s/pred pos? 'pos?)))
+
+(def PortMapping
+  "Schema for Docker Portmapping"
+  {:host-port (s/both s/Int (s/pred #(<= 0 % 65536) 'between-0-and-65536))
+   :container-port (s/both s/Int (s/pred #(<= 0 % 65536) 'between-0-and-65536))
+   (s/optional-key :protocol) s/Str})
+
+(def DockerInfo
+  "Schema for a DockerInfo"
+  {:image s/Str
+   (s/optional-key :network) s/Str
+   (s/optional-key :parameters) [{:key s/Str :value s/Str}]
+   (s/optional-key :port-mapping) [PortMapping]})
+
+(def Volume
+  "Schema for a Volume"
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -58,15 +86,76 @@
...
+                                {:docker.param/key key
+                                 :docker.param/value value})
+                              params)}))
+
+(defn- mkvolumes
+  "Helper for build-container.  Transforms volumes into the datomic schema."
+  [cid vols]
+  (when-not (empty? vols)
+    (let [vol-maps (mapv (fn [{:keys [container-path host-path mode]}]
+                           (merge (when container-path
```
leifwalsh: (into {} (filter second {:container.volume/container-path container-path
                             :container.volume/host-path host-path
                             :container.volume/mode (clojure.string/upper-case mode)}))

leifwalsh: (only if `(nil? (clojure.string/upper-case nil))`)

scheduler/src/cook/mesos/api.clj
```diff
@@ -58,15 +86,76 @@
...
+                                     {:docker.portmap/host-port host-port})
+                                   (when protocol
+                                     {:docker.portmap/protocol (clojure.string/upper-case protocol)})))
+                          ports)]
+      {:docker/port-mapping port-maps})))
+
+(defn- build-container
+  "Helper for submit-jobs, deal with container structure."
+  [id container]
+  (let [container-id (d/tempid :db.part/user)
```
leifwalsh: (when (and container (= (clojure.string/lower-case (:type container)) "docker))
      (let ...))

scheduler/src/cook/mesos/api.clj
```diff
@@ -103,6 +192,7 @@
                                   :label/key k
                                   :label/value v}]))
                             labels)
+                container (if (nil? container) [] (build-container id container))
```
leifwalsh: (or (build-container id container) [])

scheduler/src/cook/mesos/api.clj
leifwalsh: (also, optional, I'm not sure which is better)

leifwalsh:
LGTM

scheduler/src/cook/mesos/api.clj
wyegelwel:
```clojure
user=> (clojure.string/upper-case nil)

NullPointerException   clojure.string/upper-case (string.clj:209)
````

File: scheduler/src/cook/mesos/api.clj [hunk 4]
```diff
@@ -58,15 +86,76 @@
...
+(defn- mk-container-params
+  "Helper for build-container.  Transforms parameters into the datomic schema."
+  [cid params]
+  (when-not (empty? params)
+    {:docker/parameters (mapv (fn [{:keys [key value]}]
+                                {:docker.param/key key
+                                 :docker.param/value value})
+                              params)}))
+
+(defn- mkvolumes
```
icexelloss: mk-volumns? (Not important)

wyegelwel:
Will make another PR to address cosmetic concerns

PR #136, opened 2016-06-14
Add running and queue endpoints

wyegelwel:
Add endpoints to inspect the running and waiting instances in cook. The waiting jobs are returned sorted.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -366,7 +366,34 @@
                           (::results ctx)))
       ring.middleware.json/wrap-json-params))
 
+(defn waiting-jobs
+  [mesos-pending-jobs-fn]
+  (-> (resource
+        :available-media-types ["application/json"]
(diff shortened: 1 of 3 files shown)


leifwalsh:
LGTM

scheduler/src/cook/mesos/api.clj
```diff
@@ -366,7 +366,34 @@
                           (::results ctx)))
       ring.middleware.json/wrap-json-params))
 
+(defn waiting-jobs
+  [mesos-pending-jobs-fn]
+  (-> (resource
+        :available-media-types ["application/json"]
+        :allowed-methods [:get]
+        :handle-ok (fn [ctx]
+                     (->> (mesos-pending-jobs-fn)
```
icexelloss: why this is different from (def running-jobs ...)?

wyegelwel: This returns the current ordered queue of jobs whereas running just returns all jobs that are running in no particular order.

PR #145, opened 2016-06-22
Fenzo job placement, rebased and with fixes

mforsyth:


scheduler/src/cook/mesos/api.clj
```diff
@@ -243,7 +241,7 @@
                   :priority (or priority util/default-job-priority)
                   :max-retries max_retries
                   :max-runtime (or max_runtime Long/MAX_VALUE)
-                  :ports (or ports [])
+                  :ports (or ports 0)
```
wyegelwel: What happens to clusters that previously used a list of ports? Will upgrading to fenzo cause db issues?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -138,6 +139,23 @@
...
+(histograms/defhistogram [cook-mesos rebalancer pending-job-drus])
+(histograms/defhistogram [cook-mesos rebalancer nearest-task-drus])
+(histograms/defhistogram [cook-mesos rebalancer preemption-counts-for-host])
+(histograms/defhistogram [cook-mesos rebalancer positive-dru-diffs])
+(histograms/defhistogram [cook-mesos rebalancer task-counts-to-preempt])
+(histograms/defhistogram [cook-mesos rebalancer job-counts-to-run])
+
+;; If running on a cluster with very small DRU values,
+;; may need to change this scale to facilitate metrics to e.g. (math/expt 10 305)
+(def dru-scale 1)
```
wyegelwel: Should this be configuration?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -213,20 +234,29 @@
...
+
 (defn compute-preemption-decision
   "Takes state, parameters and a pending job entity, returns a preemption decision
    A preemption decision is a map that describes a possible way to perform preemption on a host. It has a hostname, a seq of tasks
    to preempt and available mem and cpus on the host after the preemption."
   [{:keys [task->scored-task host->spare-resources] :as state}
    {:keys [min-dru-diff safe-dru-threshold] :as params}
    pending-job-ent]
-  (let [{pending-job-mem :mem pending-job-cpus :cpus} (util/job-ent->resources pending-job-ent)
+  (let [timer (timers/start compute-preemption-decision-duration)
```
wyegelwel: Favor `timers/time!` over `start` `stop`

scheduler/src/cook/mesos/api.clj
mforsyth: The original fenzo branch attempted to change the the type of the Datomic attribute "port", which did cause issues.  This branch leaves the original attribute there (though it's no longer used), and introduces a new attribute "ports", which is what would populate this part of the API response.

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: Yes, it really should.  👍

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: 👍

mforsyth:
@wyegelwel I've made adjustments based on your comments.  Wrapped a block in timers/time! rather than using start & stop, and added dru-scale for metrics as a documented configuration option.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -378,16 +391,31 @@
...
-                               (select-keys (util/job-ent->resources (last candidates)) [:mem :cpus]))]
-      (if (and candidates (>= cpus (:cpus required)) (>= mem (:mem required)))
-        (recur remaining-prefixes required candidates)
-        prev))))
+  [^TaskScheduler fenzo considerable offers db fid]
+  (log/debug "Matching" (count offers) "offers to" (count considerable) "jobs with fenzo")
+  (let [t (System/currentTimeMillis)
+        leases (mapv #(->VirtualMachineLeaseAdapter % t) offers)
+        requests (mapv (fn [job]
+                         (let [job-id (:db/id job)]
```
wyegelwel: Where do we use this `job-id`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +445,232 @@
...
+                (meters/mark! matched-tasks-cpus (:cpus match-resource-requirements))
+                (meters/mark! matched-tasks-mem (:mem match-resource-requirements))
+                (doseq [{:keys [tasks leases]} matches
+                        :let [offers (mapv :offer leases)
+                              task-infos (mapv (fn process-results [^TaskAssignmentResult task]
+                                                 (reduce
+                                                   (fn add-ports-to-task-info [task-info [index port]]
+                                                     (log/debug "task-info" task-info [index port])
+                                                     (-> task-info
+                                                         (update-in [:resources :ports]
```
wyegelwel: Are we storing the ports assigned anywhere?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +445,232 @@
...
+                        (getTaskAssigner)
+                        (call (.getRequest task) (get-in (first leases) [:offer :hostname])))))
+                (:matched-head? processed-matches)))))
       (catch Throwable t
-        (decline-offers driver (:ids offer))
         (meters/mark! handle-resource-offer!-errors)
-        (log/error t "Error in match:" (ex-data t))))))
+        (async/go
+          (async/>! offers-chan @offer-stash))
+        (log/error t "Error in match:" (ex-data t)))))))
```
wyegelwel: Put this error message above the go block.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +445,232 @@
...
+         :backfill-jobs backfill-jobs
+         :matched-head? matched-head?}))))
+
+(defn handle-resource-offers!
   "Gets a list of offers from mesos. Decides what to do with them all--they should all
    be accepted or rejected at the end of the function."
-  [conn driver fid pending-jobs offer]
-  (try
+  [conn driver ^TaskScheduler fenzo fid pending-jobs num-considerable offers-chan offers]
+  (doseq [offer offers]
```
wyegelwel: I don't think this is the right place to do this. We will end up double counting in cases where there is an error and the offer stash is piped back in.

Move this expression do into `make-offer-handler`

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +445,232 @@
...
+            (let [offers (async/alt!!
+                           offers-chan ([offers] offers)
+                           timer-chan ([_] [])
+                           :priority true)
+                  matched-head? (handle-resource-offers! conn @driver-atom fenzo @fid-atom pending-jobs-atom num-considerable offers-chan offers)]
+              (when (seq offers)
+                (reset! resources-atom (view-incubating-offers fenzo)))
+              (if matched-head?
+                max-considerable
+                (max 1 (long (* 0.95 num-considerable))))) ;; With max=1000 and 1 iter/sec, this will take 88 seconds to reach 1
```
wyegelwel: What is the distribution of timing for `handle-resource-offers!`? Should we expect it is 250ms or 2 seconds?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +445,232 @@
...
+        (recur
+          (try
+            (let [offers (async/alt!!
+                           offers-chan ([offers] offers)
+                           timer-chan ([_] [])
+                           :priority true)
+                  matched-head? (handle-resource-offers! conn @driver-atom fenzo @fid-atom pending-jobs-atom num-considerable offers-chan offers)]
+              (when (seq offers)
+                (reset! resources-atom (view-incubating-offers fenzo)))
+              (if matched-head?
```
wyegelwel: Can you put a comment describing the reason for this. Essentially describe that the job list is ordered by desirability to schedule and we allow scheduling out of order but eventually want to ensure the most desirable job is scheduled. 

Need to trade off scheduling throughput and scheduling the most important jobs.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Nowhere. It looks like leftover cruft, there was probably a log statement including it at some point.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Doesn't appear so.  Actually, I'm not sure of the original motivation for the change to the ports attribute.  Do you know why this was done @wyegelwel  ?

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj
mforsyth: @wyegelwel  I believe it would vary widely depending on the cluster configuration and workload.   Perhaps we should record a metric about the number of iterations required before one is accepted?

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,238 @@
...
+         :upgrade-backfill upgrade-backfill
+         :backfill-jobs backfill-jobs
+         :matched-head? matched-head?}))))
+
+(defn handle-resource-offers!
+  "Gets a list of offers from mesos. Decides what to do with them all--they should all
+   be accepted or rejected at the end of the function."
+  [conn driver ^TaskScheduler fenzo fid pending-jobs num-considerable offers-chan offers]
+  (log/debug "invoked handle-resource-offers!")
+  (let [offer-stash (atom nil)] ;; This is a way to ensure we never lose offers fenzo assigned if an errors occures in the middle of processing
```
wyegelwel: Can you think of way to avoid making the offer-stash an atom?

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Do you mean switching it from a list to number?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,238 @@
...
+                 offer-size-cpus
+                 (get-in offer [:resources :cpus] 0))
+                (histograms/update!
+                 offer-size-mem
+                 (get-in offer [:resources :mem] 0)))
+              (when (seq offers)
+                (reset! resources-atom (view-incubating-offers fenzo)))
+              ;; This check ensures that, although we value Fenzo's optimizations,
+              ;; we also value Cook's sensibility of fairness when deciding which jobs
+              ;; to schedule.  We refuse to accept a set of matches that doesn't include
```
wyegelwel: I thought we don't refuse a match that doesn't include the highest priority but instead, restrict the set of considerable jobs for the next go around to "encourage" fenzo to pick higher priority jobs.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Yes, that's a good idea, please add a metric for number of iterations until head is matched.

We do have a timer on `handle-resource-offers!` though. I'm wondering how fast (or slow) fenzo is.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,238 @@
...
+    (log/debug "We have" (count pending-offers) "pending offers")
+    pending-offers))
+
+(defn make-offer-handler
+  [conn driver-atom fenzo fid-atom pending-jobs-atom]
+  (let [offers-chan (async/chan (async/buffer 5))
+        resources-atom (atom (view-incubating-offers fenzo))
+        timer-chan (chime-ch (periodic/periodic-seq (time/now) (time/seconds 1))
+                             {:ch (async/chan (async/sliding-buffer 1))})
+        max-considerable 1000]
```
wyegelwel: Let's make this a configuration option.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,238 @@
...
+                (reset! resources-atom (view-incubating-offers fenzo)))
+              ;; This check ensures that, although we value Fenzo's optimizations,
+              ;; we also value Cook's sensibility of fairness when deciding which jobs
+              ;; to schedule.  We refuse to accept a set of matches that doesn't include
+              ;; Cook's highest-priority job.  If Fenzo produces such a set of matches,
+              ;; we try again, giving it less freedom in the form of fewer jobs to consider.
+              ;; This is repeated until Fenzo does match our highest-priority job.
+              (if matched-head?
+                max-considerable
+                (max 1 (long (* 0.95 num-considerable))))) ;; With max=1000 and 1 iter/sec, this will take 88 seconds to reach 1
```
wyegelwel: Let's make the scalar (0.95) a configuration option as well.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Yes, the change in cardinality.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Yes, you're right.   (handle-resource-offers!) happens whether or not matched-head? is true. 👍

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj

scheduler/docs/rebalancer-config.asc
```diff
@@ -0,0 +1,52 @@
...
+
+[source,clojure]
+----
+(d/pull db ["*"] :rebalancer/config)
+----
+
+
+=== Significance of the parameters
+
+* safe-dru-threshold: Task with a DRU lower than safe-dru-threshold will not be preempted. If each DRU divisor is set to the corresponding per user share and safe-dru-threshold is set to 1.0, then tasks that consume resources in aggregate less than the user resource share will not be preempted.
```
wyegelwel: Cook currently sets the DRU divisor to the users share, correct? Or is this something different?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -266,31 +298,46 @@
     [(next-state state pending-job preemption-decision) preemption-decision]
     [state nil]))
 
+(defn taskless?
```
wyegelwel: I especially don't love my suggested name in the context it is used. 

A bit wordy but, `has-running-or-unknown-tasks?`

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+(defn process-matches-for-backfill
+  "This computes some sets:
+   
+   fully-processed: this is a set of job uuids that should be removed from the list scheduler jobs. These are not backfilled
+   upgrade-backfill: this is a set of instance datomic ids that should be upgraded to non-backfill
+   backfill-jobs: this is a set of job uuids for jobs that were matched, but are in the backfill QoS class
+   "
+  [scheduler-contents head-of-considerable matched-jobs]
+  (let [matched-job-uuids (set (mapv :job/uuid matched-jobs))
+        head-uuid (:job/uuid head-of-considerable)]
```
wyegelwel: This is partially a note to myself while I reviewing this code, however you might find it interesting.

I'm concerned that if head-of-considerable is stuck in state "unknown", all other jobs will be considered backfilled.

I'm also concerned how this logic works when we reshuffle the pending job list after rank jobs.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+    (loop [scheduler-contents scheduler-contents
+           backfilling? false
+           matched-head? false
+           fully-processed #{}
+           upgrade-backfill #{}
+           backfill-jobs #{}]
+      (if (seq scheduler-contents)
```
wyegelwel: I'm not thrilled with going through all of scheduler contents.

wyegelwel: I think we can simplify this logic a lot by finding the job in scheduler contents that begins the backfilling and keeping track of which jobs came before that (either fully processed or upgraded). 

All jobs other jobs that were matched are backfilled. That would make the backfill condition more clear as well (I looked at this code for a while before I figured out the condition.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: It appears my first concern is a problem since jobs scheduled stay in state "waiting" until they are confirmed running. 

Essentially, what I'm imagining happening is: 

A bunch of jobs are scheduled, one of the jobs, J, will not get out of state unknown. All of the jobs ahead of it in the queue will be removed from pending on the next rank jobs, however, J will remain since it is still "waiting". J is now at the head of the queue but since it is not running, it will cause all jobs after it to be considered backfilled. 

Do you agree that chain of events is possible?

wyegelwel: I'm pretty sure my second concern is a problem as well since jobs that are backfilled will not show up in scheduler contents once they are set to running. This means those jobs will never get a chance to be upgraded.

wyegelwel: I've also made an assumption that all jobs in scheduler contents are either:

- not scheduled
- backfilled
- scheduled but the instance is still in state unknown

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+                _ (reset! offer-stash (doall (for [{:keys [leases]} matches
+                                                   lease leases]
+                                               (:offer lease))))
+                matched-jobs (for [match matches
+                                   ^TaskAssignmentResult task-result (:tasks match)
+                                   :let [task-request (.getRequest task-result)]]
+                               (:job task-request))
+                processed-matches (process-matches-for-backfill scheduler-contents (first considerable) matched-jobs)
+                new-scheduler-contents (remove (fn [pending-job]
+                                                 (contains? (:fully-processed processed-matches) (:job/uuid pending-job)))
```
wyegelwel: I think we should remove upgraded jobs as well. I don't think it matters but you will end up with a bunch of upgraded jobs at the front of scheduler contents.

wyegelwel: Can you add a comment explaining why we don't remove backfilled jobs. Something along the lines of "we want to have the opportunity to upgrade backfilled jobs out of backfill status if jobs ahead of them are scheduled"

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Can we use the list form to store the ports that were assigned? 

Or better yet, can we have two fields in datomic:

1. num-ports-requested on job
2. ports-assigned on instance


That way we will have full information

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+                  (doseq [^TaskAssignmentResult task tasks]
+                    (.. fenzo
+                        (getTaskAssigner)
+                        (call (.getRequest task) (get-in (first leases) [:offer :hostname])))))
+                (:matched-head? processed-matches)))))
       (catch Throwable t
-        (decline-offers driver (:ids offer))
         (meters/mark! handle-resource-offer!-errors)
-        (log/error t "Error in match:" (ex-data t))))))
+        (log/error t "Error in match:" (ex-data t))
```
wyegelwel: oh crap. I know why the order was funny; `log/error` returns nil so `matched-head?` would be false-y, however `async/go` returns a chan so `matched-head?` will be truth-y. Can you explicitly return `false` so that the reader will think about what is going on (and re-ordering things won't break everything (= )

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+      (loop [num-considerable max-considerable]
+        ;;TODO make this cancelable (if we want to be able to restart the server w/o restarting the JVM)
+        (recur
+          (try
+            (let [offers (async/alt!!
+                           offers-chan ([offers] offers)
+                           timer-chan ([_] [])
+                           :priority true)
+                  matched-head? (handle-resource-offers! conn @driver-atom fenzo @fid-atom pending-jobs-atom num-considerable offers-chan offers)]
+              (doseq [offer offers]
```
wyegelwel: I realized we will still get double counting in there is a failure since we put offers in the stash back on this channel. Can we move this expression to where we first accept the offers from mesos?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -806,176 +951,27 @@
...
-        (log/error e "Error in start-offer-matcher!")))))
-
-(meters/defmeter [cook-mesos scheduler offer-back-pressure])
 (meters/defmeter [cook-mesos scheduler mesos-error])
 
+(defn make-fenzo-scheduler
+  [driver offer-incubate-time-ms]
+  (.. (com.netflix.fenzo.TaskScheduler$Builder.)
+      (disableShortfallEvaluation) ;; We're not using the autoscaling features
+      (withLeaseOfferExpirySecs (max (quot offer-incubate-time-ms 1000) 1)) ;; Must convert from the ms, should be at least 1 second
```
wyegelwel: Use `(-> offer-incubate-time-ms t/millis t/in-seconds)` instead of `(quot offer-incubate-time-ms 1000)`

scheduler/src/cook/mesos/schema.clj
```diff
@@ -427,20 +441,20 @@
              - task succeeded => job completed
              - task failed, no other tasks, retries exceeded => job completed
              - task failed, no other tasks, retries remaining => job waiting
-             - task failed, other tasks running => job running"
+             - task failed, other tasks running => job running
+
+             Note that backfilled running instances are treated as if they don't exist."
```
wyegelwel: A backfilled running instance will not allow the job to be moved from waiting -> running correct? 

So one of my concerns above is not a problem after all. However, now I'm concerned that this is quite complicated. Let me think about how to simplify it.

wyegelwel: Also, can you update the comment to explicitly say that backfilled running instances will not move a job from waiting to running if that is the case.

scheduler/src/cook/mesos/util.clj
```diff
@@ -150,15 +150,21 @@
 (def ^:const default-job-priority 50)
 
 (defn same-user-task-comparator
-  "Comparator to order same user's tasks"
-  [task1 task2]
-  (letfn [(task->feature-vector [task]
+  "Comparator to order same user's tasks
+
+   If consider-backfilled-jobs? is true, then we treat jobs which were backfilled as being the lowest priority"
+  [consider-backfilled-jobs?]
```
wyegelwel: I would prefer to have two functions that encapsulate `same-user-task-comparator-penalize-backfill` and `same-user-task-comparator` instead of a boolean because then it forces the reader to figure out what the boolean means.

wyegelwel: Hold on, I think I'm confused. Doesn't this make backfilled jobs the highest priority jobs instead of the lowest?

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: From looking over update-job-status, I see that my previous comment is not an issue: 

> I'm pretty sure my second concern is a problem as well since jobs that are backfilled will not show up in scheduler contents once they are set to running. This means those jobs will never get a chance to be upgraded.

However, I'm concerned about the number of unique job/instance states that exist now.

Additionally, I'm still concerned about reshuffling after a rank jobs. Consider the following case:

A job, J, is scheduled as a backfilled job and it is the second job in scheduler contents, just behind F.
Rank jobs occurs and now J is further back.
F gets scheduled


Should J still be considered backfilled? What happens to the other jobs for that user? If you change priority of jobs, what happens?

wyegelwel: For my own notes, here are all valid [job, job-modifier, instance, instance-modifier] state tuples that exist:

[waiting, none, none, none]
[waiting, backfill: false, unknown, none]
[running, backfill:false, running, none]
[completed, backfill: false, failed, preempted: true]
[completed, backfill: false, failed, preempted: false]
[completed, backfill: false, success, none]

[waiting, backfill: true, unknown, none]
[waiting, backfill: true, running, none]
[completed, backfill: true, failed, preempted: true]
[completed, backfill: true, failed, preempted: false]
[completed, backfill: true, success, none]

Unintuitive states:

[waiting, backfill: false, unknown, none]
[waiting, backfill: true, unknown, none]
[waiting, backfill: true, running, none]


I didn't like the state [waiting, unknown] before the fenzo change and was hoping to remove it. Can we think of a way to remove the need for {job/state waiting, instance/status running, job/backfilled? true} to help bug proof us.

We currently use {job/state waiting, instance/status running, job/backfilled? true} to help inform us when the job has been upgraded.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+                (log/debug "Matched tasks" task-txns)
+                ;; This launch-tasks MUST happen after the above transaction in
+                ;; order to allow a transaction failure (due to failed preconditions)
+                ;; to block the launch
+                (meters/mark! scheduler-offer-matched
+                              (->> matches
+                                   (mapcat (comp :id :offer :leases))
+                                   (distinct)
+                                   (count)))
+                (meters/mark! matched-tasks (count task-txns))
```
wyegelwel: Can you add a few more metrics:

meter for number of backfilled jobs
meter for number of upgraded jobs
meter for number of fully processed jobs

meter for number of cpus of backfilled jobs
meter for number of cpus of upgraded jobs
meter for number of cpus of fully processed jobs

meter for number of mem of backfilled jobs
meter for number of mem of upgraded jobs
meter for number of mem of fully processed jobs

histogram for number of backfilled jobs
histogram for number of upgraded jobs
histogram for number of fully processed jobs

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Yes, seems like we should be able to do that.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: The issue (and reason I believe an atom was used) is that a Try Block introduces its own lexical scope.  If we want the exception handling of the try block to reset the offer stash with a value that is set inside the try block, an atom is the simplest way to do that, since the atom (unlike vars) can acquire a value that is available outside of the scope where it is set.

I believe We could avoid using an atom, but we would have to rearrange this whole function significantly, and partially change the way exceptions are handled in order to do so.  We could have the parts that precede (reset! offer-stash) in one try block (maybe even a separate function), that returns a tuple of  "considerable" and "matches" ...  then reset the offer stash, then start a new try block that does the rest of what the existing function does (and handles exceptions in the same way).

scheduler/docs/rebalancer-config.asc
mforsyth: Correct - it's the user's share:
user->dru-divisors (->> all-users
                                (map (fn [user]
                                       [user (share/get-share db user)]))
                                (into {}))

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: One thing we could do is allow backfill jobs to move to state running, and then in `sort-jobs-by-dru`, keep backfilled jobs in the list.

I'm not sure that is the best strategy, but it would work. Let me know if you think of something else.

scheduler/docs/rebalancer-config.asc
```diff
@@ -0,0 +1,52 @@
...
+
+[source, clojure]
+----
+(let [conn (d/connect "(your-cook-database-url)")
+        db (d/db conn)]
+    @(d/transact conn [{:db/id :rebalancer/config
+                        :rebalancer.config/min-utilization-threshold 0.0
+                        :rebalancer.config/safe-dru-threshold 0.0
+                        :rebalancer.config/min-dru-diff 0.0000000001
+                        :rebalancer.config/max-preemption 64.0}]))
```
mforsyth: Actually, I never set any shares whatsoever!
So, my divisors were universally Double.MAX_VALUE.

I can add documentation about this.

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: "not-running?"

Or I could make it "running-now?"  and just filter by the negation of it where it's used?

scheduler/src/cook/mesos/util.clj
mforsyth: @wyegelwel in the vector to be compared, the first element is 1 iff backfilled? matters and the job is backfilled, 0 otherwise.   0 sorts before 1 so non-backfilled jobs will be sorted to the beginning of the list...  

To respond vaguely and in general to a lot of what you wrote, yes, this is complicated :)

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/util.clj
mforsyth: Yeah, could make this 2 separate functions.  I believe the one that cares about backfill could just use the result of the 2nd function as a "tiebreaker" comparator between jobs that share a backfill value

scheduler/src/cook/mesos/schema.clj
mforsyth: Correct, and can add to the comment 👍

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Either is good

scheduler/docs/rebalancer-config.asc
wyegelwel: Then I think you can simplify that sentence to be:

"If safe-dru-threshold is set to 1.0, then tasks that consume resources in aggregate less than the user resource share will not be preempted"

scheduler/src/cook/mesos/scheduler.clj
mforsyth: It seems like what we are dealing with is that some jobs don't fit cleanly into either of the two states "waiting" or "running".  We have code that cares about only SOME waiting jobs, and are thinking about changing that to care about only SOME running jobs.  Another alternative would be to introduce another state, a formal transitional state, e.g. "requested", or "in-flight"?

mforsyth: I agree with your definition of "backfilled" in process-matches-for-backfill

mforsyth: In this chain of comments you dismissed one of the big problems you feared, but that leaves this:

> A bunch of jobs are scheduled, one of the jobs, J, will not get out of state unknown. All of the jobs ahead of it in the queue will be removed from pending on the next rank jobs, however, J will remain since it is still "waiting". J is now at the head of the queue but since it is not running, it will cause all jobs after it to be considered backfilled.

Well... if this job is left in state "waiting" (let's say it stays there for a long time because Mesos doesn't receive the message, or something), wouldn't it possibly get matched again on a subsequent pass to Fenzo?  If it gets matched, it wouldn't cause all jobs behind it to be backfilled.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +444,237 @@
...
+       (filter #(and (= :instance.status/running (:instance/status %))
+                     (:instance/backfilled? %)))
+       (map :db/id)))
+
+(defn process-matches-for-backfill
+  "This computes some sets:
+   
+   fully-processed: this is a set of job uuids that should be removed from the list scheduler jobs. These are not backfilled
+   upgrade-backfill: this is a set of instance datomic ids that should be upgraded to non-backfill
+   backfill-jobs: this is a set of job uuids for jobs that were matched, but are in the backfill QoS class
```
mforsyth: It seems to be only used for that purpose.   There isn't any documentation external to the code about it.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: The difference in return value is used to determine whether to lower the number of jobs Fenzo can look at, right?  It's not clear to me why we want to make sure that if handle-resource-offer! errors out, we want to "penalize" Fenzo.  The previous version of this code did penalize Fenzo for this by returning something falsey though.   Do you know more about the reasoning here?

wyegelwel: Hmm, that is a good point. I think running true is valid. Just add a comment explaining why.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: > Well... if this job is left in state "waiting" (let's say it stays there for a long time because Mesos doesn't receive the message, or something), wouldn't it possibly get matched again on a subsequent pass to Fenzo? If it gets matched, it wouldn't cause all jobs behind it to be backfilled.

the function `job-allowed-to-start` would prevent that as the task is state unknown. It wouldn't get resolved until reconcile tasks ran and got the new state from mesos.

wyegelwel: > It seems like what we are dealing with is that some jobs don't fit cleanly into either of the two states "waiting" or "running". We have code that cares about only SOME waiting jobs, and are thinking about changing that to care about only SOME running jobs. Another alternative would be to introduce another state, a formal transitional state, e.g. "scheduled", or "in-flight"?

I agree with your assessment. The way I have been thinking about it is:

I think the "in-flight" state is instance state unknown. Most of the code doesn't care about the distinction between "in-flight" and actually running and the code that does can make the distinction. Similarly, most of the code doesn't care about the distinction between backfilled running and not backfilled running; however, backfilled running == waiting would break some code. Essentially, I'm trying to minimize the number of states and modifiers and put jobs and instances in the state such that if you don't care look at the modifier, it is still reasonable.

mforsyth: Ah, I didn't absorb the allowed-to-start? from schema.clj.    As an aside, job-allowed-to-start? is doing the same thing as my soon-to-be-renamed "taskless?" in scheduler.rb;  perhaps I should just   move it to util and reuse it.   

Using exceptions in a loop like this is generally frowned upon, isn't it?    Is it worth being able to reuse the code allowed-to-start?  that runs in the db?  Maybe all 3 places can use a more primitive function that lives in util?

mforsyth: So, your problem scenario seems like it would really happen.   If a desirable job gets "stuck" waiting, all other jobs will be backfilled until that is resolved.

mforsyth: Thinking out loud:

How big of a problem is this?   It means that many jobs that would otherwise be run normally will instead get marked as backfilled.   Those jobs will be the first to be preempted, if preemption happens.   From that point of view, the more stuff that is marked as backfilled, the less significant the marking is.

If it IS a big problem, it means that we would need to introduce a better mechanism for choosing which of the jobs to be run are put in the backfilled category.

If it ISN'T a big problem that a bunch of jobs might get "unjustifiably" backfilled, this seems a point in favor  of your idea that it's better to classify backfilled jobs as "running" rather than "waiting".  It SHOULD be a big deal that a job is "waiting" rather than "running", shouldn't it?

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: > Using exceptions in a loop like this is generally frowned upon, isn't it? Is it worth being able to reuse the code allowed-to-start? that runs in the db? Maybe all 3 places can use a more primitive function that lives in util?

I agree with you but it is nice to be running the same check the db runs. For now lets leave it, if it becomes a performance problem I will change it.

> How big of a problem is this? It means that many jobs that would otherwise be run normally will instead get marked as backfilled. Those jobs will be the first to be preempted, if preemption happens. From that point of view, the more stuff that is marked as backfilled, the less significant the marking is.

I think we can fix it, consider making `ids-of-backfilled-jobs` include state unknown instances.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: @wyegelwel Looking at the external documentation we already have, it seems like the best place to describe the concept of backfilling and its interaction with job & instance states would be the API docs.  Would you agree?   (Backfilled attribute on instances is already being returned by the API).

scheduler/src/cook/mesos/schema.clj
```diff
@@ -443,7 +443,9 @@
              - task failed, no other tasks, retries remaining => job waiting
              - task failed, other tasks running => job running
 
-             Note that backfilled running instances are treated as if they don't exist."
+             Note that backfilled running instances are treated as if they don't exist.
+             In other words, if a job has running tasks that are all backfilled, the
```
wyegelwel: I thought we were going to change this behavior to reduce complexity.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -962,7 +961,7 @@
       (build)))
 
 (defn create-datomic-scheduler
-  [conn set-framework-id driver-atom pending-jobs-atom heartbeat-ch offer-incubate-time-ms task-constraints]
+  [conn set-framework-id driver-atom pending-jobs-atom heartbeat-ch offer-incubate-time-ms o-max-jobs-considered fenzo-scaleback task-constraints]
```
wyegelwel: is there an "o-" in front of `max-jobs-considered` or is the diff tool just messing about?

mforsyth: oops... I renamed this parameter but this one got messed up.  pushed fixup

scheduler/src/cook/mesos/scheduler.clj
mforsyth: So about these:  I've started on this, but realized that in order to record these metrics, we're probably best off redoing the way that  process-matches-for-backfill works.   

For example, "upgraded" is a list of task id's... to get total resources requested (in terms of jobs) from that one we'd have to go back to Datomic to match the task id's to jobs, and then sum the attributes of the jobs.  If we're going to do that it would be more efficient to build up a map of jobs->tasks in the main loop in process-matches-for-backfill.

And for fully-processed and backfilled... these at least are job id's, but again, if we're just going to have to go back to datomic to get more attributes of the jobs in order to do resource totals, it probably makes more sense for the main loop  in process-matches-for-backfill to be accumulating whole jobs rather than just the ids.

Can definitely be done but I think this is worth me checking in with you to ask:  Still worth it?

scheduler/src/cook/mesos/schema.clj
mforsyth: @wyegelwel   I don't think we decided that?   Here's the convo about that:  https://github.com/twosigma/Cook/pull/145#discussion-diff-69572273  We speculated about different things that could be done to simplify the states, but no definitive conclusion.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Well, we wouldn't necessarily have to go all the way back to datomic (could just rebuild the info based on the scheduler-contents), but the main point is, just adding these metrics into the function without reworking it would be rather ugly & complex.  Would be better served by reworking the function.

mforsyth: Actually, nevermind.  I am realizing that you had wanted this function to be significantly rearranged anyway in order to make the logic more readable https://github.com/twosigma/Cook/pull/145#discussion-diff-69572273.  I'll kill 2 birds with one stone and get these metrics.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -436,112 +459,254 @@
 (gauges/defgauge [cook-mesos scheduler front-of-job-queue-mem] (fn [] @front-of-job-queue-mem-atom))
 (gauges/defgauge [cook-mesos scheduler front-of-job-queue-cpus] (fn [] @front-of-job-queue-cpus-atom))
 
-(defn handle-resource-offer!
+(defn ids-of-backfilled-instances
+  "Returns a list of the ids of running-or-unknown, backfilled instances of the given job"
```
wyegelwel: Nitpick: should be "running or unknown"

scheduler/src/cook/mesos/util.clj
```diff
@@ -98,8 +98,8 @@
          [job-ent & job-ents] job-ents]
     (if job-ent
       (let [{:keys [cpus mem]} (job-ent->resources job-ent)]
-        (recur (+ total-cpus cpus)
-               (+ total-mem mem)
+        (recur (+ total-cpus (or cpus 0))
```
wyegelwel: Does cook allow you to submit incomplete jobs?

mforsyth: No, but incomplete jobs are used in tests.   At a practical level, we either have to change the tests to pick arbitrary resource amounts, or change this function so that it doesn't throw a NullPointerException if any jobs don't have explicit cpu or mem requirements.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -713,11 +888,12 @@
...
                      (group-by util/task-ent->user)
                      (map (fn [[user task-ents]]
-                            [user (into (sorted-set-by util/same-user-task-comparator) task-ents)]))
+                            [user (into (sorted-set-by (util/same-user-task-comparator)) task-ents)]))
                      (into {})
                      (dru/sorted-task-scored-task-pairs <> user->dru-divisors)
                      (filter (fn [[task scored-task]]
-                               (contains? pending-task-ents task)))
+                               (or (contains? pending-task-ents task)
+                                   (:instance/backfilled? task))))
```
wyegelwel: Can you put a comment why we include backfilled here? I want to make sure no one removed this by accident.

dgrnbrg:
Would you mind listing me as the author of the original commit?

You can use these instructions to change the author: http://stackoverflow.com/a/28845565

If you choose to do so, you could list me as `David Greenberg <dgrnbrg.consulting@gmail.com>`

wyegelwel:
I think this PR is ready to go. When you get a chance please squash commits and I will merge.

mforsyth:
@dgrnbrg No problem 👍

dgrnbrg:
Congrats!

Issue #111, opened 2016-04-02
Design: Support for Dynamic Executor Allocation in Spark Scheduler


tnn1t1s:
The retry logic in the Coarse Grained Scheduler should support Dynamic Executor Allocation in Spark. That is, if spark.dynamicAllocation.enabled, then Cook will scale number of executors according to:
- spark.dynamicAllocation.initialExecutors
- spark.dynamicAllocation.minExecutors
- spark.dynamicAllocation.maxExecutors

Reference:
https://spark.apache.org/docs/1.6.0/job-scheduling.html#dynamic-resource-allocation
http://jerryshao.me/architecture/2015/08/22/spark-dynamic-allocation-investigation/

Questions: 
How does this interact with 'spark.executor.failures' and the failure count logic in CoarseCookSchedulerBackend?

PR #146, opened 2016-06-22
Add new spark patch for v1.6.1 with dyn-allocation support

solnic:
refs twosigma/cook#111


leifwalsh:
LGTM but this should remove the old patch probably

solnic:
Ah ok I just thought that maybe for some people who still run older spark this might be useful

leifwalsh:
This is why god gave us version control :wink:

wyegelwel:
solnic, is this ok to merge or is there more commits coming?

solnic:
@wyegelwel I'd love to hear from you re issues I described in basecamp before merging this in

spark/0001-Add-cook-support-for-spark-v1.6.1.patch
```diff
@@ -0,0 +1,769 @@
...
++  // want to get into thread management issues)
++  override def start() {
++    super.start()
++    logInfo("Starting Cook Spark Scheduler")
++    requestRemainingCores()
++  }
++
++  override def stop() {
++    super.stop()
++    abortJobs(runningJobUUIDs.toList)
```
solnic: We *probably* should not be aborting jobs here, as it results in killing all jobs that the cook scheduler thinks are still running. I've tested it w/o this line and the behavior seems to be correct.

wyegelwel: Where correct means the jobs complete successfully

wyegelwel: Spoke with Leif. I think it is ok to just remove this abort. Once that is done, I think this PR is good to go.

solnic:
@wyegelwel I updated the patch so that it longer includes the aborting of jobs

wyegelwel:
LGTM, @leifwalsh @WenboZhao unless you have any other comments I think this is good to merge.

leifwalsh:
:100:

PR #149, opened 2016-06-27
Add instance decorator to cook jobclient

WenboZhao:


jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -406,10 +405,14 @@ private Job(UUID uuid, String name, String command, Double memory, Double cpus,
         _instances = ImmutableList.copyOf(instances);
         _env = ImmutableMap.copyOf(env);
         _uris = ImmutableList.copyOf(uris);
-        /* This stringifies the JSON object and then reparses it.
-           which is inefficient but should be fine for testing. */
+        // This take the string representation of the JSON object and then parses it again which is inefficient but
```
wyegelwel: Can you reason about why this is necessary?

jobclient/src/main/java/com/twosigma/cook/jobclient/InstanceDecorator.java
```diff
@@ -0,0 +1,35 @@
...
+ * Created: June 23, 2016
+ * @author wzhao
+ */
+public interface InstanceDecorator {
+
+    /**
+     * @param builder The {@link Builder} expected to decorate.
+     * @return a decorated instance builder.
+     */
+    Builder decorate(Builder builder);
```
wyegelwel: Is there a way to import such that the type is `Instance.Builder` instead of `Builder`?

WenboZhao: Yes, will change that.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: Not sure, rusty added that and I just reformat the comment style.

wyegelwel:
lgtm, just squash the commits

jobclient/src/main/java/com/twosigma/cook/jobclient/InstanceDecorator.java
WenboZhao: Sigh~ Rusty pointed me out that we need to make Job immutable.

PR #158, opened 2016-07-12
Add "setup from scratch" doc

pjlegato:


wyegelwel:
It looks like this work is included in #159. Should we treat them as separate or close this one?

pjlegato:
@wyegelwel Yes, #160 incorporates this PR. I'll close this one.

PR #159, opened 2016-07-15
Lock down /rawscheduler endpoint

pjlegato:
This PR provides the functionality necessary to lock down the /rawscheduler endpoint, as detailled in my previous emails.

Ordinary users can only view their own jobs. Admins can view any job.

Admins are specified as a set of usernames in the config file.

Various modifications were also made to allow for easier interactive development at the REPL so that web handlers can be modified without having to restart the whole system from scratch, and more log messages were added to make it easier to inspect internal system state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ pom.xml.asc
 .lein-plugins
 .lein-repl-history
 .nrepl-port
+.minimesos/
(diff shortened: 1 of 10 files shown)


pjlegato:
I'm closing this PR pending the rework we discussed.

PR #160, opened 2016-07-20
Config file authorization system

pjlegato:
Allows user to specify an authorization function with a standard signature. Two example auth functions are provided. The new auth system is hooked up to the GET /rawscheduler endpoint.


scheduler/src/cook/mesos/api.clj
```diff
@@ -231,9 +241,25 @@
...
 
+
+(defn job-owner-by-uuid
+  "Returns the username that owns the given UUID, or nil if no job with
+  that UUID exists."
+  [db uuid]
+  (let [uuid (if (string? uuid)
+               (UUID/fromString uuid)
+               uuid)]
+    (some-> (d/q '[:find ?owner
```
wyegelwel: You can also do `(:job/user (d/entity db [:job/uuid uuid]))`

scheduler/src/cook/mesos/api.clj
```diff
@@ -419,18 +449,43 @@
...
+                            ;; whether it's allowed, and if not, why
+                            ;; it's not allowed.
+                            uuid-filter (fn [uuid]
+                                          (let [job (fetch-job-map (db conn)
+                                                                   fid
+                                                                   uuid)]
+                                            (if (or (nil? job) 
+                                                    (not (is-authorized? user :access job)))
+                                              {:is-allowed false
+                                               :message (str "No job found with UUID " uuid)
```
wyegelwel: Why not have a message that says a user is not authorized to access that job?

scheduler/src/cook/mesos/api.clj
```diff
@@ -385,7 +415,7 @@
 ;;;
 ;;; On GET; use repeated job argument
 (defn job-resource
-  [conn fid task-constraints]
+  [conn fid task-constraints admins]
```
wyegelwel: Unused reference to admins

scheduler/src/cook/mesos/api.clj
```diff
@@ -419,18 +449,43 @@
...
+                                               :message (str "No job found with UUID " uuid)
+                                               :uuid uuid}
+                                              {:is-allowed true
+                                               :uuid uuid})))
+
+                            uuids (map uuid-filter uuids)]
+                        ;; Return true if every UUID is in use, or
+                        ;; else return false with a list of
+                        ;; nonexistant UUIDs.
+                        (if (every? true? (map :is-allowed uuids))
```
wyegelwel: You can do `(every? :is-allowed uuids)`

scheduler/src/cook/mesos/api.clj
```diff
@@ -474,17 +537,25 @@
   (-> (resource
         :available-media-types ["application/json"]
         :allowed-methods [:get]
+        :allowed? (fn [ctx]
+                    (let [user  (get-in ctx [:request :authorization/user])]
+                      ;; Only allow superusers:
+                      (if (is-authorized? user :access cook.authorization/system)
+                        true
+                        (do
+                          (log/info "[waiting-jobs] Queried by non-admin" user "; denying access.")
```
wyegelwel: s/waiting-jobs/running-jobs/

scheduler/src/cook/mesos/api.clj
```diff
@@ -474,17 +537,25 @@
...
+                          [false {::error "Unauthorized"}]))))
         :handle-ok (fn [ctx]
                      (->> (util/get-running-task-ents (db conn))
                           (map d/touch)
                           cheshire/generate-string)))
       ring.middleware.json/wrap-json-params))
 
 (defn handler
-  [conn fid task-constraints mesos-pending-jobs-fn]
+  [conn fid task-constraints mesos-pending-jobs-fn admins]
```
wyegelwel: Unused reference to admins

scheduler/src/cook/mesos/api.clj
```diff
@@ -474,17 +537,25 @@
...
   (routes
     (ANY "/rawscheduler" []
-         (job-resource conn fid task-constraints))
+         (job-resource conn fid task-constraints admins))
```
wyegelwel: Unused reference to admins

scheduler/src/cook/authorization.clj
```diff
@@ -0,0 +1,180 @@
...
+(defn is-authorized?
+  "Determines whether the given user can perform the given operation
+  on the given object, according to the authorization function
+  specified in the config file.
+
+  Returns true if allowed, else false."
+  ([^String user
+     ^clojure.lang.Keyword verb
+     ^cook.authorization.Ownable object] 
+     (is-authorized? (-> @global-state :settings) user verb object))
```
wyegelwel: I prefer to create a partial function in which we explicitly pass in the config instead of using global state. Is there a compelling reason to do it this way?

An alternative to creating a partial function in components would be to have a `setup-is-authorized?` which would take the settings and return a `is-authorized?` function with is properly setup.

scheduler/src/cook/authorization.clj
```diff
@@ -0,0 +1,180 @@
...
+
+;; The defonce is a kludge to work around
+;; https://groups.google.com/forum/#!topic/clojure/SYYYwZIrFiY -
+;; reloading a protocol causes existing objects of that type to become
+;; invalid as instances of the new protocol.
+;;
+;; If you redefine this protocol at runtime, you'll have to manually
+;; remove the defonce. Other than that, this should not cause issues.
+(defonce __ownable
+  (defprotocol Ownable
```
wyegelwel: I can imagine this can be limiting if we want to have users be able to access some object types but not others or we want to have more than one class of system objects. Perhaps we could have Ownable have both an owner and object type to allow for more sophisticated auth types. What do you think? Too far in the land of overkill?

scheduler/dev-config.edn
```diff
@@ -1,6 +1,17 @@
 {:port 12321
  ;; We'll set the user to vagrant, since that's the default for many Vagrant-based Mesos setups
  :authorization {:one-user "vagrant"}
+
+ ;; What function should be used to perform user authorization?
+ ;; See the docstring in cook.authorization for details.
+ :authorization-fn cook.authorization/configfile-admins-auth
```
wyegelwel: Consider having `:authorization-config` or some better named key which is a map. That map has a `:fn` key and we pass the map into our authorization-fn instead of passing the full config file.

SETUP.md
```diff
@@ -0,0 +1,184 @@
+# Setting up your Cook dev environment
+
+This document tells you how to set up a Cook dev environment from
+scratch on a Macintosh computer.
```
wyegelwel: I really like these docs I think they are very helpful. Would it be possible to separate out the mac specific instructions and notes from the general notes?

scheduler/src/cook/mesos/api.clj
pjlegato: Such a message would leak information to the end user about whether a given UUID exists at all in the system. This might then be used as input to other unrelated processes that are attempting to hack some element of the system but which will only work when given an extant UUID.

This scheme, based on the principle of least privilege, reveals information to the user only on a need-to-know basis. They can't find any job with the UUID they provided, which is all the unauthorized user needs to know about the system state.

This way, a potentially hostile user can't differentiate between the cases where a) the given UUID exists and they are simply not authorized to use it, or b) the UUID just doesn't exist in the system at all. They aren't given a tool for narrowing the "valid job ID" search space by being able to test arbitrary UUIDs for validity.

scheduler/src/cook/mesos/api.clj
pjlegato: Ah, yes, that's cruft from an older draft of that code. I'll take that out.

scheduler/src/cook/mesos/api.clj
pjlegato: Yeah, that is better / more concise. I'll update it.

scheduler/src/cook/mesos/api.clj
pjlegato: Doh. Fixed.

scheduler/src/cook/mesos/api.clj
pjlegato: Fixed.

scheduler/src/cook/mesos/api.clj
pjlegato: Fixed.

scheduler/src/cook/authorization.clj
pjlegato: This actually is (kind of) that; the is-authorized function is overloaded, with 3- and 4-arity versions provided. The 4-arity one is a pure function that accepts its config as its first argument. The 3-arity version is purely for convenience; all it does is call the 4-arity version with the standard global state and its own arguments.

Code that doesn't care about special state can just use the 3-arity version and have the standard state assigned automatically. Code that wants custom state can use the 4-arity version. The tests use the 4-arity version to specify different config options, for example.

I find this style is a bit clearer to understand and manage than using setup functions and closures, since you don't have to keep track of setting up and updating an extra symbol in the namespace, but whatever you want to go with here is fine.

scheduler/src/cook/authorization.clj
pjlegato: I think that's not a bad idea in the long run, but overkill for this version, since the spec was just to make users allowed to do anything to all of their own owned objects.

In the future, we can easily extend the system to accomodate object types and classes by extending the protocol, if that becomes a requirement.

scheduler/dev-config.edn
pjlegato: OTOH, we don't know a priori what a user's custom authorization-fn does; it might perform any arbitrary computation. Might it need some information from the general config file to do that? Do we want to specify that it's only allowed to use data from that particular key to do authorization?

SETUP.md
pjlegato: Thanks! 

The notes are really just telling you to install a bunch of software on your system.  The procedure would be very similar on Linux, just with "apt-get" or "yum" instead of "port".  I'll add a note to that effect.

scheduler/dev-config.edn
wyegelwel: Yea, I think that is a reasonable restriction with the benefit being that we have clear modularity.

scheduler/src/cook/authorization.clj
wyegelwel: I see what you are saying, I'm just not a fan of the global state to begin with. Aside from this, is it necessary for anything else?

pjlegato: I'm not a fan of global state either, but it's already deeply embedded in the app's design. The app uses a bunch of Java libraries to interact with Mesos and the web, which are very much stateful and in the object-oriented paradigm. The global state (among other things) keeps track of those Java objects. 

Short of redesigning the entire app, there is not much that can be done now except fence the global state off into a designated "stateful" containment area so we can keep it from expanding and at least know what's in it. Over time, we can work on whittling that down -- I'm sure a lot of it can be removed and converted into non-global state and/or pure functions. For now (this version), I'm aiming at just organizing and fencing off what's already there, since a rewrite of all that was not in the remit.

Speaking more generally, all long-lived daemon apps have to have *some* kind of global state, if only a network socket or a file handle (or else they wouldn't be able to communicate with the outside world at all.) I think the best we can hope for is to isolate and manage that. 99% of the code can be purely functional and take state as an argument, but trying to carry that concept through to its logical conclusion leads to a different problem: all functions grow extra arguments for passing around database handles or things like that, which then have to be threaded through the entire call stack every single time. Adding a new item then means reworking every function in the call stack to thread that data through.

Clojure being a pragmatic language, we have the STM available for just such cases.

scheduler/dev-config.edn
pjlegato: OK, done.

scheduler/src/cook/mesos/api.clj
```diff
@@ -419,18 +449,43 @@
...
+                            uuids (map uuid-filter uuids)]
+                        ;; Return true if every UUID is in use, or
+                        ;; else return false with a list of
+                        ;; nonexistant UUIDs.
+                        (if (every? (map :is-allowed uuids))
+                          true
+                          (let [message  (->> (map :message uuids)
+                                                 (remove nil?)
+                                                 (str/join "; " ))]
+                            (log/info "Denying access:" message)
```
wyegelwel: If we are logging denying access, I would like to separate out the case when there simply was no job for that uuid.

pjlegato: Done, in latest push.

scheduler/src/cook/components.clj
```diff
@@ -261,11 +306,22 @@
...
   (graph/eager-compile
     {:server-port (fnk [[:config port]]
                        port)
+     :user-privileges (fnk [[:config {user-privileges {}}]]
+                           user-privileges)
+
+     :authorization-config (fnk [[:config authorization-config]]
+                                authorization-config)
+
+     :admins (fnk [[:config [:authorization-config admins]]]
```
wyegelwel: It looks like we never use this. Am I just missing some ref?

scheduler/src/cook/mesos/api.clj
```diff
@@ -71,26 +75,32 @@
...
+     ;; length at most 128.
+     name :- (s/both s/Str (s/pred #(re-matches #"[\.a-zA-Z0-9_-]{0,128}" %) 'under-128-characters-and-alphanum?))
+     priority :- (s/both s/Int (s/pred #(<= 0 % 100) 'between-0-and-100))
+     max-retries :- (s/both s/Int (s/pred pos? 'pos?))
+     max-runtime :- (s/both s/Int (s/pred pos? 'pos?))
+     cpus :- PosDouble
+     mem :- PosDouble
+     ;; Make sure the user name is valid. It must begin with a lower case character, end with
+     ;; a lower case character or a digit, and has length between 2 to (62 + 2).
+     user :- (s/both s/Str (s/pred #(re-matches #"\A[a-z][a-z0-9_-]{0,62}[a-z0-9]\z" %) 'lowercase-alphanum?))
```
wyegelwel: Where does this requirement come from? Posix?

scheduler/src/cook/mesos/api.clj
```diff
@@ -385,7 +415,7 @@
 ;;;
 ;;; On GET; use repeated job argument
 (defn job-resource
-  [conn fid task-constraints]
+  [conn fid task-constraints auth-config]
```
wyegelwel: Instead of passing auth-config, pass a partial of `is-authorized` with auth-config already passed in.

wyegelwel: Applies to all below.

scheduler/src/cook/components.clj
pjlegato: You're right, that is cruft leftover from an earlier draft. I removed it.

scheduler/src/cook/mesos/api.clj
pjlegato: I don't know. That requirement was in the pre-existing schema, which I just converted into an s/defrecord.

I'd guess it's probably a Mesos-level constraint, though I haven't been able to find anything specific about it after a quick look at the Mesos docs.

scheduler/src/cook/mesos/api.clj
pjlegato: Done.

pjlegato: In shares-rest-endpoints branch.

scheduler/src/cook/components.clj
wyegelwel: Is this still around or weird github issue?

scheduler/src/cook/mesos/api.clj
wyegelwel: Is this change going to be made?

scheduler/src/cook/mesos/api.clj
wyegelwel: Is this change going to be made?

SETUP.md
```diff
@@ -0,0 +1,189 @@
+# Setting up your Cook dev environment
```
wyegelwel: I vaguely remember a comment saying you had reordered this doc so that the mac specific stuff was separated or noted clearly. Is there a missing commit?

wyegelwel:
This looks ready to go, though there are a few comments that said things would be fixed but were not. Is there a missing commit?

scheduler/src/cook/components.clj
pjlegato: That got removed on the shares-rest-endpoints branch in a different commit than the one I merged back (in 82458ac780707bcd000a342383b5679a65a486c7).
I've just cherry-picked it back into this branch. Sorry about the confusion.

scheduler/src/cook/mesos/api.clj
pjlegato: Sure, if you want. Since they are equivalent forms and the comment was ambiguous, I took that as more general information rather than a change request.

pjlegato: Done.

scheduler/src/cook/mesos/api.clj
pjlegato: Done.

SETUP.md
pjlegato: The main change there was lines 7-11, which tell the reader generally the sort of changes they have to make to set it up on other non-Mac systems.

PR #164, opened 2016-08-09
Grand unified pull request

pjlegato:
@wyegelwel As requested, this PR includes all 3 of the branches I have been working on -- config-file-authorization-system, set-job-retries-endpoint, and shares-rest-endpoints.


wyegelwel:
Few comments. Once you address those, it LGTM, so can you fix the merge conflict and squash commits where it makes sense.

scheduler/src/cook/mesos/api.clj
```diff
@@ -419,18 +452,53 @@
...
+                                      (unused-uuid? datomic uuid)
+                                      false
+                                      (catch Exception e
+                                        true)))
+
+                            ;; Changes each UUID into a map describing
+                            ;; whether it's allowed, and if not, why
+                            ;; it's not allowed.
+                            uuid-filter (fn [uuid]
+                                          (let [job (fetch-job-map (db conn)
```
wyegelwel: You create `datomic` above, use that.

pjlegato: OK, good catch. Done

PR #165, opened 2016-08-15
REVIEW ONLY: Replace clj-mesos with Mesomatic.

mforsyth:
This branch fully replaces clj-mesos with Mesomatic.

At this time, it depends on a version of Mesomatic with this PR merged in:

https://github.com/pyr/mesomatic/pull/23

Thus, it can't be used without a new version of Mesomatic (or forking Mesomatic); at this stage this PR is for code review purposes only & shouldn't be merged yet.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -14,7 +14,8 @@
 ;; limitations under the License.
 ;;
 (ns cook.mesos.scheduler
-  (:require [clj-mesos.scheduler :as mesos]
+  (:require [mesomatic.scheduler :as mesos]
+            [mesomatic.types :as mtypes]
             [cook.mesos.util :as util]
             [cook.mesos.dru :as dru]
             cook.mesos.schema
@@ -29,6 +30,8 @@
             [metrics.counters :as counters]
             [metrics.gauges :as gauges]
             [clojure.core.async :as async]
+            [clojure.edn :as edn]
+            [clojure.set :as set]
             [clj-time.core :as time]
             [clj-time.periodic :as periodic]
             [clj-time.coerce :as tc]
@@ -39,6 +42,7 @@
             [clojure.core.cache :as cache]
             [cook.mesos.reason :refer :all])
   (import java.util.concurrent.TimeUnit
+          org.apache.mesos.Protos$Offer
           com.netflix.fenzo.TaskAssignmentResult
           com.netflix.fenzo.TaskScheduler
           com.netflix.fenzo.VirtualMachineLease))
@@ -47,12 +51,25 @@
   []
   (java.util.Date.))
 
+
+(defn offer-resource-values
+  [offer resource-name value-type]
+  (->> offer :resources (filter #(= (:name %) resource-name)) (map value-type)))
+
+(defn offer-resource-scalar
+  [offer resource-name]
+  (apply + (offer-resource-values offer resource-name :scalar)))
+
+(defn offer-resource-ranges
+  [offer resource-name]
+  (apply concat (offer-resource-values offer resource-name :ranges)))
+
 (defn tuplify-offer
-  "Takes an offer and converts it to a queryable format for datomic"
-  [{:as offer
-    :keys [slave-id]
-    {:keys [cpus mem]} :resources}]
-  [slave-id cpus mem])
+  "Takes an offer (from Mesomatic)  and converts it to a queryable format for datomic"
+  [offer]
+  [(-> offer :slave-id :value)
+   (offer-resource-scalar offer "cpus")
+   (offer-resource-scalar offer "mem")])
 
 (defn get-job-resource-matches
   "Given an offer and a set of pending jobs, figure out which ones match the offer"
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -47,12 +51,18 @@
   []
   (java.util.Date.))
 
+(defn offer-resource-value
+  ([offer resource-name]
+   (offer-resource-value offer resource-name :scalar))
+  ([offer resource-name value-type]
+   (->> offer :resources (filter #(= (:name %) resource-name)) first value-type)))
```
wyegelwel: Why are we taking the first? 

Suppose a single offer has "cpus" from both role "cook" and role "*", won't this mean we are throwing out one of them?

mforsyth: Hmm, that's a good point.  I didn't consider multiple roles for the same resource in the same offer....  Since Cook can only run with 1 role, and mesos is aware of the role, we can assume that any resource items will either be for Cook's role or for "*" right?   And therefore, for Cook's purposes, we can simply sum the values for each resource entry of the appropriate type... correct?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -526,10 +550,42 @@
...
+                               :scalar (:cpus resources)
+                               :role (:role t)}
+                              (when (seq (:ports resources))
+                                {:type :value-ranges,
+                                 :name "ports",
+                                 :ranges (:ports resources)
+                                 :role (:role t)})]))
+
+     :labels {:labels (mapv (fn [kv] {:key (key kv) :value (val kv)})
+                            (vec (:labels t)))}
```
wyegelwel: Is `:labels` a map or a sequence of maps?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -526,10 +550,42 @@
...
+     :command {:value (:value cmd)
+               :environment {:variables
```
wyegelwel: Perhaps use an `update-in` for environment instead of destructing and restructuring the whole map

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Also, consider making a function to do this transform, we do it at least twice.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -526,10 +550,42 @@
...
+  (let [cmd (:command t)
+        resources (:resources t)]
+    {:name (:name t)
+     :task-id [{:value (:task-id t)}]
+     :slave-id [{:value (:slave-id t)}]
+     :resources (vec (remove nil?
+                             [{:type :value-scalar,
+                               :name "mem",
+                               :scalar (:mem resources)
+                               :role (:role t)}
```
wyegelwel: I'm concerned this will not work in one of two cases.

1. Cook is started with role "cook" but gets offers with resources with role "*"
2. Cook is started with role "cook" but gets offers with resources with role "*" and role "cook"

I make a special case for the last one because it will require being a bit clever about how the resources are mapped. For example, imagine the job will fit if you add the "*" and "cook" resources but would not otherwise. 

I expect what you will end up having to do is take in the matches of offers and tasks such that you can parcel out correctly.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -648,14 +704,19 @@
...
                                                tasks)
-                              slave-id (:slave-id (first offers))]]
-                  (mesos/launch-tasks
-                    driver
-                    (mapv :id offers)
+                              slave-id (-> offers first :slave-id :value)]]
+
+                  (mesos/launch-tasks!
+                   driver
+
```
wyegelwel: Is space here intentional?

scheduler/src/cook/mesos/scheduler.clj
mforsyth: Thanks for pointing this out @wyegelwel .  I misunderstood how the "wildcard" resources are treated in allocations.  It seems that if Mesos offers your framework 8 wildcard (*) CPUS and 8 CPU's that can only be used by your framework's role, you can't simply accept 16 CPU's with your role... Mesos won't simply assign wildcard CPU's to any role. You need to tell Mesos specifically to allocate 8 wildcard CPU's and 8 role-specific CPU's.    

So, you're right, we need to change the Cook's own logic that decides which resources to accept - it's not simply a case of swapping out clj-mesos for Mesomatic.  

It seems to me that the "polite approach" in terms of impact on other frameworks would be to use up all role-specific resources before dipping into the wildcard resources.

scheduler/src/cook/mesos/scheduler.clj
mforsyth: :labels is a map, which consists of a single key :labels, whose value is a vector of maps (that each have a :key and :value).  Seems strange, but that's the structure of the protobuf.

mforsyth: 👍 for reusable function

scheduler/src/cook/mesos/scheduler.clj

PR #166, opened 2016-08-15
Small improvements on fenzo. Small features added before Fenzo

wyegelwel:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -945,22 +951,19 @@
...
-        (log/info "Found" (count (:offensive categorized-jobs)) "offensive jobs")
-        (async/>! offensive-jobs-ch (:offensive categorized-jobs))))
-    (:inoffensive categorized-jobs)))
+  (timers/time!
+    filter-offensive-jobs-duration
+    (let [max-memory-mb (* 1024.0 max-memory-gb)
+          is-offensive? (partial is-offensive? max-memory-mb max-cpus)
+          inoffensive (remove is-offensive? jobs)
+          offensive (filter is-offensive? jobs)]
+      ;; Put offensive jobs asynchronically such that it could return the
```
mforsyth: If the goal is just to make it so "offensive" and "inoffensive" are both lazy, could we accomplish that just be replacing group-by with partition-by?  https://clojuredocs.org/clojure.core/partition-by  
That way, we wouldn't have to call is-offensive twice for every job.

scheduler/src/cook/components.clj
```diff
@@ -436,7 +436,7 @@
     (println "Started cook, stored variable in user/main-graph")
     (catch Throwable t
       (log/error t "Failed to start Cook")
-      (println "Failed to start Cook")
+      (println "Failed to start Cook" t)
```
mforsyth: mo printing = mo better :)

scheduler/src/cook/mesos/reason.clj
```diff
@@ -13,78 +13,34 @@
...
-(def reason-mesos-gc-error 4002)
-(def reason-mesos-container-launch-failed 4003)
-(def reason-mesos-container-update-failed 4004)
-(def reason-mesos-slave-disconnected 4005)
-(def reason-heartbeat-lost 4006)
+(comment
 
-(def reason-mesos-framework-removed 5000)
-(def reason-mesos-master-disconected 5001)
+  (def conn (d/connect "datomic:riak://ramkv-dr.app.twosigma.com:8098/datomic1/mesos-jobs-reason-test?interface=http"))
```
mforsyth: ❌

scheduler/src/cook/mesos/reason.clj
```diff
@@ -13,78 +13,34 @@
...
-(def reason-mesos-container-limitation-disk 2001)
-(def reason-mesos-container-limitation-memory 2002)
-(def reason-max-runtime 2003)
+(defn mesos-reason->cook-reason-entity-id
+  [db mesos-reason]
+  (if-let [reason-entity-id (:db/id (d/entity db [:reason/mesos-reason mesos-reason]))]
+    reason-entity-id
+    (do
+      (log/warn "Unknown mesos reason: " mesos-reason)
+      (:db/id (d/entity db [:reason/name :mesos-unknown])))))
```
mforsyth: So the point of this is to store mesos' original reason for the job being cancelled, as well as Cook's understanding of that reason (if any), and to do that regardless of whether Cook is aware of all of the possible Mesos reasons - right?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -837,17 +836,24 @@
    :timout-hours specifies the timeout hours for lingering tasks"
   [conn driver config]
   (let [{interval-minutes :timeout-interval-minutes
-         timeout-hours :timeout-hours}
+         max-timeout-hours :max-timeout-hours
+         default-timeout-hours :default-timeout-hours
+         timeout-hours :timout-hours}
```
mforsyth: Typo?  :timot-hours

mforsyth: Maybe a sign that it's a good time to use :keys ...

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -837,17 +836,24 @@
...
+                      max-timeout-hours (or max-timeout-hours timeout-hours)
+                      default-timeout-hours (or default-timeout-hours timeout-hours)
+                      lingering-tasks (get-lingering-tasks db now max-timeout-hours default-timeout-hours)]
                   (when (seq lingering-tasks)
-                    (log/info "Starting to kill lingering jobs running more than" timeout-hours
-                              "hours. There are in total" (count lingering-tasks) "lingering tasks.")
+                    (log/info "Starting to kill lingering jobs running more than their max-runtime."
+                              {:default-timeout-hours default-timeout-hours
+                               :max-timeout-hours max-timeout-hours}
+                              "There are in total" (count lingering-tasks) "lingering tasks.")
```
mforsyth: Trying to understand why there is both a default and max timeout hours setting...

Is the idea that there is also goint to be an optional per-job timeout?  
If it's missing, default-timeout-hours is used?
If max-timeout-hours is defined, it overrides the job's configured timeout?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -774,7 +780,7 @@
     (when (seq running-tasks)
       (log/info "Preparing to reconcile" (count running-tasks) "tasks")
       (doseq [[task-id] running-tasks
-              :let [task-ent (d/entity db task-id)
+              :let [task-ent (d/entity db [:instance/task-id task-id])
```
mforsyth: Ahh, that fixes it?  Nice

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -410,14 +410,20 @@
   [^TaskScheduler fenzo considerable offers db fid]
   (log/debug "Matching" (count offers) "offers to" (count considerable) "jobs with fenzo")
   (let [t (System/currentTimeMillis)
+        _ (log/debug "offer to scheduleOnce" offers)
```
mforsyth: just for style, can't we just move these log statements above the let block?  Fewer anonymous let variables

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -410,14 +410,20 @@
...
+        _ (log/debug "tasks to scheduleOnce" considerable)
         leases (mapv #(->VirtualMachineLeaseAdapter % t) offers)
         requests (mapv (fn [job]
                          (->TaskRequestAdapter job (job->task-info db fid (:db/id job))))
                        considerable)
-        result (.scheduleOnce fenzo requests leases)
+        ;; Need to lock on fenzo when accessing scheduleOnce because scheduleOnce and
+        ;; task assigner can not be called at the same time.
+        ;; task assigner may be called when reconciling
```

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: My issue with partition-by is that I don't know which partition is which. Is there a principled way to determine that?

scheduler/src/cook/mesos/reason.clj
wyegelwel: ah good catch!

scheduler/src/cook/mesos/reason.clj
wyegelwel: ah, yes I think that is a fair characterization.

The old code was a bunch of data that was poorly organized. This groups the information together and allows us the ability to extend it in the future.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: :+1:

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Yes, exactly that. The default is so that users don't always need to specify. The max is a bit more special case and is to prevent someone from requesting such a long time out that other factors start to cause problems. In our environment this is things like kerberos tickets expiring.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: :+1:

PR #167, opened 2016-08-15
Small improvements on fenzo. Small features added before Fenzo

wyegelwel:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -585,15 +585,9 @@
             ;; "Penalization" should only be employed when Fenzo does successfully match,
             ;; but the matches don't align with Cook's priorities.
             (empty? matches) true
-            (not (compare-and-set! pending-jobs scheduler-contents new-scheduler-contents))
```
mforsyth: On the surface level, now we only have 2 branches in the cond - it might as well be an "if".

Going deeper, can you tell me a little more about this?  If we can just plow forward on atom contention rather than recycling, it's way simpler, but I don't know exactly why the contention handling was needed in the first place (what scenario it helped), so I can't feel safe declaring it's no longer needed.

wyegelwel: Yea that is a great point. We had the check for contention there because of how we processed the queue. The matching would consider the head of the queue and if that matched, then it would try the next job, so on and so forth. The invariant here was that the head of the queue would always get scheduled first. If there was atom contention, it *might* imply that the head of the queue changed. Instead of doing the processing to check if it was safe to use the new contents, we took the easier path and recycled.

Since we no longer have that invariant and are willing to go through more of the scheduler contents to remove the jobs we scheduled I don't think the recycling is as compelling and having it there has lead to weird nasty bugs.

PR #169, opened 2016-08-25
REVIEW ONLY: Scheduler: Replace clj-mesos with mesomatic.

mforsyth:
This change also enables us to properly support Mesos Roles.

At this time, it depends on a version of Mesomatic with this PR merged in:

pyr/mesomatic#23

Thus, it can't be used without a new version of Mesomatic (or forking Mesomatic); at this stage this PR is for code review purposes only & shouldn't be merged yet.

### DIFF
--- a/scheduler/src/cook/mesos/task.clj
+++ b/scheduler/src/cook/mesos/task.clj
@@ -1,0 +1,174 @@
+(ns cook.mesos.task
+  (:require [plumbing.core :refer (map-vals)]
+            [clojure.tools.logging :as log])
+  (import com.netflix.fenzo.TaskAssignmentResult))
+
+
+(defn TaskAssignmentResult->task-info
+  "Organizes the info Fenzo has already told us about the task we need to run"
+  [^TaskAssignmentResult fenzo-result]
+  (merge (:task-info (.getRequest fenzo-result))
+         {:ports-assigned (.getAssignedPorts fenzo-result)}))
+
+
+(defmulti combine-like-resources
+  (fn [list] (-> list first :type)))
+
+(defmethod combine-like-resources :value-scalar [resources]
+  (->> resources (map :scalar) (apply +)))
+
+(defmethod combine-like-resources :value-ranges [resources]
+  (->> resources (map :ranges) (apply into)))
+
+(defn resources-by-role
+  "Given a set of offers, combine all the available resources into a comprehensible
+  and easily walkable data strucuture, grouped first by resource name e.g. mem, then
+  by role."
+  [offers]
+  (->> offers
+       (map :resources)
+       flatten
+       (group-by :name)
+       (map-vals (fn [resources]
+                   (->> resources
+                        (group-by :role)
+                        (map-vals combine-like-resources))))))
+
+(defn range-contains?
+  "true iff val is contained by a mesos-style range e.g. {:begin 1 :end 10}"
+  [val mesos-range]
+  (and (<= (:begin mesos-range) val)
+       (>= (:end mesos-range) val)))
+
+(defn role-containing-port
+  "in: {\"*\" [{:begin 201 :end 202}] \"cook\" [{:begin 203 :end 204}]}, 201
+  ;; out \"*\""
+  [available-ports port]
+  (->> available-ports
+       seq
+       (filter #(some (partial range-contains? port) (val %)))
+       ffirst))
+
+(defn take-ports
+  "Given a set of available resources (in the format returned by resources-by-role),
+  returns a vector of mesos messages that will reserve the specified ports.
+  Note: unlike with scalar-resources, available-resources don't need to be
+  changed by this function, because Fenzo has already provided us with
+  specific ports to use for every task."
+  [available-resources ports-needed]
+  (mapv (fn [port]
+          {:name "ports"
+           :type :value-ranges
+           :role (role-containing-port available-resources port)
+           :ranges [{:begin port :end port}]})
+        ports-needed))
+
+(defn add-ports-to-task-info
+  "Given a set of tasks and offers that were matched together by Fenzo,
+   assigns the specific ports requested by role to each task.
+   Returns the input tasks, decorated with :ports-resource-messages"
+  [available-resources tasks]
+  (let [available-ports (available-resources "ports")]
+    (map (fn [task]
+           (let [ports (:ports-assigned task)
+                 port-env-vars (into {} (map-indexed (fn [i p] [(str "PORT" i) (str p)])
+                                                     ports))]
+             (-> task
+                 (assoc :ports-resource-messages (take-ports available-ports ports))
+                 (update-in [:command :environment] merge port-env-vars))))
+         tasks)))
+
+(defn take-resources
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,167 @@
...
+(defn resources-by-role
+  "Given a set of offers, combine all the available resources into a comprehensible
+  and easily walkable data strucuture, grouped first by resource name e.g. mem, then
+  by role."
+  [offers]
+  (->> offers
+       (map :resources)
+       flatten
+       (group-by :name)
+       (fmap (fn [resources]
```
wyegelwel: Use `map-vals` from [prismatic-plumbing](https://github.com/plumatic/plumbing/blob/master/src/plumbing/core.cljx#L68) here. I prefer to avoid fmap unless the data you are getting really can take different forms and you are unsure what you are getting. Since in this case and the one just below we know it is a map and we want to map on the values, using `map-vals` is more clear what is occurring.

scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,167 @@
...
+(defn take-resources
+  "Given a set of available resources (in the format returned by resources-by-role),
+  take the specified amount of the specified resource from the pool.
+  Return {:remaining-resources (the pool after resources were taken)
+          :mesos-messages (the Mesos messages that are necessary to reserve the resources)
+          :amount-still-needed (the amount still needed to satisfy amount... if it's not 0, something went wrong)}"
+  [available-resources resource-name amount]
+  (let [avail (available-resources resource-name)
+        ;; to be polite to other frameworks, take from role-specific pool first.
+        sorted-roles (sort-by #(= "*" %) (keys avail))
```
wyegelwel: holy shit does that work?! That is cool.

scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,167 @@
...
+        init-state {:remaining-resources avail
+                    :mesos-messages []
+                    :amount-still-needed amount}]
+    (reduce (fn [{:keys [remaining-resources mesos-messages amount-still-needed]
+                  :as state} role-name]
+              (let [amount-avail (or (remaining-resources role-name) 0)
+                    amount-to-take (min amount-still-needed amount-avail)]
+                (if (pos? amount-to-take)
```
wyegelwel: I just want to make sure I understand this. When the `available-resources` and `amount` comes from a match from fenzo, this should always be positive, right?

wyegelwel: Oh, wait ignore the first question it doesn't make sense.

`amount-to-take` *should* always be positive, right? If it is not it means either `remaining-resources` or `amount-still-needed` is negative. It is probably worth either asserting amount-to-take is positive or at least logging a warn that it is negative and should not be.

scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,167 @@
...
+    (map (fn [task]
+           (let [ports (:ports-assigned task)
+                 port-env-vars (into {} (map-indexed (fn [i p] [(str "PORT" i) (str p)])
+                                                     ports))]
+             (-> task
+                 (assoc :ports-resource-messages (take-ports available-ports ports))
+                 (update-in [:command :environment] merge port-env-vars))))
+         tasks)))
+
+(defn add-scalar-resources-to-task-info
```
wyegelwel: I think it is worthwhile to write this function without hardcoding mem and cpus as I expect gpus and disk (among others) will be added soon.

scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,167 @@
...
+        (vec m)))
+
+(defn task-info->mesos-message
+  "Given a clojure data structure (based on Cook's internal data format for jobs),
+  which has already been decorated with everything we need to know about
+  a task, return a Mesos message that will actually launch that task"
+  [t]
+  {:name (:name t)
+   :slave-id (:slave-id t)
+   :task-id [{:value (:task-id t)}]
```
wyegelwel: Why is `task-id` a list?

scheduler/src/cook/mesos/task.clj
mforsyth: @wyegelwel amount-to-take should certainly never be negative - but it's expected for it to sometimes be 0.
Let's say that the initial mem available by role is {"cook" 200 "*" 200"}.
Task 1 needs 300 MB of memory... after it is done the remaining resource pool for memory will be: {"cook" 0  "*" 100 }
So if task 2 only needs 50MB of memory, it will first look at "cook", see that amount-to-take is 0, then move onto  "*" and take 50MB from the remaining 100 there.

scheduler/src/cook/mesos/task.clj
mforsyth: Hmm, it was sent as a list in a mesomatic examples repo I had looked it... But you're right, the actual protobuf only supports a single value here.  Whether it's a single value or a list with 1 element, Mesomatic produces the exact same results, so I'll switch to single value.

scheduler/src/cook/mesos/task.clj
mforsyth: 👍

scheduler/src/cook/mesos/task.clj
mforsyth: 👍

scheduler/src/cook/mesos/task.clj
wyegelwel: Right, I always forget that `pos?` is false for 0

scheduler/src/cook/mesos/task.clj
```diff
@@ -0,0 +1,174 @@
...
+            sorted-roles)))
+
+(defn take-all-scalar-resources-for-task
+  [resources task]
+  (reduce
+   (fn [{:keys [remaining-resources mesos-messages]} [resource-keyword amount]]
+     (let [resource-name (name resource-keyword)
+           adjustment (take-resources remaining-resources resource-name amount)]
+       {:remaining-resources (assoc remaining-resources resource-name (:remaining-resources adjustment))
+        :mesos-messages (vec (concat mesos-messages (:mesos-messages adjustment)))}))
```
wyegelwel: Use `into` instead of concat here

mforsyth: Wow, I hadn't seen that!  Very interesting.  I just replaced all instances of concat with into.

wyegelwel:
I'd like to get this in tomorrow. If mesosmatic doesn't answer, we will fork and deploy a jar from the fork

dgrnbrg:
Adding myself to watch to merge.

wyegelwel:
@dgrnbrg it is merged but I need to switch the mesomatic it is pointing at to include the labels support. Will post again when that is done.

dgrnbrg:
It looks like the code doesn't compile. Could you check this error:
https://travis-ci.org/twosigma/Cook/jobs/158275827 ?

PR #177, opened 2016-09-07
Add GPU support to api & config file

dgrnbrg:
This PR adds GPUs to the Cook API, if they're enabled in the config file (by default, there is no behavior change). This should be mergeable separately from the actual GPU support, since the latter only affects the scheduler, while this change affects several aspects of the system.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -1084,7 +1084,7 @@
       (build)))
 
 (defn create-datomic-scheduler
-  [conn set-framework-id driver-atom pending-jobs-atom heartbeat-ch offer-incubate-time-ms fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset task-constraints]
+  [conn set-framework-id driver-atom pending-jobs-atom heartbeat-ch offer-incubate-time-ms fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset task-constraints gpu-enabled?]
   (let [fid (atom nil)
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/mesos/share.clj
```diff
@@ -60,14 +60,14 @@
    \"default\" user. If there is NO \"default\" value for a specific type,
    return Double.MAX_VALUE."
   [db user]
-  (->> (util/get-all-resource-types db)
+  (->> [:cpus :mem]
```
dgrnbrg: I'm not sure if you'd like to handle this differently. With the original code, a variety of tests broke, as GPU shares are not handled by the downstream consumers of this code.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1099,6 +1099,10 @@
                   (log/info "Registered with mesos with framework-id " framework-id)
                   (reset! fid framework-id)
                   (set-framework-id framework-id)
+                  (when (and gpu-enabled? (not (re-matches #"1\.\d+\.\d+" (:version master-info))))
+                    (log/error "Cannot enabled GPU support on pre-mesos 1.0. The version we found was " (:version master-info))
```
wyegelwel: nitpick: "Cannot enable GPU..." s/enabled/enable/

scheduler/src/cook/mesos/share.clj
wyegelwel: Which tests broke? I would prefer not to have to do this but on the other hand, don't want you going down a rabbit hole

dgrnbrg: I'm working on a new branch that includes the scheduler logic as well as
API. I've been developing them together as I discover how to structure the
code. The aforementioned tests have all been reconciled, and there will be
full support for GPU shares.

PR #178, opened 2016-09-09
Adding support for docker force pull image

m4ce:
This pull request adds support for force-image-pull option the the job's docker configuration.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -47,6 +47,7 @@
   "Schema for a DockerInfo"
   {:image s/Str
    (s/optional-key :network) s/Str
+   (s/optional-key :force-pull-image) s/Bool
    (s/optional-key :parameters) [{:key s/Str :value s/Str}]
    (s/optional-key :port-mapping) [PortMapping]})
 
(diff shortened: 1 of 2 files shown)


wyegelwel:
Hey Matteo, thanks for the PR! I'm excited to add this feature to Cook!

What steps did you take to test this?

m4ce:
Hi @wyegelwel,

for the testing, it's a bit manual as you need to have the image:tag already pulled on the destination system and then have a newer version of the same image:tag available in the registry. I was testing with a private registry and custom docker images.

In the job's definition, I then used:

```
 "container": {"type": "DOCKER", "docker": {"image": "mycustomimage:latest", "network": "HOST", "force-pull-image": true}}
```

wyegelwel:
Matteo, this looks good to me. With the most recent commit were you able to confirm that you are pulling the latest image? If you don't use force-pull does it use what is currently there?

m4ce:
Hi @wyegelwel,

is it does, we have been using it for a few days now and it works fine.

Thanks,
Matteo

wyegelwel:
Glad to hear it is working well! Thanks again for the pr.

PR #179, opened 2016-09-12
Backfill broken

dgrnbrg:
[This commit](https://github.com/twosigma/Cook/commit/c356e3dcf1ef6942f72774c9954777ca07d59209) broke backfilling by changing the outputs from sets to vectors. Further down, `contains?` is used on these parameters.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -552,9 +552,9 @@
       (meters/mark! upgraded-cpu (:cpus resources-upgraded))
       (meters/mark! upgraded-mem (:mem resources-upgraded)))
 
-    {:fully-processed (mapv :job/uuid jobs-fully-processed)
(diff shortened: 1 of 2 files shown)


wyegelwel:
Yes the change was to improve readability. Are you concerned about
performance? I have a change that I need to push out to send only a small
subset (10x of max jobs considerable) of the full scheduler contents to
this function so that we don't process through a potentially *very* large
queue. Once that change is in this function is a fairly insignificant
performance wise and so it is worthwhile to have the readability.

dgrnbrg:
I'm just mentioning that the performance is worse.

The actual problem is that none of the code works, because those collections need to be `set`, not `vector`.

wyegelwel:
Yea for sure worse but not problematic (yet). Broken is more bad. I noticed
we were not getting upgraded jobs often but didn't dig in. Thanks for
finding this

dgrnbrg:
No problem!

PR #183, opened 2016-09-19
scheduler: Fix error when "combining" only 1 port range.

mforsyth:


scheduler/src/cook/mesos/task.clj
```diff
@@ -18,7 +18,7 @@
   (->> resources (map :scalar) (apply +)))
 
 (defmethod combine-like-resources :value-ranges [resources]
-  (->> resources (map :ranges) (apply into)))
+  (->> resources (map :ranges) (apply (comp vec flatten conj))))
```
wyegelwel: It looks like @dgrnbrg  ran into the same bug and fixed it using `(reduce into)` instead of the solution above. Can you confirm that `(reduce into)` passes your tests and if so, just leave the tests in PR? Once @dgrnbrg's PR is in I will merge this as well so that we can have this tested.

mforsyth: @wyegelwel  Yes, (reduce into) also works (makes the test cases pass) and is definitely more elegant than (comp vec flatten conj).  Thanks @dgrnbrg !

PR #184, opened 2016-09-19
scheduler: Fix error when recycling offer-stash.

mforsyth:
offer-stash was initialized to nil and was only being reset! to its own
value, @offer-stash - meaning it could never acquire a value other than
nil.

Thus, attempting to put its value back on the channel in the exception
handling block would in turn raise an exception (can't put nil on
channel).  Since this exception was not handled, the thread would die,
and handle-resource-offers! would stop happening, resulting in no offers
to match until Cook was restarted.

I believe the var offers-scheduled contains what @offer-stash USED to
be reset! to, prior to commit a134d1f8776eb1b0627eb74a073bae125bf6d52a.


wyegelwel:
Yup, you are spot on. Good find!

PR #186, opened 2016-09-20
Add GPU support to Cook

dgrnbrg:
This adds support for:
- GPU resources on agents, which will only be accepted if enabled in the
  config file
- Scheduling GPU jobs on GPU machines and non-GPU jobs on non-GPU
  machines
- Uses DRU & preemption with GPUs as well, but they're considered
  independently
- Adds GPUs to the REST API
- If GPUs are enabled and the Mesos version doesn't support them, fails
  loudly

### DIFF
--- a/scheduler/docs/configuration.asc
+++ b/scheduler/docs/configuration.asc
@@ -98,6 +98,14 @@ We'll look at the configurable options in turn:
   This sets the role that Cook will connect to Mesos with. Default: `*`
   You can omit this property unless you've enabled security features in Mesos. The value should be in authorized list for the current `:principal` in `register_frameworks` Action in Mesos Authorization file. See http://mesos.apache.org/documentation/latest/authorization/ for details.
 
+`:enable-gpu-support`::
+  This enables GPU support for Cook.
+  It is a boolean value, with default value `false`.
+  This property will only work with Mesos 1.0 and above, since that's when GPU support was added.
+  If you enable this on an earlier version of Mesos, Cook will fail to start and print the error in the log.
+  If you enable this and your cluster doesn't have any GPU machines, Cook will accept GPU jobs, but they'll never be scheduled.
+  See https://github.com/apache/mesos/blob/master/docs/gpu-support.md for details on configuring the agents, installing external NVidia dependencies, and configuring Docker/GPU integration.
+
 [[auth_config]]
 ==== Authorization Configuration
 
--- a/scheduler/docs/scheduler-rest-api.asc
+++ b/scheduler/docs/scheduler-rest-api.asc
@@ -75,6 +75,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`max_runtime` | long | the maximum running time of the job in milliseconds. An instance that runs for more than max_runtime will be killed and job will be retried.
 |`cpus` | double | number of requested cpus.
 |`mem` | double | MB of requested memory.
+|`gpus` | integer | Number of requested GPUs. Must be a whole number and support must be enabled in the config.
 |`ports` | integer | Number of ports that should be opened for the job.
 |`uris` | list of URI objects | A list of URIs that will be fetched into the container before launch.
 |`env` | JSON map | A map of environment variables to be provided to the job.
@@ -155,6 +156,7 @@ The response body contains the following:
 |`uuid` | the job UUID
 |`cpus` | the number of CPUs requested
 |`mem` | the amount of memory requested
+|`gpus` | the number of GPUs requested
 |`framework_id` | the Mesos framework ID of cook
 |`status` |  the status of the job
 |`instances` | a list of job instance maps (see <<instance_maps>>)
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -42,7 +42,8 @@
                  [org.clojure/data.priority-map "0.0.5"]
                  [swiss-arrows "1.0.0"]
                  [riddley "0.1.10"]
-                 [com.netflix.fenzo/fenzo-core "0.8.2"
+                 [org.apache.mesos/mesos "1.0.1"]
+                 [com.netflix.fenzo/fenzo-core "0.9.4-SNAPSHOT"
                   :exclusions [org.apache.mesos/mesos
                                com.fasterxml.jackson.core/jackson-core
                                org.slf4j/slf4j-api
@@ -56,6 +57,7 @@
                  [com.draines/postal "1.11.0"
                   :exclusions [commons-codec]]
                  [prismatic/plumbing "0.1.1"]
+                 [log4j "1.2.17"]
                  [instaparse "1.4.0"]
                  [org.codehaus.jsr166-mirror/jsr166y "1.7.0"]
                  [clj-pid "0.1.1"]
@@ -84,7 +86,7 @@
 
                  ;;External system integrations
                  [me.raynes/conch "0.5.2"]
-                 [wyegelwe/mesomatic "0.28.0-r0-SNAPSHOT"]
+                 [spootnik/mesomatic "1.0.1-r0"]
                  [org.clojure/tools.nrepl "0.2.3"]
 
                  ;;Ring
@@ -134,6 +136,7 @@
 
    :dev
    {:dependencies [[clj-http-fake "1.0.1"]
+                   [org.mockito/mockito-core "1.10.19"]
                    [org.clojure/test.check "0.6.1"]]
     :jvm-opts ["-Xms2G"
                "-XX:-OmitStackTraceInFastThrow"
--- a/scheduler/src/cook/mesos/dru.clj
+++ b/scheduler/src/cook/mesos/dru.clj
@@ -67,6 +67,21 @@
                           task-resources)]
     scored-tasks))
 
+(defn compute-gpu-task-scored-task-pair
+  "Takes a sorted seq of task entities and the gpu divisor, returns a list of [task cumulative-gpus], preserving the same order of input tasks"
+  [task-ents gpu-divisor]
+  (let [task-resources (->> task-ents
+                            (map (comp #(select-keys % [:gpus]) util/job-ent->resources :job/_instance)))
+        task-cum-gpus (->> task-resources
+                           (accumulate-resources)
+                           (map (fn [{:keys [gpus]}]
+                                  (/ gpus gpu-divisor))))
+        scored-tasks (map (fn [task cum-gpus]
+                            [task cum-gpus])
+                          task-ents
+                          task-cum-gpus)]
+    scored-tasks))
+
 (defn sorted-merge
   "Accepts a seq-able datastructure `colls` where each item is seq-able.
    Each seq-able in `colls` is assumed to be sorted based on `key-fn`
@@ -90,6 +105,15 @@
    (sorted-merge key-fn compare colls))
   ([colls]
    (sorted-merge identity compare colls)))
+
+(defn gpu-task-scored-task-pairs
+  "Takes a sorted seq of task entities and the gpu divisor, returns a list of [task cumulative-gpus], preserving the same order of input tasks"
+  [user->sorted-running-task-ents user->dru-divisors]
+  (->> user->sorted-running-task-ents
+       (map (fn [[user task-ents]]
+              (compute-gpu-task-scored-task-pair
+                task-ents (get-in user->dru-divisors [user :gpus]))))
+       (sorted-merge second)))
 
 (timers/deftimer [cook-mesos dru sorted-task-scored-task-pairs-duration])
 
--- a/scheduler/src/cook/mesos/rebalancer.clj
+++ b/scheduler/src/cook/mesos/rebalancer.clj
@@ -111,6 +111,11 @@
 ;;;
 ;;; which would imply that taskC would have a DRU of 0.27.
 ;;;
+;;; For GPU preemption, the functionality works exactly the same as above, with one change: instead of computing the DRU
+;;; as (max (/ used-mem mem-divisor) (/ used-cpus cpu-divisor)), we compute the DRU as (/ used-gpus gpu-divisor). All of
+;;; the different code paths reflect this change, or the fact that GPU scored task pairs are [task cumulative-gpus] rather
+;;; than [task scored-task], as we don't need the additional data for computing the GPU preemption.
+;;;
 ;;; Parameters
 ;;; safe-dru-threshold: Task with a DRU lower than safe-dru-threshold will not be preempted. If each DRU divisor is set
 ;;;                     to the corresponding per user share and safe-dru-threshold is set to 1.0, then tasks that
@@ -136,7 +141,7 @@
 ;;; preemption-decision {:hostname String :task [task-ent] :dru Double :mem Double :cpus Double}
 ;;; preemption-candidates [{:task task-ent :dru Double :mem Double :cpus Double}]
 
-(defrecord State [task->scored-task user->sorted-running-task-ents host->spare-resources user->dru-divisors])
+(defrecord State [task->scored-task user->sorted-running-task-ents host->spare-resources user->dru-divisors compute-pending-job-dru])
 
 (timers/deftimer [cook-mesos rebalancer rebalance-duration])
 (timers/deftimer [cook-mesos rebalancer compute-preemption-decision-duration])
@@ -155,10 +160,35 @@
   [dru]
   (* dru metrics-dru-scale))
 
-(defn compute-pending-job-dru
+(defn compute-pending-gpu-job-dru
+  "Takes state and a pending gpu job entity, returns the dru of the pending-job.
+
+   This algorithm only should be used on jobs that use GPUs. It computes the GPU DRU."
+  [{:keys [task->scored-task user->sorted-running-task-ents user->dru-divisors] :as state}
+   pending-job-ent]
+  (let [user (:job/user pending-job-ent)
+        {gpu-req :gpus} (util/job-ent->resources pending-job-ent)
+        gpu-divisor (get-in user->dru-divisors [user :gpus])
+        pending-task-ent (util/create-task-ent pending-job-ent)
+        nearest-task-ent (some-> user->sorted-running-task-ents
+                                 (get user)
+                                 (rsubseq <= pending-task-ent)
+                                 (first))
+        nearest-task-dru (if nearest-task-ent
+                           (get task->scored-task nearest-task-ent)
+                           0.0)
+        pending-job-dru (+ nearest-task-dru (/ gpu-req gpu-divisor))]
+    (histograms/update! pending-job-drus (dru-at-scale pending-job-dru))
+    (histograms/update! nearest-task-drus (dru-at-scale nearest-task-dru))
+
+    pending-job-dru))
+
+(defn compute-pending-normal-job-dru
   "Takes state and a pending job entity, returns the dru of the pending-job. In the case where the pending job causes user's dominant
    resource type to change, the dru is not accurate and is only a upper bound. However, this inaccuracy won't affect the correctness
-   of the algorithm."
+   of the algorithm.
+
+   This algorithm only should be used on jobs that use cpu & mem, not gpus. It computes the cpu/mem DRU."
   [{:keys [task->scored-task user->sorted-running-task-ents user->dru-divisors] :as state}
    pending-job-ent]
   (let [user (:job/user pending-job-ent)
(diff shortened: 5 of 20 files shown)


dgrnbrg:
This branch is ready for review--I am now testing this code on actual clusters to verify behaviors end-to-end.

scheduler/docs/scheduler-rest-api.asc
```diff
@@ -75,6 +75,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`max_runtime` | long | the maximum running time of the job in milliseconds. An instance that runs for more than max_runtime will be killed and job will be retried.
 |`cpus` | double | number of requested cpus.
 |`mem` | double | MB of requested memory.
+|`gpus` | double | Number of requested GPUs. Must be a whole number and support must be enabled in the config.
```
wyegelwel: I thought gpus had to be integer

scheduler/project.clj
```diff
@@ -84,7 +86,7 @@
 
                  ;;External system integrations
                  [me.raynes/conch "0.5.2"]
-                 [wyegelwe/mesomatic "0.28.0-r0-SNAPSHOT"]
+                 [spootnik/mesomatic "1.0.1-r0"]
```
wyegelwel: :+1:

scheduler/src/cook/mesos/dru.clj
```diff
@@ -67,6 +67,21 @@
                           task-resources)]
     scored-tasks))
 
+(defn compute-gpu-task-scored-task-pair
+  "Takes a sorted seq of task entities and the gpu divisor, returns a list of [task cumulative-gpus], preserving the same order of input tasks"
+  [task-ents gpu-divisor]
+  (let [task-resources (->> task-ents
+                            (map (comp #(select-keys % [:gpus]) util/job-ent->resources :job/_instance)))
+        task-cum-gpus (->> task-resources
+                       (accumulate-resources)
```
wyegelwel: fix indenting

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,7 +155,27 @@
...
+  (let [user (:job/user pending-job-ent)
+        {gpu-req :gpus} (util/job-ent->resources pending-job-ent)
+        gpu-divisor (get-in user->dru-divisors [user :gpus])
+        pending-task-ent (util/create-task-ent pending-job-ent)
+        nearest-task-ent (some-> user->sorted-running-task-ents
+                                 (get user)
+                                 (rsubseq <= pending-task-ent)
+                                 (first))
+        nearest-task-dru (if nearest-task-ent
+                           (get-in task->scored-task [nearest-task-ent])
```
wyegelwel: why use `get-in` here?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,7 +155,27 @@
...
 
-(defn compute-pending-job-dru
+(defn compute-pending-gpu-job-dru
+  [{:keys [task->scored-task user->sorted-running-task-ents user->dru-divisors] :as state}
+   pending-job-ent]
+  (let [user (:job/user pending-job-ent)
+        {gpu-req :gpus} (util/job-ent->resources pending-job-ent)
+        gpu-divisor (get-in user->dru-divisors [user :gpus])
+        pending-task-ent (util/create-task-ent pending-job-ent)
+        nearest-task-ent (some-> user->sorted-running-task-ents
```
wyegelwel: semantically `nearest-task-ent` is the running task with dru closest to but leq the pending task correct?

wyegelwel: how is <= defined for entities?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,7 +155,27 @@
...
+        nearest-task-dru (if nearest-task-ent
+                           (get-in task->scored-task [nearest-task-ent])
+                           0.0)
+        pending-job-dru (+ nearest-task-dru (/ gpu-req gpu-divisor))]
+    (histograms/update! pending-job-drus (dru-at-scale pending-job-dru))
+    (histograms/update! nearest-task-drus (dru-at-scale nearest-task-dru))
+
+    pending-job-dru))
+
+(defn compute-pending-normal-job-dru
```
wyegelwel: normal as opposed to gpu? That seems odd since you have a case for gpu in the function.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,7 +155,27 @@
...
   "Takes state and a pending job entity, returns the dru of the pending-job. In the case where the pending job causes user's dominant
```
wyegelwel: Please update the doc string to reference the casing of `normal` vs `gpu` jobs

As an aside, "normal" seems like an odd word here. @adereth @leifwalsh can we borrow your naming expertise? We need to split the space of jobs into "gpu" jobs and... not gpu jobs, meaning they only use cpu, mem and disk. Could you recommend a name for the not-gpu job case?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,7 +155,27 @@
   [dru]
   (* dru metrics-dru-scale))
 
-(defn compute-pending-job-dru
+(defn compute-pending-gpu-job-dru
```
wyegelwel: Add docstring for this function. It is probably worth putting a mention on why it exist separate from the function below.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -385,6 +385,14 @@
   com.netflix.fenzo.VirtualMachineLease
   (cpuCores [_] (or (offer-resource-scalar offer "cpus") 0.0))
   (diskMB [_] (or (offer-resource-scalar offer "disk") 0.0))
+  (getScalarValue [_ name] (or (double (offer-resource-scalar offer name)) 0.0))
+  (getScalarValues [_]
+    (reduce (fn [result resource]
+                 (if-let [value (:scalar resource)]
+                   (update-in result [(:name resource)] (fnil + 0.0 0.0) value)
```
wyegelwel: Just want to make sure I am reading this correctly. The second 0.0 is not necessary because `value` should never be `nil` right?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -412,12 +420,48 @@
...
+
+(defrecord TaskRequestAdapter [job task-info assigned-resources]
   com.netflix.fenzo.TaskRequest
   (getCPUs [_] (get-in task-info [:resources :cpus]))
   (getDisk [_] 0.0)
-  (getHardConstraints [_] [(novel-host-constraint job)])
+  (getHardConstraints [_] [(novel-host-constraint job) (gpu-host-constraint job)])
   (getId [_] (:task-id task-info))
+  (getScalarRequests [_]
+    (reduce (fn [result resource]
```
wyegelwel: Not sure if this is better. Up to you.

```clojure
(->> (:job/resource job)
        (map (juxt :resource/type :resource/amount))
        (filter (comp nil? second))
        (into {}))
```

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -568,37 +612,58 @@
...
-                                (filter (fn [job]
-                                          (util/job-allowed-to-start? db job)))
-                                (take num-considerable))
-              _ (log/debug "We'll consider scheduling" (count considerable) "of those pending jobs (limited to " num-considerable " due to backdown)")
-              matches (match-offer-to-schedule fenzo considerable offers db fid)
+              _ (log/debug "There are" (apply + (map count scheduler-contents-by)) "pending jobs")
+              _ (log/info "scheduler-contents:" scheduler-contents-by)
+              considerable-by (->> scheduler-contents-by
+                                   (map (fn [[category jobs]]
+                                          (log/info "here2" category jobs)
```
wyegelwel: Remove or make debug

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -967,10 +1023,56 @@
...
+(defn sort-jobs-by-dru
+  "Returns a map from job category to a list of job entities, ordered by dru"
+  [filtered-db unfiltered-db]
+  ;; This function does not use the filtered db when it is not necessary in order to get better performance
+  ;; The filtered db is not necessary when an entity could only arrive at a given state if it was already committed
+  ;; e.g. running jobs or when it is always considered committed e.g. shares
+  ;; The unfiltered db can also be used on pending job entities once the filtered db is used to limit
+  ;; to only those jobs that have been committed.
+  (let [pending-job-ents-by (group-by util/categorize-job (util/get-pending-job-ents filtered-db unfiltered-db))
+        pending-task-ents-by (reduce-kv (fn [m category pending-job-ents]
```
wyegelwel: Want to make sure I'm reading this correctly. Is this the same as:

```clojure
(->> pending-job-ents-by
       (map-vals (comp set (map util/create-task-ent))))
```

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -967,10 +1023,56 @@
...
+                                      user->dru-divisors)
+     :gpu (sort-gpu-jobs-by-dru (:gpu pending-task-ents-by)
+                                (:gpu running-task-ents-by)
+                                user->dru-divisors)}))
+
 (timers/deftimer [cook-mesos scheduler filter-offensive-jobs-duration])
 
 (defn is-offensive?
   [max-memory-mb max-cpus job]
+  (log/info "offensive?" job)
```
wyegelwel: Remove or make debug

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1041,14 +1143,18 @@
     (try
       (let [jobs (->> (sort-jobs-by-dru filtered-db unfiltered-db)
                       ;; Apply the offensive job filter first before taking.
-                      offensive-job-filter)]
-        (log/debug "Total number of pending jobs is:" (count jobs)
-                   "The first 20 pending jobs:" (take 20 jobs))
+                      (map (fn [[category jobs]]
```
wyegelwel: `map-vals` instead of `map` then `into`

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1099,6 +1204,10 @@
                   (log/info "Registered with mesos with framework-id " framework-id)
                   (reset! fid framework-id)
                   (set-framework-id framework-id)
+                  (when (and gpu-enabled? (not (re-matches #"1\.\d+\.\d+" (:version master-info))))
```
wyegelwel: They only give the version as a string? If they had a major, minor, micro struct you could check if major is > 0

scheduler/src/cook/mesos/task.clj
```diff
@@ -111,14 +112,14 @@
...
-    :mesos-messages []}
-   (-> task :resources vec)))
+    (fn [{:keys [remaining-resources mesos-messages]} [resource-keyword amount]]
+      (let [resource-name (name resource-keyword)
+            adjustment (take-resources remaining-resources resource-name amount)]
+        {:remaining-resources (assoc remaining-resources resource-name (:remaining-resources adjustment))
+         :mesos-messages (into mesos-messages (:mesos-messages adjustment))}))
+    {:remaining-resources resources
+     :mesos-messages []}
+    (.getScalarRequests ^com.netflix.fenzo.TaskRequest (:task-request task))))
```
wyegelwel: why not use `(-> task :resources vec)`

scheduler/docs/configuration.asc
```diff
@@ -98,6 +98,13 @@ We'll look at the configurable options in turn:
   This sets the role that Cook will connect to Mesos with. Default: `*`
   You can omit this property unless you've enabled security features in Mesos. The value should be in authorized list for the current `:principal` in `register_frameworks` Action in Mesos Authorization file. See http://mesos.apache.org/documentation/latest/authorization/ for details.
 
+`:enable-gpu-support`::
```
wyegelwel: Can you add a doc, not sure where is the best place describing the steps necessary to get gpus working with cook and mesos. I expect it will lean heavily on Kevin's gpu doc he posted recently I just want a single place to point people if they ask.

scheduler/docs/scheduler-rest-api.asc
dgrnbrg: We can't allocate fractional gpus _yet_, but that's on the roadmap eventually. So, it's a double.

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: it's a remnant of a bygone era

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: Correct. It's defined over the comparator, which is defined by `same-user-task-comparator` in `util.clj`, which is sorted first by highest priority, then earliest submission time, then tiebreakers.

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: I don't have a case for gpu comparison in this function.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: I wish--I observed cases in the wild, although I don't recall what triggered them. `value` can be `nil`. It may have had something to do with mesomatic doing funky stuff with resources, in that it always defines the `:scalar` key even when the `:type` is `:range` or something. I definitely needed that `fnil` to fix an NPE, though.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: That's a nice way too, but if you don't mind, I like the loop I wrote.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: Yes, but without allocating an intermediate set.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: It's only a string :/

scheduler/src/cook/mesos/task.clj
dgrnbrg: The `:task-request` is constructed differently, and doesn't automatically mirror all the resources that actually need to be debited from the offer. The Fenzo object does all that accounting.

scheduler/docs/configuration.asc
dgrnbrg: I'd prefer to just link to Kevin's GPU doc--there is literally nothing special about configuring GPUs for cook vs for any Mesos framework, since everything Cook-specific is auto-enabled with the option.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Figures. My concern is that gpu support will suddenly stop working when moving to 2.x.x mesos.

scheduler/src/cook/mesos/task.clj
wyegelwel: It *doesn't* mirror the resources need? Then why are we using it?

scheduler/docs/configuration.asc
wyegelwel: That is fine. I meant put a comment somewhere (probably here, maybe somewhere else is better) that references Kevin's doc and anything else (sounds like not) needed to get mesos+cook in a position where it will schedule gpus

wyegelwel: Which I see you did in the new changes. Ignore the previous comment

scheduler/docs/scheduler-rest-api.asc
wyegelwel: I think it is worth exposing this as an int until then so users don't get unexpected behavior

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Ah, you're right. I didn't notice the fancy github break thing.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: Understandable, but I suspect that's a long way away.

scheduler/src/cook/mesos/task.clj
dgrnbrg: I think that while many changes were made moving to new data formats, the people making the changes didn't think about how far reaching they might need to be in their adaptations. I think that the move to Mesomatic & understanding resources as collections of tagged resources (with multiple CPU resources, mem resources on a single machine, etc) should have also change the existing aggregation-based code where needed. Here's a place it doesn't seem to quite line up.

scheduler/docs/scheduler-rest-api.asc
dgrnbrg: The validator makes sure it's a whole number. But the internal Mesos type is a double, and it will change, so it's a whole number double.

wyegelwel: Yea, for now let's change it.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -155,10 +155,35 @@
...
+        gpu-divisor (get-in user->dru-divisors [user :gpus])
+        pending-task-ent (util/create-task-ent pending-job-ent)
+        nearest-task-ent (some-> user->sorted-running-task-ents
+                                 (get user)
+                                 (rsubseq <= pending-task-ent)
+                                 (first))
+        nearest-task-dru (if nearest-task-ent
+                           (get task->scored-task nearest-task-ent)
+                           0.0)
+        pending-job-dru (+ nearest-task-dru (/ gpu-req gpu-divisor))]
```
wyegelwel: What is the intuition for `(+ nearest-task-dru pending-dru)`?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -186,25 +211,39 @@
...
+                                                   [user (into (sorted-set-by (case category
+                                                                                :normal (util/same-user-task-comparator-penalize-backfill)
+                                                                                :gpu (util/same-user-task-comparator))) task-ents)]))
                                             (into {}))
-        task->scored-task (into (pm/priority-map-keyfn (comp - :dru))
-                                (dru/sorted-task-scored-task-pairs user->sorted-running-task-ents user->dru-divisors))]
-    (->State task->scored-task user->sorted-running-task-ents host->spare-resources user->dru-divisors)))
+        task->scored-task (into (pm/priority-map-keyfn (case category
+                                                         :normal (comp - :dru)
+                                                         :gpu -))
```
wyegelwel: The key is dru for gpu?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -299,12 +343,14 @@
 
 (defn rebalance
   "Takes a db, a list of pending job entities, a map of spare resources and params.
-   Returns a list of pending job entities to run and a list of task entities to preempt"
-  [db pending-job-ents host->spare-resources {:keys [max-preemption] :as params}]
+   Returns a list of pending job entities to run and a list of task entities to preempt
+
+   category is :normal or :gpu, depending on which type of job we're working with"
+  [db pending-job-ents host->spare-resources {:keys [max-preemption category] :as params}]
```
wyegelwel: I can't seem to find where you are calling rebalance with different categories (except in tests). What am I missing?

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Can you add a comment about it then. Anyone that looks at it will be confused, try to remove it and then get bit.

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: That is a great question! I copied this from the previous code, so I had to reverse engineer what's going on here.

The goal of this function is to compute the DRU of a pending job w/o redoing the whole DRU sorting process. So, we first get the `nearest-task-dru`. This is lookup into the sorted set of the user's tasks (since we're computing this on a per-user basis). We find the DRU of the highest priority task for that user that's adjacent to the new pending task's priority.

Next, we add the marginal DRU contribution of the pending task to the adjacent task, since remember, DRU is cumulative.

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: the `user->dru-divisors` map maps from usernames to `dru/ScoredTask`. It turns out that we don't need all the extra metadata for gpus, so we just map from usernames to DRU for GPUs.

scheduler/src/cook/mesos/rebalancer.clj
dgrnbrg: Good catch. I forgot to `assoc` the category into the 2 calls to the rebalancer in `start-rebalancer!`.

scheduler/src/cook/mesos/scheduler.clj
dgrnbrg: I'm not sure that will reduce the amount of boilerplate, since it's important to see which category's being filtered (for debugging prod issues), and I don't think that the complexity of the operation merits hiding it (since it's a pretty simple transform).

PR #188, opened 2016-09-23
WIP Custom Executor Spec

wkf:
This is a WIP spec for Cook's imminent custom executor.

### DIFF
--- a/executor/docs/spec.md
+++ b/executor/docs/spec.md
@@ -1,0 +1,276 @@
+# Cook Executor
+
+The document outlines a specification for Cook's custom executor.
+
+It is a modular design allowing users to add functionality by writing reusable, single-responsibility processes. Declarative configuration would determine when then processes would be executed. An HTTP API exposed by the executor itself would enable communication between the executor and the tasks it starts.
+
+
+## Scheduler HTTP API
+
+The following enhancements to the existing Cook scheduler API will:
+
+1. Allow users to specify multiple commands when creating jobs
+2. Allow users to query the status code or codes of a completed job
+
+### Launch a Job
+
+It should be possible to specify multiple commands when launching a job. Any command failure will fail the whole job. In the future, it may be desirable to have finer-grained control over how failure will be handled (for example, ignoring failures from asynchronous monitoring commands).
+
+Request body schema additions:
+
+| JSON Key | Type | Description |
+|----------|------|-------------|
+| commands | array of command objects | the commands to run, as well as information about how they are to be run |
+
+`command` object schema:
+
+| JSON Key | Type | Description |
+|----------|------|-------------|
+| command  | string | the command to run |
+| async | boolean | a flag determining whether the command is run asynchronously.  defaults to `false` |
+| guard | boolean | a flag determining whether failure of the command will stop execution of the whole job. defaults to `false` |
+
+Example `commands`:
+
+
+    commands: [
+      {
+        async: true,
+        command: “tail_logs.py”
+      },
+      {
+        guard: true,
+        command: “ensure_only_one_job.sh”
+      },
+      {
+        command: "actual_command.py"
+      },
+      {
+        command: "send_email.pl"
+      }
+    ]
+
+Most return codes won’t change, but since `command` and `commands` are mutually exclusive, there’s a new way to get a `400` response.
+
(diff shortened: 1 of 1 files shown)


executor/docs/spec.md
```diff
@@ -0,0 +1,276 @@
...
+
+Schema additions to support multiple commands per job:
+
+| Attribute | Type | Cardinality | Description |
+|-----------|------|-------------|-------------|
+| :job/command | ref | many | reference to a `command` entity |
+| :command/line | string | one | command line to run |
+| :command/code | double | one | exit code of command |
+| :command/params | ref | many | reference to a `command.param` entity |
+| :command.param/key | string | one | a command param key, like `async` |
```
wyegelwel: I would prefer to be explicit about these params (i.e. `:command/async`) since we are likely to only have a few of them and it is better from a datomic standpoint.

executor/docs/spec.md
```diff
@@ -0,0 +1,276 @@
...
+| JSON Key | Type | Description |
+|----------|------|-------------|
+| command  | string | the command to run |
+| async | boolean | a flag determining whether the command is run asynchronously.  defaults to `false` |
+| guard | boolean | a flag determining whether failure of the command will stop execution of the whole job. defaults to `false` |
+
+Example `commands`:
+
+
+    commands: [
```
wyegelwel: How is this different then bash "&&" and "||"?

wyegelwel:
How does the executor work in relation to running with docker?

wyegelwel:
How are we going to pass info from executor to cook scheduler? Framework messages? http requests directly to scheduler?

executor/docs/spec.md
wkf: It's not much different, but I think just using bash's `&&` and `||` would make it harder to capture exit codes for the individual processes. I could also imagine this being extended to (for example) restart asynchronous processes, which would not be easy to express with bash syntax.

wkf:
Re: Docker... if a docker container is specified for a job, the executor is launched in that container. This means the container should have python.

wkf:
Re: passing info from executor to the scheduler... right now, the only information we're sending back to the scheduler (progress, custom failure reason) can be put into the `data` field of the `TaskInfo` structure. This has the desirable property of keeping the executor completely decoupled from the scheduler.

wyegelwel:
> Re: Docker... if a docker container is specified for a job, the executor is launched in that container. This means the container should have python.

Does that mean that the executor must be included in the container and that the executor will be tied to the code running?

executor/docs/spec.md
wyegelwel: Why is this better than having "pre", "post", "parallel" and "command" commands so that "pre", "post" and "parallel" are add-ons to the current behavior?

Also, please consider the case when I have a script to set certain environment vars before the main command runs.

wkf:
> Does that mean that the executor must be included in the container

No, the agent should download it from a `uri` on the task before the Docker containerizer mounts it inside the container and runs it.

executor/docs/spec.md
wkf: > Why is this better than having "pre", "post", "parallel" and "command" commands so that "pre", "post" and "parallel" are add-ons to the current behavior?

Even with `pre`, `post`, and `parallel`, we'd still need the `guard` option (or equivalent). Having one `commands` structure seemed, and I believe makes retrieving exit statuses for all commands more straightforward. Beyond that, they are mostly equivalent.

> Also, please consider the case when I have a script to set certain environment vars before the main command runs.

I have given this some thought. There are a few ways I can think of: 

1. There could be an additional HTTP endpoint for manipulating env variables
2. There could be a `.env` file on disk, to which scripts would add lines like `NAME=value`.
3. There could be a special type of `wrapper` script that's passed the command. It could set any kind of env variables, but it would also be responsible for running the command.

For now, I'll plan on adding the API endpoint, but using a file would allow you to pull in env variables via a `uri`, which might be desirable.

wyegelwel: Can you include your current thinking on environment in the spec?

Have you thought about adding a way for an operator of cook to provide default (both mandatory and optional) commands to the list? This would allow things like:

1. Setting environment variables
2. default log4j processes

PR #189, opened 2016-09-28
Mf/rebalancer sim on travis

mforsyth:


scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -375,10 +375,35 @@
                               (apply max))]
     utilization))
 
+(defn read-datomic-params
+  [conn]
+  (-<>>
+   (d/pull (mt/db conn) ["*"] :rebalancer/config)
+   (dissoc <> ":db/id" ":db/ident")
+   (map (fn [[k v]]
+          [(keyword (name (keyword k))) v]))
```
wyegelwel: `(-> k keyword name keyword)` is to strip the `rebalancer/` from the keyword?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -375,10 +375,35 @@
...
+   (into {})))
+
+(defn update-datomic-params-from-config!
+  [conn config]
+  (let [recognized-params (select-keys config [:min-utilization-threshold
```
wyegelwel: It is probably worth putting these as a variable at the top or pulling them from the schema so that if we add another rebalancer setting it is picked up.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -375,10 +375,35 @@
...
+                                               :safe-dru-threshold
+                                               :min-dru-diff
+                                               :max-preemption])]
+    (if (not-empty recognized-params)
+      (do
+        (log/info "Updating rebalancer params to" recognized-params)
+        @(d/transact conn [(into {:db/id :rebalancer/config}
+                                 (zipmap (map #(keyword "rebalancer.config" (name %))
```
wyegelwel: Use `map-keys` https://github.com/plumatic/plumbing/blob/master/src/plumbing/core.cljx#L79

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: Yes.  FYI this is actually existing code.  It's just been extracted from a previously inline function.

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: 👍

scheduler/src/cook/mesos/rebalancer.clj
mforsyth: 👍 Much better thanks

simulator/config/two-sigma-settings.edn
```diff
@@ -0,0 +1,65 @@
+{:sim-db-uri "datomic:free://localhost:4334/cook-sim"
+ :cook-db-uri "datomic:free://localhost:4334/cook-jobs"
+ :cook-api-uri "http://localhost:12321"
+ :process-count 10
+ :sim-model {:label "Two Sigma's cluster usage"
```
wyegelwel: Can we scrub this of Two Sigma?

Just call it something like large-simulation. Change simulation running to video encoding (or something similar) and fcat to python users.

simulator/travis/prepare_simulation.sh
```diff
@@ -0,0 +1,12 @@
...
+set -ev
+
+lein deps
+
+cd ../scheduler
+lein deps
+lein uberjar
+
+cd ../simulator/travis
+unzip datomic-free-0.9.5394.zip
```
mforsyth: This doesn't include the metatransaction jar, but the jar is added to a spot which will be included in Datomic's classpath on the next line.

PR #193, opened 2016-10-06
scheduler: don't NullPointerException on nil values when rebalancing

mforsyth:


scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -325,8 +325,8 @@
                                                          (conj (:task aggregation) task)
                                                          (:task aggregation))
                                                  :gpus (+ (:gpus aggregation) (or gpus 0.0))
-                                                 :mem (+ (:mem aggregation) mem)
-                                                 :cpus (+ (:cpus aggregation) cpus)})
+                                                 :mem (+ (:mem aggregation) (or mem 0))
+                                                 :cpus (+ (:cpus aggregation) (or cpus 0))})
```
wyegelwel: Why are these ever nil?

mforsyth: @wyegelwel It appears that mem and cpus are nil whenever task is nil. You can see that in this function, there is some code to handle the case where task is nil, but mem and cpus are assumed not to be nil.

mforsyth: It appears that an "empty" entry is often at the beginning of the sequence passed to (compute-aggregations). 

Example:
({:dru 1.7976931348623157E308, :task nil, :mem nil, :cpus nil, :gpus nil} #cook.mesos.dru.ScoredTask{:task {:db/id 17592186045586, :job/_instance {:db/id 17592186045566, :job/resource #{{:db/id 17592186045568, :resource/type :resource.type/mem, :resource/amount 1768.0} {:db/id 17592186045567, :resource/type :resource.type/cpus, :resource/amount 5.0}}, :job/user "light5", :job/priority 29}, :instance/backfilled? false, :instance/start-time #inst "2016-10-10T13:30:27.157-00:00", :instance/hostname "172.17.0.8"}, :dru 1.2148903267449322E-305, :mem 1768.0, :cpus 5.0} #cook.mesos.dru.ScoredTask{:task {:db/id 17592186045587, :job/_instance {:db/id 17592186045554, :job/resource #{{:db/id 17592186045555, :resource/type :resource.type/cpus, :resource/amount 3.0} {:db/id 17592186045556, :resource/type :resource.type/mem, :resource/amount 778.0}}, :job/user "heavy5", :job/priority 10}, :instance/backfilled? false, :instance/start-time #inst "2016-10-10T13:30:27.158-00:00", :instance/hostname "172.17.0.8"}, :dru 1.0624727674371888E-305, :mem 778.0, :cpus 3.0})

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -232,8 +232,8 @@
                                                                                 :gpu (util/same-user-task-comparator))) task-ents)]))
                                             (into {}))
         task->scored-task (into (pm/priority-map-keyfn (case category
-                                                         :normal (comp - :dru)
-                                                         :gpu -))
+                                                         :normal (comp (fnil - 0) :dru)
```
mforsyth: Good question.  I only added this for consistency with the change to handle nil gpu's below it.  DRU should never be nil, I'll remove that check.

PR #195, opened 2016-10-06
[WIP] Custom executor

wkf:


executor/docs/spec.md
```diff
@@ -0,0 +1,286 @@
...
+## Scheduler Datomic API
+
+It will also be necessary to change the Datomic schema in order to in order to support users of Cook’s “Datomic API”.
+
+Schema additions to support multiple commands per job:
+
+| Attribute | Type | Cardinality | Description |
+|-----------|------|-------------|-------------|
+| :job/command | ref | many | reference to a `command` entity |
+| :instance/command | ref | many | reference to a `command` entity |
```
wyegelwel: Why does instance need a reference to command?

executor/docs/spec.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Scheduler Datomic API
+
+It will also be necessary to change the Datomic schema in order to in order to support users of Cook’s “Datomic API”.
+
+Schema additions to support multiple commands per job:
+
+| Attribute | Type | Cardinality | Description |
+|-----------|------|-------------|-------------|
+| :job/command | ref | many | reference to a `command` entity |
```
wyegelwel: Does this override the current `:job/command` which is a string?

wyegelwel: Yes, it is important to maintain backwards compatibility here. For example, if we need to rollback when deploying this we would end up in a bad state.

wyegelwel:
I'm generally concerned that these changes introduce both backwards incompatible changes as well as generally different behavior if a use switches from the single command to the command list. The latter is problematic if for example, an operator wants to set default commands to set the environment before the "real" command runs.

wkf:
@wyegelwel, understood, thanks for the feedback! I've pushed a change to the spec that should preserve backward compatibility and keep behavior more consistent.

Also, is there any more detail you can provide about operator use cases?

executor/docs/spec.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Launch a Job
+
+It should be possible to specify additional commands when launching a job. Additional commands should run either before, after, or during the primary command. Any failures (non-zero exit codes) will cause the entire job to fail. In the future, it may be desirable to have finer-grained control over how failure will be handled (for example, ignoring failures from asynchronous monitoring commands).
+
+Request body schema additions:
+
+| JSON Key | Type | Description |
+|----------|------|-------------|
+| before_commands | array of strings | the commands to run before the real command |
```
wyegelwel: Instead of an array of strings, make it an array of maps which on field is the command. That will allow in the future to add async, retries and timeouts to individual commands.

executor/cook/command.py
```diff
@@ -0,0 +1,83 @@
...
+
+    try:
+        while not (stop and stop.isSet()):
+            should_await = [c for c in rcommands if is_running(c) and is_sync(c)]
+            should_abort = [c for c in rcommands if is_crashed(c) and is_guard(c)]
+
+            if should_abort:
+                break
+            elif not should_await:
+                if len(commands) == len(rcommands):
```
wyegelwel: What am I missing.. 

If the last command is_running and async, then won't we end up exiting while it is still running?

executor/cook/executor.py
```diff
@@ -0,0 +1,142 @@
...
+                'reason': str,
+                'progress': float,
+                'env': {str: str}
+            }})
+        self.stop_event = Event()
+        self.server_thread = None
+        self.commands_thread = None
+
+    def registered(self, driver, executorInfo, frameworkInfo, slaveInfo):
+        print "CookExecutor registered"
```
wyegelwel: please use a logging library instead of prints.

Also, please use python 3 instead of 2.

executor/cook/executor.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if not self.server_thread:
+            def sync_task_status(action, type, id, entity):
```
wyegelwel: Where is this function used?

executor/cook/executor.py
```diff
@@ -0,0 +1,142 @@
...
+                                        async = False)
+                                for i, c in enumerate(payload['before_commands'] or [])]
+
+                during_commands = [Command(id = '.during.%i' % i,
+                                        value = c,
+                                        guard = False,
+                                        async = True)
+                                for i, c in enumerate(payload['during_commands'] or [])]
+
+                primary_commands = [Command(id = '',
```
wyegelwel: Primary command should be singular

wyegelwel:
General comments: 

Please add doc string at the type of python files describing what they are and why they exist (some are more obvious than others).

Please add doc strings to the functions

wyegelwel:
Operator use cases:

1. We know that certain environment vars need to be set for all jobs that are going to run, so we put a default pre job to set the env
2. We want to run a script at the end of a job to compress all output except stderr and stdout
3. We want to run a process while the command is running to push logs to elk

wyegelwel:
Another operator use case:

4. If the job fails, run some post analysis 


The general assumption is that there is a primary job and all other commands are doing *something* on behalf of that job.

executor/cook/command.py
wkf: You're not missing anything, that's the behavior in the spec. Long-running monitoring processes seems like the primary use-case for asynchronous commands, and I think gracefully stopping them when the "real" command finishes running is reasonable behavior.

executor/cook/executor.py
wkf: Good question. I had temporarily removed a call to `store.add_watch`. It'll be back when I next push.

wkf:
Thanks for the review, and for providing some operator use cases, I really appreciate it.

I'll add more doc strings, particularly to the `store`, as it is not immediately obvious why it exists.

executor/cook/executor.py
wkf: @wyegelwel I realized after starting to go from python 2 to 3 that the libmesos python bindings only work with python 2 (https://issues.apache.org/jira/browse/MESOS-355). If supporting python 3 is important, we'll have to start using the mesos agent HTTP API. What do you think?

executor/cook/executor.py
```diff
@@ -0,0 +1,105 @@
...
+from store import WATCH_ACTION_PUT
+
+def new_task_info(id, state, data = None):
+    """
+    Create a new TaskInfo dict.
+    """
+    return {
+        'state': state,
+        'task_id': {'value': id},
+        # FIXME: setting uuid manually is a workaround for a bug in pymesos
```
wyegelwel: Can you link to issue or pr so we can keep track of it?

executor/cook/__main__.py
```diff
@@ -0,0 +1,66 @@
...
+                'value': str,
+                'async': bool,
+                'guard': bool,
+            }],
+            'codes': [int]
+        }})
+
+    threads = [
+        Thread(target = run_server, args = (store, event)),
+        # TODO: rename this to driver in order to avoid confusion?
```
wyegelwel: I think it is ok

wkf:
This PR enables scheduling jobs that use the cook custom executor (in fact, that is the new default). 

I've been running the following to schedule jobs with before/after commands. You should be able to see the outputs of the separate `echo`s in the sandbox. Keep in mind that you'll currently need to update the `:executor-command` config option to point to the executor directory.

```bash
#!/usr/bin/env bash

cat <<EOF |
{
    "jobs": [
        {
            "uuid": "$(uuidgen)",
            "command": "echo hello",
            "name": "a-test-job",
            "priority": 0,
            "max_retries": 1,
            "max_runtime": 200,
            "cpus": 1,
            "mem": 512,
            "before_commands": [{"value": "echo before"}],
            "after_commands": [{"value": "echo after"}]
        }
    ]
}
EOF
curl -XPOST --data-binary @- -H 'Content-Type: application/json' 127.0.0.1:12321/rawscheduler
```

executor/docs/spec.md
```diff
@@ -0,0 +1,274 @@
...
+All tasks are killed. The processes for each task are sent a SIGTERM, followed in 10s by a SIGKILL. Each task status is updated to “killed”.
+
+`error` :
+
+No action taken, just logged.
+
+
+## Executor HTTP API
+
+There must be some API between the executor and the processes it start. At the very least, it must enable the ability for a task to report its progress to the executor. There are a number of potential API designs that could satisfy these constraints. An HTTP API on the executor itself seems like a good balance of complexity, since even though it makes the executor more complicated (by requiring an HTTP server), it provides a flexible interface that is simple to consume by the task. Some alternate approaches (along with their pros and cons) are outlined below.
```
mforsyth: "it start" - typo

scheduler/src/cook/components.clj
```diff
@@ -320,6 +321,8 @@
                                :memory-gb 12
                                :cpus 4}
                               task-constraints))
+     :executor-command (fnk [[:config [:scheduler {executor-command "echo NO EXECUTOR CONFIGURED"}]]]
```
mforsyth: :)

scheduler/dev-config.edn
```diff
@@ -22,4 +23,6 @@
        :levels {"datomic.db" :warn
                 "datomic.peer" :warn
                 "datomic.kv-cluster" :warn
+                "cook.mesos.task" :debug
+                "cook.mesos.util" :debug
```
wyegelwel: Remove this before we merge

scheduler/src/cook/mesos/api.clj
```diff
@@ -435,29 +486,27 @@
...
-                    base (if end
-                           (assoc base :end_time (.getTime end))
-                           base)
-                    base (if reason
-                           (assoc base
-                                  :reason_code (:reason/code reason)
-                                  :reason_string (:reason/string reason))
-                           base)]
-                base))
+                    code (:instance/code instance)
```
wyegelwel: Can we rename this (and other uses of it) from code to exit-code so there is no confusion what is being referred to?

executor/docs/spec.md
```diff
@@ -0,0 +1,276 @@
...
+### Launch a Job
+
+It should be possible to specify additional commands when launching a job. Additional commands should run either before, after, or during the primary command. Any failures (non-zero exit codes) will cause the entire job to fail. In the future, it may be desirable to have finer-grained control over how failure will be handled (for example, ignoring failures from asynchronous monitoring commands).
+
+Request body schema additions:
+
+| JSON Key        | Type                     | Description                              |
+| --------------- | ------------------------ | ---------------------------------------- |
+| before_commands | array of command objects | the commands to run before the real command |
+| after_commands  | array of command objects | the commands to run after the real command |
```
wyegelwel: Do the after commands run regardless of whether the real command succeeded? Is there a way to access the exit code?

wyegelwel:
Can you add another field to the instance in datomic for the working directory of the instance (MESOS_SANDBOX) and have the executor send it to cook? We currently need to query the mesos status endpoint on the agents to get this information and it causes a lot of problems

executor/docs/spec.md
wkf: Yes, after commands run regardless of whether or not the real command succeeds. `guard` can also be set for individual after commands, to stop execution on failure, but the default is to run all of them. The exit codes for the primary command, the before commands, and the after commands are all store in datomic, and I've made them available via the rawscheduler API as `exit-code` (primary exit code), `before-exit-codes` (array of exit codes for before commands), `after-exit-codes` (array of exit codes for after commands). The before/after arrays will include exit codes for any default commands set by the operator.

scheduler/src/cook/mesos/task.clj
```diff
@@ -1,50 +1,57 @@
...
 (defonce custom-executor-name "cook_agent_executor")
 (defonce custom-executor-source "cook_scheduler")
 
+(defn build-task-status-data [job-ent]
+  (.getBytes
+   (if (:job/custom-executor job-ent true)
+     ;;TODO this data is a race-condition
+     (pr-str
+      {:instance (str (count (:job/instance job-ent)))})
+     (json/generate-string
```
wyegelwel: Why do we send the commands when custom-executor is false?

wkf: To maintain backwards compatibility with existing clojure executor that expects a serialized clojure data structure.

scheduler/src/cook/mesos/task.clj
```diff
@@ -1,50 +1,57 @@
...
+  [db fid executor-command job-ent task-id]
   (let [resources (util/job-ent->resources job-ent)
         ;; If the custom-executor attr isn't set, we default to using a custom
         ;; executor in order to support jobs submitted before we added this field
         container (util/job-ent->container db job-ent)
         custom-executor (:job/custom-executor job-ent true)
         environment (util/job-ent->env job-ent)
         labels (util/job-ent->label job-ent)
-        command {:value (:job/command job-ent)
+        command {:value (if custom-executor
```
wyegelwel: Want to make sure I'm understanding this correctly:

If the custom executor is false, that used to mean we are using the command executor. Now it means we are using the new cook executor you built. Is that correct?

wkf: Yes, that's correct. The new executor replaces the command executor; it should be transparent to end users. This way we're able to maintain backward compatibility with existing clojure executor.

wyegelwel:
This is looking good. Would you mind adding two docs about the executor. 

One  from a users perspective. Basically explain what the executor is, and then show cases of when you would use it with examples of submission process and what to expect back.

One from an operators perspective, again what the executor is and show how to provide default commands. Finally, note failure cases so operators know what to expect.

scheduler/src/cook/components.clj
```diff
@@ -323,7 +323,7 @@
                                :memory-gb 12
                                :cpus 4}
                               task-constraints))
-     :executor-command (fnk [[:config [:scheduler {executor-command "echo NO EXECUTOR CONFIGURED"}]]]
+     :executor-command (fnk [[:config [:scheduler {executor-command "cook-executor"}]]]
```
wyegelwel: I'm confused, where is this being pulled from?

executor/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+#### Conditionally preventing command execution
+
+In cases where only one copy of a certain job should be running, before and after commands could be used to acquire and release a lock. Assuming you've written some scripts for managing your lock, you could execute them by passing the additional data to the cook scheduler's job creation API:
+
+```json
+{
+	"command": "my-actual-command",
+	"before_commands": [{
+		"value": "./acquire-zk-lock.sh my-service"
```
wyegelwel: Where does this script come from?

executor/README.md
```diff
@@ -0,0 +1,226 @@
...
+		"guard": true
+	}],
+	"after_commands":[{
+		"value": "./release-zk-lock.sh my-service"
+	}]
+}
+```
+
+The `guard` flag is used to prevent execution of later commands if we can't acquire the lock (in this case, `acquire-zk-lock.sh` script will return a non-zero exit code). Note that the scripts for manging the lock could be completely reusable, and that the primary command did not need to change.
```
wyegelwel: manging -> managing

executor/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+#### Monitoring a running command
+
+Perhaps you'd like to tail a log created by your primary command. You could use an asynchronous before command to so. It will be started before your primary command, and continue running in the background until the primary command (and all other synchronous commands) finish.
```
wyegelwel: "before command to so" -> "before command to do so"

wyegelwel: This means there command needs to be robust to main command not running yet.

executor/README.md
```diff
@@ -0,0 +1,226 @@
...
+```
+
+#### Sending an email on completion
+
+In order to send an email once your primary command has finished, you can use an after command:
+
+```json
+{
+	"after_commands": [{
+		"value": "mail -s 'your job failed, sorry' bob@hello.com"
```
wyegelwel: Do the after commands have access to exit codes of commands run before? i.e. if this script actually wanted to notify the user on failure, how would it know the command failed?

executor/README.md
```diff
@@ -0,0 +1,226 @@
...
+	}]
+}
+```
+
+For Operators
+---------
+
+The cook executor replaces the default command executor. It must be installed on each agent and requires python 3. As the executor communicates with Mesos agent via HTTP, the agent must support the HTTP API and have it enabled. Primarily, the operator will interact with the executor by setting default commands that run either before or after the commands specified by users when submitting a job to cook. For guidance on configuring these commands, see the configuration section below.
```
wyegelwel: "It must be installed on each agent" -- what happen to supporting it through a uri in task-info?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -301,7 +301,11 @@
                                              (when progress
                                                [[:db/add instance :instance/progress progress]])
                                              (when message
-                                               [[:db/add instance :instance/message message]])
+                                               (if (> (count message) executor-message-limit)
+                                                 (log/warn "Executor message too large for task" task-id "as instance" instance
```
wyegelwel: Truncate to the executor-message-limit.

Also, it might be worthwhile doing the check on the executor side (instead of or in addition to this check) so that we don't have infinity data going over the wire

executor/README.md
wkf: It must be provided by a user. The name was meant to convey an idea of what the script would do.

executor/README.md
wkf: That's correct.

executor/README.md
wkf: I ran into some problems with this approach after adding python 3 support and pulling in pymesos. However, I got it working again using `pyinstaller` to package the executor and its dependencies. I'll update the doc.

executor/cook/server.py
```diff
@@ -18,12 +18,20 @@ def do_GET(self):
...
+          => returns 200 and task data if task_id is valid
+          => returns 404 if task_id is unknown or invalid
         """
         if self.path is '/health':
-            self.send_response(200)
-            self.send_header('Content-type', 'application/json')
-            self.end_headers()
-            self.wfile.write(json.dumps('{"response": "okay"}'))
+            self.send_json_response(200, {'response': 'okay'})
+        elif re.match('\/task\/[^/]+', self.path):
```
wyegelwel: I think I might be missing something.

There should only exist 1 task data for the executor right? Or is this per command?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -185,35 +186,30 @@
...
-(defn munge-exit-codes
-  [job-ent codes]
-  (let [n (count (:job/before-command job-ent))
-        m (count (:job/after-command job-ent))]
-    {:before-exit-codes (take n codes)
-     :exit-code (->> codes (drop n) first)
-     :after-exit-codes (->> codes (drop n) rest (take m))}))
-
 (defn exit-code->tx
   [instance attr i c]
```
wyegelwel: please rename `c` to `exit-code`

executor/README.md
```diff
@@ -151,10 +169,20 @@ In order to send an email once your primary command has finished, you can use an
 }
 ```
 
+You could also take advantage of the read API to check the exit code of the primary command in order to only send an email on error. Here's an example script using `bash` and `jq`:
+
+```bash
+exit_code=$(curl $EXECUTOR_ENDPOINT | jq '.exit_code')
```
wyegelwel: Nice!

executor/README.md
```diff
@@ -166,45 +194,57 @@ Then, to install the executor executable (`cook-executor`) and all dependencies,
 ./setup.py install
 ```
 
+Alternatively, `pyinstaller` can be used to build a single executable that is distributed via a URI. For example, to build the executable, run the following command from the `executor` folder:
```
wyegelwel: I haven't heard of pyinstaller. Is it a well-known tool? 

Can you add a link to it in the doc?

executor/cook/server.py
wkf: No, you're not missing anything, there's only 1 task per executor, and each executor has its own API. I did it this way in case the API was ever moved into the scheduler, or centralized in some way. From the end user's perspective, it doesn't matter much, as the $EXECUTOR_ENDPOINT variable has the whole URL, including the task id.

wyegelwel:
This looks mostly ready to go. Can you rebase and I will do another pass.

executor/README.md
wyegelwel: How does that script get to that path to be called?

wkf: Most likely via a uri, but it could be placed on the machine some other way (with ansible for instance).

wyegelwel:
I realized for dev-ing this requires making sure the executor is on each host which may be a pain if you are using something like minimesos. 

Could you add script to create a virtualenv, add in the dependencies and the scripts, and zip it. We could then add the zip file to the jar as a resource and expose an endpoint in cook to get the it as a uri. This will making getting a dev cluster up easier and then just recommend people not use it for prod. Does that make sense?

To be clear, my goal is to make it no harder to set up a dev cluster than it is today, if you think of a better way than the virtualenv, please let me know. I was thinking virtualenv since it is standard tech.

executor/README.md
wyegelwel: Can you add the uri into the job spec so there is no magic?

wkf:
@wyegelwel that sounds very similar to the approach I took with pyinstaller, and for the same reason. There's some info on it in the operator doc, but basically it packages up the script and its dependencies into a single executable file, which you can pass around via a URI. I also added a way to set default URIs via the operator config file for the scheduler. Is there a way to serve static files from the scheduler now? If not, I think that's the only piece of the puzzle that's missing. Unless we could get away with using `file` URIs?

README.md
```diff
@@ -27,6 +27,17 @@ In this repository, you'll find several subprojects, each of which has its own d
...
 
+## Quickstart
+
+The quickest way to Mesos and Cook running locally is with [docker](https://www.docker.com/) and [minmesos](https://minimesos.org/). 
+
+1. Install `docker`
+2. Install `minimesos`
+3. Clone down this repo
+4. Run `make` from the project root to build all images
+5. Run `minimesos up` from the project root to start a Mesos and Cook
```
wyegelwel: When running this, I get:
```
Failed to run command 'up'. Condition with alias 'Mesos master not start responding' didn't complete within 60 seconds because condition com.containersol.minimesos.mesos.MesosMasterContainer$MesosClusterStateResponse was not fulfilled.
```

What version of minimesos are you using?

wyegelwel:
Looks like this code is failing the simulator test. Can you look into it?

https://travis-ci.org/twosigma/Cook/jobs/194609565

README.md
wkf: I'm using master.

wkf:
I'll look into it @wyegelwel.

executor/README.md
wyegelwel: Ping

executor/README.md
wyegelwel: Can you update the example to check for the exit code?

README.md
wyegelwel: Ok, figured out the problem. mapAgentSandboxVolume was causing it not to start up. When I set it to false everything (seemed) to work fine. Is there a strong reason you wanted it set to true?

minimesosFile
```diff
@@ -0,0 +1,75 @@
+minimesos {
+    clusterName = "cook"
+    loggingLevel = "INFO"
+    mapAgentSandboxVolume = true
```
wyegelwel: Is there a reason to set this to true? It was causing problems on my linux box when I was trying to follow this.

scheduler/dev-config.edn
```diff
@@ -1,6 +1,6 @@
 {:port 12321
  ;; We'll set the user to vagrant, since that's the default for many Vagrant-based Mesos setups
- :authorization {:one-user "vagrant"}
+ :authorization {:one-user "will"}
```
wyegelwel: Does this need to be "will"?

simulator/travis/minimesosFile
```diff
@@ -132,18 +132,18 @@ minimesos {
 
             ports {
                 role = "*"
-                value = "[31000-32000]"
+                value = "[31400-31500]"
             }
             ports {
-                role = "*"
-                value = "[32000-33000]"
+                role = "cook"
```
wyegelwel: Please leave at least one of these as "*" as I want to make sure nothing weird happens with roles.

wkf: Is the "*" on line 44 not sufficient? Do you want both of the "ports" blocks for this agent to apply to the wildcard role?

wyegelwel: Oh, my bad I missed that one.

executor/cook/__main__.py
```diff
@@ -0,0 +1,92 @@
...
+            'sandbox': str,
+            'progress': float,
+            'env': {str: str},
+            'command': str,
+            'before_commands': [{
+                'name': str,
+                'value': str,
+                'async': bool,
+                'guard': bool,
+                'default': bool
```
wyegelwel: If you haven't started working on this, don't worry about it. Let's get this merged and we can add it as a feature after.

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+You could also take advantage of the read API to check the exit code of the primary command in order to only send an email on error. Here's an example script using `bash` and `jq`:
+
+```bash
+exit_code=$(curl $EXECUTOR_ENDPOINT | jq '.exit_code')
+
+if [[ "$exit_code" != "0" ]]; then
+    mail -s 'your job failed, sorry!' bob@hello.com
+fi
+```
+
```
DiegoAlbertoTorres: Would be very useful to have an endpoint to kill the main process. For example, consider the case in which a before_command is a progress monitor, which is supposed to kill the main command when no progress is made for x amount of time. Then it could just hit such endpoint. @wyegelwel what do you think?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+```json
+{
+	"command": "my-actual-command",
+	"before_commands": [{
+		"value": "./acquire-zk-lock.sh my-service"
+		"guard": true
+	}],
+	"after_commands":[{
+		"value": "./release-zk-lock.sh my-service"
```
wkf: Yes, `after_commands` execute even if the main command fails. There's no way to set `guard` for the main command. This was to allow, for example, releasing a lock even if the main command  fails. Additionally, it's possible to have the `after_commands` check the exit code of the main command, in case you want to bail if the main command failed.

executor/README.md
wkf: Is this something you think should block merging this PR?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+  		"executable": true
+  	}]
+}
+```
+
+The `guard` flag is used to prevent execution of later commands if we can't acquire the lock (in this case, `acquire-zk-lock.sh` script will return a non-zero exit code). Note that the scripts for managing the lock could be completely reusable, and that the primary command did not need to change.
+
+#### Monitoring a running command
+
+Perhaps you'd like to tail a log created by your primary command. You could use an asynchronous before command to do so. It will be started before your primary command, and continue running in the background until the primary command (and all other synchronous commands) finish.
```
DiegoAlbertoTorres: Can we have a guarantee as to how the async commands will be killed? Such as, trying to stop with a SIGINT before using SIGKILL?

scheduler/src/cook/mesos/api.clj
```diff
@@ -148,22 +148,20 @@
...
+  "Schema for a command"
+  {:value s/Str
+   (s/optional-key :async?) s/Bool
+   (s/optional-key :guard?) s/Bool
+   (s/optional-key :default?) s/Bool})
+
+(def CommandRequest
+  "Schema for a command in request"
+  {:value s/Str
+   (s/optional-key :name) s/Str
```
wyegelwel: I can't find where the name is used

wkf: The name is generated on submit, combining the command order with the command suffix (before or after). It's used to create log files per command. The executor sends it back when updating task status, so it's included in the response from the scheduler API.

Here's where it's generated:
https://github.com/mojotech/Cook/blob/7c0c8162a136f032d37fe01b8686ee0bb7b9edce/scheduler/src/cook/mesos/util.clj#L168

executor/README.md
wkf: That is the current behavior: SIGINT, wait ~10 seconds, SIGKILL.

https://github.com/mojotech/Cook/blob/7c0c8162a136f032d37fe01b8686ee0bb7b9edce/executor/cook/launcher.py#L52

executor/docs/spec.md
```diff
@@ -0,0 +1,276 @@
+# Cook Executor
```
wkf: That's a good idea, though I don't think it should hold up a merge?

scheduler/src/cook/mesos/api.clj
wyegelwel: My confusion is that it seems a user could add a "name" to their command map but it wouldn't be used.

This is fine though because I think a worthwhile enhancement will be to allow naming the commands and have the stderr and stdout use the name instead of the order so it would be like stdout.upload_to_elk which is a lot more interpret-able than stdout.1

README.md
```diff
@@ -27,6 +27,17 @@ In this repository, you'll find several subprojects, each of which has its own d
 
 Please visit the `scheduler` subproject first to get started.
 
+## Quickstart
+
+The quickest way to Mesos and Cook running locally is with [docker](https://www.docker.com/) and [minmesos](https://minimesos.org/).
```
shamsimam: nitpick: typo in `minimesos`

executor/README.md
```diff
@@ -0,0 +1,270 @@
+Cook Executor
+=============
+
+The cook executor is a custom executor written in python. It replaces the default command executor in order to enable a number of features for both operators and end users.
```
shamsimam: Out of curiosity, why was Python chosen vs other languages such as Clojure/Java?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+| key | type | description |
+|-----|------|-------------|
+| `before_commands` | array | An array of `command`s to run before all user specified commands. |
+| `after_commands` | array | An array of `command`s to run after all user specified commands. |
+
+Commands have the following structure:
+
+| key | type | description |
+|-----|------|-------------|
+| `value` | string | A string representing the command to be run, ex: `echo hello`. |
```
shamsimam: Mitpick: `e.g.` instead of `ex`

wyegelwel:
Fast start up time, python more likely to be on a box. The custom executor
is fairly simple and largely managing processes so python seemed the best
fit.

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+=============
+
+The cook executor is a custom executor written in python. It replaces the default command executor in order to enable a number of features for both operators and end users.
+
+For Users
+-----
+
+The cook executor supports a number of (completely optional) features beyond the standard Mesos command executor. It allows users to specify additional commands that run either before or after a job's primary command. This enables the use of existing utilities, and promotes sharing and reuse. It also provides an HTTP API that allows commands to expose progress updates and custom failure messages, as well as update environment variables.
+
+### Scheduler HTTP API
```
shamsimam: Should this be `Executor HTTP API`? If so, should the endpoints below be `raw-executor`?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### `POST /rawscheduler`
```
shamsimam: Are the `before_commands` and `after_comamnds` job specific or do they apply to all jobs? If they apply to all jobs, should this endpoint be in the scheduler HTTP api section?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### `PATCH /task/:taskid`
+
+This request also requires a body. Since it is a `PATCH` request, a valid request may contain only some of the following keys:
+
+| key | type   | description                              |
+| -------- | ------ | ---------------------------------------- |
+| `progress` | float  | A number between 0 and 100, representing the progress of the task |
+| `message`   | string | A custom failure reason, expected to be accompanied by a non-zero exit code. |
+| `env`      | object | A map of env variable names to values, a null value will unset the variable. This is really only useful from a before command, since it will only affect commands that run afterward.  |
```
shamsimam: Are there restrictions on the length of the names or values of the environment variables?

executor/cook/server.py
```diff
@@ -0,0 +1,109 @@
...
+            task_id = self.path.split('/')[2]
+            task_data = self.server.store.get('task', task_id)
+
+            if task_data:
+                self.send_json_response(200, task_data)
+            else:
+                self.send_json_response(404, {'response': 'not found'})
+
+    def do_PATCH(self):
```
sradack: Shouldn't we have some checking as to what keys are being updated in this method?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+The job creation endpoint can be used to specify before and after commands, with the following keys:
+
+| key | type | description |
+|-----|------|-------------|
+| `before_commands` | array | An array of `command`s to run before all user specified commands. |
+| `after_commands` | array | An array of `command`s to run after all user specified commands. |
+
+Commands have the following structure:
+
+| key | type | description |
```
shamsimam: It would be nice to have an `id` for each command, that way the return codes can be returned in a map instead of an array.

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+| `async` | boolean | An optional flag. If set, the command runs asynchronously, in parallel with other future commands. It will be killed when all synchronous commands for a job are finished. |
+
+#### `GET /rawscheduler?job=:job`
+
+The job status endpoint can be used to view a job's progress, custom failure message, as well as the exit codes for all commands, by examining the following additional keys:
+
+| key | type    | description                              |
+| -------- | ------- | ---------------------------------------- |
+| `exit_code`     | integer | Status code when job specifies `command`  (will be absent if job has yet to finish). |
+| `before_exit_codes` | array of integers | Status codes for `before_commands`. |
```
shamsimam: Please see comment above regarding using an `id` for command. Then we can return a map instead of an array of exit codes.

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+| 503  | Service Unavailable | The request was valid, and received, but there was an (ostensibly temporary) failure with mesos itself. |
+
+#### `GET /task/:taskid`
+
+For successful requests, the response body could contain:
+
+| key | type | description |
+|-----|------|-------------|
+| `exit_code` | int | The exit code of the primary command. |
+| `before_codes` | array of ints | The exit codes of the before commands. |
```
shamsimam: Please see comment above regarding using an `id` for command. Then we can return a map instead of an array of exit codes.

executor/README.md
shamsimam: Do we have an example of preventing execution of after commands if the main command fails in the document?

executor/README.md
```diff
@@ -0,0 +1,270 @@
...
+| option | type | description |
+|--------|------|-------------|
+| `:command` | string | A string containing the command executed on the mesos agent to launch the cook executor. If the executor is installed using the instructions above, the default value of `cook-executor` should not need to be changed. |
+| `:log-level` | string | The log level for the executor process. Defaults to "INFO" |
+| `:max-message-length` | long | The maximum length for the custom failure message set by a task via the executor HTTP API. The default is 512. |
+| `:before-commands` | vector | An optional vector of `command`s to run before all user specified commands. |
+| `:after-commands` | vector | An optional vector of `command`s to run after all user specified commands. |
+| `:uris` | vector | An optional vector of `uri`s to be associated with every job |
+
+`command`s have the following structure:
```
shamsimam: Please refer to the comment above about associating an `id` with a command.

scheduler/src/cook/components.clj
```diff
@@ -64,26 +65,55 @@
     (require (symbol ns))
     (resolve var-sym)))
 
+(defn tell-jetty-about-usename [h]
```
sradack: sp? username

executor/cook/executor.py
```diff
@@ -0,0 +1,115 @@
...
+
+        def sync_task_status(action, type, id, entity):
+            if action is WATCH_ACTION_PUT and type is 'task':
+                state = entity.get('state')
+
+                entity.update({'sandbox': self.sandbox})
+
+                if state:
+                    if state == 'TASK_FAILED' or state == 'TASK_FINISHED':
+                        self.server.stop()
```
shamsimam: Nitpick: Should we add logs that server is being stopped as the task finished/failed?

executor/cook/executor.py
```diff
@@ -0,0 +1,115 @@
...
+            self.store.merge('task', task_id, {'state': 'TASK_STARTING'})
+        except Exception as e:
+            logging.exception("Exception in CookExecutor:launchTask")
+
+            driver.sendStatusUpdate(
+                new_task_info(task['task_id']['value'], 'TASK_FAILED'))
+            raise e
+
+    def killTask(self, driver, taskId):
+        logging.info("CookExecutor:killTask")
```
shamsimam: Nitpick: we should probably include the `taskId` in the log

executor/cook/executor.py
```diff
@@ -0,0 +1,115 @@
...
+        self.event.set()
+
+    def frameworkMessage(self, driver, message):
+        logging.info("CookExecutor:frameworkMessage")
```
shamsimam: Does this message need to be stored somewhere and published? At the very least it should be logged?

executor/cook/executor.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def shutdown(self, driver):
+        logging.info("CookExecutor:shutdown")
+        self.event.set()
+
+    def error(self, error, message):
+        logging.info("CookExecutor:error")
```
shamsimam: Should we include the message in the log?

executor/cook/executor.py
```diff
@@ -0,0 +1,115 @@
...
+
+def run_executor(store, stop, server, sandbox = ''):
+    """
+    Run an executor driver until the stop event is set.
+    """
+    driver = MesosExecutorDriver(CookExecutor(store, stop, server, sandbox))
+    driver.start()
+
+    # TODO: check the status of the driver and bail if it's crashed
+    while not stop.isSet():
```
shamsimam: Can we use `stop.wait()` instead to avoid polling?

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+        'RunningCommand', Command._fields + ('stdout', 'stderr', 'process'))
+):
+    """
+    An internal representation of a running command. Using a namedtuple helps catch
+    issues with data integrity.
+    """
+
+    def kill(self):
+        self.stdout.close()
+        self.stderr.close()
```
shamsimam: Should we close `stdout` and `stderr` after the process has been killed?

scheduler/src/cook/mesos/api.clj
```diff
@@ -686,32 +752,30 @@
...
-                                  :reason_code (:reason/code reason)
-                                  :reason_string (:reason/string reason))
-                           base)]
-                base))
+                    exit-code (:instance/exit-code instance)
+                    message (:instance/message instance)
+                    sandbox (:instance/sandbox instance)
+                    before-exit-codes (util/instance-ent->before-exit-codes instance)
+                    after-exit-codes (util/instance-ent->after-exit-codes instance)]
+                (cond-> {:task_id (:instance/task-id instance)
```
shamsimam: :+1:

shamsimam:
Completed my first pass.

executor/README.md
wkf: @shamsimam there's an example in the doc of checking the main command's return value from an after command, which is currently how you'd need to handle preventing execution.

executor/README.md
wkf: `before_commands` and `after_commands` provided here are job specific. It's possible to set before/after commands for all jobs, in the cook scheduler configuration file.

executor/README.md
wkf: You're right, that would make it easier to interpret return codes, especially when a cook operator has set default commands for every job.

executor/README.md
wkf: None are currently enforced.

executor/cook/server.py
wkf: The call to `store.merge` below will validate the keys/values against the 'task' schema described in `__main__.py`. If validation fails, we consider the request malformed and return a `400`.

scheduler/src/cook/components.clj
wkf: This commit just moves the function to a different place in the same file, but I agree that `usename` probably wasn't intentional.

executor/cook/executor.py
wkf: That seems more than reasonable.

executor/cook/executor.py
wkf: As of right now, the executor shouldn't be receiving any framework messages, and if it did, it wouldn't know what to do with it. Logging it could be useful for debugging though.

executor/cook/executor.py
wkf: That seems reasonable.

executor/cook/executor.py
wkf: We likely could, for now, but I think it might be better to check the status of the driver here, which I believe would require polling. What do you think?

shamsimam: Okay, it might help with debugging too with some additional logging

executor/cook/server.py
sradack: The documentation specifies that PATCH can update certain keys:

> Since it is a `PATCH` request, a valid request may contain only some of the following keys: ... progress ... message ... env

Where is it enforced that only those keys may be updated?

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+    exit if the stop event is set.
+
+    Returns a list of exit codes for the command processes.
+    """
+    did_abort = True
+    running_commands = []
+    iterator_commands = (Command(**c) for c in commands)
+
+    try:
+        while not (stop and stop.isSet()):
```
sradack: Exiting immediately results in not running an after-command that the user may be expecting to run, such as the lock-releasing script shown as an example in the docs.

sradack: Also it seems that having a lock-releasing script is incompatible with also using a command that sets the guard flag.

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+            should_await = [c for c in running_commands if c.is_running() and c.is_sync()]
+            should_abort = [c for c in running_commands if c.is_crashed() and c.is_guard()]
```
sradack: Should an async command be allowed to also set the guard flag?

sradack: Should a singleton async after-command be allowed?  Should async be allowed at all for after-commands?

executor/tests/test_launcher.py
```diff
@@ -0,0 +1,96 @@
...
+    A failing guard should prevent the execution of additional commands
+    """
+    codes = run_commands([
+        {'name': 0, 'value': '/bin/sh -c "exit 1"', 'guard': True},
+        {'name': 1, 'value': 'echo 1'}
+    ])
+
+    assert codes == [1]
+
+def test_run_commands_with_async():
```
sradack: Should we add some more unit tests testing more complex scenarios?  Multiple async, multiple guards, guards combined with asyncs, etc?

executor/tests/test_server.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_run_server():
+    event = Event()
+    store = FakeStore()
+    entity =  {'progress': 1.0}
+
+    server = CookExecutorHTTPServer(store, event)
+    server.start()
+
+    request('PATCH', '/task/123', entity)
```
sradack: Can we test updating the message and malformed updates?

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+    issues with data integrity.
+    """
+
+    def kill(self):
+        self.stdout.close()
+        self.stderr.close()
+
+        if self.process.returncode is None:
+            self.process.terminate()
+            for _ in range(0, 10):
```
sradack: Should 10 be configurable somewhere?

executor/README.md
```diff
@@ -0,0 +1,242 @@
...
+| `before_commands` | array | An array of `command`s to run before all user specified commands. |
+| `after_commands` | array | An array of `command`s to run after all user specified commands. |
+
+Commands have the following structure:
+
+| key | type | description |
+|-----|------|-------------|
+| `value` | string | A string representing the command to be run, ex: `echo hello`. |
+| `guard` | boolean | An optional flag. If set, and the command returns a non-zero exit code, no further commands for the job will be executed. |
+| `async` | boolean | An optional flag. If set, the command runs asynchronously, in parallel with other future commands. It will be killed when all synchronous commands for a job are finished. |
```
sradack: What guidance is given to clients as to how much memory that they should configure for their task given that the scheduler can specify async before-commands that co-exist in their process space?

executor/cook/launcher.py
sradack: Should guard be disallowed on the last item in the list of after/before commands?

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+    n = len(codes)
+    m = len(task.get('before_commands'))
+
+    d = {}
+
+    if n > m and codes[m] is not None:
+        d.update({'exit_code': codes[m]})
+
+    if n >= m:
+        d.update({'before_exit_codes': codes[0:m]})
```
sradack: Does `None` creep in as a value in the array here?  Do we care?

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+    An internal representation of a runnable command. Using a namedtuple helps catch
+    issues with data integrity.
+    """
+
+    def run(self, env):
+        if self.name is None:
+            stdout = open('stdout', 'w+')
+            stderr = open('stderr', 'w+')
+        else:
+            stdout = open('stdout.' + str(self.name), 'w+')
```
sradack: With name being optional, I can't seem to find where a value gets set if one isn't specified.  Maybe I missed it.

executor/cook/server.py
```diff
@@ -0,0 +1,110 @@
...
+        self.store = store
+        self.event = event
+
+    def start(self):
+        self.server = HTTPServer(('', self.port), CookExecutorHTTPRequestHandler)
+        self.server.store = self.store
+        self.server.timeout = 1
+
+        def run_server(server, event = None):
+            while not (event and event.isSet()):
```
sradack: Stopping immediately preventing any running commands from sending error/progress messages.

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+                else:
+                    running_commands.append(next(iterator_commands).run(get_env()))
+            else:
+                time.sleep(1)
+
+            update_task([c.poll() for c in running_commands], False, should_abort)
+    except Exception as e:
+        logging.exception("Exception in run_commands")
+    finally:
+        update_task([c.kill() for c in running_commands], True, did_abort)
```
sradack: Should async commands that are still running be first sent SIGTERM before we send SIGKILL.  The old documentation indicates that was the desired behavior.

executor/cook/launcher.py
```diff
@@ -0,0 +1,181 @@
...
+        while not stop.isSet() and not store.all('task'):
+            time.sleep(1)
+
+        if not stop.isSet():
+            id, task = store.all('task').popitem()
+
+            logging.info('Launching task %s', id)
+
+            def get_env():
+                return store.get('task', id).get('env', {})
```
sradack: Is the environment only set up to include what was explicitly specified in the task configuration?  Should we be setting standard env variables like HOME and USER?

scheduler/src/cook/mesos/api.clj
```diff
@@ -199,9 +199,37 @@
...
+            (s/optional-key :before-commands) [CommandRequest]
+            (s/optional-key :after-commands) [CommandRequest]
             :cpus PosNum
             :mem PosNum})))
 
+(def JobUuids
+  "Schema for one or more job uuids in params"
+  {:job [s/Uuid]})
+
+(def Instance
```
sradack: Do we have any sense on how these additions affect scheduler performance/query times?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -263,7 +300,21 @@
                            :instance.status/failed} instance-status)
                     [[:db/add instance :instance/end-time (now)]])
                   (when progress
-                    [[:db/add instance :instance/progress progress]])]))))
+                    [[:db/add instance :instance/progress progress]])
```
sradack: Do we update Datomic every time we receive a status update with progress?  How often can a task update the status?  How many status updates per unit time can we handle with the current architecture?

PR #197, opened 2016-10-19
WIP: Create initial job group spec

wyegelwel:


scheduler/docs/design-docs/job-groups.md
```diff
@@ -0,0 +1,162 @@
...
+
+###Open question: Should jobs be able to set a group id without explicitly creating it?
+
+####Choice: Yes
+
+This makes it easier for clients to group jobs when they don't intend to set any preferences about the group. 
+
+It also makes the failure cases (create a group, then client fails) a little more managable. 
+
+However, this means we either need to allow group preferences to be mutable since a client could first submit jobs then change preferences
```
DiegoAlbertoTorres: Allowing this might reinforce the paradigm of submitting then setting preferences. This can lead to problematic cases during the submission of large job-groups: The first few jobs in the sequence might manage to start running before the last job is submitted. After the last is submitted, job preferences are changed, potentially requiring the cancellation of the jobs that are running. To avoid this we would require job-groups to have a state indicating whether the group is fully submitted or not, adding a layer of complexity.

wyegelwel: Yea, I agree with you. Here is what I propose:

1. Job groups are immutable 
2. The group a job belongs to is immutable
3. If a job submits with a group id that hasn't been created yet, we create it with the default settings. 
4. If a group wants non-default settings the group must be explicitly created first.

DiegoAlbertoTorres: A similar problem could still occur under that spec: A job-group is submitted with a few jobs, some jobs are scheduled, then the user adds a few more jobs, potentially requiring cancellations under the job-groups rules. Seems like the solution would be to make even the members of job-groups immutable, and so requiring users to batch everything in a single request. But maybe that is too rigid?

wyegelwel: That is a good point and agree we should make the job-groups immutable (need to submit in single request).

Once we have the commit latch, we can have a workflow such as:

create commit latch
create group
submit batch 1
submit batch 2
...
submit batch N
Set the commit latch to committed

scheduler/docs/design-docs/job-groups.md
```diff
@@ -0,0 +1,162 @@
...
+
+and is more heavy weight for clients that just want to use the defaults and use groups for querying.
+
+ 
+
+###Open question: Should groups be part of a single batch submission (to /rawscheduler)?
+
+####Choice: Yes
+
+This reduces the number of requests a client needs to make (and reduces the failure surface) and allows for a simplier submission protocol in the case that a group of jobs spans only one submission
```
wyegelwel: Yea, it means the server has more to process and there is a max to the size of a post body.

DiegoAlbertoTorres:
For immutable job-group creation, we have 2 options now:
1. Use the /rawscheduler endpoint. To create job-group, a request can be made with multiple jobs. Each job can specify a job-group-uuid in its settings, and jobs with the same job-group-uuid would become members of a single new group, with all settings set to default.
2. Use a new job-group creation endpoint. Through a single request to this endpoint, users can set the group's settings and create all of its single jobs.

I propose getting rid of option 1 and only allow option 2 for job-group creation. Although option 1 would allow for a speedy transition for users, I think most users would eventually want to fine tune their job-groups' settings and will thus end up using option 2. Cutting down option 1 would allow us to have a single workflow for job-group submission from the start, which should make everything easier to maintain.

@maxrcho, what do you think?

maxrcho:
What does the transition look like if you eliminate option 1 for users?

wyegelwel:
I'm not sure I understand the advantage you mentioned above to adding the job-group creation endpoint (that lets you also assign child jobs) as opposed to having an endpoint to create just the group and a second endpoint to create the children. 

The workflows I imagine are:
1. A user submits 1 or more jobs without group info
2. A user submits 1 or more jobs with a group that they use just for querying (use the default settings)
3. A users creates a group with non-default settings and then submit 1 or more jobs using the created group
4. A user creates a group with non-default settings and begins a "transaction" to submit jobs over multiple batches. For each batch, they include the transaction id (commit-latch) and group. Once they have submitted all the batches, they complete the transaction (commit the latch).

In these workflows, if we keep the current job submission endpoint, they would all use that endpoint. If instead we had a non-grouped job endpoint and a grouped job endpoint (which is what I think you have proposed, please correct me if I'm wrong),  1 would be done differently than 2 and 3 and it isn't clear where 4 would fall under. Keeping the same endpoint for submission means there is no transition for current users and groups is an add on as opposed to a sharp break.

tl;dr I'm in favor of having a group endpoint that is just for setting up group metadata and a separate endpoint (rawscheduler) for submitting jobs.

DiegoAlbertoTorres:
@wyegelwel: How do you envision job grouping to work on the rawscheduler endpoint? Are we assuming a hit to the endpoint can only create 1 group? Or can multiple be created in a single request?

Also, I think it is strange to have different endpoints for what is basically an atomic operation (create a group and populate it):
1. Groups are never created and not populated. A request to create groups is always accompanied by a request to create jobs.
2. We would force all users of non-default job-groups into using commit-latches.
But also I am sure commit latches exist for a reason: Are we at risk of hitting the max limit of a POST if we limit all group creation to a single request?

wyegelwel:
@DiegoAlbertoTorres @maxrcho  Here is what I propose. The /rawscheduler endpoint takes a map with only one key at the moment, "jobs". We add a new key to that map "groups" in which users can define they groups they want created. We add a new field on job, group_uuid in the request. If that group_uuid exists in "groups" then just use that group. If not, then we create a group with default params for the user. The latter option is just a convenience for users. 

We also add a group endpoint to query group data (so we don't overload the rawscheduler TOO much). 

I want to give you something to think on, Spark when using cook submits an initial set of jobs but then occasionally will submit more for a few reasons. This pattern of growing and shrinking your compute dynamically will likely become more common. We should have some way of expanding the members of a group after it is initially created. Perhaps something like a "force" option to say I know what I'm doing.

DiegoAlbertoTorres:
Ok, and then we can let jobs have a group uuid that is not included in the same request IFF:
1. The group was created in a request that is part of the same commit
2. A 'force' flag was given
Does this sound good to? @wyegelwel @maxrcho

wyegelwel:
I think if either of those conditions holds, allow the job to be a member of the group.

DiegoAlbertoTorres:
Woops yeah, I meant that to be an IFF either conditions holds.

DiegoAlbertoTorres:
So far the schema for a Group is as defined in the spec. Do we want to add anything else? I think at least a user (although this could be taken from any of the jobs, assuming that all groups only have jobs from a single group) and a group name are worth considering. @wyegelwel @maxrcho please advise

wyegelwel:
@DiegoAlbertoTorres I think it is ok to grab the user from a job in the group. A group should only contain jobs from the same user. 
Sure to adding name, but please restrict it to say 128 chars.

@DiegoAlbertoTorres on the api I think a job should have a group_uuid field to refer to its group, but in the db, I think the group should have a list of jobs. The reason is at some point I expect we will want to have a hierarchy of groups and jobs and having the ownership being on the group entity simplifies that a bit.

PR #198, opened 2016-10-20
Mf/swagger

mforsyth:


scheduler/src/cook/mesos/api.clj
```diff
@@ -499,6 +533,49 @@
                           cheshire/generate-string)))
       ring.middleware.json/wrap-json-params))
 
+(def SwaggerApiDocs
```
wyegelwel: I would prefer to have the paths come from annotation on the liberator resources so that it is harder for them to fall out of sync. 

I've looked and there doesn't seem to be any automated (or partially automated) way to go from liberator to swagger (there really should be, they are documenting very similar things), however, any way to tie the path description closer to the actual resource will be welcome.

mforsyth: OK 👍 If there's not a library to do it, we can at least generate both from the same data ourselves.  I'll look into this.

PR #202, opened 2016-10-28
Serve API using Compojure-API, thus enabling Swagger docs

mforsyth:


scheduler/src/cook/mesos/api.clj
```diff
@@ -403,21 +452,16 @@
...
-                              [false {::jobs (mapv #(UUID/fromString %) jobs)}]
-                              (catch Exception e
-                                [true {::error e}])))
-                          [true {::error "must supply at least one job query param"}])
+  (-> (liberator/resource
+       :available-media-types ["application/json"]
+       :allowed-methods [:post :get :delete]
+       :malformed? (fn [ctx]
+                     (condp contains? (get-in ctx [:request :request-method])
+                       ;; schema will already validate.
```
wyegelwel: Why will it already validate?

wyegelwel:
This LGTM! I'm really excited about this.

scheduler/src/cook/mesos/api.clj
mforsyth: Before the request arrives at the "job-resource" handler, the request body will have been coerced to a specific schema.  Coercing to that schema already does everything that the inline validation here used to do (prove that the job parameters are an array of UUID's with at least 1 entry), and provides better error messages to boot.

mforsyth: (Basically, if the coercion fails, you'll get a 400 response with a JSON structure describing why the coercion failed).

PR #203, opened 2016-10-29
Support for Spark-2.0.1

prasincs:
Notable changes from the patch for 1.6.1:
- The method signatures for doRequestTotalExecutors,doKillExecutors
  changed to return Future, corresponding changes have also been changed
- Moved the calculateTotalmemory method to CoarseCookSchedulerBackend
  since it's not implemented in MesosBackend anymore

The naming scheme seems to have changed since version 1.6.1
and I haven't been able to actually execute jobs with it yet but I'm not
sure if that's because of the patch or I'm not starting Cook right.

### DIFF
--- a/spark/0001-Add-cook-support-for-spark-v2.0.1.patch
+++ b/spark/0001-Add-cook-support-for-spark-v2.0.1.patch
@@ -1,0 +1,686 @@
+From 395b510432893b4d83ec85dd8b5c6968655532d9 Mon Sep 17 00:00:00 2001
+From: Prasanna Gautam <prasannagautam@gmail.com>
+Date: Fri, 28 Oct 2016 14:45:28 -0700
+Subject: [PATCH] This updates Spark-2 to support Cook scheduler
+
+At high level, this keeps the behavior for cook support the same.
+You can connect to a cook cluster using
+cook://<username>@<cook host>
+similar to using mesos schema.
+
+Notable changes from the patch for 1.6.1:
+
+- The method signatures for doRequestTotalExecutors,doKillExecutors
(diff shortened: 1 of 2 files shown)


wyegelwel:
Hey @prasincs! How can I help to verify that things are working? One thing I recommend to rule out cook is submit a job, say "sleep 10" to the raw scheduler and make sure that gets scheduled.

leifwalsh:
Actually, development on this has moved to https://github.com/twosigma/spark, we haven't removed the patches here yet but we need to. We can create a branch of 2.0 and work on this there.

prasincs:
@wyegelwel @leifwalsh ah, thanks. We're fully on Spark 2+, so using Cook means adding this patch. Do I have to send a PR to the spark project to `branch-2.0`? What's a good way to setup a sane environment for testing? I didn't find a Vagrantfile in the projects to try on Vagrant for instance.

leifwalsh:
Yes, we need to move some things around and then we'll give you instructions in this thread. 

Have you signed our CLA? https://github.com/twosigma/Cook/blob/master/CONTRIBUTING.md

prasincs:
Yep, I signed the CLA and sent to the email address.

hkothari:
Hi @prasincs we've got 2.0.2-cook support at this branch: https://github.com/twosigma/spark/tree/2.0.2-cook If you need 2.0.1 you should be able to cherrypick the cook changes back to 2.0.1 without any issues and compile as you normally would with Spark.

This branch has a lot more changes outside of just the previous patch you probably saw to improve handling of when executors are lost, etc. If you run into any issues, feel free to create a PR to that branch.

Thanks!

PR #213, opened 2016-12-22
[WIP] Cook CLI and python client

wkf:
First pass on Cook CLI and python client. Looking for high level feedback; more documentation to follow.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -1,0 +1,447 @@
+# Cook Client
+This is the code for the Cook API client. There’s a command line client as well as a Python client.
+
+## CLI Client
+The Cook CLI client was designed with “end user ergonomics” in mind. Input can be piped via `stdin` and is sent to `stdout` in order to facilitate chaining commands together. Custom defaults may be provided for all commands. Multiple clusters are supported via configuration. Basic and key based authentication are supported out of the box,  it’s also possible to implement your own.
+
+### Installation
+
+To install the Cook CLI, clone this repo and from this folder run:
+
+```bash
+python3 setup.py install
+```
+
+This will install the `cook` command on your system.
+
+### Configuration
+
+In order to use the Cook CLI, you’ll need a configuration file. `cook` looks first for a `.cook.json` file in the current directory, and then for a `cook.json` file in your home directory. This file may also be provided manually via the command line with the `—config` option.
+
+The file looks something like this:
+
+```json
+{
+  "defaults":{
+    "cluster": "dev0",
+    "create-job": {
+      "mem": 128,
+      "cpus": 1
+    }
+  },
+  "clusters": [
+    {
+      "name": "dev0",
+      "url": "http://127.0.0.1:12321/swagger-docs"
+    },
+    {
+      "name": "dev1",
+      "url": "http://127.0.0.1:12322/swagger-docs",
+      "http_client": {
+        "type": "basic_auth",
+        "params": {
+          "host": "127.0.0.1:12322",
+          "username": "u",
+          "password": "p"
+        }
+      }
+    }
+  ]
+}
+```
+
+The `defaults` map contains default values that are passed each time a command is run. For instance, the `create-job` map above is used to set the default values for `mem` and `cpus` when calling the `create-job` command. These can always be overridden directly from the command line. The default `cluster` is used to determine which cluster (specified in the `clusters` array) `cook` connects to by default. If no default is provided, all available clusters are tried in order.
+
+Each entry in the `clusters` array conforms to a cluster specification (“spec”). A cluster spec requires a name and a url pointing to the swagger spec for a Cook scheduler. Additionally, an optional `http_client` can be provided in order to add custom behavior, like authentication. If the `http_client` type takes the form of `module:function` (note the colon), the module is loaded and the function is called with the provided params. The function should return an `http_client`, and can be used (for instance) to provide custom authentication mechanisms.
+
+### Commands
+
+The fastest way to learn more about `cook` is with the `-h` (or `—help`) option.
+
+```
+usage: cook [-h] [--cluster CLUSTER] [--config CONFIG] [--dry-run] [--no-dry-run]
+                   {create-job,create-group,kill-job,kill-instance,retry-job,await-job,await-instance,query-job,query-instance,query-group}
+                   ...
+
+positional arguments:
+  {create-job,create-group,kill-job,kill-instance,retry-job,await-job,await-instance,query-job,query-instance,query-group}
+    create-job          create job for command
+    create-group        create group
+    kill-job            kill job related to uuid
+    kill-instance       kill instance related to uuid
+    retry-job           retry job related to uuid
+    await-job           await job related to uuid
+    await-instance      await instance related to uuid
+    query-job           query job related to uuid
+    query-instance      query instance related to uuid
+    query-group         query group related to uuid
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --cluster CLUSTER, -c CLUSTER
+                        cluster to use
+  --config CONFIG, -C CONFIG
+                        config file to use
+  --dry-run, -d         print action and arguments without submitting
+  --no-dry-run
+
+```
+
+All global options (`—cluster`, `—config`, `—dry-run`) can be provided when using subcommands. `—dry-run` is useful for testing commands without actually attempting to submit.
+
+In cases where a sub-command can take multiple positional arguments, it’s possible to pass a single dash (`-`) instead. This will cause `cook` to read input from `stdin`, with each line corresponding to one positional argument. In this way, you can pipe the output from one command into the input of another. It also makes it possible to operate on multiple entities at once. For instance, you could start multiple jobs by piping one command per line to `cook create-job`.
+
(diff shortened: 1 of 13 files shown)


client/cook/__main__.py
```diff
@@ -0,0 +1,225 @@
...
+import importlib
+
+from cook.client import CookMultiClient, basic_auth_http_client, key_auth_http_client
+
+DEFAULT_CONFIG_PATHS = [
+    ".cook.json",
+    os.path.expanduser("~/.cook.json")
+]
+
+def generate_job_uuid():
```
wyegelwel: Please use numpy style doc strings https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

wyegelwel:
Please create job and instance classes in the cook scheduler api that will re-pull data when needed (i.e. a call to get status will need to re-query, a request for cpus wouldn't).The goal is to make the api a little more user friendly. You can leave the more functional version of the api for users that need finer grained control of when requests are sent.

wyegelwel:
Are there features in click that you specifically needed to build the api or are you just most comfortable in it?

I should have mentioned before, we generally choose to use argparse. If there is no specific feature that the api relies on, please switch to use argparse for the sake of conformity.

client/cook/client.py
```diff
@@ -0,0 +1,201 @@
...
+        return self.poll_jobs([uuid])[0]
+
+    def poll_jobs(self, uuids):
+        _, result = self.client.rawscheduler.get_rawscheduler(
+            job=uuids
+        ).result()
+
+        return result.json()
+
+    def await_job(self, uuid, t=30):
```
wyegelwel: Use datetime timedeltas (instead of t) to determine when to break.

wyegelwel: Add a parameter for how frequently to poll. For many jobs, 1 second is probably way too frequent, 30 seconds is probably a good default.

client/cook/client.py
```diff
@@ -0,0 +1,201 @@
...
+    def __init__(self, specs):
+        if specs:
+            self.specs = [hashabledict(s) for s in specs]
+        else:
+            raise ValueError('At least one client spec must be provided.')
+
+    def __getattr__(self, name):
+        return functools.partial(self.request, name)
+
+    @functools.lru_cache(maxsize=None)
```
wyegelwel: `maxsize=None` makes me nervous. Please put a bound on it, even if it is large.

client/cook/client.py
```diff
@@ -0,0 +1,201 @@
...
+    def request(self, name, *args):
+        for s in self.specs:
+            client = self.client(s)
+
+            if client:
+                try:
+                    return getattr(client, name)(*args)
+                # TODO: be more specific?
+                except Exception as e:
+                    print(e)
```
wyegelwel: Please use logging instead of printing.

client/example-cook.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "clusters": [
+        {
+            "name": "dev",
+            "url": "http://127.0.0.1:12321/swagger-docs"
+        },
+        {
+            "name": "fake-dev0",
+            "url": "http://127.0.0.1:12322/swagger-docs",
+            "http_client": ["basic_auth", "127.0.0.1:12322", "u", "p"]
```
wyegelwel: Please use a map so that the values are self documenting. i.e. I assume the values in the list are [auth-type, url, user, password] but I'm not sure. Having it be:
```
{"auth" : {"type" : "basic_auth",
               "params" : {"user" : "u", "password" : "p"}}
 "url" : "127.0.0.1:12322",
```

wkf: I mentioned this in more detail in the basecamp post, but it's not `[auth-type, url, user, password]`, it's `[function-that-returns-an-http-client, *args]`. I can still make it a map, but it's not just for `auth`.

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+All global options (`—cluster`, `—config`, `—dry-run`) can be provided when using subcommands. `—dry-run` is useful for testing commands without actually attempting to submit.
+
+In cases where a sub-command can take multiple positional arguments, it’s possible to pass a single dash (`-`) instead. This will cause `cook` to read input from `stdin`, with each line corresponding to one positional argument. In this way, you can pipe the output from one command into the input of another. It also makes it possible to operate on multiple entities at once. For instance, you could start multiple jobs by piping one command per line to `cook create-job`.
+
+All options can also be provided from a file by prefixing the file path with an `@` sign. For instance, to start a job with a command read from a file: `cook create-job @command.txt`
+
+#### `create-job`
+
+You can create one or more jobs with `create-job`. The command is the only required option, and can be provided as a positional argument, via `stdin`, or from a file. Jobs can also be passed as “raw” JSON data by setting the `—raw` flag. This command returns UUIDs, one per line, for each created job.
```
wyegelwel: What cluster does this submit to?

wkf: It'll submit to the default `cluster`... if you were using something like the config file toward the top of the README, it would be the `dev0` cluster.

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+  --raw, -r             raw job spec in json format
+  --no-raw
+```
+
+#### `create-group`
+
+You can create one or more groups with `create-group`. Groups can also be passed as “raw” JSON data as positional arguments, via `stdin`, or from a file. You can also pass multiple params (`—placement-param`, `—straggler-param`) as name value pairs, like so: `—placement-param name0=value0 —placement-param name1=value1`. This command returns UUIDs, one per line, for each created group.
+
+```
+usage: cook create-job [-h] [--uuid UUID] [--name NAME] [--priority PRIORITY]
```
sradack: Why is the `create-job` usage being shown here again?

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+  --max-retries MAX_RETRIES
+                        maximum retries for job
+  --max-runtime MAX_RUNTIME
+                        maximum runtime for job
+  --cpus CPUS           cpus to reserve for job
+  --mem MEM             memory to reserve for job
+  --raw, -r             raw job spec in json format
+  --no-raw
+ λ  python3 -m cook create-group -h
+usage: __main__.py create-group [-h] [--uuid UUID] [--name NAME] [--placement-type PLACEMENT_TYPE]
```
dposada: `cook` instead of `__main__.py`?

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+```
+usage: cook query-group [-h] [--detailed] [--no-detailed] uuid [uuid ...]
+
+positional arguments:
+  uuid
+
+optional arguments:
+  -h, --help      show this help message and exit
+  --detailed, -r  show additional group statistics
+  --no-detailed
```
dposada: Should we add a description for `--no-detailed`?

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+  --config CONFIG, -C CONFIG
+                        config file to use
+  --dry-run, -d         print action and arguments without submitting
+  --no-dry-run
+
+```
+
+All global options (`—cluster`, `—config`, `—dry-run`) can be provided when using subcommands. `—dry-run` is useful for testing commands without actually attempting to submit.
+
+In cases where a sub-command can take multiple positional arguments, it’s possible to pass a single dash (`-`) instead. This will cause `cook` to read input from `stdin`, with each line corresponding to one positional argument. In this way, you can pipe the output from one command into the input of another. It also makes it possible to operate on multiple entities at once. For instance, you could start multiple jobs by piping one command per line to `cook create-job`.
```
dposada: Is it possible to accomplish this without the dash (e.g. `printf 'echo 1' | cook create-job`)?

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+```
+
+## Python Client
+The Cook python client is really two different interfaces. One is a low-level client (`CookClient`, and the `CookMultiClient` wrapper), and the other is a set of convenience classes for working with Cook’s primary entities: `Group`, `Job`, and `Instance`.
+
+### Classes
+
+#### `CookClient`
+
+The `CookClient` class is a low-level interface to the cook API. It’s constructor accepts a swagger spec url:
```
dposada: "Its" instead of "It's"

client/README.md
```diff
@@ -0,0 +1,447 @@
...
+          "username": "u",
+          "password": "p"
+        }
+      }
+    }
+  ]
+}
+```
+
+The `defaults` map contains default values that are passed each time a command is run. For instance, the `create-job` map above is used to set the default values for `mem` and `cpus` when calling the `create-job` command. These can always be overridden directly from the command line. The default `cluster` is used to determine which cluster (specified in the `clusters` array) `cook` connects to by default. If no default is provided, all available clusters are tried in order.
```
dposada: @wyegelwel Do you know if this is the behavior we want?

tnn1t1s:
is this PR still in progress? can we prioritize getting a first pass merged then pushing to another Cook dev to complete style & docs comments from DP.

dposada:
I'm working on getting this functioning and merged

dposada:
Please ignore the "approved these changes" - wrong PR

dposada:
Moved these changes into PR #357 and others to come

PR #220, opened 2017-01-05
Straggler handling

wyegelwel:
Add concept of straggler handling to groups. 

Clean up processing in api

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -709,8 +709,7 @@
                 base))
             (:job/instance job))}
       (when groups
-        {:groups (map #(str (:group/uuid %)) groups)})
-      )))
+        {:groups (map #(str (:group/uuid %)) groups)}))))
 
 (defn fetch-group-job-details
   [db guuid]
@@ -724,40 +723,14 @@
 (defn fetch-group-map
   [db guuid]
   (let [group (d/entity db [:group/uuid guuid])
-        hp (:group/host-placement group)
-        hp-params (:host-placement/parameters hp)
-        hp-type (->> hp :host-placement/type :host-placement.type/name)
-        jobs (:group/job group)
-        straggler-handling (:group/straggler-handling group)]
-    {:uuid (:group/uuid group)
-     :name (:group/name group)
-     ;; TODO: strip everything before slash in parameters so we can just do:
-     ;; {:type hp-type
-     ;;  :parameters (map-keys remove-datomic-namepace hp-params
-     ;; See cook.mesos.util/deep-transduce-kv
-     :host-placement {:type (util/without-ns hp-type)
-                      :parameters (merge {}
-                                         (when (= hp-type :host-placement.type/attribute-equals)
-                                           {:attribute (:host-placement.attribute-equals/attribute hp-params)}))}
-     :straggler-handling {:type (util/without-ns (:straggler-handling/type straggler-handling))
-                          :parameters (map-keys util/without-ns 
-                                                (:straggler-handling/parameters straggler-handling))}
-     :jobs (->> jobs
-                (map :job/uuid))}))
-
-#_(defn fetch-group-map
-  [db guuid]
-  (let [group (d/entity db [:group/uuid guuid])
         jobs (:group/job group)]
-    (println (d/touch group))
-    (println (merge
-               (util/remove-datomic-namespacing (dissoc (into {} group) :group/job))
-               {:jobs (->> jobs
-                           (map :job/uuid))}))
-    (merge
-      (util/remove-datomic-namespacing (dissoc (into {} group) :group/job))
-      {:jobs (->> jobs
-                  (map :job/uuid))})))
+    (-> (into {} group)
+        ;; Remove job because we don't want to walk entire job list
+        (dissoc :group/job)
+        util/remove-datomic-namespacing
+        (assoc :jobs (map :job/uuid jobs))
+        (update-in [:host-placement :type] (comp util/without-ns :name))
+        (update-in [:straggler-handling :type] util/without-ns))))
 
 (defn instance-uuid->job-uuid
   "Queries for the job uuid from an instance uuid.
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -460,7 +460,7 @@
   [straggler-handling :- [StragglerHandling]]
   (merge {:straggler-handling/type (keyword "straggler-handling.type"
                                             (name (:type straggler-handling)))}
-         (when (:parameters straggler-handling)
+         (when (seq (:parameters straggler-handling))
            {:straggler-handling/parameters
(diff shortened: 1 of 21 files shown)


wyegelwel:
Things left to do:

1. Add straggler handler to job client
2. Add straggler agent to simulation tests
3. Tests on a cluster
4. Clean up fetch-group-map in api.clj
5. Document changes

wyegelwel:
@DiegoAlbertoTorres Can you take a look at this when you get a chance? I think it is good to go

scheduler/src/cook/mesos/schema.clj
```diff
@@ -558,6 +591,11 @@
     :db/ident :instance/create
     :db/doc "Creates an instance for a job"}])
 
+(def straggler-handling-types 
+  (->> schema-attributes
+       (map :db/ident)
+       (filter #(= "straggler-handling.type" (namespace %)))))
+
```
DiegoAlbertoTorres: Can we keep this consistent with how hp-types are installed in the database?

scheduler/src/cook/mesos/util.clj
```diff
@@ -44,24 +46,70 @@
...
+                x))
+            m))
+
+(defn remove-datomic-namespacing
+  "Takes a map from datomic (pull) and removes the namespace
+   as well as :db/id keys"
+  [datomic-map]
+  (->> datomic-map
+       (deep-transduce-kv (comp
+                            (filter (comp (partial not= :db/id) first))
```
DiegoAlbertoTorres: You said dissoc does not work?

scheduler/test/cook/test/mesos/group.clj
```diff
@@ -0,0 +1,74 @@
...
+            _ (create-dummy-instance conn job-a :instance-status :instance.status/success
+                                     :start-time (tc/to-date (t/ago (t/hours 3)))
+                                     :end-time (tc/to-date (t/ago (t/hours 2))))
+            job-b (create-dummy-job conn :group group-ent-id)
+            straggler (create-dummy-instance conn job-b :instance-status :instance.status/running
+                                             :start-time (tc/to-date (t/ago (t/minutes 190))))
+            db (d/db conn)
+            group-ent (d/entity db group-ent-id)
+            straggler-task (d/entity db straggler)]
+        (is (= [straggler-task](group/find-stragglers group-ent)))))
```
DiegoAlbertoTorres: Can we have a test with more jobs?

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -59,6 +61,34 @@
             (concat (take 100 (filter odd? (range)))
                     (drop 100 (filter even? (range)))))))
 
+(deftest test-task-run-time
+  (let [uri "datomic:mem://test-task-run-time"
+        conn (restore-fresh-database! uri)]
+    (testing "task complete"
+      (let [expected-run-time (t/hours 3)
+            start (t/ago expected-run-time)
+            end (t/now)
```
DiegoAlbertoTorres: Some milliseconds may occur between the calls to t/ago and t/now, which can throw off this test randomly

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -59,6 +61,34 @@
...
+    (testing "task running"
+      (let [expected-run-time (t/hours 3)
+            start (t/ago expected-run-time)
+            job (create-dummy-job conn :user "tsram" :job-state :job.state/completed :retry-count 3)
+            task-entid (create-dummy-instance conn 
+                                              job 
+                                              :instance-status :instance.status/running
+                                              :start-time (tc/to-date start))
+            task-ent (d/entity (d/db conn) task-entid)]
+        (is (= (t/in-seconds expected-run-time)
```
DiegoAlbertoTorres: Same rounding problem as above?

scheduler/src/cook/mesos/api.clj
```diff
@@ -424,11 +473,14 @@
...
-                    :group/host-placement hp-txn-id
-                    :group/job job-db-ids}]
-      (conj hp-txns group-txn)))
+                   :group/uuid uuid
+                   :group/name (:name group)
+                   :group/host-placement hp-txn-id
+                   :group/job job-db-ids
+                   :group/straggler-handling (-> group
+                                                 :straggler-handling
+                                                 (make-straggler-handling-txn))}]
```
wyegelwel: > (Didn't realize you could inline them like that!)

This requires making the ref a component (isComponent in the schema)

scheduler/src/cook/mesos/schema.clj
wyegelwel: I think we should switch hp-types to match straggler-handling. The ident as enum is a pretty standard pattern in datomic.

scheduler/test/cook/test/mesos/util.clj
wyegelwel: Yea, but it is ok, because in-seconds takes the floor on milliseconds

scheduler/src/cook/mesos/util.clj
wyegelwel: dissoc isn't a transducer

scheduler/test/cook/test/mesos/group.clj
wyegelwel: Added another job to this test.

Is there something specific you want to test for?

scheduler/src/cook/mesos/util.clj
DiegoAlbertoTorres: gotcha

scheduler/src/cook/mesos/api.clj
DiegoAlbertoTorres: Ah, interesting. I just read about components, seem like HostPlacement should be component entities. Should this change be added to this PR?

wyegelwel: It already is ;)

scheduler/src/cook/mesos/schema.clj
DiegoAlbertoTorres: Sounds good. Should we do it in this PR?

scheduler/test/cook/test/mesos/group.clj
DiegoAlbertoTorres: Sorry, I meant the 'stragglers found' test. I feel like having a test with only two jobs doesn't really test the 'quantile' part of detection. I would rest easier if the test (or an additional test) worked with a quantile-index that is not in the extremes of the list of jobs, for example, quantile 0.75 in a set of 10 jobs.

scheduler/test/cook/test/mesos/util.clj
DiegoAlbertoTorres: I see. Sounds reasonable.

scheduler/src/cook/mesos/api.clj
DiegoAlbertoTorres: Sorry, gitlab vs github CR confuses me.

PR #228, opened 2017-01-20
scheduler:  API supports retrying multiple jobs with one request.

mforsyth:


scheduler/src/cook/mesos/api.clj
```diff
@@ -1015,39 +1015,72 @@
...
 
 (defn display-retries
   [conn ctx]
-  (:job/max-retries (d/entity (db conn) [:job/uuid (::job ctx)])))
+  (:job/max-retries (d/entity (db conn) [:job/uuid (-> ctx ::jobs first)])))
 
-(defn check-job-exists
+(defn jobs-from-request
+  [ctx]
+  (let [single-job (or (get-in ctx [:request :query-params :job])
```
wyegelwel: Can we make this in line with how rawscheduler handles multiple jobs? I think it has a single "job" field that may be set multiple times.

mforsyth: @wyegelwel you're referring to the repeated job query parameters when requesting the status of multiple jobs.   That is one way of obtaining an array-like data structure from query parameters "&job=x&job=y".  It's not a strategy that's really applicable to this here, because we're talking about the JSON payload of a POST request.   You can't repeat the same key in a JSON hash... e.g {"foo": 1, "foo": 2} isn't valid JSON.  

So, there were a few choices:

1. make it so that "job" parameter can be either a string or an array, either "job": "uuid" or "job": ["uuid1", "uuid2"]
2. get rid of the "job" parameter and only support "jobs" - so only "jobs": ["uuid1", "uuid2"]
3. Just allow either parameter, {"job": "uuid1"} or  {"jobs": ["uuid1", "uuid2"]}

My thoughts were: #1  makes for kind of a weird, non-specific schema.  #2 would break backwards compatibility.

wyegelwel: Sorry, didn't explain that well. I've used this endpoint in the past as follows:

curl cook.twosigma.com/retry?job=<uuid>&retries=5 

I was hoping that the following would also work:

curl cook.twosigma.com/retry?job=<uuid1>&job=<uuid2>&retries=5

mforsyth: Ohhhhh @wyegelwel  I smell what you are cooking now.  Yes, I can change it to work like that.   Simpler.  Thanks

wyegelwel: Haha, thanks. May still be worth keeping the "jobs" key for the body if there is a LOT of jobs to modify.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1082,9 +1121,17 @@
...
-    :exists? (partial check-job-exists conn)
+    :exists? (partial check-jobs-exist conn)
     :handle-ok (partial display-retries conn)}))
 
+(defn retry-jobs!
+  [conn ctx]
+  (doseq [job (distinct (::jobs ctx))]
+    (let [new-retries (or (::retries ctx)
+                          (+ (:job/max-retries (d/entity (d/db conn) [:job/uuid job]))
+                             (::increment ctx)))]
```
wyegelwel: Can you check through incrementing that we don't exceed the retry-limit?

mforsyth: 👍

PR #231, opened 2017-01-25
Allow cancelling instances via api

mforsyth:


wyegelwel:
Fix conflict but looks good to go.

Can you make sure the reason for the failed instance reflects that it was killed

wyegelwel:
LGTM. Before I merge, what is the reason that is returned for the killed instance?

mforsyth:
@wyegelwel Hold off on merging for now!  I've been trying to verify the reason code after having rebased, but have actually been having some issues getting Cook back up and running jobs with the new master.   Details to follow.

mforsyth:
@wyegelwel Good to go.  I found that the reason Mesos was assigning to these tasks was 99003,  which we call :mesos-command-executor-failed.   So, I added some code to override this to 99002, :mesos-executor-terminated

wyegelwel:
It occurred to me as I took a final pass that from a user experience perspective this will be a little odd.

A user may request that an instance be killed, get an ack, then query for the instance and see it is still running (since the cancelled loop hadn't run yet).

wyegelwel:
Perhaps we can have a kill channel where processes put things on the channel to be killed. When processing normally, it will be pulled off the channel and killed. On fail over, reconciliation will bring up instances that are running and then [this](https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/scheduler.clj#L240) check will kill the ones that should be down.

Can you think of another way to get the experience to be more in line with what a user would expect?

mforsyth:
@wyegelwel Hmm, I don't think that we should try to make the users believe that cancelling a task is synchronous (that the task process will no longer be running when they get the ack).  Even if we go with a channel monitor, the actual request from Cook to Mesos to kill the task is asynchronous, and the speed at which the task will be killed (and its killed status reported) depends entirely on the task executor.    

The Ack means that Cook INTENDS to kill the task, not that the task has already been killed.  If the user makes another request to see the task's status within seconds of receiving the cancellation ack, they will see in the API response that the "cancelled" flag is now set on the task... and if it was still running at the time of the get request, they will also see that it is still running.  This doesn't seem like a bad user experience to me unless the user has incorrect ideas about how Mesos works.

Let me know your thoughts...

wyegelwel:
@mforsyth, that is a good point. We have generally taken the tact thus far of hiding the async nature from our users since most of the time it is *fast enough*. It has always irked me a bit, perhaps we should revisit it. 

This does however mean that if a user cancels a job it can take 10 seconds + time for mesos to kill task before a new task is started. Can you drop it down to 5 or 3 seconds and add a timer around it so we can track how long it takes?

mforsyth:
@wyegelwel Yes, can def do that.  I don't think this will actually take a lot of extra work to run the killer more frequently.  Most of the time, the total number of jobs it needs to kill will be the same regardless of how frequently it runs, so the only additional load is that the query to find cancellable instances will run more often.  I don't think that should be a significant query, since it is primarily constrained by instances that have been cancelled via the api (which should be a relatively tiny number of instances).

PR #232, opened 2017-01-25
API explicitly indicates whether a completed job was successful

mforsyth:


scheduler/src/cook/mesos/api.clj
```diff
@@ -198,6 +198,7 @@
               :status s/Str
               :submit-time PosInt
               :retries-remaining NonNegInt
+              (s/optional-key :successful) s/Bool
```
mforsyth: @wyegelwel What would you think about simply adding to the possible values returned by the existing "status" field?  If we have both "status" and e.g. "descriptive_state" that makes the API hard to explain.   There isn't necessarily a mandate to keep the Datomic schema resembling the API schema, when we actually do want to convey different information in each context.

wyegelwel: Yea, I agree, my concern is that it would break backwards compatibility. We can deprecate the use of status or something else but I'm hesitant to change the behavior of it.

PR #236, opened 2017-02-06
scheduler: Eventually count mea-culpa failures towards job retry limit.

mforsyth:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1259,8 +1259,19 @@
                                      (log/error "Unable to decline offer; no current driver"))))))
       (build)))
 
+(defn persist-mea-culpa-failure-limit!
+  "The Datomic transactor needs to be able to access this part of the
+  configuration, so on cook startup we transact the configured value into Datomic."
+  [conn limit]
+  (when limit
+    @(d/transact conn [{:db/id :scheduler/config
+                        :scheduler.config/mea-culpa-failure-limit (biginteger limit)}])))
```
wyegelwel: Why is this a biginteger?

mforsyth: Goooooood question.   It should probably just be a long in the db schema rather than a bigint, and then this type conversion wouldn't be necessary.

wyegelwel:
This LGTM. I will wait to merge until the switch from biginteger -> long and the attempt to reduce the code duplication of all-attempts-consumed?

mforsyth:
Hi @wyegelwel Yes, sounds good.  I'm working on that refactor now.

wyegelwel:
This LGTM. Is there anything else you were planning to do before I merge?

mforsyth:
Good to go @wyegelwel !

PR #238, opened 2017-02-13
Support null token in spnego

pschorf:


shamsmahmood:
lgtm

scheduler/src/cook/spnego.clj
```diff
@@ -78,13 +78,14 @@
   [rh]
   (fn [req]
     (if (get-in req [:headers "authorization"])
-      (let [gss_context (gss-context-init)]
-        (if-let [token (do-gss-auth-check gss_context req)]
+      (let [gss_context (gss-context-init)
+            token (do-gss-auth-check gss_context req)]
+        (if (.isEstablished gss_context)
```
wyegelwel: Can you add a comment what is going on here?

scheduler/src/cook/spnego.clj
```diff
@@ -78,13 +78,14 @@
...
-            (-> req
-                (assoc :krb5-authenticated-princ princ
-                       :authorization/user (first (str/split princ #"@" 2)))
-                (rh)
-                (header "WWW-Authenticate" token)))
+            (cond-> (-> req
+                        (assoc :krb5-authenticated-princ princ
+                               :authorization/user (first (str/split princ #"@" 2)))
+                        (rh))
+              token (header "WWW-Authenticate" token)))
```
wyegelwel: What is the implication of not including this WWW-Authenticate when the token doesn't exist?

pschorf: As I understand the spec, SPNEGO can have an arbitrary number of back and forth challenges. The way it's used in Cook right now just expects a challenge from the client, then a response from the server authenticating the server.

It's possible for a client not to request the server to authenticate itself, in which case the token is nil and there's no need to send a WWW-Authenticate header back to the client: just treat the session as authenticated (in Java, you can accomplish this by calling GSSContext.requestMutualAuth(false) on the client GSSContext before making the request.)

scheduler/src/cook/spnego.clj
pschorf: Added a comment

dposada:
LGTM, can we merge?

pschorf:
Yes I think it should be good.

PR #241, opened 2017-02-28
Remove the distinction of backfilled jobs

wyegelwel:
The concept of backfilled jobs was to allow Cook to schedule out of
order (specifically, skip the head of the queue) to oppurtunistically
improve utilization. However, this can lead to cases where the job at
the head of the queue is starved and not able to get scheduled. When
this occurs, we rely on the rebalancer to free up space such that the
head of the queue can run. Instances were explicitly marked as
backfilled when Cook scheduled out of order and were then upgraded out
of backfilled status once the jobs ahead of them had been scheduled.
Instances that were marked as backfilled were treated as the first
jobs to be preempted.

This caused a variety of problems:
1. Preempting backfilled jobs first caused unexpected behavior when a
backfilled, high priority job was preempted instead of a low priority
job
2. Upgrading jobs from backfilled to fully processed is error prone and
causes issues the ordering of the queue. Specifically, jobs that were
backfilled would remain in the queue (generally high in the queue) and
would be removed when the jobs ahead of them were scheduled. However, in
the event that the backfilled job failed, it would "jump the queue"
which means that jobs that fail very quickly that get backfilled can
wreck havoc on the utilization of the cluster

We are choosing to remove the backfill distinction (while still allowing
cook to schedule out of order) because it of the problems above and
because the rebalancer will still be able to preempt jobs to make room
for starved jobs, it just may not be the backfilled jobs.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -689,7 +689,7 @@
                     base {:task_id (:instance/task-id instance)
                           :hostname hostname
                           :ports (:instance/ports instance)
-                          :backfilled (:instance/backfilled? instance false)
+                          :backfilled false ;; Backfill has been deprecated
                           :preempted (:instance/preempted? instance false)
                           :slave_id (:instance/slave-id instance)
                           :executor_id (:instance/executor-id instance)
--- a/scheduler/src/cook/mesos/rebalancer.clj
+++ b/scheduler/src/cook/mesos/rebalancer.clj
@@ -229,7 +229,7 @@
                                             (group-by util/task-ent->user)
                                             (map (fn [[user task-ents]]
                                                    [user (into (sorted-set-by (case category
-                                                                                :normal (util/same-user-task-comparator-penalize-backfill)
+                                                                                :normal (util/same-user-task-comparator)
                                                                                 :gpu (util/same-user-task-comparator))) task-ents)]))
                                             (into {}))
         task->scored-task (into (pm/priority-map-keyfn (case category
@@ -262,7 +262,7 @@
         (reduce (fn [task-ents-by-user task-ent]
                   (let [user (util/task-ent->user task-ent)
                         f (if (= new-running-task-ent task-ent)
-                            (fnil conj (sorted-set-by (util/same-user-task-comparator-penalize-backfill)))
+                            (fnil conj (sorted-set-by (util/same-user-task-comparator)))
                             disj)]
                     (update-in task-ents-by-user [user] f task-ent)))
                 user->sorted-running-task-ents
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -112,21 +112,6 @@
 (histograms/defhistogram [cook-mesos scheduler hist-task-fail-times])
 (meters/defmeter [cook-mesos scheduler task-fail-times])
 
-(meters/defmeter [cook-mesos scheduler backfilled-count])
-(histograms/defhistogram [cook-mesos scheduler hist-backfilled-count])
-(meters/defmeter [cook-mesos scheduler backfilled-cpu])
-(meters/defmeter [cook-mesos scheduler backfilled-mem])
-
-(meters/defmeter [cook-mesos scheduler upgraded-count])
-(histograms/defhistogram [cook-mesos scheduler hist-upgraded-count])
-(meters/defmeter [cook-mesos scheduler upgraded-cpu])
-(meters/defmeter [cook-mesos scheduler upgraded-mem])
-
-(meters/defmeter [cook-mesos scheduler fully-processed-count])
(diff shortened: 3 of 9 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -687,7 +687,7 @@
                     base {:task_id (:instance/task-id instance)
                           :hostname hostname
                           :ports (:instance/ports instance)
-                          :backfilled (:instance/backfilled? instance false)
+                          :backfilled false ;; Backfill has been deprecated
```
wyegelwel: I'm hesitant to do that because their may be clients that expect it. The joys(?) of backwards compatibility.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -262,7 +262,7 @@
         (reduce (fn [task-ents-by-user task-ent]
                   (let [user (util/task-ent->user task-ent)
                         f (if (= new-running-task-ent task-ent)
-                            (fnil conj (sorted-set-by (util/same-user-task-comparator-penalize-backfill)))
+                            (fnil conj (sorted-set-by (util/same-user-task-comparator)))
```
shamsimam: Should the `nil` be handled by the function generating `user->sorted-running-task-ents` who initializes to a sorted set instead of nil?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -646,35 +555,24 @@
                                         :let [task-request (.getRequest task-result)]
                                         :when (= :normal (util/categorize-job (:job task-request)))]
                                     (:job task-request))
+              matched-normal-job-uuids (set (mapv :job/uuid matched-normal-jobs))
```
shamsimam: Would `map` work just as well as `mapv` here?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -646,35 +555,24 @@
...
-              ;; prioritized above them are also scheduled.
-              first-considerable-resources (-> considerable-by :normal first util/job-ent->resources)
+                                              (update-in [:gpu] 
+                                                         #(remove (fn [{pending-job-uuid :job/uuid}]
+                                                                    (contains? matched-gpu-job-uuids pending-job-uuid))
+                                                                  %))
+                                              (update-in [:normal] 
+                                                         #(remove (fn [{pending-job-uuid :job/uuid}]
+                                                                    (contains? matched-normal-job-uuids pending-job-uuid))
+                                                                  %))))
```
shamsimam: Should we extract the common code in the two remove calls in the `let`:
```
remove-job (fn [existing-jobs matched-job-uuids] (remove (fn [{job-uuid :job/uuid}] (contains? matched-job-uuids job-uuid)) existing-jobs))
...
(update-in [:gpu] remove-job matched-gpu-job-uuids)...
`

File: scheduler/src/cook/mesos/scheduler.clj [hunk 5]
```diff
@@ -489,72 +474,6 @@
 (gauges/defgauge [cook-mesos scheduler front-of-job-queue-mem] (fn [] @front-of-job-queue-mem-atom))
 (gauges/defgauge [cook-mesos scheduler front-of-job-queue-cpus] (fn [] @front-of-job-queue-cpus-atom))
 
-(defn ids-of-backfilled-instances
```
shamsimam: Do these changes need to be backed by unit tests in `scheduler_test.clj` to ensure everything still works as expected?

wyegelwel: I don't think so, the non-backfill functionality already had tests and then there was backfill specific tests that were added. I just removed the backfill ones.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: That seems awfully reasonable =), will do

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: I don't think so. The `user->sorted-running-task-ents` will only have a user if it has at least one running job. In the rebalancer, we are considering "what would happen" if we let job X run. Job X may be from a user who has nothing currently running and therefore wouldn't exist in `user->sorted-running-task-ents`. 

Let me know if you had something else in mind.

wyegelwel:
@shamsimam @DiegoAlbertoTorres I just pushed an updated version with your changes, please let me know if you have any other comments or if it is ready to merge.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Ah, got it! For some reason I (incorrectly) felt that `user` was one of the keys in `user->sorted-running-task-ents`. 

The `update-in` below is either initializing the entry against a user to a sorted set or adding elements into the `sorted-running-task-ents'` set.

shamsimam:
lgtm

PR #246, opened 2017-03-08
scheduler: Fix error when only one job in query for retries api.

mforsyth:


mforsyth:
@wyegelwel Thanks for the prompting.  Actually, there was sort of a similar issue with the handling of jobs in body-params as well :(  It seems like in the top level tests in api.clj,  we should look for a way to pass a more raw form request data into the handler functions somehow, so that the unit tests can be covering Compojure-API's coercion behavior as well.  Right now it is quite easy to have unit tests for a handler function passing, but then they fail with a real browser request because the coercion was handled differently than the person writing the unit tests expected.     

Anyway, i've fixed this and re-pushed.  I've curled successfully every combination you requested...
curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{"retries": 4, "job": ["58b5e0aa-78b6-4c9f-8f32-aebd140f7229"]}' --user name:password 'http://localhost:12321/retry'
4~/projects/cook $ curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{"retries": 4, "job": ["58b5e0aa-78b6-4c9f-8f32-aebd140f7229", "58b5e0a3-d65c-4aa4-997c-af5850484440"]}' --user name:password 'http://localhost:12321/retry?job=58b5e0a7-87ad-4ad5-9cfd-63119a36dda0'
4~/projects/cook $ curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{"retries": 4}' --user name:password 'http://localhost:12321/retry?job=58b5e0a3-d65c-4aa4-997c-af5850484440&job=58b5e0a7-87ad-4ad5-9cfd-63119a36dda0'
4~/projects/cook $ curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{"retries": 4}' --user name:password 'http://localhost:12321/retry?job=58b5e0a3-d65c-4aa4-997c-af5850484440
> ''
>   C-c C-c
~/projects/cook $ curl -X PUT --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{"retries": 4}' --user name:password 'http://localhost:12321/retry?job=58b5e0a3-d65c-4aa4-997c-af5850484440'
4~/projects/cook $ 

One caveat:  you listed
curl -d '{"retries" : 5, "job": }' :12321/retry?job=

If you meant to include a plain string rather than a single-element array in the "job" in the body, that isn't supported.  I don't really see a reason to support "either string or array" in the body; the only reason we are effectively doing so with query-params is an implementation detail (compojure seemingly cannot be easily told to coerce a single query param to a single-element vector).   Let me know what you think.

wyegelwel:
> If you meant to include a plain string rather than a single-element array in the "job" in the body, that isn't supported. I don't really see a reason to support "either string or array" in the body; the only reason we are effectively doing so with query-params is an implementation detail (compojure seemingly cannot be easily told to coerce a single query param to a single-element vector). Let me know what you think.

Makes sense, though in that case, can we rename the body param to "jobs" so it is obvious we expect a list?

mforsyth:
Can do 👍

PR #247, opened 2017-03-08
Use commit latch with get-pending-job-ents to speed up ranking

wyegelwel:


scheduler/test/cook/test/mesos/util.clj
```diff
@@ -160,16 +160,65 @@
...
+                                :committed? false
+                                :submit-time (tc/to-date (-> 6 t/hours t/ago))))
+          uncommitted-count 25
+          _ (dotimes [_ uncommitted-count]
+              (create-dummy-job conn
+                                :committed? false
+                                :submit-time (tc/to-date (-> 48 t/hours t/ago))))]
+      (is (= (count (util/clear-uncommitted-jobs conn (t/yesterday) true))
+             uncommitted-count))))
+  (testing "uncommitted with retract"
```
shamsimam: Should this one be merged with the previous test and the extra assertion at the end with a corresponding message.

scheduler/src/cook/mesos/util.clj
```diff
@@ -167,22 +167,34 @@
 
 (timers/deftimer [cook-mesos scheduler get-pending-jobs-duration])
 
+(defn get-pending-job-ents*
```
shamsimam: Should we make `get-pending-job-ents*` a private function? I assume the * in the name means that you want to encourage users to use `get-pending-job-ents`

scheduler/src/cook/mesos/util.clj
```diff
@@ -342,3 +354,31 @@
    (keyword (name name-space) (name value)))
   ([name-space subspace value]
    (namespace-datomic (str (name name-space) "." (name subspace)) value)))
+
+(defn clear-uncommitted-jobs
```
shamsimam: Is `clear-uncommitted-jobs` used anywhere other than test code?

scheduler/test/cook/test/mesos/util.clj
wyegelwel: I'm indifferent. I split them so that we could test that dry run did the thing we expect.

scheduler/src/cook/mesos/util.clj
wyegelwel: Yea, perhaps it could just use a better name (in additional to making it private). Any suggestions on name?

scheduler/src/cook/mesos/util.clj
wyegelwel: At the moment, no. My plan was to eventually have a thread call it periodically to clean things up. However, I didn't want to add too much to this PR. For the moment this function could be called manually using the repl.

scheduler/src/cook/mesos/util.clj
shamsimam: I'm okay with the name, I like the convention of `*` suffixes to ask users to use the function with care.

scheduler/src/cook/mesos/util.clj
shamsimam: Okay.

scheduler/test/cook/test/mesos/util.clj
shamsimam: Okay, was just trying to reduce some test code :D

shamsimam:
lgtm

PR #248, opened 2017-03-08
Cache feature vectors to speed ranking

wyegelwel:


scheduler/src/cook/mesos/util.clj
```diff
@@ -267,9 +267,22 @@
...
+   ;; This is done because accessing fields in datomic entities is much slower than
+   ;; a map access, even when accessing multiple times. 
+   ;; Don't want to complicate the function by caching new values in the event we see them
+   (let [task-ent->feature-vector (into {} 
+                                        (map #(vector % 
+                                                      (task->feature-vector %))) 
+                                        tasks)]
+     (fn [task1 task2]
+       (compare (or (task-ent->feature-vector task1)
+                    (task->feature-vector task1))
```
shamsimam: Why is the `or` necessary?

scheduler/src/cook/mesos/util.clj
```diff
@@ -267,9 +267,22 @@
...
-  (fn [task1 task2]
-      (compare (task->feature-vector task1) (task->feature-vector task2))))
+  ([]
+   (same-user-task-comparator []))
+  ([tasks]
+   ;; Pre-compute the feature-vector for tasks we expect to see to improve performance
+   ;; This is done because accessing fields in datomic entities is much slower than
+   ;; a map access, even when accessing multiple times. 
+   ;; Don't want to complicate the function by caching new values in the event we see them
+   (let [task-ent->feature-vector (into {}
```
shamsimam: would `plumbing.core/map-from-keys` be a better fit here?

scheduler/src/cook/mesos/util.clj
wyegelwel: In the event that this comparator is called without pre-loading the tasks we still want it to work.

scheduler/src/cook/mesos/util.clj
wyegelwel: Yes, good point. Pushed a new version, it is much better thanks to your suggestion

scheduler/src/cook/mesos/util.clj
shamsimam: Just for my understanding, in the current context it is used (`sort-normal-jobs-by-dru`), it is not necessary, right?

shamsimam:
lgtm

PR #249, opened 2017-03-09
Allow 0 retries

pschorf:


wyegelwel:
Can we add a unit test for this? 

Can you document what happens when you use 0 retries? I expect the job will not be retried regardless of the failure, including a mea culpa failure. Is that the case?

wyegelwel:
Also, I believe you will need to rework how "retries-remaining" is calculated since it will now be negative when max-retries = 0

scheduler/src/cook/mesos/schema.clj
```diff
@@ -673,7 +678,9 @@
                    :params [db job-ent]
                    :code
                    (let [done-statuses #{:instance.status/success :instance.status/failed}
-                         mea-culpa-limit (or (:scheduler.config/mea-culpa-failure-limit
+                         mea-culpa-limit (or (when (:job/disable-mea-culpa-retries job-ent)
```
shamsimam: Would `cond` be a better fit here?

scheduler/src/cook/mesos/schema.clj
```diff
@@ -773,6 +780,13 @@
                                any-unknown? (some #{:instance.status/unknown} instance-states)
                                all-failed? (every? #{:instance.status/failed} instance-states)
                                prior-state (:job/state job-ent)
+                               reason (->> (q '[:find ?e
```
shamsimam: formatting seems off.

scheduler/src/cook/mesos/api.clj
```diff
@@ -173,14 +173,15 @@
    ;; length at most 128.
    :name (s/both s/Str (s/pred #(re-matches #"[\.a-zA-Z0-9_-]{0,128}" %) 'under-128-characters-and-alphanum?))
    :priority (s/both s/Int (s/pred #(<= 0 % 100) 'between-0-and-100))
-   :max-retries NonNegInt
+   :max-retries PosInt
```
shamsimam: This breaks backwards compatibility by being more strict. Do we know if there are any clients who set `max-retries` to zero?

shamsimam: Should there be a unit tests to validate that passing 0 or negative fails the schema validation?

scheduler/src/cook/mesos/schema.clj
pschorf: It's better for the first case (disable-mea-culpa-retries) but worse for the second (mea-culpa-failure-limit) so to me, it's a toss-up. I picked or because (presumably) you'd want to introduce a let for mea-culpa-failure-limit.

scheduler/src/cook/mesos/schema.clj
pschorf: Changed this code

scheduler/src/cook/mesos/api.clj
pschorf: It was originally PosInt, first I changed it to NonNegInt, then changed it back.

pschorf:
@wyegelwel I added a new field instead of using 0 for max-retries: that's more clear to me.

Also added a unit test to verify the behavior

wyegelwel:
I think it is more clear once a user is aware of mea culpa reasons (and the flag). Retries == 0 is more discover-able and clients built on top don't need special logic for the 0 retries case.

I'm not sold either way. Anyone else have an opinion?

shamsimam:
I like differentiating between user configured retries/attempts and cook triggered (mea culpa) retries.The latter we can configure it to be whatever value in cook configs and user can choose to disable or override.

wyegelwel:
> I like differentiating between user configured retries/attempts and cook triggered (mea culpa) retries.The latter we can configure it to be whatever value in cook configs and user can choose to disable or override.

I think in this regard, renaming max-retries to something like "attempts" would be more informative. I'm not sure I care enough to fight for retries == 0 if you feel strongly in this direction. Please document this flag and explicitly note the 0 retries use case.

scheduler/src/cook/mesos/schema.clj
```diff
@@ -117,6 +117,11 @@
     :db/valueType :db.type/ref
     :db/cardinality :db.cardinality/one
     :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/ident :job/disable-mea-culpa-retries
+    :db/valueType :db.type/boolean
+    :db/cardinality :db.cardinality/one
+    :db.install/_attribute :db.part/db}
```
wyegelwel: please add a :db.doc key here describing what this is used for

pschorf: Done

pschorf:
Also added documentation to scheduler-rest-api.asc

PR #250, opened 2017-03-09
Adds ignores for IntelliJ IDEA generated files

dposada:
(and removes 3 unrelated duplicate entries)


dposada:
@shamsimam can you please take a look?

.gitignore
```diff
@@ -23,8 +22,8 @@ pom.xml.asc
 .lein-deps-sum
 .lein-failures
 .lein-plugins
-.lein-repl-history
-.nrepl-port
```
wyegelwel: These two lein ones seem as though they should stay in ignore. Why are you removing them?

dposada:
Agree, they were duplicated elsewhere in the file

wyegelwel:
It was really important to ignore them :)

wyegelwel:
LGTM, will wait on @shamsimam to look as well

.gitignore
shamsimam: I agree.

shamsimam:
Can we sort the contents of the file?

dposada:
@shamsimam .gitignore is now lexicographically sorted :)

PR #260, opened 2017-03-17
Mf/409 on dupe job UUID

mforsyth:


scheduler/docs/scheduler-rest-api.asc
```diff
@@ -123,9 +123,10 @@ The request body is list of job maps with each entry containing a map of the fol
 |====
 |Code | HTTP Meaning | Possible reason
 |201 | Created | Job has been successfully created.
-|400 | Malformed | This will be returned if the UUID is already used or the requst syntax is not correct.
+|400 | Malformed | This will be returned if the request syntax is not correct.
 |401 | Not Authorized | Returned if authentication fails or user is not authorized to run jobs on the system.
-|422 | Unprocessable Entity | Returned if there is an error committing jobs to the Cook database.
+|409 | Not Authorized | Returned if one or more of the job UUIDs was already in use.
```
shamsimam: `|409 | Conflict | Returned if one or more of the job UUIDs is already in use.`

scheduler/src/cook/mesos/api.clj
```diff
@@ -895,50 +937,30 @@
...
+    :exists? (fn [ctx]
+               (let [db (d/db conn)
+                     existing (filter (partial job-exists? db) (map :uuid (::jobs ctx)))]
+                 [(seq existing) {::existing existing}]))
+
+    ;; To ensure compatibility with existing clients,
+    ;; we need to return 409 (conflict) when a client POSTs a Job UUID that already exists.
+    ;; Liberator normally only supports 409 responses to PUT requests, so we need to override
+    ;; the specific decisions that will lead to a 409 response on a POST request as well.
+    ;; (see https://clojure-liberator.github.io/liberator/tutorial/decision-graph.html)
```
shamsimam: :+1:  thanks for the link, helped explain a lot.

scheduler/src/cook/mesos/api.clj
```diff
@@ -858,6 +856,50 @@
 
 (def override-group-str "override-group-immutability")
 
+(defn create-jobs!
```
shamsimam: Missing docstring.

scheduler/src/cook/mesos/api.clj
```diff
@@ -858,6 +856,50 @@
...
+  [conn ctx]
```
shamsimam: Can we use destructuring `{:keys [::groups ::jobs] :as ctx}`

scheduler/src/cook/mesos/api.clj
```diff
@@ -1385,7 +1407,9 @@
              :handler (read-jobs-handler conn fid task-constraints gpu-enabled? is-authorized-fn)}
        :post {:summary "Schedules one or more jobs."
               :parameters {:body-params RawSchedulerRequest}
-              :responses {200 {:schema [JobResponse]}}
+              :responses {201 {:description "The jobs were successfully scheduled."}
+                          400 {:description "One or more of the jobs were incorrectly specified."}
+                          409 {:description "One or more of the jobs UUIDs were already in use."}}
```
shamsimam: Nitpick: `One or more of the jobs UUIDs *are* already in use`

shamsimam: Instead of saying one or more, can we include the exact count? E.g. `2 job(s) have UUIDs which are already in use.`

shamsimam:
Done with initial pass.

scheduler/src/cook/mesos/api.clj
mforsyth: @shamsimam This string is used in the Swagger documentation for the endpoint, but is never included in a response generated by the endpoint.  So, it's static - can't mention the number of jobs in a particular request here.

mforsyth:
@shamsimam I made changes in accordance with your suggestions.  This was a perfect opportunity to destructure function arguments, thanks!

scheduler/src/cook/mesos/api.clj
```diff
@@ -858,6 +856,52 @@
...
+(defn create-jobs!
+  "Based on the context, persists the specified jobs, along with their groups,
+  to Datomic.
+  Preconditions:  The context must already have been populated with both
+  ::jobs and ::groups, which specify the jobs and job groups."
+  [conn {:keys [::groups ::jobs] :as ctx}]
+  (try
+    (log/info "Submitting jobs through raw api:" (::jobs ctx))
+    (let [group-uuids (set (map :uuid groups))
+                                        ; Create new implicit groups (with all default settings)
```
shamsimam: Formatting seems off.

shamsimam:
One minor formatting comment, else lgtm.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1385,7 +1409,9 @@
              :handler (read-jobs-handler conn fid task-constraints gpu-enabled? is-authorized-fn)}
        :post {:summary "Schedules one or more jobs."
               :parameters {:body-params RawSchedulerRequest}
-              :responses {200 {:schema [JobResponse]}}
+              :responses {201 {:description "The jobs were successfully scheduled."}
+                          400 {:description "One or more of the jobs were incorrectly specified."}
+                          409 {:description "One or more of the jobs UUIDs are already in use."}}
```
wyegelwel: Do we also return 500 or 422? I think I've seen both at times.

mforsyth: @wyegelwel This PR removes the possibility of a 422.  422 was really being used when 500 would have been more appropriate (it was only used to report that an unknown error occurred, it was never indicating a specific correctable problem with the content of the request).   

500 could theoretically occur on basically ANY endpoint depending on where an unpredictable error occurs...  I can add that here but it really applies to the whole API (and basically any HTTP API).

mforsyth:
Fixed the weird whitespace before the comment.  👍

PR #266, opened 2017-03-27
Add basic integration test

pschorf:


integration/test.py
```diff
@@ -0,0 +1,30 @@
...
+from retrying import retry
+import time
+import unittest
+import uuid
+
+class CookTest(unittest.TestCase):
+
+    @retry(stop_max_delay=120000, wait_fixed=5000)
+    def wait_for_job(self, job_id, status):
+        job = requests.get('%s/rawscheduler?job=%s' % (self.cook_url, job_id))
```
wyegelwel: We should use request's session http://docs.python-requests.org/en/master/user/advanced/ as it makes it easier to sub in configuration like kerberos.

wyegelwel: We can make a single session for all of the tests by using https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUp

wyegelwel:
Questions for a pythonista:

1. What is the suggested way to make a test configurable (set the cook url)

integration/tests/cook/test_basic.py
```diff
@@ -0,0 +1,200 @@
...
+    #         self.assertEqual('failed', instance['status'])
+
+    def test_max_runtime_exceeded(self):
+        job_spec = self.minimal_job(command='sleep 60', max_runtime=5000)
+        resp = self.session.post('%s/rawscheduler' % self.cook_url,
+                                 json={'jobs': [job_spec]})
+        self.assertEqual(201, resp.status_code)
+        job = self.wait_for_job(job_spec['uuid'], 'completed')
+        self.assertEqual(1, len(job['instances']))
+        self.assertEqual('failed', job['instances'][0]['status'])
```
wyegelwel: It is also worth checking that the reason code is what is expected. I expect this test is "passing" because the job completes after a minute even though it hasn't been killed for exceeding max run time

pschorf: Done

wyegelwel: Can you push the change when you get a chance?

pschorf: Done, accidentally pushed to wrong repo

integration/tests/cook/test_basic.py
```diff
@@ -0,0 +1,303 @@
...
+        job.update(kwargs)
+        return job
+
+    def minimal_group(self, **kwargs):
+        group = {"uuid" : str(uuid.uuid4())}
+        group.update(kwargs)
+        return group
+
+    def setUp(self):
+        self.cook_url = 'http://localhost:12321'
```
sradack: Is there a standard environment variable for referring to the cook scheduler URL?  Maybe just use COOK_SCHEDULER_URL.  Use http://localhost:12321 if it's not defined.  Also maybe add some helpful print statements for the user.

sradack: Also the docs should be updated to reflect the new setting

wyegelwel:
Am I ok to merge as is or are you ( @pschorf @sradack ) still working on something?

integration/tests/cook/test_basic.py
pschorf: Done

pschorf:
@sradack can merge if my last commit addressed his comment

sradack:
Should we squash?

pschorf:
Squashed

PR #276, opened 2017-03-29
Integration test for cook scheduler / rebalancer using mock mesos

wyegelwel:
This PR includes a mock of mesos and a simulator that starts the scheduling aspects of cook to allow back testing and auditing behavior. The mock of mesos allows us to simulate arbitrarily large clusters (until computation or memory limits on the machine running the simulation are reached). The intent is this simulator will be run "faster than real time" so that running over a day or week of historical jobs can be done in minutes and hours.

### DIFF
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -86,36 +86,40 @@
 
 (def mesos-scheduler
   {:mesos-scheduler (fnk [[:settings mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset task-constraints riemann mesos-gpu-enabled rebalancer good-enough-fitness] mesos-datomic mesos-datomic-mult curator-framework mesos-pending-jobs-atom mesos-offer-cache]
-                         (try
-                           (Class/forName "org.apache.mesos.Scheduler")
-                           ((lazy-load-var 'cook.mesos/start-mesos-scheduler)
-                            mesos-master
-                            mesos-master-hosts
-                            curator-framework
-                            mesos-datomic
-                            mesos-datomic-mult
-                            mesos-leader-path
-                            mesos-failover-timeout
-                            mesos-principal
-                            mesos-role
-                            mesos-framework-name
-                            offer-incubate-time-ms
-                            mea-culpa-failure-limit
-                            task-constraints
-                            (:host riemann)
-                            (:port riemann)
-                            mesos-pending-jobs-atom
-                            mesos-offer-cache
-                            mesos-gpu-enabled
-                            rebalancer
-                            {:fenzo-max-jobs-considered fenzo-max-jobs-considered
-                             :fenzo-scaleback fenzo-scaleback
-                             :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
-                             :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
-                             :good-enough-fitness good-enough-fitness})
-                           (catch ClassNotFoundException e
-                             (log/warn e "Not loading mesos support...")
-                             nil)))})
+                         (let [make-mesos-driver-fn (partial (lazy-load-var 'cook.mesos/make-mesos-driver)
+                                                             {:mesos-master mesos-master
+                                                              :mesos-failover-timeout mesos-failover-timeout
+                                                              :mesos-principal mesos-principal
+                                                              :mesos-role mesos-role
+                                                              :mesos-framework-name mesos-framework-name
+                                                              :gpus-enabled? mesos-gpu-enabled})
+                               get-mesos-utilization-fn (partial (lazy-load-var 'cook.mesos/get-mesos-utilization) mesos-master-hosts)]
+                           (try
+                             (Class/forName "org.apache.mesos.Scheduler")
+                             ((lazy-load-var 'cook.mesos/start-mesos-scheduler)
+                              make-mesos-driver-fn
+                              get-mesos-utilization-fn
+                              curator-framework
+                              mesos-datomic
+                              mesos-datomic-mult
+                              mesos-leader-path
+                              offer-incubate-time-ms
+                              mea-culpa-failure-limit
+                              task-constraints
+                              (:host riemann)
+                              (:port riemann)
+                              mesos-pending-jobs-atom
+                              mesos-offer-cache
+                              mesos-gpu-enabled
+                              rebalancer
+                              {:fenzo-max-jobs-considered fenzo-max-jobs-considered
+                               :fenzo-scaleback fenzo-scaleback
+                               :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
+                               :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
+                               :good-enough-fitness good-enough-fitness})
+                             (catch ClassNotFoundException e
+                               (log/warn e "Not loading mesos support...")
+                               nil))))})
 
 (defn health-check-middleware
   "This adds /debug to return 200 OK"
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -14,7 +14,9 @@
 ;; limitations under the License.
 ;;
 (ns cook.mesos
-  (:require [clojure.core.async :as async]
+  (:require [clj-http.client :as http]
+            [clojure.core.async :as async]
+            [clojure.data.json :as json]
             [clojure.tools.logging :as log]
             [cook.curator :as curator]
             [cook.datomic :refer (transact-with-retries)]
@@ -28,7 +30,8 @@
             [mesomatic.types]
             [metatransaction.core :as mt :refer (db)]
             [metatransaction.utils :as dutils]
-            [metrics.counters :as counters])
+            [metrics.counters :as counters]
+            [swiss.arrows :refer :all])
   (:import [org.apache.curator.framework.recipes.leader LeaderSelector LeaderSelectorListener]
            org.apache.curator.framework.state.ConnectionState))
 
@@ -75,9 +78,93 @@
 
 (counters/defcounter [cook-mesos mesos mesos-leader])
 
+(defn make-mesos-driver
+  "Creates a mesos driver
+
+   Parameters:
+   mesos-framework-name     -- string, name to use when connecting to mesos.
+                               Will be appended with version of cook
+   mesos-role               -- string, (optional) The role to connect to mesos with
+   mesos-principal          -- string, (optional) principal to connect to mesos with
+   gpu-enabled?             -- boolean, (optional) whether cook will schedule gpu jobs
+   mesos-failover-timeout   -- long, (optional) time in milliseconds mesos will wait
+                               for framework to reconnect.
+                               See http://mesos.apache.org/documentation/latest/high-availability-framework-guide/
+                               and search for failover_timeout
+   mesos-master             -- str, (optional) connection string for mesos masters
+   scheduler                -- mesos scheduler implementation
+   framework-id             -- str, (optional) Id of framework if it has connected to mesos before
+   "
+  ([config scheduler]
+   (make-mesos-driver config scheduler nil))
+  ([{:keys [mesos-framework-name mesos-role mesos-principal gpu-enabled?
+            mesos-failover-timeout mesos-master] :as config}
+    scheduler framework-id]
+   (apply mesomatic.scheduler/scheduler-driver
+          scheduler
+          (merge
+            {:user ""
+             :name (str mesos-framework-name
+                        "-"
+                        cook.util/version)
+             :checkpoint true}
+            (when mesos-role
+              {:role mesos-role})
+            (when mesos-principal
+              {:principal mesos-principal})
+            (when gpu-enabled?
+              {:capabilities [{:type :framework-capability-gpu-resources}]})
+            (when mesos-failover-timeout
+              {:failover-timeout mesos-failover-timeout})
+            (when framework-id
+              {:id {:value framework-id}}))
+          mesos-master
+          (when mesos-principal
+            [{:principal mesos-principal}]))))
+
+(defn get-mesos-utilization
+  "Queries the mesos master to get the utilization.
+   Returns the max of cpu and mem utilization as a decimal (e.g. 0.92)"
+  [mesos-master-hosts]
+  (let [mesos-master-urls (map #(str "http://" % ":5050/metrics/snapshot") mesos-master-hosts)
+        get-stats (fn [url] (some->> url
+                                     (http/get)
+                                     (:body)
+                                     (json/read-str)))
+        utilization (some-<>> mesos-master-urls
+                              (map get-stats)
+                              (filter #(pos? (get % "master/elected")))
+                              (first)
+                              (select-keys <> ["master/cpus_percent" "master/mem_percent"])
+                              (vals)
+                              (apply max))]
+    utilization))
+
 (defn start-mesos-scheduler
-  "Starts a leader elector that runs a mesos."
-  [mesos-master mesos-master-hosts curator-framework mesos-datomic-conn mesos-datomic-mult zk-prefix mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit task-constraints riemann-host riemann-port mesos-pending-jobs-atom offer-cache gpu-enabled? rebalancer-config
+  "Starts a leader elector. When the process is leader, it starts the mesos
+   scheduler and associated threads to interact with mesos.
+
+   Parameters
+   make-mesos-driver-fn     -- fn, function that accepts a mesos scheduler and framework id
+                                   and returns a mesos driver
+   get-mesos-utilization    -- fn, function with no parameters, returns utilization of cluster [0,1]
+   mesos-master-hosts       -- seq[strings], url of mesos masters to query for cluster info
+   curator-framework        -- curator object, object for interacting with zk
+   mesos-datomic-conn       -- datomic conn, connection to datomic db for interacting with datomic
+   mesos-datomic-mult       -- async channel, feed of db writes
+   zk-prefix                -- str, prefix in zk for cook data
+   offer-incubate-time-ms   -- long, time in millis that offers are allowed to sit before they are declined
+   mea-culpa-failure-limit  -- long, max failures of mea culpa reason before it is considered a 'real' failure
+                                     see scheduler/docs/configuration.asc for more details
+   task-constraints         -- map, constraints on task. See scheduler/docs/configuration.asc for more details
+   riemann-host             -- str, dns name of riemann
+   riemann-port             -- int, port for riemann
+   mesos-pending-jobs-atom  -- atom, Populate (and update) list of pending jobs into atom
+   offer-cache              -- atom, map from host to most recent offer. Used to get attributes
+   gpu-enabled?             -- boolean, whether cook will schedule gpus
+   rebalancer-config        -- map, config for rebalancer. See scheduler/docs/rebalancer-config.asc for details
+   fenzo-config             -- map, config for fenzo, See scheduler/docs/configuration.asc for more details"
+  [make-mesos-driver-fn get-mesos-utilization curator-framework mesos-datomic-conn mesos-datomic-mult zk-prefix offer-incubate-time-ms mea-culpa-failure-limit task-constraints riemann-host riemann-port mesos-pending-jobs-atom offer-cache gpu-enabled? rebalancer-config
    {:keys [fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn
            fenzo-floor-iterations-before-reset good-enough-fitness]
     :as fenzo-config}]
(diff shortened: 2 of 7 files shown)


scheduler/src/cook/components.clj
```diff
@@ -84,35 +84,38 @@
...
-                            rebalancer
-                            {:fenzo-max-jobs-considered fenzo-max-jobs-considered
-                             :fenzo-scaleback fenzo-scaleback
-                             :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
-                             :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
-                             :good-enough-fitness good-enough-fitness})
-                           (catch ClassNotFoundException e
-                             (log/warn e "Not loading mesos support...")
-                             nil)))})
+                         (let [make-mesos-driver-fn (partial (lazy-load-var 'cook.mesos/make-mesos-driver)
```
dposada: Why is the `lazy-load-var` necessary here?

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,458 @@
...
+                    [state' new-offer-chan]))]
+            (when state'
+              (recur state' new-offer-chan))))
+        (catch Exception ex
+          (log/fatal ex "Error while simulating mesos")
+          (println "Exception occurred: " ex))))
+    (mesos/wrap-driver driver)))
+
+
+(comment
```
dposada: Can we remove the comment?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,749 @@
+(ns cook.test.mesos.mesos-mock
+  (:use clojure.test)
+  (:require [cook.mesos :as c]
```
dposada: Sort requires

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,749 @@
...
+                       )
+    (disconnected [this driver]
+                  )
+    ;; We don't care about losing slaves or executors--only tasks
+    (slave-lost [this driver slave-id])
+    (executor-lost [this driver executor-id slave-id status])
+    (error [this driver message]
+           )
+    (resource-offers [this driver offers]
+                     )
```
dposada: Several `)` that can be on the previous line

dposada:
@wyegelwel I see several `TODO`s in the code - are those things you want to do as part of this PR or just some time in the future?

scheduler/src/cook/components.clj
wyegelwel: The `lazy-load-var` are a way to avoid requiring namespaces initially which has two consequences (I explicitly didn't use "benefits" here):
1. AOT compilation can pass even if the lazy loaded namespace can't be compiled (since it is done at run time)
2. It makes initial start up time faster since not everything is loaded immediately

I'm not convinced either of these are worth the headache but I don't think this CR is the place to address it.

scheduler/src/cook/mesos.clj
```diff
@@ -75,9 +75,73 @@
...
+            (when framework-id
+              {:id {:value framework-id}}))
+          mesos-master
+          (when mesos-principal
+            [{:principal mesos-principal}]))))
+
 (defn start-mesos-scheduler
-  "Starts a leader elector that runs a mesos."
-  [mesos-master mesos-master-hosts curator-framework mesos-datomic-conn mesos-datomic-mult zk-prefix mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit task-constraints riemann-host riemann-port mesos-pending-jobs-atom gpu-enabled? rebalancer-config
+  "Starts a leader elector which while leader starts the mesos scheduler and
```
wyegelwel: me either, will fix

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,458 @@
+(ns cook.mesos.mesos-mock
+  (:require [mesomatic.scheduler :as mesos]
```
wyegelwel: Just kidding, of course

wyegelwel:
Ok, this PR includes code to mock mesos and have the cook scheduler connect to it. It includes a test of the basic schedule / rebalance workflow that was previously missing from our tests. 

There will be additional changes to make a full-fledge simulator but I think this is a milestone to merge. 

@dposada when you get a moment, could you take another pass? 

The todos that are remaining are for future work

dposada:
Will do

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,454 @@
...
+        slave-id (:value (:slave-id (first tasks)))]
+    (when-not (every? #(= slave-id (:value (:slave-id %))) tasks)
+      (throw (ex-info "All tasks must have same slave-id" 
+                      {:slave-ids (map (comp :value :slave-id) tasks)})))
+    (when-not (get slave-id->host slave-id)
+      (throw (ex-info "Unknown slave-id" {:slave-id slave-id :known-slave-ids (keys slave-id->host)})))
+    (when (some nil? offers) 
+      (throw (ex-info "Unknown offers" 
+                      {:unknown-offers 
+                       (remove #(contains? (set (keys offer-id->offer)))
```
dposada: Should this be 

```clj
#(contains? (set (keys offer-id->offer)) %) 
```
?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+              update-status-error-ms 200
+              ]
+          ;; Transact user "a"s jobs
+          (doall user-a-job-ent-ids)
+          ;; Let them get scheduled
+          (Thread/sleep 5000)
+          ;; Transact user "b"s jobs
+          (doall user-b-job-ent-ids)
+          ;; Let the system complete
+          (Thread/sleep 30000)
```
dposada: Is it possible to refactor this to avoid the `sleep`s and instead poll for expected outcomes with a timeout?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+     (.start curator-framework)
+     [zookeeper-server curator-framework])))
+
+(defmacro with-cook-scheduler
+  [conn make-mesos-driver-fn
+   {:keys [get-mesos-utilization mesos-datomic-mult zk-prefix offer-incubate-time-ms
+           mea-culpa-failure-limit task-constraints riemann-host riemann-port
+           mesos-pending-jobs-atom gpu-enabled? rebalancer-config fenzo-config]
+    :or {get-mesos-utilization (fn [] 0.9) ; can get this from mesos mock later
+;         mesos-datomic-mult (async/mult (async/chan)) ; don't use mult for now
```
dposada: Can we delete this commented line?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+                  instance (first instances)]
+              (is (= (count instances) 1))
+              (is (= (:job/state job-ent) :job.state/completed ))
+              (log/info "job i: " (d/touch job-ent))
+              (is (< 0
+                     (- (.getTime (:instance/end-time instance))
+                        (.getTime (:instance/start-time instance))
+                        (read-string (:job/command job-ent)))
+                     update-status-error-ms))))
+          (dump-jobs-to-csv (d/db mesos-datomic-conn) "trace.csv"))))))
```
dposada: Comments/questions on the csv creation:

- Should we `log/debug` the trace info as well?
- Can we make the csv file optional? I suspect we usually won't care about the csv file
- Can we name the file more descriptively e.g. `trace-basic-scheduling-rebalancing-test.csv`
- Can we add an ignore for `scheduler/trace*.csv`

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+            (let [job-ent (d/entity (d/db mesos-datomic-conn) job-ent-id)
+                  instances (:job/instance job-ent)
+                  instance (first instances)]
+              (is (= (count instances) 1))
+              (is (= (:job/state job-ent) :job.state/completed ))
+              (log/info "job i: " (d/touch job-ent))
+              (is (< 0
+                     (- (.getTime (:instance/end-time instance))
+                        (.getTime (:instance/start-time instance))
+                        (read-string (:job/command job-ent)))
```
dposada: Why are we calling `read-string` on the command?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+                                     :mem {}
+                                     :ports {}}
+               :slave-id "a5cb3212-3023-4546-888d-67cc13eac3dd"}}))
+      (is (= (count (vals (:offer-id->offer new-state))) 2)))))
+
+(deftest combine-ranges
+  (testing "basic test"
+    (is (= (mm/combine-ranges [{:begin 16 :end 20} {:begin 0 :end 2}
+                               {:begin 3 :end 5} {:begin 8 :end 10}]))
+        [{:begin 0 :end 5} {:begin 8 :end 10} {:begin 16 :end 20}])))
```
wyegelwel: That is super embarassing, good catch

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+          scheduler (mesos/scheduler
+                      (registered [this driver framework-id master-info]
+                                  (reset! registered-atom true))
+                      (resource-offers [this driver offers]
+                                       (swap! offer-atom into offers )))
+          hosts [(make-random-host)]
+          speed-multiplier 20
+          mock-driver (mm/mesos-driver-mock hosts speed-multiplier scheduler)]
+      (mesos/start! mock-driver)
+      (Thread/sleep 300) ;; Wait for offers to be sent
```
wyegelwel: Done, thanks this was a great suggestion

wyegelwel: (and by done, I mean when I push next it will be there)

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: This one is a little trickier to remove because of the polling that would need to be done. I used the tracing csv to make close bounds so it shouldn't be too much slower. In tests, I was seeing things complete in ~22-25 seconds and so gave a little bit of a buffer. Can we push this off for a subsequent PR?

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: The int of the command is the run time. It is the default way to set the run time of a task with the mesos mock. I agree this is a little sketchy.

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: Yea, I added this because there was a bug with the integration and I couldn't figure it out without the trace. I will have it write only when log is set debug, rename, and add the ignore. 

Going to skip log/debug because I think it will look noisy in the log.

scheduler/src/cook/mesos.clj
```diff
@@ -75,9 +78,90 @@
...
+              {:failover-timeout mesos-failover-timeout})
+            (when framework-id
+              {:id {:value framework-id}}))
+          mesos-master
+          (when mesos-principal
+            [{:principal mesos-principal}]))))
+
+(defn get-mesos-utilization
+  [mesos-master-hosts]
+  (let [mesos-master-urls (map #(str "http://" % ":5050/metrics/snapshot") mesos-master-hosts)
```
wyegelwel: Yes but I copied this from cook.mesos.scheduler as is. Let's make it configurable in a different PR

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: Yes, thanks for catching that. I added these checks to save me grief when there was a bug upstream but I guess I haven't exercised that one yet.

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: Never mind, I did it

wyegelwel:
Ok, just pushed updates. Thanks for the suggestions, seeing big improvements.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,430 @@
...
+                                                    :version "1.0.1"}))
+         (loop [state {:config {:task->runtime-ms task->runtime-ms
+                                :task->complete-status task->complete-status
+                                :speed-multiplier speed-multiplier}
+                       :slave-id->host (map-from-vals :slave-id hosts)
+                       :offer-id->offer {}
+                       :task-id->task {}
+                       :task-id->completed-task {} ;;TODO test this in kill
+                       :complete-chan->task-id {}}
+                new-offer-chan (async/timeout (/ 1000 speed-multiplier))]
```
dposada: Since we have `(/ 1000 speed-multiplier)` duplicated here and below, can we move it up into the outer `let` and call it something like `new-offer-delay-ms`?

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,430 @@
...
+   `scheduler` is an implementation of the mesomatic scheduler protocol"
+  ([hosts speed-multiplier scheduler]
+   (mesos-driver-mock hosts speed-multiplier default-task->runtime-ms scheduler))
+  ([hosts speed-multiplier task->runtime-ms scheduler]
+   (mesos-driver-mock hosts speed-multiplier task->runtime-ms default-task->complete-status scheduler))
+  ([hosts speed-multiplier task->runtime-ms task->complete-status scheduler]
+   (let [action-chan (async/chan 10) ; This will block calls to the driver. This may be problematic..
+         exit-chan (async/chan)
+         start-chan (async/chan)
+         driver (reify SchedulerDriver
```
dposada: Perhaps move this `reify` out into a separate function

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+    (framework-message [this driver executor-id slave-id data])
+    (disconnected [this driver])
+    (slave-lost [this driver slave-id])
+    (executor-lost [this driver executor-id slave-id status])
+    (error [this driver message])
+    (resource-offers [this driver offers])
+    (status-update [this driver status])))
+
+(deftest make-offer
+  (let [basic-host-info {:hostname "hosta"
```
dposada: Should we create a helper function for creating these host info maps? We've got them in several places and they're kind of big.

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+            :attributes {}
+            :resources {:cpus {"*" 12}
+                        :mem {"*" 120000}
+                        :ports {"*" [{:begin 10000 :end 30000}]}}
+            :available-resources {:cpus {"*" 4.0}
+                                  :mem {"*" 40000.0}
+                                  :ports {"*" [{:begin 10000 :end 12000} {:begin 25001 :end 30000}]}}
+            :slave-id "a5cb3212-3023-4546-888d-67cc13eac3dd"}))))
+
+(defn make-random-host
```
dposada: `make-random-host` sort of implies that the contents of the host are random, when really it's only the `uuid` and `slave-id` that are. Maybe `dummy-host`?

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+          (poll-until #(= (count @task-complete-atom) 2) 20 1000)
+          (mesos/stop! mock-driver)
+          (is @registered-atom)
+          (is (= (count @launched-atom) 2))
+          (is (= (count @task-running-atom) 2))
+          (is (= (count @task-complete-atom) 2)))
+        (catch Throwable t
+          (throw (ex-info "Error while testing launch" atom-map t)))))))
+
+
```
dposada: Nitpick: extra newline

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+  ([zk-port]
+   ;; Copied from src/cook/components
+   ;; TODO: don't copy from components
+   (let [retry-policy (BoundedExponentialBackoffRetry. 100 120000 10)
+         ;; The 180s session and 30s connection timeouts were pulled from a google group
+         ;; recommendation
+         zookeeper-server (org.apache.curator.test.TestingServer. zk-port false)
+         zk-str (str "localhost:" zk-port)
+         curator-framework (CuratorFrameworkFactory/newClient zk-str 180000 30000 retry-policy)
+         ]
```
dposada: `]` can be on previous line

scheduler/test/cook/test/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,844 @@
...
+                                                     :user "b"
+                                                     :command "1000"
+                                                     :custom-executor? false
+                                                     :memory mem
+                                                     :ncpus cpus))
+              update-status-error-ms 500]
+          ;; Transact user "a"s jobs
+          (doall user-a-job-ent-ids)
+          ;; Let them get scheduled
+          (Thread/sleep 5000)
```
dposada: Can we use `poll-until` here as well?

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: I will change where it makes sense. For this function, the offer is built based on the values of the host so we would need to specify everything about the host anyway.

scheduler/src/cook/mesos.clj
```diff
@@ -75,9 +78,92 @@
...
+              {:id {:value framework-id}}))
+          mesos-master
+          (when mesos-principal
+            [{:principal mesos-principal}]))))
+
+(defn get-mesos-utilization
+  "Queries the mesos master to get the utilization.
+   Returns the max of cpu and mem utilization as a decimal (e.g. 0.92)"
+  [mesos-master-hosts]
+  (let [mesos-master-urls (map #(str "http://" % ":5050/metrics/snapshot") mesos-master-hosts)
```
wyegelwel: Sorry, thought I responded to this last time. Yes, but I don't think we should in this PR. This function was moved from another namespace (cook.mesos.rebalancer) because it was nicer from an injection standpoint to have it in cook.mesos. However, changing the functionality of it is unrelated to this pr.

scheduler/test/cook/test/mesos/mesos_mock.clj
wyegelwel: I ended up not changing this anywhere. My reasoning is, for the tests that have the structure shown explicitly, the test will later make assertions based on the particular data in the structure. Specifically, the problem is all the functions are mutating the structure in some way so writing a generic "thing is what I expect function" would be effectively the same as the function I'm trying to test. 

The alternative would be to def the host map but in that case, a change to a value for one test will end up breaking every other test. Also, it would be less clear why the values are the way the are because they are defined elsewhere.

I'm not sure I care so much that if you feel strong about def-ing the host I will do it.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -0,0 +1,430 @@
...
+(defn mesos-driver-mock
+  "A mock mesos implementation.
+
+   It stores the 'hosts' in the cluster, how much resources are available on
+   each host, what jobs are running on each 'host' and how much longer each
+   job is expected to run for.
+
+   Parameters:
+   `hosts` is a seq of maps that contain a hostname, agent id, resources and attributes
+   `speed-multiplier` is a number >1 which is how much to speed up 'time'
```
wyegelwel: The speed multiplier isn't really used yet but the intent with all of this is to write a faster than real time simulator where we can play back what happened in a production cluster in order to test improvements to the scheduling algorithm. The fact that it helps us improve our testing is an added bonus.

In the simulator, being about to pick the speed up factor is important to compare different runs of the simulation.

scheduler/test/cook/test/mesos/mesos_mock.clj
dposada: Response from @wyegelwel:

> Yea, probably but I'm not sure it is worth changing. For the simulator, I will set something that isn't the default and the simulator is really the only use case for this

scheduler/test/cook/test/mesos/mesos_mock.clj
dposada: I completely agree with your points. 

I was thinking something more along the lines of:

```clj
(defn host-info
  "Constructs a host-info map with the given data"
  ([hostname slave-id cpus mem ports]
   (host-info hostname slave-id cpus mem ports nil nil nil))
  ([hostname slave-id cpus mem [ports-begin ports-end] cpus-available
    mem-available [ports-available-begin ports-available-end]]
   {:hostname hostname
    :attributes {}
    :resources {:cpus {"*" cpus}
                :mem {"*" mem}
                :ports {"*" [{:begin ports-begin :end ports-end}]}}
    :available-resources (cond-> {:cpus {}, :mem {}, :ports {}}
                                 cpus-available (assoc :cpus {"*" cpus-available})
                                 mem-available (assoc :mem {"*" mem-available})
                                 (and ports-available-begin
                                      ports-available-end) (assoc :ports {"*" [{:begin ports-available-begin
                                                                                :end ports-available-end}]}))
    :slave-id slave-id}))
```

which you could then use in tests, either with available resources or without:

```clj
;; with available resources
(host-info "hosta" "e8ab3892-3023-4546-888d-67cc13eac3cc" 10 100000 [10000 20000] 10 100000 [10000 20000])

;; without
(host-info "hostb" "a5cb3212-3023-4546-888d-67cc13eac3dd" 12 120000 [10000 20000])
```

This avoids the "thing that I expect" function and the host map `def`, while also preventing a bunch of boilerplate and making the tests proper easier to read. What do you think?

wyegelwel: This removes the boilerplate but then requires looking back at this function to remember what the args mean. I first tried switching to use the `dummy-host` function but it just changed one type of boilerplate for another. 

I propose we merge as is so that we can get the benefit of the tests and I can avoid the code drifting further. I will be continuing to work on this code to build out the simulator and will keep the idea of reducing the boilerplate in mind. Perhaps you can help me brainstorm how to do this out of band.

Issue #192, opened 2016-10-04
Track metadata indicating which application scheduled cook jobs


sgopalan:
Today, the process of figuring out which applications launched jobs running on cook amounts to pattern matching against the name and/or command.

I would like to enable the application scheduling jobs to be able to provide optional "user-agent" metadata that could be used to analyze the source of requests after the fact.

rustyconover:
You should just be able to add this information in a label of the job when you send the job to Cook.  Cook should pass the label or labels through to Mesos.

wyegelwel:
Yes this could be done with a label but it is probably worth making it first class so that systems that are on top of cook can expect it.

@rustyconover, were you imagining a case where a downstream consumer (mesos executor for example) might want to know the "user agent"?

dposada:
@wyegelwel Are you thinking we should add a top-level field to the job description (for example, `application`)? What type of data do we want to know about the application? Maybe `name`, `owner`, `version` as sub-fields within the `application` map?

wyegelwel:
Initially I was thinking just {"application" : "spark", ...} though, you are right it might be useful to have other info. 

@sgopalan, what do you think?

sgopalan:
My initial thoughts were around capturing application and version.  That
would help us to determine 1) the variety of applications launched, and 2)
within one application, what versions are running.

Thanks.
Sri

On Mar 28, 2017 5:42 PM, "wyegelwel" <notifications@github.com> wrote:

Initially I was thinking just {"application" : "spark", ...} though, you
are right it might be useful to have other info.

@sgopalan <https://github.com/sgopalan>, what do you think?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/twosigma/Cook/issues/192#issuecomment-289913904>, or mute
the thread

dposada:
Is application owner useful?

```clj
{
  ...
  "application": {
    "name": "foo",
    "version": "0.1.0",
    "owner": "dposada"
  }
  ...
}
```

sgopalan:
I think application owner is less useful because depending on when you look
at the data, it will be wrong (e.g owner left the company yesterday).

Instead, I envision this feature being utilized in concert with a different
tracking system that manages the details of the application (owner,
department, company, etc).

Thanks.
Sri

On Mar 28, 2017 5:57 PM, "Daniel Posada" <notifications@github.com> wrote:

Is application owner useful?

{
  ...
  "application": {
    "name": "foo",
[...]

PR #278, opened 2017-03-29
Adds an optional application field to the job description

dposada:
Resolves issue https://github.com/twosigma/Cook/issues/192

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -365,15 +365,6 @@
                 (mk-docker-ports docker-id port-mappings))])
       {})))
 
-(defn- build-application
-  "Returns the datoms for adding the provided application"
-  [id {:keys [name version]}]
-  (let [application-id (d/tempid :db.part/user)]
-    [[:db/add id :job/application application-id]
-     {:db/id application-id
-      :application/name name
-      :application/version version}]))
-
 (s/defn make-job-txn
   "Creates the necessary txn data to insert a job into the database"
   [job :- Job]
@@ -415,7 +406,6 @@
                           :label/value v}]))
                     labels)
         container (if (nil? container) [] (build-container user db-id container))
-        application (if (nil? application) [] (build-application db-id application))
         ;; These are optionally set datoms w/ default values
         maybe-datoms (reduce into
                              []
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -170,13 +170,28 @@
...
+   :version (s/constrained s/Str non-empty-max-128-characters-and-alphanum?)})
+
 (def Job
   "A schema for a job"
   {:uuid s/Uuid
    :command s/Str
-   ;; Make sure the job name is a valid string which can only contain '.', '_', '-' or any word characters and has
-   ;; length at most 128.
-   :name (s/both s/Str (s/pred #(re-matches #"[\.a-zA-Z0-9_-]{0,128}" %) 'under-128-characters-and-alphanum?))
+   :name (s/both s/Str (s/pred max-128-characters-and-alphanum? 'max-128-characters-and-alphanum?))
```
wyegelwel: I think it makes sense for name to be non-empty (I know it wasn't)

scheduler/src/cook/mesos/api.clj
```diff
@@ -349,12 +365,21 @@
                 (mk-docker-ports docker-id port-mappings))])
       {})))
 
+(defn- build-application
```
wyegelwel: If you make application a component (), then you can set the value of the key :job/application to the following map

```clojure
{:application/name name
 :application/version version}
```

wyegelwel: For example, see the resource list in make-job-txn

scheduler/src/cook/mesos/api.clj
```diff
@@ -650,79 +677,81 @@
...
+                              base (if reason
+                                     (assoc base
+                                       :reason_code (:reason/code reason)
+                                       :reason_string (:reason/string reason))
+                                     base)]
+                          base))
+                      (:job/instance job))}]
+    (cond-> job-map
+            groups (assoc :groups (map #(str (:group/uuid %)) groups))
+            application (assoc :application {:name (:application/name application)
```
wyegelwel: See [remove-datomic-namespacing](https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/util.clj#L89) so you don't have to do this manually.

scheduler/src/cook/mesos/schema.clj
```diff
@@ -117,6 +117,11 @@
     :db/valueType :db.type/ref
     :db/cardinality :db.cardinality/one
     :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/ident :job/application
+    :db/valueType :db.type/ref
+    :db/cardinality :db.cardinality/one
```
wyegelwel: set `:db/isComponent true` (see comment above)

scheduler/src/cook/mesos/schema.clj
```diff
@@ -332,27 +337,40 @@
...
+    :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/ident :label/value
+    :db/valueType :db.type/string
+    :db/cardinality :db.cardinality/one
+    :db.install/_attribute :db.part/db}
+   ;; Application attributes
+   {:db/id (d/tempid :db.part/db)
+    :db/doc "TODO(DPO)"
```
wyegelwel: TODO

scheduler/src/cook/mesos/schema.clj
```diff
@@ -332,27 +337,40 @@
...
+    :db/ident :application/name
+    :db/valueType :db.type/string
+    :db/cardinality :db.cardinality/one
+    :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/doc "TODO(DPO)"
```
wyegelwel: TODO

scheduler/src/cook/mesos/api.clj
dposada: Agreed, but it seems like we shouldn't break people who are sending empty string names

scheduler/src/cook/mesos/api.clj
dposada: Thanks, will take a look

scheduler/src/cook/mesos/api.clj
dposada: Nice! Will do

scheduler/src/cook/mesos/api.clj
dposada: Done, thanks

dposada:
@wyegelwel I assume we want to update the `JobClient` to allow for specifying the application info as well?

wyegelwel:
Yea it is probably worth doing. A lot of clients use it.

PR #279, opened 2017-03-30
scheduler: Require a reason for changing shares/quotas.

### LINKED ISSUE #272: Add back 'job' as a post option for retry
By switching from 'job' to 'jobs' in /retry we broke backwards compatibility. We should support to avoid breaking use cases as well as supporting retrying multiple jobs

mforsyth:
Fixes ##211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ## [Unreleased]
 ### Added
 ### Changed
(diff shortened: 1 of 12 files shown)


wyegelwel:
@mforsyth , can you update the changelog (https://github.com/twosigma/Cook/blob/master/CHANGELOG.md) to include this feature and note that it intentionally breaks backwards compatibility.

wyegelwel:
Hey Matt, can you update the CHANGELOG to note that this feature intentionally breaks backward compatibility.

PR #280, opened 2017-03-30
Small cleanup in integration tests

dposada:


integration/.gitignore
```diff
@@ -2,3 +2,4 @@
 __pycache__
 *.egg-info
 *.egg*
+venv/
```
dposada: This is a common directory for virtualenv

integration/requirements_dev.txt
```diff
@@ -7,4 +7,5 @@ coverage==4.1
 Sphinx==1.4.8
 cryptography==1.7
 PyYAML==3.11
-
+requests==2.13.0
+retrying==1.3.3
```
dposada: Seems as though we need these since `basic_test.py` imports them

integration/tests/cook/test_basic.py
```diff
@@ -77,13 +80,12 @@ def test_max_runtime_exceeded(self):
         self.assertEqual(2003, job['instances'][0]['reason_code'])
 
     # load a job by UUID using GET /rawscheduler
-    def get_job(self, uuid):
-        return self.session.get('%s/rawscheduler?job=%s' % (self.cook_url, uuid)).json()[0]
+    def get_job(self, job_uuid):
```
dposada: `uuid` shadows the `uuid` library we import

PR #281, opened 2017-03-31
Adds basic docker and minimesos support

dposada:
This is almost entirely copied from PR #195, with some small modifications. I figure it will be good to pull out the Docker parts of that PR into this standalone PR so we can go ahead and start using them.

### DIFF
--- a/scheduler/Dockerfile
+++ b/scheduler/Dockerfile
@@ -1,0 +1,28 @@
+FROM mesosphere/mesos:1.0.1-rc1
+
+RUN apt-get -y update
+RUN apt-get -y install curl
+
+# Env setup
+ENV HOME "/root/"
+ENV LEIN_ROOT true
+
+ENV MESOS_NATIVE_JAVA_LIBRARY /usr/lib/libmesos.so
+
+# Lein Setup
+RUN mkdir $HOME/bin
+ENV PATH $PATH:$HOME/bin
+RUN curl -o $HOME/bin/lein https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein
+RUN chmod a+x $HOME/bin/lein
+RUN lein
+
+RUN mkdir -p /opt/cook
+WORKDIR /opt/cook
+COPY project.clj /opt/cook/
+RUN lein deps
+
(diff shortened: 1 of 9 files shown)


dposada:
@wyegelwel Thoughts on this?

scheduler/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Lein Setup
+RUN mkdir $HOME/bin
+ENV PATH $PATH:$HOME/bin
+RUN curl -o $HOME/bin/lein https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein
+RUN chmod a+x $HOME/bin/lein
+RUN lein
+
+RUN mkdir -p /opt/cook
+WORKDIR /opt/cook
+COPY project.clj /opt/cook/
```
wyegelwel: Why only copy the project.clj file instead of the entire directory?

scheduler/minimesosFile
```diff
@@ -0,0 +1,50 @@
+minimesos {
+    clusterName = "cook"
+    loggingLevel = "INFO"
+    mapAgentSandboxVolume = false
+    mapPortsToHost = false
```
wyegelwel: Can probably set this one to true so that the mesos master is viewable at localhost:5050

scheduler/src/cook/components.clj
```diff
@@ -473,7 +484,7 @@
           literal-config (try
                            {:config
                             (case config-format
-                              "edn" (read-string (slurp config))
```
wyegelwel: Oh god we were eval-ing this the whole time?!

scheduler/Dockerfile
dposada: Fixed

scheduler/minimesosFile
dposada: Done, thanks

scheduler/src/cook/components.clj
dposada: It seems that way!

dposada:
@wyegelwel I addressed your feedback

PR #282, opened 2017-04-03
scheduler: Allow "job" in addition to "jobs" in /retry API

### LINKED ISSUE #272: Add back 'job' as a post option for retry

mforsyth:
Fixes #272

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1093,7 +1093,8 @@
   {:job [s/Uuid]})
 
 (def UpdateRetriesRequest
-  {(s/optional-key :jobs) [s/Uuid]
+  {(s/optional-key :job) s/Uuid
+   (s/optional-key :jobs) [s/Uuid]
    (s/optional-key :retries) PosNum
    (s/optional-key :increment) PosNum})
 
@@ -1102,17 +1103,18 @@
   (:job/max-retries (d/entity (db conn) [:job/uuid (-> ctx ::jobs first)])))
 
 (defn jobs-from-request
-  "compojure-api doesn't support applying a schema to a combination of
-   query parameters and body parameters.  So some manual work is needed to combine
-   ?job=a&job=b with the retries specified in the JSON body payload."
+  "Reads a set of job UUIDs from the request, supporting \"job\" in the query,
+  and either \"job\" or \"jobs\" in the body, while accommodating some aspects of
+  liberator and compojure-api. For example, compojure-api doesn't support
+  applying a schema to a combination of query parameters and body parameters."
   [ctx]
-  (let [query-jobs-coerced (vectorize (get-in ctx [:request :query-params :job]))
-        query-jobs-uncoerced (vectorize (get-in ctx [:request :query-params "job"]))
-        ;; if the query params are coerceable, both :job and "job" keys will be
-        ;; present, but in that case we only want to read the coerced version
-        query-jobs (or query-jobs-coerced (mapv #(UUID/fromString %) query-jobs-uncoerced))
-        body-jobs (get-in ctx [:request :body-params :jobs])]
-    (reduce conj query-jobs body-jobs)))
+  (or (vectorize (get-in ctx [:request :query-params :job]))
+      ;; if the query params are coerceable, both :job and "job" keys will be
+      ;; present, but in that case we only want to read the coerced version
+      (if-let [uncoerced-query-params (get-in ctx [:request :query-params "job"])]
+        (mapv #(UUID/fromString %) (vectorize uncoerced-query-params)))
+      (get-in ctx [:request :body-params :jobs])
+      [(get-in ctx [:request :body-params :job])]))
 
 (defn check-jobs-exist
   [conn ctx]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -1093,7 +1093,8 @@
   {:job [s/Uuid]})
 
 (def UpdateRetriesRequest
-  {(s/optional-key :jobs) [s/Uuid]
+  {(s/optional-key :job) s/Uuid
```
shamsimam: Is there a reason we want to support both `job` and `jobs`? Can't the same be achieved by passing a single job in `jobs` field?

mforsyth: @shamsimam you are right about the design... this is for backwards compatibility.  See the referenced issue https://github.com/twosigma/Cook/issues/272

shamsimam: Ah okay.

I think we should allow only one or the other then and report an error if both fields are specified.

mforsyth: @shamsimam 👍  done.  Added a unit test to prove that "jobs" in conjunction with "job" is considered malformed.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1111,8 +1112,9 @@
         ;; if the query params are coerceable, both :job and "job" keys will be
         ;; present, but in that case we only want to read the coerced version
         query-jobs (or query-jobs-coerced (mapv #(UUID/fromString %) query-jobs-uncoerced))
+        body-job (vectorize (get-in ctx [:request :body-params :job]))
```
shamsimam: Does vectorize belong in `util`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1134,6 +1136,10 @@
       (empty? jobs)
       [true {:error "Need to specify at least 1 job."}]
 
+      (and (get-in ctx [:request :body-params :job])
+           (get-in ctx [:request :body-params :jobs]))
+      [true {:error "Can't specify both \"job\" and \"jobs\"."}]
```
shamsimam: Do we also need to check for job specified in query params?

shamsimam:
Couple of minor comments, else LGTM

scheduler/src/cook/mesos/api.clj
mforsyth: Hmm.  As far as I'm aware it hasn't yet been needed in another namespace.  That is not the only barometer for a function should be moved to a util namespace, but in general, in my opinion, treating a piece of data as "could be either an array or a single value" is not something that we'd want to encourage more of (at least not as a first option) - it's a sign of weak schema (or stretching for backwards compatibility as is being done here).

mforsyth: Thinking more, I think vectorize is particularly related to this api namespace (and to compojure-api in particular) because it was created primarily to interact with the way Compojure coerces http query parameters.   If someone passes "job=x", Compojure will serialize it to a single value, whereas "job=x&job=y" will be serialized to a Vector.

mforsyth: So, since the param I'm handling here is actually part of the body, I probably shouldn't use vectorize here (we know it is going to be a single value).  I will change it.  Thanks for the review, it got me thinking!

scheduler/src/cook/mesos/api.clj
mforsyth: I suppose to be consistent we should also do that.  I can't think of a really good reason why someone would want to specify job UUID's in multiple ways...  so, yes, I'll add checks to make sure it's only specified in 1 of the 3 possible ways

mforsyth:
OK, I re-worked it.

I realized that, if we are going to disallow more than 1 way of specifying job UUIDs from being used in the same request, rather than combining multiple ways into a single list, it lets us simplify (jobs-from-request), so that's what I did.

shamsimam:
LGTM

scheduler/src/cook/mesos/api.clj
```diff
@@ -1102,17 +1103,18 @@
...
-        query-jobs (or query-jobs-coerced (mapv #(UUID/fromString %) query-jobs-uncoerced))
-        body-jobs (get-in ctx [:request :body-params :jobs])]
-    (reduce conj query-jobs body-jobs)))
+  (or (vectorize (get-in ctx [:request :query-params :job]))
+      ;; if the query params are coerceable, both :job and "job" keys will be
+      ;; present, but in that case we only want to read the coerced version
+      (if-let [uncoerced-query-params (get-in ctx [:request :query-params "job"])]
+        (mapv #(UUID/fromString %) (vectorize uncoerced-query-params)))
+      (get-in ctx [:request :body-params :jobs])
+      [(get-in ctx [:request :body-params :job])]))
```
shamsimam: (y)

scheduler/src/cook/mesos/api.clj
```diff
@@ -1102,17 +1103,18 @@
...
-        query-jobs-uncoerced (vectorize (get-in ctx [:request :query-params "job"]))
-        ;; if the query params are coerceable, both :job and "job" keys will be
-        ;; present, but in that case we only want to read the coerced version
-        query-jobs (or query-jobs-coerced (mapv #(UUID/fromString %) query-jobs-uncoerced))
-        body-jobs (get-in ctx [:request :body-params :jobs])]
-    (reduce conj query-jobs body-jobs)))
+  (or (vectorize (get-in ctx [:request :query-params :job]))
+      ;; if the query params are coerceable, both :job and "job" keys will be
+      ;; present, but in that case we only want to read the coerced version
+      (if-let [uncoerced-query-params (get-in ctx [:request :query-params "job"])]
```
wyegelwel: Use `when-let` here since there is only one case

PR #283, opened 2017-04-03
scheduler: Make job name and priority optional in API.

### LINKED ISSUE #268: Make "name" and "priority" in job schema optional
When we schema-ized job, we incorrectly made "name" and "priority" required fields which was a backwards incompatible change. There is no strong reason to make these required and breaking backwards compatibility for it isn't worth it. We should change these back to being optional

mforsyth:
They are still required in the database - if missing from API request, default
values are assigned.

Fixes #268

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -187,12 +187,18 @@
   {:name (s/constrained s/Str non-empty-max-128-characters-and-alphanum?)
    :version (s/constrained s/Str non-empty-max-128-characters-and-alphanum?)})
 
+(s/defschema JobName
+  (s/both s/Str (s/both s/Str (s/pred max-128-characters-and-alphanum? 'max-128-characters-and-alphanum?))))
+
+(s/defschema JobPriority
+  (s/both s/Int (s/pred #(<= 0 % 100) 'between-0-and-100)))
+
 (def Job
   "A schema for a job"
   {:uuid s/Uuid
    :command s/Str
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -198,9 +204,13 @@
   (-> Job
     ;; make max-runtime optional.
     ;; It is *not* optional internally but don't want to force users to set it
+    (dissoc :name)
```
shamsimam: We should have unit tests verifying this behavior change of optional name and priority fields.

mforsyth: @shamsimam done (pushed yesterday)

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -193,6 +193,24 @@
...
+                                :body-params {"jobs" [job-with-defaults]}})
+            get-response (h {:scheme :http
+                             :uri "/rawscheduler"
+                             :authorization/user "mforsyth"
+                             :request-method :get
+                             :query-params {"job" (job-with-defaults "uuid")}})
+            new-job (-> get-response response->body-data first)]
+        (is (= (:status create-response) 201))
+        (is (= (:status get-response) 200))
+        (is (= (new-job "name") "cookjob"))
```
shamsimam: Nitpick: we should co-locate all the defaults for job attributes like name, priority, max runtime, etc. This may be done in a separate PR.

shamsimam:
One minor comment, else LGTM

PR #292, opened 2017-04-11
Run integration tests on travis.

mforsyth:
Where possible, reuse existing code that set up a mesos cluster on Travis for the purpose of running a Simulation.


integration/travis/run_integration.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ev
+
+export PROJECT_DIR=`pwd`
+$PROJECT_DIR/../travis/start_scheduler.sh
+sleep 60
```
dposada: Would it be possible to poll the scheduler to check if it's ready (with a timeout) rather than `sleep`ing for 60 seconds? That way if it's ready after 30 seconds, we can start faster.

me-code-pretty: What is necessitating this sleep call?

mforsyth: @me-code-pretty Cook (running as an external process during this test) just takes a while to start up.  @dposada made a good suggestion of just polling to check the api's availability before starting the tests (this is what is done in the Simulator test).

mforsyth: FYI @dposada  - I believe that the failure you see there was a temporarily inability to download the public python 3 docker image.  Probably temporary network/bandwidth issues, I see that happen on Travis from time to time when downloading dependencies.  It's kind of funny because in this particular run, the process was able to download the same docker image later on in the test run (it's just the attempt to pre-load it that failed), so the actual simulation ran to completion normally...

mforsyth:
I've replaced the sleep with a poll until the connection is no longer refused when attempting to connect to the cook API.

mforsyth:
@dposada Renamed the function.  However, it isn't supposed to be an instance method.  In fact the @retry decorator doesn't appear to be capable of calling an instance method with retry_on_exception.  I moved is_connection_error to avoid confusion

dposada:
LGTM. @wyegelwel should we merge?

Issue #289, opened 2017-04-07
Expected run time


dposada:
If we had the expected run time of jobs, we could make better scheduling decisions

PR #295, opened 2017-04-13
Adds expected-runtime to the Job description

dposada:
Resolves issue #289

### DIFF
--- a/scheduler/docs/scheduler-rest-api.asc
+++ b/scheduler/docs/scheduler-rest-api.asc
@@ -73,6 +73,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`priority` | integer | priority of the job. Should be between 0 and 100, inclusive.
 |`max_retries` | integer | the maximum number of retries
 |`max_runtime` | long | the maximum running time of the job in milliseconds. An instance that runs for more than max_runtime will be killed and job will be retried.
+|`expected_runtime` | long | the (optional) expected running time of the job in milliseconds. If provided, `expected_runtime` must be less than or equal to `max_runtime`. Currently, this value is being collected but not used by the scheduler. In the future, this will allow us to make better scheduling decisions (e.g. don't preempt a job that is expected to complete very soon).
 |`cpus` | double | number of requested cpus.
 |`mem` | double | MB of requested memory.
 |`gpus` | integer | Number of requested GPUs. Must be a whole number and support must be enabled in the config.
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -566,7 +566,7 @@
    new-groups (submitted in the same request as the job). Returns proper Job
    objects, or else throws an exception"
   [db user task-constraints gpu-enabled? new-group-uuids
-   {:keys [cpus mem gpus uuid command priority max-retries max-runtime name
+   {:keys [cpus mem gpus uuid command priority max-retries max-runtime expected-runtime name
            uris ports env labels container group]
     :or {group nil}
     :as job}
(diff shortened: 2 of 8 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -213,31 +221,40 @@
...
+
+(def Job
+  "Full schema for a job"
+  (s/constrained JobMap valid-runtimes?))
+
+(def JobRequestMap
+  "Schema for the fields of a job request"
+  (-> JobMap
+      ;; make max-runtime optional.
+      ;; It is *not* optional internally but don't want to force users to set it
```
pschorf: I know it was already here, but should we remove the "internally" comment?

scheduler/src/cook/mesos/schema.clj
```diff
@@ -56,6 +56,11 @@
     :db/cardinality :db.cardinality/one
     :db.install/_attribute :db.part/db}
    {:db/id (d/tempid :db.part/db)
+    :db/ident :job/expected-runtime
```
pschorf: Can you add a db/doc?

dposada:
@wyegelwel any feedback on this?

scheduler/docs/scheduler-rest-api.asc
```diff
@@ -73,6 +73,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`priority` | integer | priority of the job. Should be between 0 and 100, inclusive.
 |`max_retries` | integer | the maximum number of retries
 |`max_runtime` | long | the maximum running time of the job in milliseconds. An instance that runs for more than max_runtime will be killed and job will be retried.
+|`expected_runtime` | long | the (optional) expected running time of the job in milliseconds. If provided, `expected_runtime` must be less than or equal to `max_runtime`. Currently, this value is being collected but not used by the scheduler. In the future, this will allow us to make better scheduling decisions (e.g. don't preempt a job that is expected to complete very soon).
```
wyegelwel: Consider naming this `expected_runtime_ms` so that it is self documenting what the unit it is

scheduler/src/cook/mesos/api.clj
```diff
@@ -213,31 +221,38 @@
...
-    (dissoc :user)
-    (merge {(s/optional-key :name) JobName
-            (s/optional-key :priority) JobPriority
-            (s/optional-key :uris) [UriRequest]
-            (s/optional-key :env) {s/Keyword s/Str}
-            (s/optional-key :labels) {s/Keyword s/Str}
-            (s/optional-key :max-runtime) PosInt
-            :cpus PosNum
-            :mem PosNum})))
+  (s/constrained JobRequestMap valid-runtimes?))
```
wyegelwel: Why is the separation of JobRequestMap and JobRequest necessary?

scheduler/docs/scheduler-rest-api.asc
dposada: Is that worth it being inconsistent with `max_runtime`, `start_time`, and `end_time`?

scheduler/src/cook/mesos/api.clj
dposada: Because in the `JobResponse` schema def below, we are `dissoc`ing and `merge`ing with the `JobRequestMap`, so the additional `valid-runtimes?` constraint shouldn't be present there.

scheduler/docs/scheduler-rest-api.asc
wyegelwel: I'm not sure. I think at some point it will be worth deprecating those three and adding new ones with units. I'm fairly indifferent to whether we add the unit to expected_runtime now or wait for when we deprecate and rename. Up to you.

dposada: Cool. I would favor adding it now without the `_ms` so that we are consistent.

PR #296, opened 2017-04-13
Avoids lein deps when project.clj hasn't changed

dposada:
This makes building the docker image much faster when you haven't changed `project.clj`.


dposada:
@wyegelwel any feedback?

PR #297, opened 2017-04-14
scheduler: Add /failure_reasons endpoint describing known reasons.

### LINKED ISSUE #209: Create an endpoint to expose the reasons a job may fail
Basically return a json blob of the reasons in the schema.clj file.

mforsyth:
Fixes #209

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1331,6 +1331,36 @@
     :post! (fn [ctx]
              (apply set-limit-fn conn (get-in ctx [:request :body-params :user]) (reduce into [] (::limits ctx))))}))
 
+
+;; /failure-reasons
+
+(s/defschema FailureReasonsResponse
+  [{:code s/Int
+    :name s/Str
+    :description s/Str
+    :mea_culpa s/Bool
+    (s/optional-key :failure_limit) s/Int}])
+
+(defn reason-entity->consumable-map
+  [default-failure-limit e]
+  (cond-> {:code (:reason/code e)
+           :name (name (:reason/name e))
+           :description (:reason/string e)
+           :mea_culpa (:reason/mea-culpa? e)}
+    (:reason/mea-culpa? e)
+    (assoc :failure_limit (or (:reason/failure-limit e) default-failure-limit))))
+
+(defn failure-reasons-handler
+  [conn is-authorized-fn]
+  (base-cook-handler
+   {:allowed-methods [:get]
+    :handle-ok (fn [_]
+                 (let [db (d/db conn)
+                       default-limit (reason/default-failure-limit db)]
+                   (->> (reason/all-known-reasons db)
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -1331,6 +1331,36 @@
...
+    (s/optional-key :failure_limit) s/Int}])
+
+(defn reason-entity->consumable-map
+  [default-failure-limit e]
+  (cond-> {:code (:reason/code e)
+           :name (name (:reason/name e))
+           :description (:reason/string e)
+           :mea_culpa (:reason/mea-culpa? e)}
+    (:reason/mea-culpa? e)
+    (assoc :failure_limit (or (:reason/failure-limit e) default-failure-limit))))
```
dposada: Can we use `remove-datomic-namespacing` in this function?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -670,6 +671,65 @@
...
+                :failure_limit 8}))
+        (is (= (api/reason-entity->consumable-map default-limit mea-culpa-with-default)
+               {:code 4003
+                :name "mesos-container-launch-failed"
+                :description "Container launch failed"
+                :mea_culpa true
+                :failure_limit default-limit}))))
+
+    (testing "reasons-handler"
+      (let [h (basic-handler conn :retry-limit 200)
```
dposada: Maybe `handler` instead of `h`?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -670,6 +671,65 @@
...
+            reasons-response (h {:request-method :get
+                                 :scheme :http
+                                 :uri "/failure_reasons"
+                                 :headers {"Content-Type" "application/json"}
+                                 :authorization/user "dgrnbrg"})
```
dposada: Maybe `"user"`?

scheduler/src/cook/mesos/reason.clj
```diff
@@ -37,3 +37,16 @@
   [db instance]
   (or (:instance/reason instance)
       (reason-code->reason-entity db (:instance/reason-code instance))))
+
+(defn all-known-reasons
```
dposada: Docstring here would be good

scheduler/src/cook/mesos/api.clj
mforsyth: Not sure it's a great match...   of the 6 keys in the schema, only 2 are transformed directly in the manner that (remove-datomic-namespacing) would do it (simply removing the namespace).  I'm not including "mesos-reason" at all (don't think it's relevant to consumers of this API), I'm renaming "string" to "description" (more consumer friendly name for this attribute), and I'm converting "me-culpa?" to "mea_culpa" and "failure-limit" to "failure_limit" (these are more normal for JSON api's)

PR #298, opened 2017-04-17
Add marathon to minimesosFile

pschorf:
Minimesos HEAD expects a marathon to be running and gets an NPE otherwise:
https://github.com/ContainerSolutions/minimesos/blob/master/minimesos/src/main/java/com/containersol/minimesos/cluster/MesosCluster.java#L109

### DIFF
--- a/scheduler/minimesosFile
+++ b/scheduler/minimesosFile
@@ -72,6 +72,8 @@ minimesos {
         loggingLevel = "# INHERIT FROM CLUSTER"
     }
 
+    // Minimesos currently has an NPE without marathon
+    // https://github.com/ContainerSolutions/minimesos/issues/532
     marathon {
         imageName = "mesosphere/marathon"
         imageTag = "v1.3.5"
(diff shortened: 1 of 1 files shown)


scheduler/minimesosFile
```diff
@@ -72,6 +72,16 @@ minimesos {
         loggingLevel = "# INHERIT FROM CLUSTER"
     }
 
+    marathon {
+        imageName = "mesosphere/marathon"
+        imageTag = "v1.3.5"
+
+        // Add 'app { marathonJson = "<path or URL to JSON file>" }' for every task you want to execute
+        app {
+            marathonJson = "https://raw.githubusercontent.com/ContainerSolutions/minimesos/master/opt/apps/weave-scope.json"
```
dposada: Can we avoid starting up weave-scope?

scheduler/minimesosFile
```diff
@@ -72,6 +72,16 @@ minimesos {
         loggingLevel = "# INHERIT FROM CLUSTER"
     }
 
+    marathon {
```
dposada: Can we add a comment about why this is needed even though we don't use it?

scheduler/minimesosFile
pschorf: Done

scheduler/minimesosFile
pschorf: Done

PR #305, opened 2017-04-20
Adds /settings, for retrieving cook's configuration settings

dposada:


dposada:
Do you mean use the path `/config`, or use the "literal config" instead of the parsed settings that are output from `config-settings`? If you mean the literal config, I went with the parsed settings because that seems to be what is consumed by the various parts of the system, and therefore seemed more useful for troubleshooting/inspection.

integration/tests/cook/test_basic.py
```diff
@@ -79,7 +82,8 @@ def test_failing_submit(self):
         self.assertEqual('failed', job['instances'][0]['status'])
 
     def test_max_runtime_exceeded(self):
-        job_uuid, resp = self.submit_job(command='sleep 60', max_runtime=5000)
+        timeout_interval_seconds = self.settings()['task-constraints']['timeout-interval-minutes'] * 60
```
shamsimam: no `get_in(self.settings() ['task-constraints' 'timeout-interval-minutes'])` available :(

shamsimam: Should we be more careful on setting this timeout, as we can end up sleeping for a really long time.

scheduler/src/cook/components.clj
```diff
@@ -67,14 +67,15 @@
...
+  {:scheduler (fnk [mesos-datomic framework-id mesos-pending-jobs-atom [:settings task-constraints mesos-gpu-enabled is-authorized-fn :as settings]]
                 ((lazy-load-var 'cook.mesos.api/main-handler)
                   mesos-datomic
                   framework-id
                   task-constraints
                   mesos-gpu-enabled
                   (fn [] @mesos-pending-jobs-atom)
-                  is-authorized-fn))
+                  is-authorized-fn
+                  settings))
```
shamsimam: We should avoid duplicating sending settings values as parameters, e.g. `task-constraints`, `mesos-gpu-enabled`, `is-authorized-fn`

scheduler/src/cook/mesos/api.clj
```diff
@@ -1397,6 +1400,26 @@
...
                                        default-limit)))))}))
 
+;; /settings
+
+(defn stringify
+  "Converts values to strings as needed for conversion to JSON"
+  [v]
+  (cond
+    (fn? v) (str v)
+    (map? v) (map-vals stringify v)
```
shamsimam: Shouldn't we also stringify the keys?

integration/tests/cook/test_basic.py
dposada: re: `get_in`, I'm going to add it as a util function, coming soon
re: the timeout, `wait_for_job` has `stop_max_delay=120000` so the most we should wait is about 2 minutes

scheduler/src/cook/mesos/api.clj
dposada: cheshire will do that for us:

```clj
(cheshire/generate-string {:foo 1})
=> "{\"foo\":1}"
```

dposada:
@shamsimam I addressed your feedback, thanks

dposada:
@wyegelwel @shamsimam Any additional feedback on this?

wyegelwel:
@dposada, your reasoning for using settings instead of config makes sense. LGTM

dposada:
@shamsimam Any additional feedback on this?

PR #306, opened 2017-04-21
Explicitly configures :mesos :master-hosts

dposada:
Without this, cook will attempt to derive `mesos-master-hosts` from `:mesos` `:master`, which in the case of our minimesos setup, is incorrect, and causes:

```bash
2017-04-21 21:16:05,589 INFO  cook.mesos.rebalancer [async-thread-macro-6] - Rebalance cycle starting
2017-04-21 21:16:05,651 ERROR cook.mesos.rebalancer [async-thread-macro-6] - Rebalance failed
java.net.ConnectException: Connection refused
```

### DIFF
--- a/scheduler/bin/run-docker.sh
+++ b/scheduler/bin/run-docker.sh
@@ -8,12 +8,13 @@ if [ "$(docker ps -aq -f name=${NAME})" ]; then
     docker rm ${NAME}
 fi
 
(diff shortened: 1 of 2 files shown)


shamsimam:
lgtm

dposada:
Can we merge?

Issue #173, opened 2016-09-01
Provide executor for raw scheduler API instead of using default command line executor


WenboZhao:
A few importance features could be implemented by using the customized executor
(1) progress report
(2) consolidated and nice formatted log
(3) reliable outputURL for job
(4) expose outputURL to Spark executor and thus the Spark UI in the driver for easy debugging.

wyegelwel:
I agree, we should move to this. 

What do you mean by (2)?

WenboZhao:
For (2), I meant merge stderr, stdout and possible other logs into one single file for ease of browsing and adding timestamp etc.

wyegelwel:
Another feature could be to capture and send back the exit code of the job. The mesos status update currently doesn't provide this.

sgopalan:
Many users have also requested SFv1-like email notification options for failed, completed jobs.

This is very helpful if you have a long-running job and only want to know if it has broken or it has completed.

sgopalan:
Could this also incorporate the functionality of the task proxy (http://dft-task-proxy.app.twosigma.com:28880) which flattens the raw output into human-readable form?

shamsimam:
The current plan with the custom executor:
1. Provide progress update messages to the scheduler
2. Send the exit code of the job to the scheduler
3. Send the sandbox location of the job to the scheduler

shamsimam:
The Cook Executor is being successfully used as the default executor.

PR #308, opened 2017-04-27
Issue #173 - Cook Executor

shamsimam:
Adds support for the Cook executor that sends sandbox location, exit code and progress updates to the scheduler.

I am still finishing up on the travis config to get all the tests running. Wanted to have this out there to get early reviews on the code.


sradack:
Looks pretty solid after a brief skim.  I will look more tomorrow.

shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/228118292

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+        :param max_message_length: The allowed max message length after encoding
+        :return: whether the message was successfully sent
+    """
+    logging.debug('Sending framework message {}'.format(message))
+    encoded_message = encode_data(str(message).encode('utf8'))
+    if len(encoded_message) < max_message_length:
+        driver.sendFrameworkMessage(encoded_message)
+        return True
+    else:
+        log_message_template = 'Unable to send message {} as it exceeds allowed max length of {}'
```
wyegelwel: Why error instead of truncate?

executor/README.MD
```diff
@@ -0,0 +1,104 @@
...
+ ...}
+```
+
+Supported configuration options:
+
+| option | type | description |
+|--------|------|-------------|
+| `:command` | string | A string containing the command executed on the mesos agent to launch the cook executor. If the executor is installed using the instructions above, the default value of `cook-executor` should not need to be changed. |
+| `:log-level` | string | The log level for the executor process. Defaults to "INFO" |
+| `:max-message-length` | long | The maximum length for the custom failure message set by a task via the executor HTTP API. The default is 512. |
```
sradack: There is no HTTP API

executor/README.MD
```diff
@@ -0,0 +1,104 @@
...
+| `:progress-poll-interval-ms` | long | The interval in ms after which to send progress updates. The default is 1000. |
```
sradack: The name of this implies that we are polling the process for progress.  Suggestion for rename: `:progress-sample-interval-ms`

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+        update_status(driver, task, 'TASK_FAILED')
+
+
+def await_process_completion(driver, task, stop_signal, process_info):
+    """
+        :param driver: The mesos driver to use.
+        :param task: The task to execute.
+        :param stop_signal: Event that determines if an interrupt was sent
+        :param process_info: Tuple of (process, stdout_file, stderr_file)
+        :return: Truw if the process was killed, False if it terminated naturally.
```
wyegelwel: s/Truw/True

executor/cook/executor.py
shamsimam: The message is encoded, truncating the message will lead to error at the receiver's end.

executor/README.MD
shamsimam: I meant Mesos's executor HTTP api which we are using to interact with the scheduler.

executor/README.MD
shamsimam: Will do.

executor/cook/executor.py
shamsimam: Will fix.

executor/cook/executor.py
sradack: We're sending JSON across, presumably to be parsed.  Truncating is going to cause a parse error in the framework.

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+        :param driver: The driver to send the message to.
+        :param message: The raw message to send.
+        :param max_message_length: The allowed max message length after encoding
+        :return: whether the message was successfully sent
+    """
+    logging.debug('Sending framework message {}'.format(message))
+    encoded_message = encode_data(str(message).encode('utf8'))
+    if len(encoded_message) < max_message_length:
+        driver.sendFrameworkMessage(encoded_message)
```
sradack: Is this convention in Mesos?  To encode data as JSON?

shamsimam: Cook already uses json to handle framework messages from the custom executor (fyi, custom executor is different from cook executor).

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+        :param stdout_file: The file to use to redirect stdout.
+        :param stderr_file: The file to use to redirect stderr.
+        :return: The tuple containing the process launched, stdout file, and stderr file.
+    """
+    try:
+        update_status(driver, task, 'TASK_RUNNING')
+
+        command = decode_data(task['data']).decode('utf8')
+        logging.info('Command: {}'.format(command))
+        stdout = open(stdout_file, 'a+')
```
sradack: Is `w+` a more appropriate mode?

shamsimam: During debugging I also used echo to generate output, these get written to stdout. Using the append mode avoided deleting those messages.

In general, the append mode should hurt and will avoid gobbling messages generated from other sources to stdout/stderr files.

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class ProgressWatcher(object):
+    """
+        This class tails the output from the target file listening for progress messages.
+        The retrieve_progress_states generates all progress messages iteratively.
+    """
+    @staticmethod
+    def match_progress_update(input_string):
+        result = re.findall(r'\^\^\^\^JOB-PROGRESS: (\d*)(?: )?(.*)', input_string)
```
wyegelwel: I propose either making this configurable or make it something less crazy and then patching over it for specific formats

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    return process.poll() is None
+
+
+def has_crashed(process):
+    """
+        :param process:
+        :return:
+    """
+    return not is_running(process) and process.poll() is not 0
```
sradack: `is_running` is based upon the result of `process.poll()`.  Should we not call `process.poll()` directly inside `has_crashed` to avoid calling it swiace?

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def tail(self, sleep_time_ms):
+        while True:
+            if os.path.isfile(self.target_file):
+                logging.info('{} has been created, reading contents'.format(self.target_file))
+                break
+            if self.completed_signal.isSet():
+                logging.info('No output has been read to parse progress messages')
+                return
+            logging.info('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
```
wyegelwel: Either make this debug or print it less frequently. This could be very chatty

executor/cook/executor.py
shamsimam: Okay, I can send the regex as a config from the scheduler.

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+        :param driver: The mesos driver to use.
+        :param task: The task to execute.
+        :param process_info: Tuple of (process, stdout_file, stderr_file)
+        :return: Nothing
+    """
+    try:
+        process, _, _ = process_info
+        if is_running(process):
+            logging.info('Killing task')
+            process.kill()
```
sradack: Should we give processes the chance to handle a SIGTERM before sending them SIGKILL?

executor/cook/executor.py
shamsimam: Will do

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    try:
+        task_id = get_task_id(task)
+        sandbox_message = json.dumps({'sandbox-location': sandbox_location, 'task-id': task_id})
+        send_message(driver, sandbox_message, max_message_length)
+
+        process_info = launch_task(driver, task, stdout_file, stderr_file)
+        process, _, _ = process_info
+
+        progress_watcher = ProgressWatcher(stdout_file, completed_signal)
```
wyegelwel: A job should be able to configure where to read progress from as it might not make sense to write to stdout always.

executor/cook/executor.py
shamsimam: Will convert it to debug.

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def __init__(self, target_file, completed_signal):
+        self.target_file = target_file
+        self.progress = None
+        self.completed_signal = completed_signal
+
+    def current_progress(self):
+        return self.progress
+
+    def tail(self, sleep_time_ms):
```
wyegelwel: docstring here

executor/cook/executor.py
shamsimam: okay

executor/cook/executor.py
shamsimam: will add

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+            line = target_file_obj.readline()
+            if not line:
+                if self.completed_signal.isSet():
+                    target_file_obj.close()
+                    break
+                time.sleep(sleep_time_ms / 1000.0)  # sleep briefly
+                continue
+            yield line
+
+    def retrieve_progress_states(self):
```
wyegelwel: docstring here too

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+    def has_enough_time_elapsed_since_last_update(self):
+        if self.last_reported_time is None:
+            return True
+        else:
+            current_time = time.time()
+            time_diff_ms = (current_time - self.last_reported_time) * 1000
+            return time_diff_ms >= self.poll_interval_ms
+
+    def send_progress_update(self, driver, task_id, max_message_length, progress_data, force_send=False):
+        with self.lock:
```
wyegelwel: why is this lock here?

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+            logging.info('Awaiting progress updater completion with timeout of one second')
+            progress_complete_event.wait(1)
+
+        logging.info('Process completed with exit code {}'.format(exit_code))
+        exit_message = json.dumps({'exit-code': exit_code, 'task-id': task_id})
+        send_message(driver, exit_message, max_message_length)
+
+        # send the latest progress state if available
+        latest_progress = progress_watcher.current_progress()
+        progress_updater.send_progress_update(driver, task_id, max_message_length, latest_progress, True)
```
wyegelwel: instead of just `True`, use `force_send=True` so we know what is being set

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+                    if type(progress_data) is dict:
+                        logging.info('Sending progress message {}'.format(progress_data))
+                        message_dict = dict(progress_data)
+                        message_dict['task-id'] = task_id
+                        progress_message = json.dumps(message_dict)
+                        send_message(driver, progress_message, max_message_length)
+                        self.last_reported_time = time.time()
+                        self.last_progress_data = progress_data
+                    else:
+                        logging.error('Progress data {} is not a dictionary!'.format(progress_data))
```
wyegelwel: Where is progress data coming from? Is it correct to say it is a bug in the executor if it isn't a dict?

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+            time_diff_ms = (current_time - self.last_reported_time) * 1000
+            return time_diff_ms >= self.poll_interval_ms
+
+    def send_progress_update(self, driver, task_id, max_message_length, progress_data, force_send=False):
+        with self.lock:
+            if progress_data is not None and self.last_progress_data != progress_data:
+                if force_send or self.has_enough_time_elapsed_since_last_update():
+                    if type(progress_data) is dict:
+                        logging.info('Sending progress message {}'.format(progress_data))
+                        message_dict = dict(progress_data)
```
wyegelwel: why are you making it into a dict if you have just checked it is a dict above?

wyegelwel: Oh, is it a copy?

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+                        message_dict['task-id'] = task_id
+                        progress_message = json.dumps(message_dict)
+                        send_message(driver, progress_message, max_message_length)
+                        self.last_reported_time = time.time()
+                        self.last_progress_data = progress_data
+                    else:
+                        logging.error('Progress data {} is not a dictionary!'.format(progress_data))
+                else:
+                    logging.error('Not sending progress data as enough time has not elapsed since last update')
```
wyegelwel: This will be very chatty if the process produces progress updates much more frequently than we poll_interval_ms

executor/cook/executor.py
```diff
@@ -0,0 +1,364 @@
...
+    def send_progress_update(self, driver, task_id, max_message_length, progress_data, force_send=False):
+        with self.lock:
+            if progress_data is not None and self.last_progress_data != progress_data:
+                if force_send or self.has_enough_time_elapsed_since_last_update():
+                    if type(progress_data) is dict:
+                        logging.info('Sending progress message {}'.format(progress_data))
+                        message_dict = dict(progress_data)
+                        message_dict['task-id'] = task_id
+                        progress_message = json.dumps(message_dict)
+                        send_message(driver, progress_message, max_message_length)
```
wyegelwel: Can we add a way to limit which progress values we send? My concern is that we can potentially get 100 progress updates per process. Perhaps have a way to say which progress values will get sent as part of the config.

executor/cook/executor.py
shamsimam: `send_progress_update` can potentially be called from two threads - the one doing the progress update, the one handling termination and force sending the current progress state.

executor/cook/executor.py
shamsimam: Will do

executor/cook/executor.py
shamsimam: True, I'll remove the check.

integration/travis/scheduler_config.edn
```diff
@@ -16,6 +16,10 @@
          :failover-timeout-ms nil
          :leader-path "/cook-scheduler"
          :role "cook"}
+ :executor {:command "/home/travis/build/twosigma/Cook/travis/cook-executor"
```
wyegelwel: Is it possible to have travis use the uri instead of have it deployed to the machines? This will ensure that code path always works.

executor/cook/executor.py
shamsimam: Yes, it's a copy

executor/cook/executor.py
shamsimam: Changed to debug

executor/cook/executor.py
shamsimam: The `poll_interval_ms` will limit us being too spammy.

integration/travis/scheduler_config.edn
shamsimam: Yes, but this mimics a use case where we want to use a command available on the agent.

The simulator uses an example of downloading it (I couldn't get http uri working, so I manually curl/wget it there).

executor/README.MD
shamsimam: Done.

executor/README.MD
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Done.

shamsimam:
@sradack  @wyegelwel incorporated your feedback into the executor.

shamsimam:
rebased.

integration/tests/cook/test_basic.py
```diff
@@ -237,7 +254,8 @@ def test_change_retries(self):
             '%s/rawscheduler?job=%s' % (self.cook_url, job_uuid)).json()[0]
         self.assertEqual('waiting', job['status'])
         job = self.wait_for_job(job_uuid, 'completed')
-        self.assertEqual('success', job['state'])
+        self.assertEqual('success', job['state'],
```
pschorf: Is it worth refactoring these "assert with messages" calls out to a common method?

integration/travis/scheduler_config.edn
pschorf: Have you gotten http working locally?

integration/tests/cook/test_basic.py
shamsimam: I'm okay extracting these methods, but in general, we have an aversion to extracting the asserts from the unit tests.

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+        If the process does not terminate inside a second, it is then sent a SIGKILL.
+        :param driver: The mesos driver to use.
+        :param task: The task to execute.
+        :param process_info: Tuple of (process, stdout_file, stderr_file)
+        :return: Nothing
+    """
+    try:
+        process, _, _ = process_info
+        if is_running(process):
+            logging.info('Waiting up to one second for process to terminate')
```
sradack: Seems a bit aggressive but OK

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+    except Exception:
+        logging.exception('Error in executing task')
+        update_status(driver, task, 'TASK_FAILED')
+
+    finally:
+        # ensure completed_signal is set
+        completed_signal.set()
+
+
+def run_mesos_driver(stop_signal, max_message_length, progress_sample_interval_ms, sandbox_location, progress_output_name,
```
sradack: The argument list is long.  Is that pythonic?  Is there a more idiomatic way to pass this config data around?

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+    return progress_complete_event
+
+
+class ProgressWatcher(object):
+    """
+        This class tails the output from the target file listening for progress messages.
+        The retrieve_progress_states generates all progress messages iteratively.
+    """
+    @staticmethod
+    def match_progress_update(progress_regex_str, input_string):
```
sradack: Why did you make this static?

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+            logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
+            time.sleep(sleep_time_ms / 1000.0)  # sleep briefly
+
+        target_file_obj = open(self.target_file, 'r')
+        while True:
+            line = target_file_obj.readline()
+            if not line:
+                # no new line available, sleep before trying again
+                if self.completed_signal.isSet():
+                    target_file_obj.close()
```
sradack: Does python automatically reopen the file at the previous seek position?

sradack: Never mind.  The comment placement confused me.

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+                    break
+                time.sleep(sleep_time_ms / 1000.0)  # sleep briefly
```
sradack: How did you get confidence that calling `readline()` again after it previously returned an empty string would give you the next line after data had been added?

executor/cook/executor.py
```diff
@@ -0,0 +1,384 @@
...
+                logging.info('{} has been created, reading contents'.format(self.target_file))
+                break
+            if self.completed_signal.isSet():
+                logging.info('No output has been read to parse progress messages')
+                return
+            logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
+            time.sleep(sleep_time_ms / 1000.0)  # sleep briefly
+
+        target_file_obj = open(self.target_file, 'r')
+        while True:
```
sradack: If I spam you with progress updates, you're going to build up a huge queue of lines to read.  Shouldn't we be ensuring we're reading to the last progress update line during each interval?

sradack: Again I'm wrong.  Disregard.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1334,6 +1365,7 @@
                         ;; TODO: Rescind the offer in fenzo
                         )
        (framework-message [this driver executor-id slave-id message]
+                          (future (handle-framework-message conn message))
```
sradack: How important is `future` here?  Is it not important to ensure framework messages from about the same task a processed in order?

sradack:
I'm done with my comments.

executor/cook/executor.py
shamsimam: I updated two unit tests to verify this behavior: `test_progress_watcher_tail` and `test_manage_task_successful_exit_with_progress_message`

executor/cook/executor.py
shamsimam: For simplified testability

executor/cook/executor.py
shamsimam: I could create a config object and pass that around. Is that preferred?

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Avoided the chance of out-of-order updates by removing the `future`.

executor/cook/executor.py
shamsimam: Mesos uses `MESOS_EXECUTOR_SHUTDOWN_GRACE_PERIOD` to forcefully kill a task. I will change the code to use this value before force killing.

executor/cook/executor.py
shamsimam: Created a `ExecutorConfig` class

executor/cook/executor.py
shamsimam: Use the configured value for grace period during shutdown, defaulting to 1 second.

shamsimam:
I also implemented additional interface methods to log different events which will simplify debugging.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1346,5 +1378,5 @@
        (resource-offers [this driver offers]
                         (receive-offers offers-chan driver offers))
        (status-update [this driver status]
-                      (future (handle-status-update conn driver fenzo status))))
+                      (handle-status-update conn driver fenzo status)))
```
shamsimam: @wyegelwel How important is that the `status-update` processing returns quickly? I checked the metrics, and it takes less than 1 ms except rare spikes around 1ms (I noticed three spikes on the dashboard). If we really need to do it asynchronously, then I'll need to introduce an agent to ensure the in order processing.

wyegelwel:
The reason the future is there is actually a classloader problem. For
reasons I don't know how to investigate, the thread the mesos client
launches on doesn't pick up the full class path but launching a new thread
does. Last time I say the problem it showed up with an exception
clojure/Lang/reflector not found

.travis.yml
```diff
@@ -18,6 +19,7 @@ env:
     - MESOS_NATIVE_JAVA_LIBRARY=/usr/lib/libmesos.so
     - CLJ_HTTP_ASYNC_POOL_TEST_DURATION_MULTIPLIER=5
   matrix:
+    - TEST_DIR=executor DEPS_CMD='python3 setup.py install' TEST_CMD='python3 setup.py nosetests'
```
dposada: For the `TEST_CMD`, would just `nosetests` suffice?

executor/.gitignore
```diff
@@ -0,0 +1,13 @@
+
```
dposada: Nitpick: blank line at the top of the file

executor/README.MD
```diff
@@ -0,0 +1,108 @@
+Cook Executor
```
dposada: Why is the file extension in uppercase?

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+
+The Cook executor is a custom executor written in Python.
+It replaces the default command executor in order to enable a number of features for both operators and end users.
+
+For Users
+---------
+
+The cook executor supports a number of (completely optional) features beyond the standard Mesos command executor.
+The executor sends the following information to the scheduler:
+1. progress update messages of the task (at intervals of `progress-sample-interval-ms` ms, see below)
```
dposada: Maybe "from the task" instead of "of the task"

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+                  :value "file:///path/to/cook-executor"}}
+ ...}
+```
+
+Supported configuration options:
+
+| option | type | description |
+|--------|------|-------------|
+| `:command` | string | A string containing the command executed on the mesos agent to launch the cook executor. If the executor is installed using the instructions above, the default value of `cook-executor` should not need to be changed. |
+| `:log-level` | string | The log level for the executor process. Defaults to "INFO" |
```
dposada: Nitpick: missing period at end of second sentence in description

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+| `:max-message-length` | long | The maximum length for the custom failure message set by a task via the Mesos executor HTTP API. The default is 512. |
+| `:progress-output-name` | string | The file to track for progress updates. The default is stdout. |
+| `:progress-regex-string` | string | The regex used to identify progress update messages. The regex should generates two components in the match, the first being an integer representing the progress percent. The second being a message about the progress. When not provided, progress updates will not be tracked. |
```
dposada: "should generate" instead of "should generates"

dposada: Maybe "should have two capture groups" instead of "should generate two components in the match"

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+| `:progress-sample-interval-ms` | long | The interval in ms after which to send progress updates. The default is 1000. |
+| `:uri` | map | A map of the `uri` to be associated with every the job. |
```
dposada: I don't understand the description for `uri`

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+| `:extract` | boolean | Should the URI be extracted (must be a tar.gz, zipfile, or similar).
+| `:value` | string | The URI to fetch. Supports everything the Mesos fetcher supports, i.e. http://, https://, ftp://, file://, hdfs:// |
+
+### Tests
+
+The cook executor uses `nose`.
+To install test dependencies and run the executor test suite, run:
+
+```bash
+$ python setup.py nosetests
```
dposada: Would just `nosetests` suffice? I believe that's what we have documented for the integration tests

executor/README.MD
```diff
@@ -0,0 +1,108 @@
...
+
+If the executor is not correctly installed on an agent (or if `:executor-command` is not set correctly), all tasks will fail, and there will be a message in the `stderr` file for each task indicating the command the agent attempted to run.
+Verify that the command is the correct command, and that it is available on each agent.
+If not, either fix the `:executor-command` config, or install the executor on the agent.
+
+If the executor is installed correctly, it will log info and errors to the `executor.log` file in the task sandbox.
+This should be the first place to look for stack traces and other error information.
+Each lifecycle callback is logged (`registered`, `launchTask`, etc), which can help to narrow down the issue.
+
+When troubleshooting any issues with running the executor on the agent, it may be helpful to remove the `:job-defaults` entirely until you confirm that executor is being launched correctly.
```
dposada: What are the `:job-defaults`?

executor/bin/build-cook-executor.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ "$(docker ps -aq -f name=${NAME})" ]; then
+    # Cleanup
+    docker rm ${NAME}
+fi
+
+EXECUTOR_DIR="$(dirname ${DIR})"
+
+mkdir -p ${EXECUTOR_DIR}/dist
+docker build -t cook-executor-build -f ${EXECUTOR_DIR}/Dockerfile.build .
```
dposada: Should this use `${NAME}`?

executor/bin/build-cook-executor.sh
```diff
@@ -0,0 +1,16 @@
...
+docker run --name cook-executor-build cook-executor-build
```
dposada: Should this use `${NAME}`?

executor/Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+
+RUN mkdir /opt/cook
+WORKDIR /opt/cook
+
+COPY requirements.txt /opt/cook/
+RUN pip install -r requirements.txt
+
+COPY . /opt/cook
+
+CMD ["pyinstaller", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
```
dposada: Why does this need to happen in a container?

executor/cook/__main__.py
```diff
@@ -0,0 +1,71 @@
...
+
+"""
+
+The primary entry point for Cook's custom executor.
+This module configures logging and starts the executor's driver thread.
+
+"""
+
+import cook.executor as ce
+# https://github.com/pyinstaller/pyinstaller/issues/1113
```
dposada: Would be nice to have a sentence or two explaining what this URL is about

executor/cook/__main__.py
```diff
@@ -0,0 +1,71 @@
...
+
+    executor_id = os.environ.get('MESOS_EXECUTOR_ID', '1')
+    log_level = os.environ.get('EXECUTOR_LOG_LEVEL', 'INFO')
+    max_bytes_read_per_line = 4 * 1024
+    max_message_length = int(os.environ.get('EXECUTOR_MAX_MESSAGE_LENGTH', 512))
+    progress_output_name = os.environ.get('PROGRESS_OUTPUT_FILE', 'stdout')
+    progress_regex_string = os.environ.get('PROGRESS_REGEX_STRING', '')
+    progress_sample_interval_ms = os.environ.get('PROGRESS_SAMPLE_INTERVAL_MS', 1000)
+    sandbox_location = os.environ.get('MESOS_SANDBOX', '')
+    shutdown_grace_period_secs = os.environ.get('MESOS_EXECUTOR_SHUTDOWN_GRACE_PERIOD', '2secs')
```
dposada: Why is "secs" in both the variable name and the value?

.travis.yml
shamsimam: The integration test uses `python setup.py nosetests`. I'm being more explicit with `python3`

executor/.gitignore
shamsimam: Fixed.

executor/README.MD
shamsimam: changed to lower-case.

executor/README.MD
shamsimam: changed as suggested.

executor/README.MD
shamsimam: changed as suggested.

executor/README.MD
shamsimam: changed as suggested.

shamsimam: changed as suggested.

executor/README.MD
shamsimam: Rewritten and referenced the next paragraph.

executor/README.MD
shamsimam: Copied this from the integration readme, which also uses `python setup.py nosetests`

executor/README.MD
shamsimam: removed.

executor/bin/build-cook-executor.sh
shamsimam: Done.

executor/bin/build-cook-executor.sh
shamsimam: Done.

executor/Dockerfile.build
shamsimam: The executable needs to be compiled using the same architecture as where it will be run - on the agents inside docker.

executor/cook/__main__.py
shamsimam: Added, it's a bug in the CPython base_library.zip

executor/cook/__main__.py
shamsimam: removed the `_secs`.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Wil: The reason the future is there is actually a classloader problem. For
reasons I don't know how to investigate, the thread the mesos client
launches on doesn't pick up the full class path but launching a new thread
does. Last time I say the problem it showed up with an exception
clojure/Lang/reflector not found.

@wyegelwel  Did this used to happen on the integration/simulation tests also? Or did it happen only in production?

scheduler/bin/build-docker-image.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
dposada: Nitpick: this file could be its own PR, no?

travis/start_scheduler.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -ev
 
+cd $PROJECT_DIR/../executor
+bin/build-cook-executor.sh
+cp -fv $PROJECT_DIR/../executor/dist/cook-executor $PROJECT_DIR/../travis/
+mkdir -p $PROJECT_DIR/../scheduler/resources/public/
+cp -fv $PROJECT_DIR/../executor/dist/cook-executor $PROJECT_DIR/../scheduler/resources/public/
```
dposada: Why does this need file need to go to both `$PROJECT_DIR/../travis/` and `$PROJECT_DIR/../scheduler/resources/public/`?

scheduler/bin/run-docker.sh
```diff
@@ -23,6 +23,20 @@ else
...
 
+SCHEDULER_DIR="$( dirname ${DIR} )"
+EXECUTOR_DIR="$(dirname ${SCHEDULER_DIR})/executor"
+COOK_EXECUTOR_FILE=${EXECUTOR_DIR}/dist/cook-executor
+SCHEDULER_EXECUTOR_DIR=${SCHEDULER_DIR}/resources/public
+
+if [ ! -f ${COOK_EXECUTOR_FILE} ]; then
+    echo "cook-executor not found at ${COOK_EXECUTOR_FILE}!"
+    echo "Please ensure cook-executor file exists before proceeding."
+    exit 1
```
dposada: Instead of exiting, could we run `build-cook-executor.sh`?

scheduler/src/cook/components.clj
```diff
@@ -328,6 +332,22 @@
                           :or {user-limit-per-m 600}} rate-limit]
                      {:user-limit (->UserRateLimit :user-limit user-limit-per-m (t/minutes 1))}))
      :sim-agent-path (fnk [] "/usr/bin/sim-agent")
+     :executor (fnk [[:config {executor {}}]]
+                 (when (str/blank? (:command executor))
+                   (throw (ex-info "Executor command is missing!" {:executor executor})))
```
dposada: Seems like we shouldn't require the `:executor` config to be present, for two reasons:

1. It breaks backwards compatibility
1. It couples deployment of this change to deployment of the executor on all agents in the Mesos cluster(s)

scheduler/src/cook/mesos/api.clj
```diff
@@ -173,7 +173,11 @@
    (s/optional-key :reason_code) s/Int
    (s/optional-key :output_url) s/Str
    (s/optional-key :cancelled) s/Bool
-   (s/optional-key :reason_string) s/Str})
+   (s/optional-key :reason_string) s/Str
+   (s/optional-key :exit_code) s/Int
+   (s/optional-key :progress) s/Int
+   (s/optional-key :progress_message) s/Str
```
dposada: Should these be `:last_progress` and `:last_progress_message`?

shamsimam: I followed how `status`, another field that can mutate over time, was named.

integration/tests/cook/test_basic.py
```diff
@@ -77,13 +77,29 @@ def test_basic_submit(self):
...
         job_uuid, resp = self.submit_job(command='exit 1')
         self.assertEqual(201, resp.status_code)
         job = self.wait_for_job(job_uuid, 'completed')
         self.assertEqual(1, len(job['instances']))
         self.assertEqual('failed', job['instances'][0]['status'])
+        self.assertEqual(1, job['instances'][0]['exit_code'])
+        self.assertTrue(bool(job['instances'][0]['sandbox']))
+
+    def test_progress_update_submit(self):
+        command = 'echo "^^^^JOB-PROGRESS: 25 Twenty-five percent"; sleep 1; exit 0'
```
shamsimam: Will do.

executor/cook/executor.py
```diff
@@ -0,0 +1,432 @@
...
+        :return: whether the message was successfully sent
+    """
+    logging.debug('Sending framework message {}'.format(message))
+    encoded_message = encode_data(str(message).encode('utf8'))
+    if len(encoded_message) < max_message_length:
+        driver.sendFrameworkMessage(encoded_message)
+        return True
+    else:
+        log_message_template = 'Unable to send message {} as it exceeds allowed max length of {}'
+        logging.info(log_message_template.format(message, max_message_length))
```
shamsimam: The message is actually a json string, so stripping to the first few characters won't work.

executor/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+| option | type | description |
+|--------|------|-------------|
+| `:command` | string | A string containing the command executed on the mesos agent to launch the cook executor. If the executor is installed using the instructions above, the default value of `cook-executor` should not need to be changed.|
+| `:log-level` | string | The log level for the executor process. Defaults to "INFO".|
+| `:max-message-length` | long | The maximum length for the custom failure message set by a task via the Mesos executor HTTP API. The default is 512.|
+| `:progress-output-name` | string | The file to track for progress updates. The default is stdout.|
+| `:progress-regex-string` | string | The regex used to identify progress update messages. The regex should have two capture groups, the first being an integer representing the progress percent. The second being a message about the progress. When not provided, progress updates will not be tracked.|
+| `:progress-sample-interval-ms` | long | The interval in ms after which to send progress updates. The default is 1000.|
+| `:uri` | map | A description of the `uri` to be associated with every the job. The `uri` structure is defined below.|
```
shamsimam: Rewritten.

shamsimam: Rewritten.

executor/cook/executor.py
```diff
@@ -0,0 +1,432 @@
...
+    logging.info('Closing {}'.format(stdout.name))
+    stdout.close()
+    logging.info('Closing {}'.format(stderr.name))
+    stderr.close()
+
+
+def kill_task(driver, task, process_info, shutdown_grace_period_ms):
+    """
+        Attempts to kill a process by sending it a SIGTERM.
+        If the process does not terminate inside a second, it is then sent a SIGKILL.
```
shamsimam: Done.

executor/cook/executor.py
```diff
@@ -0,0 +1,432 @@
...
+        progress_updater.send_progress_update(driver, task_id, config.max_message_length, latest_progress,
+                                              force_send=True)
+
+        if not process_killed:
+            task_state = 'TASK_FINISHED' if exit_code == 0 else 'TASK_FAILED'
+            update_status(driver, task_id, task_state)
+
+    except Exception:
+        logging.exception('Error in executing task')
+        update_status(driver, task_id, 'TASK_FAILED')
```
shamsimam: Done.

executor/cook/executor.py
```diff
@@ -0,0 +1,432 @@
...
+                if force_send or self.has_enough_time_elapsed_since_last_update():
+                    logging.info('Sending progress message {}'.format(progress_data))
+                    message_dict = dict(progress_data)
+                    message_dict['task-id'] = task_id
+                    progress_message = json.dumps(message_dict)
+                    send_message(driver, progress_message, max_message_length)
+                    self.last_reported_time = time.time()
+                    self.last_progress_data = progress_data
+                else:
+                    logging.debug('Not sending progress data as enough time has not elapsed since last update')
```
shamsimam: Done.

executor/tests/test_executor.py
```diff
@@ -0,0 +1,787 @@
...
+        os.remove(stdout_name)
+    if os.path.isfile(stderr_name):
+        with open(stderr_name) as f:
+            logging.debug('==========================================')
+            logging.debug('Contents of {}:'.format(stderr_name))
+            logging.debug(f.read())
+        os.remove(stderr_name)
+
+
+def test_get_task_id():
```
shamsimam: Used a class that inherits from unittest.TestCase

executor/tests/test_executor.py
```diff
@@ -0,0 +1,787 @@
...
+        actual_encoded_message_0 = driver.messages[0]
+        expected_message_0 = {'sandbox-location': sandbox_location, 'task-id': task_id}
+        assert_message(expected_message_0, actual_encoded_message_0)
+
+        actual_encoded_message_1 = driver.messages[1]
+        expected_message_1 = {'exit-code': 1, 'task-id': task_id}
+        assert_message(expected_message_1, actual_encoded_message_1)
+
+    command = 'echo "Hello World"; exit 1'
+    manage_task_runner(command, assertions)
```
shamsimam: Done.

scheduler/bin/build-docker-image.sh
shamsimam: Created a separate PR: https://github.com/twosigma/Cook/pull/327

scheduler/bin/run-docker.sh
shamsimam: Done.

travis/start_scheduler.sh
shamsimam: It is to test both variants of usage. The integration tests use the path to command directly while the simulation tests download the executor from the cook-scheduler server.

scheduler/src/cook/components.clj
shamsimam: Added changes to make the executor backwards compatible.

executor/cook/executor.py
dposada: I mean in the `logging.info` call, not in the framework message

travis/start_scheduler.sh
dposada: :+1: -- can we add a comment indicating this?

shamsimam:
Passing executor tests: https://travis-ci.org/twosigma/Cook/builds/230797079

integration/travis/run_integration.sh
```diff
@@ -2,16 +2,7 @@
...
-# Start minimesos
-cd ${PROJECT_DIR}/../travis
-./minimesos up
-
-# Start cook scheduler
-cd ${PROJECT_DIR}/../scheduler
-## on travis, ports on 172.17.0.1 are bindable from the host OS, and are also
-## available for processes inside minimesos containers to connect to
-LIBPROCESS_IP=172.17.0.1 lein run ${PROJECT_DIR}/travis/scheduler_config.edn &
+$PROJECT_DIR/../travis/start_scheduler.sh
```
dposada: Doesn't `$PROJECT_DIR/../travis/start_scheduler.sh` start a datomic transactor? If so, we don't want that here. We want datomic to run in memory for the integration tests.

shamsimam: I'll include support for a command line argument to explicitly avoif starting a datomic trasactor

shamsimam: Done.

integration/travis/run_integration.sh
```diff
@@ -2,7 +2,7 @@
 set -ev
 
 export PROJECT_DIR=`pwd`
-$PROJECT_DIR/../travis/start_scheduler.sh
+$PROJECT_DIR/../travis/start_scheduler.sh no-transactor
```
dposada: Should we make this `--no-transactor`?

shamsimam:
@dposada added the change where `output_url` uses `sandbox` location when available.

shamsimam:
Ready for another round of reviews, passing build: https://travis-ci.org/twosigma/Cook/builds/241902140

integration/tests/cook/util.py
```diff
@@ -20,6 +21,28 @@ def get_in(dct, *keys):
     return dct
 
 
+def dict_as_string(dct):
```
dposada: Why not just use `json.dumps`?

scheduler/example-prod-config.edn
```diff
@@ -19,6 +19,11 @@
                                 :memory-gb 96
                                 :retry-limit 15
                                 :cpus 20}}
+ :executor {:command "./cook-executor"
+            :uri {:cache true
+                  :executable true
+                  :extract false
+                  :value #config/env "COOK_EXECUTOR"}}
```
dposada: Elsewhere in this file, we use actual (fake) values, instead of environment variables. Maybe we should follow that pattern here

scheduler/src/cook/mesos/api.clj
```diff
@@ -716,16 +720,51 @@
   "Takes the executor-id->sandbox-directory from the agent json and constructs a URL to query it. Hardcodes fun
    stuff like the port we run the agent on. Users will need to add the file path & offset to their query. Refer to
    the 'Using the output_url' section in docs/scheduler-rest-api.asc for further details."
-  [fid agent-hostname executor-id]
+  [framework-id agent-hostname executor-id sandbox-location]
```
dposada: Why not `sandbox-directory`? Also, should we add a comment to the docstring about why this may or may not be `nil`?

scheduler/src/cook/mesos/task.clj
```diff
@@ -14,33 +14,57 @@
...
+                       cook-executor? :cook-executor
+                       ;; use mesos' command executor by default
+                       :else :command-executor)
         data (.getBytes
-               (pr-str
-                 {:instance (str (count (:job/instance job-ent)))})
+               (if cook-executor?
+                 (json/write-str {"command" (:job/command job-ent)})
+                 (pr-str
+                   ;;TODO this data is a race-condition
```
dposada: Do we need to resolve this TODO?

simulator/travis/simulator_config.edn
```diff
@@ -23,7 +23,7 @@
                               :ceiling 120}
                :job-memory {:mean 512
                             :std-dev 200
-                            :floor 1
+                            :floor 256
```
dposada: Why was this change needed?

integration/tests/cook/test_basic.py
```diff
@@ -27,12 +27,18 @@ def get_job(self, job_uuid):
         """Loads a job by UUID using GET /rawscheduler"""
         return util.query_jobs(self.cook_url, job=[job_uuid]).json()[0]
 
+    def is_using_cook_executor(self):
+        return util.get_in(self.settings(), 'executor', 'command') is not None
```
shamsimam: The cook executor is triggered only if the executor command is available.

Checks at:
https://github.com/twosigma/Cook/pull/308/files#diff-74ded9cfa2ff7c2832fa8c4b22ea40c6R340
https://github.com/twosigma/Cook/pull/308/files#diff-04ee7917c0d3616697cecca6d5c4c071R44

integration/tests/cook/test_basic.py
```diff
@@ -51,6 +57,39 @@ def test_failing_submit(self):
...
+        self.assertEqual(201, resp.status_code)
+        job = util.wait_for_job(self.cook_url,
+                                job_uuid, 'completed')
+        self.assertEqual(1, len(job['instances']))
+        self.assertEqual('success', job['instances'][0]['status'])
+        if self.is_using_cook_executor():
+            self.assertEqual(0, job['instances'][0]['exit_code'])
+            self.assertEqual(75, job['instances'][0]['progress'])
+            self.assertEqual('Seventy-five percent', job['instances'][0]['progress_message'])
+            self.assertTrue(bool(job['instances'][0]['sandbox']))
```
shamsimam: Okay, I will add tests for these cases.

integration/tests/cook/util.py
shamsimam: Thanks for pointing this out, `dumps` supports a `sort_keys` flag too :)

scheduler/example-prod-config.edn
shamsimam: Done.

scheduler/src/cook/mesos/api.clj
shamsimam: Done.

scheduler/src/cook/mesos/task.clj
shamsimam: This TODO is in existing code and is unaffected by the cook-executor changes.

simulator/travis/simulator_config.edn
shamsimam: The cook-executor needs some amount of memory to run, 1mb is too little for it.

integration/tests/cook/util.py
shamsimam: Done.

integration/tests/cook/test_basic.py
shamsimam: Added tests for the first three cases

shamsimam:
This branch is ready for another round of review :)

scheduler/src/cook/mesos/task.clj
```diff
@@ -14,33 +14,57 @@
...
+(def cook-executor-source "cook_scheduler_executor")
 (defonce custom-executor-name "cook_agent_executor")
 (defonce custom-executor-source "cook_scheduler")
 
+(defn build-executor-environment
+  "Build the environment for the cook executor."
+  [{:keys [log-level max-message-length progress-output-name progress-regex-string progress-sample-interval-ms]}]
+  {"EXECUTOR_LOG_LEVEL" log-level
+   "EXECUTOR_MAX_MESSAGE_LENGTH" max-message-length
+   "PROGRESS_OUTPUT_FILE" progress-output-name
```
shamsimam: renamed to `progress-output-file`

scheduler/src/cook/mesos/schema.clj
```diff
@@ -548,6 +553,18 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
     :db/valueType :db.type/boolean
     :db/cardinality :db.cardinality/one
     :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/ident :instance/sandbox-directory
+    :db/doc "represents the sandbox directory of the instance on the executor"
```
shamsimam: Done

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1351,13 +1383,13 @@
...
    fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset fenzo-fitness-calculator
-   task-constraints gpu-enabled? good-enough-fitness {:keys [match-trigger-chan rank-trigger-chan] :as trigger-chans}]
+   task-constraints executor gpu-enabled? good-enough-fitness {:keys [match-trigger-chan rank-trigger-chan] :as trigger-chans}]
 
   (persist-mea-culpa-failure-limit! conn mea-culpa-failure-limit)
 
   (let [fid (atom nil)
         fenzo (make-fenzo-scheduler driver-atom offer-incubate-time-ms fenzo-fitness-calculator good-enough-fitness)
-        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset match-trigger-chan)]
+        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid executor pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset match-trigger-chan)]
```

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -271,6 +273,35 @@
...
+(defn handle-framework-message
+  "Processes a framework message from Mesos."
+  [conn framework-message]
+  (timers/time!
+    handle-framework-message-duration
+    (try
+      (let [db (db conn)
+            {:strs [exit-code progress-message progress-percent sandbox-directory task-id] :as message}
+            (-> (String. ^bytes framework-message "UTF-8") (json/read-str))
+            _ (log/info "Received framework message:" {:task-id task-id, :message message})
```
shamsimam: Changed to `debug`

scheduler/src/cook/mesos.clj
```diff
@@ -211,29 +214,30 @@
...
-                                          fenzo-scaleback
-                                          fenzo-floor-iterations-before-warn
-                                          fenzo-floor-iterations-before-reset
-                                          fenzo-fitness-calculator
-                                          task-constraints
-                                          gpu-enabled?
-                                          good-enough-fitness
-                                          trigger-chans)
-                                        driver (make-mesos-driver-fn scheduler framework-id)]
+        (sched/create-datomic-scheduler
```
shamsimam: Formatted

scheduler/src/cook/components.clj
```diff
@@ -330,6 +334,27 @@
...
+                 (if (seq executor)
+                   (do
+                     (when (str/blank? (:command executor))
+                       (throw (ex-info "Executor command is missing!" {:executor executor})))
+                     (when (and (:uri executor) (nil? (get-in executor [:uri :value])))
+                       (throw (ex-info "Executor uri value is missing!" {:executor executor})))
+                     (let [default-executor-config {:log-level "INFO"
+                                                    :max-message-length 512
+                                                    :progress-output-name "stdout"
+                                                    :progress-regex-string "\\^\\^\\^\\^JOB-PROGRESS: (\\d*)(?: )?(.*)"
```
shamsimam: Done

scheduler/src/cook/components.clj
```diff
@@ -330,6 +334,27 @@
...
+     :executor (fnk [[:config {executor {}}]]
+                 (if (seq executor)
+                   (do
+                     (when (str/blank? (:command executor))
+                       (throw (ex-info "Executor command is missing!" {:executor executor})))
+                     (when (and (:uri executor) (nil? (get-in executor [:uri :value])))
+                       (throw (ex-info "Executor uri value is missing!" {:executor executor})))
+                     (let [default-executor-config {:log-level "INFO"
+                                                    :max-message-length 512
+                                                    :progress-output-name "stdout"
```
shamsimam: renamed to `progress-output-file`

scheduler/src/cook/components.clj
```diff
@@ -82,7 +83,7 @@
                    (route/not-found "<h1>Not a valid route</h1>")))})
 
 (def mesos-scheduler
-  {:mesos-scheduler (fnk [[:settings mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset fenzo-fitness-calculator task-constraints riemann mesos-gpu-enabled rebalancer good-enough-fitness] mesos-datomic mesos-datomic-mult curator-framework mesos-pending-jobs-atom mesos-offer-cache]
+  {:mesos-scheduler (fnk [[:settings executor mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset fenzo-fitness-calculator task-constraints riemann mesos-gpu-enabled rebalancer good-enough-fitness] mesos-datomic mesos-datomic-mult curator-framework mesos-pending-jobs-atom mesos-offer-cache]
```
shamsimam: Done

scheduler/example-prod-config.edn
```diff
@@ -19,6 +19,11 @@
                                 :memory-gb 96
                                 :retry-limit 15
                                 :cpus 20}}
+ :executor {:command "./cook-executor"
+            :uri {:cache true
+                  :executable true
+                  :extract false
+                  :value "http://example.com:12321  /resources/cook-executor"}}
```
shamsimam: Done

integration/tests/cook/test_basic.py
```diff
@@ -60,11 +122,16 @@ def test_max_runtime_exceeded(self):
         job_timeout_interval_seconds = (2 * settings_timeout_interval_minutes * 60) + 15
         job_uuid, resp = util.submit_job(self.cook_url, command='sleep %s' % job_timeout_interval_seconds, max_runtime=5000)
         self.assertEqual(201, resp.status_code)
-        job = util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_seconds * 1000)
+        util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_seconds * 1000)
+        time.sleep(2) # allow update for exit_code
```
shamsimam: Added `wait_for_exit_code` flag to wait_for_job.

dposada:
A couple of things we should consider adding:

1. The ability to gradually roll out the cook executor (e.g. day 1, 2% of instances use it; day 2, 20%; day 3, 80%; day 4, 100%)
1. The ability to force using (or not using) the cook executor from the client

shamsimam:
This branch is ready for another round of review :)

shamsimam:
This branch is ready for another round of review :)

Passing build: https://travis-ci.org/twosigma/Cook/builds/245675568

shamsimam:
Rebased, this branch is ready for another round of review :)

Passing build: https://travis-ci.org/twosigma/Cook/builds/246074240

integration/tests/cook/test_basic.py
```diff
@@ -48,7 +52,62 @@ def test_failing_submit(self):
...
         self.assertEqual(1, len(job['instances']))
-        self.assertEqual('failed', job['instances'][0]['status'])
+        message = json.dumps(job['instances'][0], sort_keys=True)
+        self.assertEqual('failed', job['instances'][0]['status'], message)
+        if util.is_using_cook_executor(self.cook_url):
+            self.assertEqual(1, job['instances'][0]['exit_code'], message)
+            self.assertTrue(bool(job['instances'][0]['sandbox_directory']), message)
+
+    def test_progress_update_submit(self):
+        command = 'echo "^^^^JOB-PROGRESS: 25 Twenty-five percent"; sleep 1; exit 0'
```
sradack: Please use the default status pattern instead of a custom pattern.

shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/246493771

shamsimam:
@dposada @pschorf @wyegelwel any feedback on this?

scheduler/src/cook/components.clj
```diff
@@ -83,8 +81,11 @@
                    (route/not-found "<h1>Not a valid route</h1>")))})
 
 (def mesos-scheduler
-  {:mesos-scheduler (fnk [[:settings mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset fenzo-fitness-calculator task-constraints riemann mesos-gpu-enabled rebalancer good-enough-fitness] mesos-datomic mesos-datomic-mult curator-framework mesos-pending-jobs-atom mesos-offer-cache mesos-leadership-atom]
-
+  {:mesos-scheduler (fnk [[:settings executor fenzo-fitness-calculator fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-warn
```
wyegelwel: Was this just a sort and organizing or were things added to this list? It is hard to tell.

scheduler/src/cook/mesos.clj
```diff
@@ -178,14 +178,17 @@
...
    riemann-host             -- str, dns name of riemann
    riemann-port             -- int, port for riemann
    mesos-pending-jobs-atom  -- atom, Populate (and update) list of pending jobs into atom
    offer-cache              -- atom, map from host to most recent offer. Used to get attributes
    gpu-enabled?             -- boolean, whether cook will schedule gpus
    rebalancer-config        -- map, config for rebalancer. See scheduler/docs/rebalancer-config.asc for details
    fenzo-config             -- map, config for fenzo, See scheduler/docs/configuration.asc for more details"
-  [make-mesos-driver-fn get-mesos-utilization curator-framework mesos-datomic-conn mesos-datomic-mult zk-prefix offer-incubate-time-ms mea-culpa-failure-limit task-constraints riemann-host riemann-port mesos-pending-jobs-atom offer-cache gpu-enabled? rebalancer-config mesos-leadership-atom
+  [make-mesos-driver-fn get-mesos-utilization curator-framework mesos-datomic-conn mesos-datomic-mult zk-prefix offer-incubate-time-ms
+   mea-culpa-failure-limit task-constraints executor riemann-host riemann-port mesos-pending-jobs-atom offer-cache gpu-enabled? rebalancer-config mesos-leadership-atom
```
wyegelwel: Why put `executor` after `task-constraints`? It took me a while to find where the executor was added

scheduler/src/cook/components.clj
shamsimam: Executor was added.

organizing: I broke it up across multiple lines as @sradack requested it.

sort: I introduced the sort so that it is easier to find it the next time.

scheduler/src/cook/mesos.clj
shamsimam: `executor` was added. Where would you prefer to place the parameter?

organizing: I broke it up across multiple lines as @sradack requested it.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -272,6 +274,35 @@
...
+          (let [txns (cond-> []
+                              exit-code (conj [:db/add instance :instance/exit-code (int exit-code)])
+                              progress-message (conj [:db/add instance :instance/progress-message (str progress-message)])
+                              progress-percent (conj [:db/add instance :instance/progress (int progress-percent)])
+                              sandbox-directory (conj [:db/add instance :instance/sandbox-directory (str sandbox-directory)]))]
+            (when (seq txns)
+              (log/info "Updating instance" instance "to" txns)
+              (transact-with-retries conn txns)))))
+      (catch Exception e
+        (log/error e "Mesos scheduler framework message error")))))
```
wyegelwel: This can cause a lot of noise if this is an "expected" error. When would this come up?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -272,6 +274,35 @@
...
+  (timers/time!
+    handle-framework-message-duration
+    (try
+      (let [db (db conn)
+            {:strs [exit-code progress-message progress-percent sandbox-directory task-id] :as message}
+            (-> (String. ^bytes framework-message "UTF-8") (json/read-str))
+            _ (log/debug "Received framework message:" {:task-id task-id, :message message})
+            _ (when (str/blank? task-id)
+                (throw (ex-info "task-id is empty in framework message" {:message message})))
+            instance (ffirst (q '[:find ?i
```
wyegelwel: You can use a look-up ref (`[:instance/task-id task-id]`) with an entity like `(d/entity db [:instance/task-id task-id])` which I think is cleaner the query. Alternatively (or also) consider making this a function.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: The error should be a rare scenario, e.g. if the transaction failed. We do not expect to have issues with parsing and validating the json data.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Sweet, thanks! Done.

shamsimam:
@dpo I agree with https://github.com/twosigma/Cook/pull/308#issuecomment-309789414

Will implement them in a separate PR.

.travis.yml
```diff
@@ -19,15 +19,16 @@ env:
...
+    - COOK_EXECUTOR='1' TEST_DIR=integration DEPS_CMD='travis/prepare_integration.sh' TEST_CMD='travis/run_integration.sh'
+    - COOK_EXECUTOR='0' TEST_DIR=integration DEPS_CMD='travis/prepare_integration.sh' TEST_CMD='travis/run_integration.sh'
+    - COOK_EXECUTOR='1' TEST_DIR=integration DEPS_CMD='travis/prepare_integration.sh' TEST_CMD='travis/run_integration.sh explicit'
 matrix:
     fast_finish: true
     allow_failures:
       - env: TEST_DIR=integration DEPS_CMD='travis/prepare_integration.sh' TEST_CMD='travis/run_integration.sh explicit'
 before_script: (cd $TEST_DIR && $DEPS_CMD)
-script: cd $TEST_DIR && $TEST_CMD
+script: cd $TEST_DIR && COOK_EXECUTOR=$COOK_EXECUTOR $TEST_CMD
```
shamsimam: removed.

integration/tests/cook/test_basic.py
```diff
@@ -33,6 +34,9 @@ def test_basic_submit(self):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertEqual('success', job['instances'][0]['status'])
         self.assertEqual(False, job['disable_mea_culpa_retries'])
+        if util.is_using_cook_executor(self.cook_url):
+            self.assertEqual(0, job['instances'][0]['exit_code'])
+            self.assertTrue(bool(job['instances'][0]['sandbox_directory']))
```
shamsimam: Used `assertIsNotNone` to make the intent clearer

integration/tests/cook/test_basic.py
```diff
@@ -50,7 +54,62 @@ def test_failing_submit(self):
...
+    def test_multiple_progress_updates_submit(self):
+        command = 'echo "progress: 25 Twenty-five percent" && sleep 2 && '\
+                  'echo "progress: 50 Fifty percent" && sleep 2 && ' \
+                  'echo "progress: 75 Seventy-five percent" && sleep 2 && ' \
+                  'echo "progress: Eighty percent invalid format" && sleep 2 && ' \
+                  'echo "Done" && exit 0'
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, max_runtime=60000)
+        self.assertEqual(201, resp.status_code)
+        job = util.wait_for_job(self.cook_url,
+                                job_uuid, 'completed')
```
shamsimam: Done.

integration/tests/cook/util.py
```diff
@@ -100,21 +105,41 @@ def submit_job(cook_url, **kwargs):
     return job_spec['uuid'], resp
 
 
-def wait_for_job(cook_url, job_id, status, max_delay=120000):
+def load_job(cook_url, job_id):
```
shamsimam: `get_job` is in `test_basic`. I have removed `get_job` and used `util.load_job` instead.

integration/tests/cook/util.py
```diff
@@ -100,21 +105,41 @@ def submit_job(cook_url, **kwargs):
...
-            logger.info('Job %s has status %s - %s', job_id, status, job)
+            logger.info('Job {} has exit_code {}'.format(job_id, job['instances'][0]['exit_code']))
             return job
 
-    return wait_for_job_inner()
+    if wait_for_exit_code:
+        wait_for_job_inner()
+        return wait_for_exit_code_inner()
+    else:
+        return wait_for_job_inner()
```
shamsimam: Created `wait_for_exit_code`

integration/travis/run_integration.sh
```diff
@@ -1,10 +1,24 @@
...
 
 NOSE_ATTRIBUTES=${1:-'!explicit'}
 
 export PROJECT_DIR=`pwd`
 
+CONFIG_FILE="scheduler_config_command_executor.edn"
+if [ "${COOK_EXECUTOR}" = "1" ]
+then
+  CONFIG_FILE="scheduler_config_cook_executor.edn"
+fi
```
shamsimam: The issue is that we don't want an entry for `:executor` at all when the variable is set to false.

Also, the `scheduler_config_command_executor.edn` file is temporary until we switch to the cook executor completely.

shamsimam:
@dposada and @wyegelwel addressed your feedback.
Passing build: https://travis-ci.org/twosigma/Cook/builds/248154438

dposada:
Needs rebase

integration/travis/run_integration.sh
dposada: Would having the `:executor` entry with `:command nil` have the same effect as not having the `:executor` entry at all? If so, I think we could get away with only one file and using `#config/env`

integration/tests/cook/test_basic.py
wyegelwel: @shamsimam did you push the update, it looks like this comment is not addressed.

scheduler/src/cook/mesos.clj
wyegelwel: Put it at the end so it is easier to find the change

scheduler/src/cook/components.clj
```diff
@@ -334,6 +338,27 @@
...
+     :executor (fnk [[:config {executor {}}]]
+                 (if (seq executor)
+                   (do
+                     (when (str/blank? (:command executor))
+                       (throw (ex-info "Executor command is missing!" {:executor executor})))
+                     (when (and (:uri executor) (nil? (get-in executor [:uri :value])))
+                       (throw (ex-info "Executor uri value is missing!" {:executor executor})))
+                     (let [default-executor-config {:log-level "INFO"
+                                                    :max-message-length 512
+                                                    :progress-output-file "stdout"
```
wyegelwel: Consider naming this `default-progress-output-file` and `default-progress-regex-string` to signal we intend to allow users to configure this in the future.

wyegelwel:
Please Add the executor stanza to the configuration doc.

wyegelwel:
I'm still concerned that bad actors could write a LOT of progresses to cook over the course of a job run which would be a problem both from a processing the messages on cook perspective and on a lots of writes to datomic perspective.

I know we will only accept one update per `progress-sample-interval-ms` but at the default of 1 second a job could still send 60 updates a minute which is a lot of noise.

sradack:
@wyegelwel Are you arguing for a different default or a different strategy like having a local rate limit?

integration/travis/run_integration.sh
shamsimam: Okay, sounds good.

shamsimam:
Another todo in a follow-up PR:

Throttle progress updates from jobs on the scheduler.

integration/travis/run_integration.sh
shamsimam: Done.

scheduler/src/cook/components.clj
shamsimam: Done.

shamsimam:
@wyegelwel 

> Please Add the executor stanza to the configuration doc.

Done.

shamsimam:
Ready for review, passing build: https://travis-ci.org/twosigma/Cook/builds/252286363

scheduler/docs/configuration.asc
```diff
@@ -266,6 +266,68 @@ Optionally, you can include a "rebalancer" stanza.  If you do, on startup, Cook
 `:retry-limit`::
   This limits the number of retries a job is allowed to request. Something in the low tens is often more than sufficient.
 
+[[cook_executor]]
+==== Cook Executor
+
+The Cook executor is a custom executor written in Python.
+It replaces the default command executor in order to enable a number of features for both operators and end users.
```
sradack: Why do we not still support using the standard Mesos executor?

scheduler/docs/configuration.asc
```diff
@@ -266,6 +266,68 @@ Optionally, you can include a "rebalancer" stanza.  If you do, on startup, Cook
...
+`:log-level`::
+  The log level for the executor process.
+  Defaults to "INFO".
+
+`:max-message-length`::
+  The maximum length for the unencoded string messages sent from a task via the Mesos executor HTTP API.
+  The default is 512.
+
+`:progress-sample-interval-ms`::
+  The interval in ms after which to send progress updates.
```
wyegelwel: Put a note of warning about setting this too low.

shamsimam: Updated documentation.

scheduler/docs/configuration.asc
shamsimam: Updated documentation.

shamsimam:
Updated documentation, build passing, ready for another round.

shamsimam:
Can we get this merged?

shamsimam:
@pschorf @sradack any more feedback on this?

shamsimam:
Can we get this merged?

shamsimam:
Summary of upcoming follow-up PRs:

1. Throttle progress updates from jobs on the scheduler.
1. The ability to gradually roll out the cook executor (e.g. day 1, 2% of instances use it; day 2, 20%; day 3, 80%; day 4, 100%)
1. The ability to force using (or not using) the cook executor from the client

PR #313, opened 2017-05-02
Remove redundant calls to constraint functions in matching

wyegelwel:
This is two changes:

1. Cache the list of constraints for a job since getHardConstraints is
called once for each host with available resources
2. Cache the call to needs-gpus? because it is called once for each job
running on each host (for non-gpu hosts)

### DIFF
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -63,15 +63,12 @@
   [object]
   (.getSimpleName (type object)))
 
-(defrecord novel-host-constraint [job]
+(defrecord novel-host-constraint [job previous-hosts]
   JobConstraint
   (job-constraint-name [this] (get-class-name this))
   (job-constraint-evaluate
     [this _ vm-attributes]
     (let [job (:job this)
-          previous-hosts (->> (:job/instance job)
-                              (remove #(true? (:instance/preempted? %)))
-                              (mapv :instance/hostname))
           target-hostname (get vm-attributes "HOSTNAME")]
       [(not-any? #(= target-hostname %) previous-hosts)
        (str "Can't run on " target-hostname
@@ -82,7 +79,16 @@
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
 
-(defrecord gpu-host-constraint [job]
+(defn build-novel-host-constraint
+  "Constructs a novel-host-constraint. 
+  The constraint prevents the job from running on hosts it has already run on"
+  [job]
+  (let [previous-hosts (->> (:job/instance job)
+                            (remove #(true? (:instance/preempted? %)))
+                            (mapv :instance/hostname))]
+    (->novel-host-constraint job previous-hosts)))
+
+(defrecord gpu-host-constraint [job needs-gpus?]
   JobConstraint
   (job-constraint-name [this] (get-class-name this))
   (job-constraint-evaluate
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/mesos/constraints.clj
```diff
@@ -82,7 +79,15 @@
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
 
-(defrecord gpu-host-constraint [job]
+(defn build-novel-host-constraint
+  "Constructs a novel-host-constraint"
```
shamsimam: Should we also add that the constraint is meant to exclude any hosts the job previously ran on?

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -82,7 +79,15 @@
...
+  [job]
+  (let [previous-hosts (->> (:job/instance job)
+                            (remove #(true? (:instance/preempted? %)))
+                            (mapv :instance/hostname))]
+    (->novel-host-constraint job previous-hosts)))
+
+(defrecord gpu-host-constraint [job needs-gpus?]
```
shamsimam: Would we ever add a `gpu-host-constraint` for a job that does not need gpus?

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -93,17 +98,22 @@
...
           job (:job this)
-          needs-gpus? (job-needs-gpus? job)
           passes? (or (and needs-gpus? vm-has-gpus?)
                       (and (not needs-gpus?) (not vm-has-gpus?)))]
       [passes? (when-not passes? (if (and needs-gpus? (not vm-has-gpus?))
                                    "Job needs gpus, host does not have gpus."
                                    "Job does not need gpus, host has gpus."))])))
 
-(def job-constraint-constructors [->novel-host-constraint ->gpu-host-constraint])
+(defn build-gpu-host-constraint
```
shamsimam: +1 for factory methods.

scheduler/src/cook/mesos/constraints.clj
wyegelwel: Yes, just updated the help message on build-gpu-host-constraint. It is meant to prevent gpu jobs from going on non-gpu hosts and non-gpu jobs from going on gpu hosts

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -82,7 +79,16 @@
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
 
-(defrecord gpu-host-constraint [job]
+(defn build-novel-host-constraint
+  "Constructs a novel-host-constraint. 
+  The constraint prevent the job from running on hosts it has already run on"
```
dposada: prevent -> prevent**s**

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -93,17 +99,24 @@
     (let [; Look at attribute and running jobs to determine if vm has gpus
           vm-has-gpus? (or (get vm-attributes "COOK_GPU?") ; Set when putting attributes in cache
                            (some (fn gpu-task? [req]
-                                   (job-needs-gpus? (:job req)))
+                                   (:needs-gpus? req))
```
dposada: Where is `:needs-gpus?` being set on the `req`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -392,14 +392,11 @@
               (not has-gpus?))
             (str "The machine " (.getHostname target-vm) (if @needs-gpus? " doesn't have" " has") " gpus")))))))
 
-(defrecord TaskRequestAdapter [job resources task-id assigned-resources guuid->considerable-cotask-ids]
+(defrecord TaskRequestAdapter [job resources task-id assigned-resources guuid->considerable-cotask-ids constraints needs-gpus?]
```
dposada: Is `needs-gpus?` used anywhere?

scheduler/src/cook/mesos/constraints.clj
wyegelwel: The req is a TaskRequest, coming from the TaskRequestAdaptor. It is the needs-gpus? argument on the record.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: Yes, see the comment above

scheduler/src/cook/mesos/constraints.clj
shamsimam: Can we annotate the `req` with its type?

sradack: Also maybe `{:keys [needs-gpus?]}`

sradack:
LGTM

shamsimam:
Needs rebasing.

wyegelwel:
rebased. I added two more small changes. The first caches another field (scalar-requests) the second adds type hints to a function that is called for every task,host,constraint so optimizing it is important

PR #315, opened 2017-05-03
mesos_mock: minor refactoring and reformatting of mesos_mock

shamsimam:
- move execution time into an environment variable.

Independent changes to declutter diff in cook-executor PR.


shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/228490183

PR #316, opened 2017-05-03
Stream JSON responses

pschorf:
Sometimes, responses from Cook can be quite large (16 MB.) In order to reduce stress on the GC from allocating large objects to store the responses, we can stream them back to the client.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -45,11 +45,11 @@
             [ring.util.response :as res]
             [schema.core :as s]
             [swiss.arrows :refer :all])
-  (:import (java.io OutputStreamWriter)
+  (:import clojure.lang.Atom
+           (java.io OutputStreamWriter)
            (java.net ServerSocket URLEncoder)
            (java.util Date UUID)
            javax.servlet.ServletResponse
-           clojure.lang.Atom
            org.apache.curator.test.TestingServer
            org.joda.time.Minutes
            schema.core.OptionalKey))
@@ -1556,12 +1556,18 @@
                        (::jobs ctx)))}))
 
 (defn- streaming-json-encoder
+  "Takes as input the response body which can be converted into JSON,
+  and returns a function which takes a ServletResponse and writes the JSON
+  encoded response data. This is suitable for use with jet's server API."
   [data]
   (fn [^ServletResponse resp]
     (cheshire/generate-stream data
                               (OutputStreamWriter. (.getOutputStream resp)))))
 
-(defn- streaming-json-middleware [handler]
+(defn- streaming-json-middleware
+  "Ring middleware which encodes JSON responses with streaming-json-encoder.
+  All other response types are unaffected."
+  [handler]
   (fn streaming-json-handler [req]
     (let [{:keys [headers body] :as resp} (handler req)
           json-response (= "application/json" (get headers "Content-Type" "application/json"))]
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1721,9 +1721,9 @@
                           400 {:description "Invalid request format."}
                           404 {:description "The UUID doesn't correspond to a job."}}}})))
     (ANY "/queue" []
-      (waiting-jobs mesos-pending-jobs-fn is-authorized-fn))
+         (waiting-jobs mesos-pending-jobs-fn is-authorized-fn))
     (ANY "/running" []
-      (running-jobs conn is-authorized-fn))
+         (running-jobs conn is-authorized-fn))
     (ANY "/list" []
          (list-resource (db conn) fid is-authorized-fn)))
    (format-params/wrap-restful-params {:formats [:json-kw]
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1570,7 +1570,8 @@
   [handler]
   (fn streaming-json-handler [req]
     (let [{:keys [headers body] :as resp} (handler req)
-          json-response (= "application/json" (get headers "Content-Type" "application/json"))]
+          json-response (or (= "application/json" (get headers "Content-Type"))
+                            (coll? body))]
       (cond-> resp
         json-response (res/content-type "application/json")
         (and json-response body) (assoc :body (streaming-json-encoder body))))))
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -42,11 +42,15 @@
...
+            [ring.middleware.format-params :as format-params]
             [ring.middleware.json]
+            [ring.util.response :as res]
             [schema.core :as s :refer (defschema)]
             [swiss.arrows :refer :all])
-  (:import (java.net ServerSocket URLEncoder)
+  (:import (java.io OutputStreamWriter)
+           (java.net ServerSocket URLEncoder)
            (java.util Date UUID)
+           javax.servlet.ServletResponse
```
dposada: Nitpick: this one is not sorted

scheduler/src/cook/mesos/api.clj
```diff
@@ -1094,36 +1098,35 @@
...
-                                     (->> queue
-                                          (take (::limit ctx))
-                                          (map d/touch)))
-                                   (mesos-pending-jobs-fn))
-                         cheshire/generate-string)))
-      ring.middleware.json/wrap-json-params))
+  (liberator/resource
+   :available-media-types ["application/json"]
+   :allowed-methods [:get]
+   :as-response (fn [data ctx] {:body data})
```
dposada: Should we extract this to a named function since we use it in a few places?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1571,138 +1572,153 @@
...
+    (cheshire/generate-stream data
+                              (OutputStreamWriter. (.getOutputStream resp)))))
+
+(defn- streaming-json-middleware [handler]
+  (fn streaming-json-handler [req]
+    (let [{:keys [headers body] :as resp} (handler req)
+          json-response (= "application/json" (get headers "Content-Type" "application/json"))]
+      (cond-> resp
+        json-response (res/content-type "application/json")
+        (and json-response body) (assoc :body (streaming-json-encoder body))))))
```
dposada: Would be nice to have a docstring on one or both of these

scheduler/src/cook/mesos/api.clj
```diff
@@ -1715,10 +1731,13 @@
...
                           404 {:description "The UUID doesn't correspond to a job."}}}})))
-
     (ANY "/queue" []
          (waiting-jobs mesos-pending-jobs-fn is-authorized-fn))
     (ANY "/running" []
          (running-jobs conn is-authorized-fn))
     (ANY "/list" []
-         (list-resource (db conn) fid is-authorized-fn))))
+         (list-resource (db conn) fid is-authorized-fn)))
+   (format-params/wrap-restful-params {:formats [:json-kw]
```
dposada: Why are we adding `wrap-restful-params` here?

scheduler/src/cook/mesos/api.clj
pschorf: Done

scheduler/src/cook/mesos/api.clj
pschorf: I'd vote not to, since it's really just a no-op and wouldn't have much other utility.

scheduler/src/cook/mesos/api.clj
pschorf: Done

scheduler/src/cook/mesos/api.clj
pschorf: With compojure api, you can't turn off output processing without also disabling input processing. This is basically re-inserting the input processing they provide by default. It also lets us remove it from the liberator resources since they're now wrapped by this middleware.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1556,12 +1556,18 @@
                        (::jobs ctx)))}))
 
 (defn- streaming-json-encoder
+  "Takes as input an the response body which can be converted into JSON,
```
dposada: "an" can be removed

pschorf: Done

scheduler/src/cook/mesos/api.clj
```diff
@@ -1556,138 +1555,159 @@
...
+    (cheshire/generate-stream data
+                              (OutputStreamWriter. (.getOutputStream resp)))))
+
+(defn- streaming-json-middleware
+  "Ring middleware which encodes JSON responses with streaming-json-encoder.
+  All other response types are unaffected."
+  [handler]
+  (fn streaming-json-handler [req]
+    (let [{:keys [headers body] :as resp} (handler req)
+          json-response (= "application/json" (get headers "Content-Type" "application/json"))]
```
shamsimam: Is defaulting the content type to `application/json` safe?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1700,10 +1720,13 @@
...
                           404 {:description "The UUID doesn't correspond to a job."}}}})))
-
     (ANY "/queue" []
       (waiting-jobs mesos-pending-jobs-fn is-authorized-fn))
     (ANY "/running" []
       (running-jobs conn is-authorized-fn))
     (ANY "/list" []
-      (list-resource (db conn) fid is-authorized-fn))))
\ No newline at end of file
+         (list-resource (db conn) fid is-authorized-fn)))
```
shamsimam: nitpick: formatting is off

scheduler/src/cook/mesos/api.clj
pschorf: So I added this because the Swagger UI was broken when serving JSON to it's frontend. I think the way it was working before was that it was automatically converting maps and lists to JSON. Would you prefer that?

scheduler/src/cook/mesos/api.clj
pschorf: Done

scheduler/src/cook/mesos/api.clj
pschorf: I think I like that better. Changed and tested with the swagger-ui

shamsimam: Okay.

shamsimam:
LGTM, ready to merge?

pschorf:
No objection here, build is passing: https://travis-ci.org/twosigma/Cook/builds/228740698

PR #317, opened 2017-05-03
Adds optional flag to allow partial results when querying for jobs

dposada:
We want to be able to query a particular cook cluster for a set of uuids, where some of them may not belong to that cluster, and get back the data for those that do match. This change allows for that, instead of always returning 404 if any of the uuids is invalid.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -92,9 +92,12 @@ class CookTest(unittest.TestCase):
         self.assertEqual('failed', job['instances'][0]['status'])
         self.assertEqual(2003, job['instances'][0]['reason_code'])
 
-    # load a job by UUID using GET /rawscheduler
+    def query_jobs(self, **kwargs):
+        return self.session.get('%s/rawscheduler' % self.cook_url, params=kwargs)
+
     def get_job(self, job_uuid):
-        return self.session.get('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid)).json()[0]
+        """Loads a job by UUID using GET /rawscheduler"""
+        return self.query_jobs(job=[job_uuid]).json()[0]
 
     def test_get_job(self):
         # schedule a job
@@ -337,3 +340,42 @@ class CookTest(unittest.TestCase):
         # Should require application version
         _, resp = self.submit_job(application={'name': 'foo-app'})
         self.assertEqual(resp.status_code, 400)
+
+    def test_allow_partial(self):
+        job_uuid, resp = self.submit_job()
+        self.assertEqual(201, resp.status_code)
+
+        # Single, valid job uuid
+        resp = self.query_jobs(job=[job_uuid])
+        self.assertEqual(200, resp.status_code)
+
+        # One valid job uuid, one invalid job uuid
+        bogus_uuid = str(uuid.uuid4())
+        resp = self.query_jobs(job=[job_uuid, bogus_uuid])
+        self.assertEqual(404, resp.status_code)
+        resp = self.query_jobs(job=[job_uuid, bogus_uuid], partial='false')
+        self.assertEqual(404, resp.status_code, resp.json())
+
+        # Partial results with one valid job uuid, one invalid job uuid
+        resp = self.query_jobs(job=[job_uuid, bogus_uuid], partial='true')
+        self.assertEqual(200, resp.status_code, resp.json())
+        self.assertEqual(1, len(resp.json()))
+        self.assertEqual(job_uuid, resp.json()[0]['uuid'])
+
+        # Single, valid instance uuid
+        job = self.wait_for_job(job_uuid, 'completed')
+        instance_uuid = job['instances'][0]['task_id']
+        resp = self.query_jobs(instance=[instance_uuid])
+        self.assertEqual(200, resp.status_code)
+
+        # One valid instance uuid, one invalid instance uuid
+        resp = self.query_jobs(instance=[instance_uuid, bogus_uuid])
+        self.assertEqual(404, resp.status_code)
+        resp = self.query_jobs(instance=[instance_uuid, bogus_uuid], partial='false')
+        self.assertEqual(404, resp.status_code)
+
+        # Partial results with one valid instance uuid, one invalid instance uuid
+        resp = self.query_jobs(instance=[instance_uuid, bogus_uuid], partial='true')
+        self.assertEqual(200, resp.status_code)
+        self.assertEqual(1, len(resp.json()))
+        self.assertEqual(job_uuid, resp.json()[0]['uuid'])
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -7,7 +7,7 @@ import uuid
 
 from retrying import retry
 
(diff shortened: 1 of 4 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -92,9 +92,12 @@ def test_max_runtime_exceeded(self):
         self.assertEqual('failed', job['instances'][0]['status'])
         self.assertEqual(2003, job['instances'][0]['reason_code'])
 
-    # load a job by UUID using GET /rawscheduler
+    def query_jobs(self, **kwargs):
```
shamsimam: missing docstring, it's not clear to me how the function parameters are converted to url query parameters.

scheduler/src/cook/mesos/api.clj
```diff
@@ -822,23 +828,34 @@
...
+             (every? used? jobs)
+             (every? some? instance-jobs))
         [true {::jobs (into jobs instance-jobs)
                ::jobs-requested jobs
                ::instances-requested instances}]
+
+        (and allow-partial-results
+             (or (some used? jobs)
+                 (some some? instance-jobs)))
+        [true {::jobs (into (filter used? jobs) (filter some? instance-jobs))
```
shamsimam: `(remove nil? instance-jobs)` is cleared to me than `(filter some? instance-jobs)`

scheduler/src/cook/mesos/api.clj
```diff
@@ -822,23 +828,34 @@
 (defn retrieve-jobs
   [conn ctx]
   (let [jobs (get-in ctx [:request :query-params :job])
-        instances (get-in ctx [:request :query-params :instance])]
+        instances (get-in ctx [:request :query-params :instance])
+        allow-partial-results (get-in ctx [:request :query-params :partial])]
     (let [instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
           instance-jobs (mapv instance-uuid->job-uuid instances)
           used? (partial job-exists? (db conn))]
```
shamsimam: is `exists?` a better name?

shamsimam: Also, can we use `fn` instead of `partial`

scheduler/src/cook/mesos/api.clj
```diff
@@ -822,23 +828,34 @@
 (defn retrieve-jobs
```
shamsimam: Can we have a docstring explaining the returned value?

integration/tests/cook/test_basic.py
```diff
@@ -337,3 +340,42 @@ def test_application_field(self):
         # Should require application version
         _, resp = self.submit_job(application={'name': 'foo-app'})
         self.assertEqual(resp.status_code, 400)
+
+    def test_allow_partial(self):
```
shamsimam: Do we have a test that tests the non-partial variant?

dposada: Yes, the scenarios in this test include both absence of the `partial` param and setting it explicitly to false. Let me know what you think.

scheduler/src/cook/mesos/api.clj
dposada: Why?

integration/tests/cook/test_basic.py
dposada: Added docstring that explains this

scheduler/src/cook/mesos/api.clj
dposada: Added docstring

scheduler/src/cook/mesos/api.clj
```diff
@@ -820,30 +826,59 @@
...
+        instances (get-in ctx [:request :query-params :instance])
+        allow-partial-results (get-in ctx [:request :query-params :partial])]
     (let [instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
           instance-jobs (mapv instance-uuid->job-uuid instances)
-          used? (partial job-exists? (db conn))]
+          exists? #(job-exists? (db conn) %)]
       (cond
-        (and (every? used? jobs)
-             (every? (complement nil?) instance-jobs))
+        (and (not allow-partial-results)
```
pschorf: Can you remove `allow-partial-results` here? I get that it works because it will fall into the next clause, but it's kind of confusing.

dposada: Removed

integration/tests/cook/test_basic.py
shamsimam: I should have been clearer. I meant to say the non-partial variant that loads multiple jobs. The one in this test handles only a single job.

scheduler/src/cook/mesos/api.clj
```diff
@@ -809,30 +815,58 @@
...
+        instances (get-in ctx [:request :query-params :instance])
+        allow-partial-results (get-in ctx [:request :query-params :partial])]
     (let [instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
           instance-jobs (mapv instance-uuid->job-uuid instances)
-          used? (partial job-exists? (db conn))]
+          exists? #(job-exists? (db conn) %)]
       (cond
-        (and (every? used? jobs)
-             (every? (complement nil?) instance-jobs))
+        (and (every? exists? jobs)
```
shamsimam: nitpick: do we care about potential multiple calls to `exists?` over the execution of this function. Do we want to filter the jobs eagerly and work on the lengths of the collections in the conditionals?

integration/tests/cook/test_basic.py
dposada: Ah got it, will add

scheduler/src/cook/mesos/api.clj
pschorf: 👍

integration/tests/cook/test_basic.py
dposada: Added an additional valid uuid to each of the scenarios

scheduler/src/cook/mesos/api.clj
```diff
@@ -839,18 +839,19 @@
...
-        (and (every? exists? jobs)
+        (and (= (count existing-jobs) (count jobs))
              (every? some? instance-jobs))
         [true {::jobs (into jobs instance-jobs)
                ::jobs-requested jobs
                ::instances-requested instances}]
 
         (and allow-partial-results
-             (or (some exists? jobs)
+             (or (> (count existing-jobs) 0)
```
shamsimam: A good place to use `pos?`

scheduler/src/cook/mesos/api.clj
dposada: Done

scheduler/src/cook/mesos/api.clj
dposada: Done

dposada:
Passing build: https://travis-ci.org/twosigma/Cook/builds/229260098

dposada:
@shamsimam Any more feedback on this?

dposada:
Can we merge this?

dposada:
@shamsimam @pschorf Any more feedback on this?

PR #322, opened 2017-05-08
Parameterizes run-docker.sh to allow for multiple local cooks

dposada:


dposada:
Passing build: https://travis-ci.org/twosigma/Cook/builds/230103079

PR #326, opened 2017-05-09
scheduler: use agent to ensure order of processing

shamsimam:
processing happens on non-mesos client thread to avoid class loader problems previously encountered.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -1299,7 +1299,13 @@
 
   (let [fid (atom nil)
         fenzo (make-fenzo-scheduler driver-atom offer-incubate-time-ms fenzo-fitness-calculator good-enough-fitness)
-        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset)]
+        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset)
+        order-enforcer-agent (agent nil)
+        agent-processor (fn agent-processor [_ body-fn]
+                          (try
+                            (body-fn)
+                            (catch Exception e
+                              (log/error e "Error processing mesos status/message."))))]
     (start-jobs-prioritizer! conn pending-jobs-atom task-constraints)
     {:scheduler
      (mesos/scheduler
@@ -1346,5 +1352,5 @@
        (resource-offers [this driver offers]
                         (receive-offers offers-chan driver offers))
        (status-update [this driver status]
-                      (future (handle-status-update conn driver fenzo status))))
+                      (send order-enforcer-agent agent-processor #(handle-status-update conn driver fenzo status))))
      :view-incubating-offers (fn get-resources-atom [] @resources-atom)}))
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -1290,6 +1290,61 @@
             (catch Exception e
               (log/error e "Unable to decline offers!")))))))
 
+(defn create-mesos-scheduler
+  "Creates the mesos scheduler which offloads processing of status updates to an agent."
+  [fid set-framework-id gpu-enabled? conn heartbeat-ch fenzo offers-chan]
+  (let [processor-agent (agent nil)
+        safe-call (fn agent-processor [_ body-fn]
+                    (try
+                      (body-fn)
+                      (catch Exception e
+                        (log/error e "Error processing mesos status/message."))))]
+    (mesos/scheduler
+      (registered [this driver framework-id master-info]
+                  (log/info "Registered with mesos with framework-id " framework-id)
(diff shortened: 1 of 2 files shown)


shamsimam:
From @wyegelwel  
"The reason the future is there is actually a classloader problem. For
reasons I don't know how to investigate, the thread the mesos client
launches on doesn't pick up the full class path but launching a new thread
does. Last time I say the problem it showed up with an exception
clojure/Lang/reflector not found"

shamsimam:
passing build: https://travis-ci.org/twosigma/Cook/builds/230435255

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1299,7 +1299,13 @@
 
   (let [fid (atom nil)
         fenzo (make-fenzo-scheduler driver-atom offer-incubate-time-ms fenzo-fitness-calculator good-enough-fitness)
-        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset)]
+        [offers-chan resources-atom] (make-offer-handler conn driver-atom fenzo fid pending-jobs-atom offer-cache fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset)
+        order-enforcer-agent (agent nil)
```
dposada: maybe `status-update-agent`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1299,7 +1299,13 @@
...
+        agent-processor (fn agent-processor [_ body-fn]
```
dposada: maybe `safe-call`?

scheduler/src/cook/mesos/scheduler.clj
shamsimam: It will also handle framework messages in the cook-executor branch.

shamsimam:
Will add unit tests.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: renamed

shamsimam:
@dpo added unit test

dpo:
Please get your names straight...

shamsimam:
@dpo sorry for the spamming.

shamsimam:
@wyegelwel how do you feel about this change?

shamsimam:
@sradack @wyegelwel how do you feel about the executor? this PR doesn't integrate the executor with the scheduler.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1290,6 +1290,61 @@
             (catch Exception e
               (log/error e "Unable to decline offers!")))))))
 
+(defn create-mesos-scheduler
+  "Creates the mesos scheduler which offloads processing of status updates to an agent."
+  [fid set-framework-id gpu-enabled? conn heartbeat-ch fenzo offers-chan]
+  (let [processor-agent (agent nil)
```
wyegelwel: Can you put a comment why we are using an agent? I *think* it is just to exploit that fact that the agent will linearize calls to it. However, it is a little weird because we are basically ignoring the agent otherwise

shamsimam: Yes, will add the comment.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1290,6 +1290,62 @@
             (catch Exception e
               (log/error e "Unable to decline offers!")))))))
 
+(defn create-mesos-scheduler
+  "Creates the mesos scheduler which offloads processing of status updates to an agent.
+   The agent is used to ensure in-order asynchronous processing of status updates."
```
wyegelwel: nitpick: put this line as a comment above the `processor-agent`

Larger comment though is that since we are just using the agent as in order async processing perhaps it is worth making a function / interface that makes that clear and then the implementation is the agent. This way, the thing we are trying to achieve is upfront.

shamsimam: okay will do.

wyegelwel: Something like: `(async-in-order (...))`

Also, how does `agent` handle the queue of requests? Is it unbounded? Is there back pressure?

shamsimam: It is unbounded (limited by heap memory), no back pressure.

I thought about using multiple agents to get benefits of parallelization, but that requires parsing the message and using the status id to still ensure in-order processing of messages from a given task.

shamsimam: Added the `async-in-order-processing` function that abstracts out use of the agent further.

wyegelwel: > It is unbounded (limited by heap memory), no back pressure.

This makes me worried. I know we effectively had this with `future` before but can you add a way to monitor how large the queue is getting (have a counter and inc when `async-in-order-processing` and dec when safe call is called) as well as a time spent in queue so we can monitor it?

shamsimam: Will do

shamsimam: added metrics support

shamsimam:
@wyegelwel addressed your previous feedback.

shamsimam:
Any more feedback on this?

PR #328, opened 2017-05-10
executor: cook executor that adds exit code, progress and sandbox location

shamsimam:
This PR introduces the cook-executor without integrating it with the scheduler.

The previous PR (https://github.com/twosigma/Cook/pull/308) had comments from too many different modules.


executor/Dockerfile.build
```diff
@@ -0,0 +1,13 @@
...
+
+RUN mkdir /opt/cook
+WORKDIR /opt/cook
+
+COPY requirements.txt /opt/cook/
+RUN pip install -r requirements.txt
+
+COPY . /opt/cook
+
+CMD ["pyinstaller", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
```
dposada: Copying over comments from the original PR (PR #308):

@dposada 2 days ago
> Why does this need to happen in a container?

@shamsimam a day ago
> The executable needs to be compiled using the same architecture as where it will be run - on the agents inside docker.

dposada: How does this accomplish that? Don't we need to ensure that it uses the same base image as the minimesos agent?

executor/README.md
```diff
@@ -0,0 +1,114 @@
+Cook Executor
+=============
+
+The Cook executor is a custom executor written in Python.
+It replaces the default command executor in order to enable a number of features for both operators and end users.
```
dposada: Should we add something like:

"for more information on Mesos executors, see "Working with Executors" at http://mesos.apache.org/documentation/latest/app-framework-development-guide/"?

executor/README.md
```diff
@@ -0,0 +1,114 @@
...
+1. progress update messages from the task (at intervals of `progress-sample-interval-ms` ms, see below),
+2. the exit code of the task, and
+3. the sandbox location of the task.
+
+For Operators
+---------
+
+The Cook executor replaces the default command executor.
+It can either be installed on each agent, or downloaded on each agent via a `uri`.
+As the executor communicates with Mesos agent via HTTP, the agent must support the HTTP API and have it enabled.
```
dposada: Nitpick: I would favor "Because" over "As"

executor/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+If the executor is not correctly installed on an agent (or if `:executor-command` is not set correctly), all tasks will fail, and there will be a message in the `stderr` file for each task indicating the command the agent attempted to run.
+Verify that the command is the correct command, and that it is available on each agent.
+If not, either fix the `:executor-command` config, or install the executor on the agent.
+
+If the executor is installed correctly, it will log info and errors to the `executor.log` file in the task sandbox.
+This should be the first place to look for stack traces and other error information.
+Each lifecycle callback is logged (`registered`, `launchTask`, etc), which can help to narrow down the issue.
+
+When troubleshooting any issues with running the executor on the agent, it may be helpful to use a simple job command until you confirm that executor is being launched correctly.
```
dposada: "until you confirm that **the** executor is being launched correctly"

executor/README.md
shamsimam: Done.

executor/README.md
shamsimam: Done.

executor/README.md
shamsimam: Done.

executor/Dockerfile.build
shamsimam: We worked on this online. Using docker to build the executable avoids different architecture and python environment hassles.

shamsimam:
Build with passing unit tests: https://travis-ci.org/twosigma/Cook/builds/230900419

shamsimam:
@pschorf @sradack @wyegelwel how do you feel about the executor? this PR doesn't integrate the executor with the scheduler.

executor/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Supported configuration options:
+
+| option | type | description |
+|--------|------|-------------|
+| `:command` | string | A string containing the command executed on the mesos agent to launch the cook executor. If the executor is installed using the instructions above, the default value of `cook-executor` should not need to be changed.|
+| `:log-level` | string | The log level for the executor process. Defaults to "INFO".|
+| `:max-message-length` | long | The maximum length for the custom failure message set by a task via the Mesos executor HTTP API. The default is 512.|
+| `:progress-output-name` | string | The file to track for progress updates. The default is stdout.|
+| `:progress-regex-string` | string | The regex used to identify progress update messages. The regex should have two capture groups, the first being an integer representing the progress percent. The second being a message about the progress. When not provided, progress updates will not be tracked.|
```
wyegelwel: Perhaps we should provide a sensible default like "<progress>, <message>"

shamsimam: I'll add support for something like "progress: 10 percent, processing foo"

executor/cook/progress.py
```diff
@@ -0,0 +1,148 @@
...
+    def tail(self, sleep_time_ms):
+        """
+        This method incrementally generates lines from a file by waiting for new content from a file.
+        It behaves like the 'tail -f' shell command.
+        :param sleep_time_ms: The unit of time in ms to repetitively sleep when the file has not been created or no new
+          content is available in the file being tailed.
+        :return: an incrementally generated list of lines in the file being tailed.
+        """
+        try:
+            sleep_param = (1.0 * sleep_time_ms) / 1000
```
wyegelwel: why are you multiplying by 1.0 here?

shamsimam: My bad, habit of forcing double operations on int arguments. Will remove.

executor/tests/test_progress.py
```diff
@@ -0,0 +1,288 @@
...
+        time.sleep(poll_interval_ms / 1000.0)
+        progress_data_2 = {'progress-message': 'Progress message-2'}
+        progress_updater.send_progress_update(driver, task_id, max_message_length, progress_data_2, ce.send_message)
+
+        self.assertEqual(2, driver.method_call_count())
+        actual_encoded_message_2 = driver.messages[1]
+        expected_message_2 = {'progress-message': 'Progress message-2', 'task-id': task_id}
+        assert_message(expected_message_2, actual_encoded_message_2)
+
+    def test_progress_watcher_tail(self):
```
wyegelwel: Can you add a test case where you send a LOT of lines? I think the yields will handle this gracefully but want to be sure. Also, if a LOT of lines were sent out, would the progress updater just be way behind?

wyegelwel: or would it get the latest update?

executor/README.md
```diff
@@ -0,0 +1,115 @@
...
+Configuration for the executor should go under the `:executor` key of the edn configuration file for the cook scheduler.
+
+For example:
+
+```clojure
+{...
+ :executor {:command "cook-executor"
+            :log-level "INFO"
+            :max-message-length 512
+            :progress-output-name "stdout"
```
wyegelwel: Can we make it possible to override this on a per job basis? (I know it wouldn't be in this commit, just want to get the idea out there)

executor/tests/test_progress.py
shamsimam: Yes, the progress updater would lag behind. We currently wait for one second after the job has completed for progress updater to catch up.

executor/README.md
shamsimam: We can override the values sent from the scheduler on a per task basis, we will just need to update the job schema to store these fields. 

Maybe we do this in a follow-up PR after the executor+scheduler changes are merged?

executor/tests/test_progress.py
shamsimam: Will add the test.

executor/cook/progress.py
shamsimam: Fixed.

executor/tests/test_progress.py
shamsimam: Added tests that write lots of data and also verify parsing progress updates from those data.

executor/README.md
shamsimam: Added default progress string support.

shamsimam:
@wyegelwel addressed your previous feedback.

shamsimam:
Any more feedback on this?

executor/README.md
```diff
@@ -0,0 +1,115 @@
...
+For more information on Mesos executors, see the "Working with Executors" section at http://mesos.apache.org/documentation/latest/app-framework-development-guide/
+
+For Users
+---------
+
+The cook executor supports a number of (completely optional) features beyond the standard Mesos command executor.
+The executor sends the following information to the scheduler:
+1. progress update messages from the task (at intervals of `progress-sample-interval-ms` ms, see below),
+2. the exit code of the task, and
+3. the sandbox location of the task.
```
wyegelwel: This one may be less optional. We can remove a bunch of crummy code if we have the sandbox location and speed up queries.

executor/cook/executor.py
```diff
@@ -0,0 +1,344 @@
...
+
+    Returns
+    -------
+    :return: The tuple containing the process launched, stdout file, and stderr file.
+    """
+    task_id = get_task_id(task)
+    try:
+        update_status(driver, task_id, cook.TASK_RUNNING)
+
+        command = decode_data(task['data']).decode('utf8')
```
wyegelwel: Instead of sending the command as a string, consider sending a json map where one of the keys is "command". That way later on we can add other fields if needed without breaking backwards compatibility.

executor/cook/executor.py
```diff
@@ -0,0 +1,344 @@
...
+     First attempt is made by sending the process a SIGTERM.
+     If the process does not terminate inside (shutdown_grace_period_ms - 100) ms, it is then sent a SIGKILL.
+     The 100 ms grace period is allocated for the executor to perform its other cleanup actions.
+
+    Parameters
+    ----------
+    :param driver: MesosExecutorDriver
+        The mesos driver to use.
+    :param task: map
+        The task to execute.
```
wyegelwel: nitpick: executing a task can mean both running the task and killing the task, s/execute/kill/

executor/cook/executor.py
```diff
@@ -0,0 +1,344 @@
...
+    :return: Nothing
+    """
+    task_id = get_task_id(task)
+    try:
+        shutdown_grace_period_ms = max(shutdown_grace_period_ms - 100, 0)
+        process, _, _ = process_info
+        if is_running(process):
+            logging.info('Waiting up to {} ms for process to terminate'.format(shutdown_grace_period_ms))
+            process.terminate()
+            loop_limit = int(shutdown_grace_period_ms / 10)
```
wyegelwel: Is it ok that if shutdown_grace_period_ms is not divisible by 10 it will miss the last remainder(shutdown_grace_period_ms, 10) millis to complete. I imagine the answer is "no" but wanted to check

executor/cook/progress.py
```diff
@@ -0,0 +1,207 @@
...
+            The unit of time in ms to repetitively sleep when the file has not been created or no new
+            content is available in the file being tailed.
+        
+        Returns
+        -------
+        :return: an incrementally generated list of lines in the file being tailed.
+        """
+        try:
+            sleep_param = sleep_time_ms / 1000
+            while True:
```
wyegelwel: Can this be re-written as:

```python
while not os.path.isFile(self.target_file) and not self.completed_signal.isSet():
   logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
   time.sleep(sleep_param)

if self.completed_signal.isSet():
   logging.info('No output has been read to parse progress messages')

if os.path.isfile(self.target_file):
   logging.info('{} has been created, reading contents'.format(self.target_file))
```

I don't like while true as it hides the intent of the loop

executor/cook/progress.py
```diff
@@ -0,0 +1,207 @@
...
+                    logging.info('{} has been created, reading contents'.format(self.target_file))
+                    break
+                if self.completed_signal.isSet():
+                    logging.info('No output has been read to parse progress messages')
+                    return
+                logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
+                time.sleep(sleep_param)
+
+            target_file_obj = open(self.target_file, 'r')
+            while True:
```
wyegelwel: Similarly:

```python
while not self.completed_signal.isSet():
  line = target_file_obj...
  if not line:
    time.sleep(sleep_param)
    continue
  yield line
```

executor/cook/progress.py
```diff
@@ -0,0 +1,207 @@
...
+                else:
+                    logging.debug('Not sending progress data as enough time has not elapsed since last update')
+
+
+class ProgressWatcher(object):
+    """This class tails the output from the target file listening for progress messages.
+    The retrieve_progress_states generates all progress messages iteratively.
+    """
+
+    @staticmethod
```
wyegelwel: Why is this static?

executor/cook/progress.py
```diff
@@ -0,0 +1,207 @@
...
+        Parameters
+        ----------
+        :param progress_regex_str: string
+            The progress regex to match against, it must return two capture groups.
+        :param input_string: string
+            The input string.
+        
+        Returns
+        -------
+        :return: whether the input string matches the provided regex.
```
wyegelwel: This implies it is returning a boolean but it doesn't appear to be.

executor/cook/progress.py
```diff
@@ -0,0 +1,207 @@
...
+        yields the current progress as a dictionary.
+        
+        Returns
+        -------
+        :return: an incrementally generated list of progress states.
+        """
+        if self.progress_regex_string:
+            sleep_time_ms = 50
+            for line in self.tail(sleep_time_ms):
+                progress_report = ProgressWatcher.match_progress_update(self.progress_regex_string, line)
```
wyegelwel: It doesn't look like this function returns a tuple

executor/README.md
shamsimam: Actually, none of these are now optional. The executor will perform all of the following actions.

executor/cook/executor.py
shamsimam: okay, sounds good.

executor/README.md
shamsimam: Updated the README

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: It is okay, as it means that we will wait 10ms while the task is shutting down. The grace period is usually configured to seconds, so missing out 10 ms isn;t that big a deal.

executor/cook/progress.py
shamsimam: Done.

executor/cook/progress.py
shamsimam: Done.

executor/cook/progress.py
shamsimam: To simplify testability.

executor/cook/progress.py
shamsimam: Updated the doc.

executor/cook/progress.py
shamsimam: Updated the documentation for `match_progress_update`

shamsimam:
Updated as per @wyegelwel's comments

shamsimam:
ready for another round of reviews :)

wyegelwel:
Super nitpick, in the comments remove :param: and :returns: since they are redundant. Otherwise, looks good

shamsimam:
@wyegelwel removed `:param` and `:return:` from docstrings.

wyegelwel:
This LGTM. @sradack did you want to take a final pass?

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+            stop_signal.set()
+
+            result = ce.await_process_completion(driver, task, stop_signal, process_info, shutdown_grace_period_ms)
+
+            self.assertTrue(result)
+
+            self.assertEqual(1, driver.method_call_count())
+            self.assertEqual(1, len(driver.statuses))
+            actual_status = driver.statuses[0]
+            expected_status = {'task_id': {'value': task_id},
```
sradack: It's possible the task finishes before `await_process_completion` kills it because the `stop_signal` is set, and therefore we wouldn't see a `TASK_KILLED`.

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+                               'state': cook.TASK_KILLED}
+            assert_status(expected_status, actual_status)
+
+            self.assertTrue(stdout.closed)
+            self.assertTrue(stderr.closed)
+
+        finally:
+            cleanup_output(stdout_name, stderr_name)
+
+    def test_await_process_completion(self):
```
sradack: This test doesn't test the "await" part.  It only tests the case where `stop_signal` is set.

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+            self.assertTrue(completed_signal.isSet())
+            assertions_fn(driver, task_id, sandbox_location)
+
+        finally:
+            cleanup_output(stdout_name, stderr_name)
+
+    def test_manage_task_successful_exit(self):
+        def assertions(driver, task_id, sandbox_location):
+            logging.info('driver.method_call_count() = {}'.format(driver.method_call_count()))
+            self.assertEqual(4, driver.method_call_count())
```
sradack: Did you arrive at this constant after running the test?

executor/tests/test_progress.py
```diff
@@ -0,0 +1,404 @@
...
+        max_message_length = 100
+        poll_interval_ms = 10
+
+        progress_updater = cp.ProgressUpdater(poll_interval_ms, ce.send_message)
+        progress_data_0 = {'progress-message': 'Progress message-0 is really long lorem ipsum dolor sit amet text'}
+        progress_updater.send_progress_update(driver, task_id, max_message_length, progress_data_0)
+
+        self.assertEqual(1, driver.method_call_count())
+        actual_encoded_message_0 = driver.messages[0]
+        expected_message_0 = {'progress-message': 'Progress message-0 is really long lorem ipsum dolor...',
```
sradack: Did you arrive at the expected value after running the test?

executor/cook/executor.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # set the completed signal to trigger progress updater termination and await progress
+        # updater termination if executor is terminating normally
+        completed_signal.set()
+        if not stop_signal.isSet():
+            logging.info('Awaiting progress updater completion with timeout of one second')
+            progress_complete_event.wait(1)
+
+        logging.info('Process completed with exit code {}'.format(exit_code))
+        exit_message = json.dumps({'exit-code': abs(exit_code), 'task-id': task_id})
```
shamsimam: From http://docs.python.org/library/subprocess.html#subprocess.Popen.returncode:
> A negative value -N indicates that the child was terminated by signal N (Unix only).

executor/cook/progress.py
```diff
@@ -0,0 +1,224 @@
...
+                    logging.info('Sending progress message {}'.format(progress_data))
+                    message_dict = dict(progress_data)
+                    message_dict['task-id'] = task_id
+                    progress_message = json.dumps(message_dict)
+
+                    if len(progress_message) > max_message_length and cook.PROGRESS_MESSAGE_KEY in message_dict:
+                        progress_str = message_dict[cook.PROGRESS_MESSAGE_KEY]
+                        num_extra_chars = len(progress_message) - max_message_length
+                        allowed_progress_message_length = max(len(progress_str) - num_extra_chars - 3, 0)
+                        new_progress_str = progress_str[:allowed_progress_message_length].strip() + '...'
```
shamsimam: The strip is just so that the ellipsis appears after a non-whitespace character. Do you prefer it to be removed?

executor/cook/executor.py
```diff
@@ -0,0 +1,350 @@
...
+        send_message(driver, exit_message, config.max_message_length)
+
+        # send the latest progress state if available
+        latest_progress = progress_watcher.current_progress()
```
shamsimam: `ProgressUpdater.send_progress_update` already has a force send flag. I need these two lines (and `current_progress`) as the `ProgressUpdater` and `ProgressWatcher` are decoupled.

executor/cook/progress.py
```diff
@@ -0,0 +1,224 @@
...
+
+            while not os.path.isfile(self.target_file) and not self.completed_signal.isSet():
+                logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
+                time.sleep(sleep_param)
+
+            if self.completed_signal.isSet():
+                logging.info('No output has been read to parse progress messages')
+                return
+
+            if os.path.isfile(self.target_file):
```
shamsimam: removed the if clause.

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+        assert_status(expected_status, actual_status)
+
+    def test_update_status(self):
+        driver = FakeMesosExecutorDriver()
+        task_id = get_random_task_id()
+        task_state = "TEST_TASK_STATE"
+
+        ce.update_status(driver, task_id, task_state)
+
+        self.assertEqual(1, driver.method_call_count())
```
shamsimam: Okay, removing them won't hurt as there are already assertions on the statuses.

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+
+        stdout = open(stdout_name, 'w+')
+        stderr = open(stderr_name, 'w+')
+
+        try:
+            command = 'sleep 100'
+            process = subprocess.Popen(command, shell=True, stdout=stdout, stderr=stderr)
+            process_info = process, stdout, stderr
+            shutdown_grace_period_ms = 2000
+            ce.kill_task(driver, task, process_info, shutdown_grace_period_ms)
```
shamsimam: `sleep 100` will run for 100 seconds, we are killing the task immediately after launching.

executor/tests/test_executor.py
shamsimam: Ensured testing of the await part.

executor/tests/test_executor.py
shamsimam: `sleep 100` will run for 100 seconds, we are killing the task using `stop_signal`.

executor/tests/test_executor.py
shamsimam: This count is the bare minimum messages/statuses sent from the executor: exit-code, sandbox location, running and finished. Didn't need to run it to get this value.

executor/tests/test_progress.py
shamsimam: Yes, I could also have programmatically pre-computed it by substring-ing `progress_data_0['progress-message']` at length 100.

executor/cook/executor.py
shamsimam: I've added a helper function named `force_send_progress_update`

executor/tests/test_executor.py
```diff
@@ -0,0 +1,417 @@
...
+            stdout.close()
+            stderr.close()
+
+            if process.poll() is None:
+                process.kill()
+
+            self.assertEqual(1, driver.method_call_count())
+            self.assertEqual(1, len(driver.statuses))
+            actual_status = driver.statuses[0]
+            expected_status = {'task_id': {'value': task_id},
```
shamsimam: Changed the implementation so that `manage_task` sends the task status updates.

executor/tests/test_executor.py
sradack: I was thinking it was ms.

executor/cook/executor.py
sradack: By using `abs` you may collide with an exit code that has meaning with the program.  Let's just leave them negative.

executor/cook/executor.py
```diff
@@ -281,15 +293,15 @@ def manage_task(driver, task, stop_signal, completed_signal, config, stdout_name
...
 
         # send the latest progress state if available
-        latest_progress = progress_watcher.current_progress()
-        progress_updater.send_progress_update(driver, task_id, config.max_message_length, latest_progress,
-                                              force_send=True)
+        cp.force_send_progress_update(progress_watcher, progress_updater)
 
         if not process_killed:
+            # task either completed successfully or aborted with an error
             task_state = cook.TASK_FINISHED if exit_code == 0 else cook.TASK_FAILED
```
sradack: What about when a task is killed? Should we be using `TASK_KILLED`?

shamsimam: Moved task_killed status updates from `kill_task` to `manage_task`

executor/cook/executor.py
shamsimam: Removed the abs, using the exit code returned by python `subprocess`.

shamsimam:
@sradack  addressed feedback. Please have a look.

executor/cook/executor.py
```diff
@@ -0,0 +1,363 @@
...
+        logging.info('Process completed with exit code {}'.format(exit_code))
+        exit_message = json.dumps({'exit-code': exit_code, 'task-id': task_id})
+        send_message(driver, exit_message, config.max_message_length)
+
+        # send the latest progress state if available
+        cp.force_send_progress_update(progress_watcher, progress_updater)
+
+        # task either completed successfully or aborted with an error
+        task_state = get_task_state(exit_code)
+        update_status(driver, task_id, task_state)
```
sradack: Should we be calling `cleanup_process` here?

executor/cook/progress.py
```diff
@@ -0,0 +1,248 @@
...
+                logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
+                time.sleep(sleep_param)
+
+            if self.completed_signal.isSet():
+                logging.info('No output has been read to parse progress messages')
+                return
+
+            logging.info('{} has been created, reading contents'.format(self.target_file))
+            target_file_obj = open(self.target_file, 'r')
+            while not self.completed_signal.isSet():
```
sradack: Is it not possible for the process to have completed yet there is still data to read?

executor/cook/executor.py
```diff
@@ -0,0 +1,363 @@
...
+                                              config.progress_sample_interval_ms, send_message)
+        progress_complete_event = cp.launch_progress_tracker(progress_watcher, progress_updater)
+
+        await_process_completion(stop_signal, process_info, config.shutdown_grace_period_ms)
+        exit_code = process.returncode
+
+        # set the completed signal to trigger progress updater termination and await progress
+        # updater termination if executor is terminating normally
+        completed_signal.set()
+        if not stop_signal.isSet():
```
sradack: Should we not wait for `progress_complete_event` unconditionally, given that the final progress/status might be important?

executor/cook/executor.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def killTask(self, driver, task_id):
+        logging.info('Task {} has been killed by Mesos'.format(task_id))
+        self.stop_signal.set()
+
+    def shutdown(self, driver):
+        logging.info('Mesos requested executor to shutdown!')
+        self.stop_signal.set()
+
+    def has_stopped(self):
```
sradack: Reanme to `allTasksCompleted`?  Also why the difference in naming convention?

shamsimam: The other methods override `Executor` methods which use camelcase.

Will rename using underscores.

executor/cook/__main__.py
```diff
@@ -0,0 +1,55 @@
...
+
+    config = cc.initialize_config(environment)
+    stop_signal = Event()
+    driver_thread = Thread(target=ce.run_mesos_driver, args=(stop_signal, config))
+    driver_thread.start()
+
+    def handle_interrupt(interrupt_code, _):
+        logging.info('Received interrupt code {}, preparing to terminate executor'.format(interrupt_code))
+        stop_signal.set()
+    signal.signal(signal.SIGINT, handle_interrupt)
```
shamsimam: I'm not sure it changes anything, but I'll move them above the thread creation.

executor/cook/config.py
```diff
@@ -0,0 +1,74 @@
...
+        self.progress_sample_interval_ms = progress_sample_interval_ms
+        self.sandbox_location = sandbox_location
+        self.shutdown_grace_period_ms = ExecutorConfig.parse_time_ms(shutdown_grace_period)
+
+
+def initialize_config(environment):
+    """Initializes the config using the environment.
+    Populates the default values for missing environment variables.
+    """
+    max_bytes_read_per_line = 4 * 1024
```
shamsimam: Making it configurable.

executor/cook/executor.py
shamsimam: Okay.

executor/cook/executor.py
shamsimam: Yes. Will do cleanup as part of `await_process_completion`

executor/cook/progress.py
shamsimam: Yes, using a separate event to track command execution completion.

executor/cook/__main__.py
shamsimam: Done.

executor/cook/config.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Created a new `task_completion_signal`

executor/cook/executor.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Renamed to `has_task_completed` (the executor runs a single task).

executor/cook/progress.py
shamsimam: Done, using `task_completed_signal`.

shamsimam:
@sradack addressed your latest feedback, ready for another round.

executor/cook/executor.py
```diff
@@ -0,0 +1,363 @@
...
+        progress_watcher = cp.ProgressWatcher(config, completed_signal)
+        progress_updater = cp.ProgressUpdater(driver, task_id, config.max_message_length,
+                                              config.progress_sample_interval_ms, send_message)
+        progress_complete_event = cp.launch_progress_tracker(progress_watcher, progress_updater)
+
+        await_process_completion(stop_signal, process_info, config.shutdown_grace_period_ms)
+        exit_code = process.returncode
+
+        # set the completed signal to trigger progress updater termination and await progress
+        # updater termination if executor is terminating normally
```
shamsimam: Done.

executor/cook/progress.py
sradack: I don't see how the change you made addresses the original concern.  I'm concerned the `task_completed_signal` will be set immediately upon the completion the subprocess, not allowing the `ProgressWatcher` to report progress updates that are remaining.

shamsimam: Ah okay. I've now used `task_completed_signal` to notify the `tail` that there are no more upcoming contents and `stop_signal` to trigger early termination from the signal handler.

shamsimam:
@sradack please merge if you have no other feedback

PR #330, opened 2017-05-11
WIP: Scheduler simulator

wyegelwel:


wyegelwel:
I need to add a test confirming that the example output provided is the same as what is produced by the simulator.

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,565 @@
...
+            (org.joda.time.DateTimeUtils/setCurrentMillisFixed (inc (.getTime (tc/to-date (t/now)))))
+            (async/>!! rebalancer-trigger-chan rebalancer-complete-chan)
+            (async/<!! rebalancer-complete-chan)
+            (log/info "Rebalance complete")
+
+            ;; Request for jobs that are complete to have cook be notified
+            (log/info "Send completion status to scheduler")
+            (org.joda.time.DateTimeUtils/setCurrentMillisFixed (inc (.getTime (tc/to-date (t/now)))))
+            (async/>!! complete-trigger-chan :trigger)
+            (async/>!! complete-trigger-chan :trigger)
```
shamsimam: Please add a comment that we are relying on the processing of this second trigger to be a no-op.

wyegelwel:
@shamsimam Thanks for the feedback, splitting the PR into those pieces is a great suggestion. I will do that now.

scheduler/src/cook/components.clj
```diff
@@ -83,41 +85,47 @@
...
+                              mesos-gpu-enabled
+                              rebalancer
+                              {:fenzo-max-jobs-considered fenzo-max-jobs-considered
+                               :fenzo-scaleback fenzo-scaleback
+                               :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
+                               :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
+                               :fenzo-fitness-calculator fenzo-fitness-calculator
+                               :good-enough-fitness good-enough-fitness}
+                              ; TODO Make these configurable..
+                              {:ranker-trigger-chan (chime-ch  (periodic/periodic-seq (t/now) (t/seconds 5)))
```
wyegelwel: WHAT?!?!

PR #335, opened 2017-05-12
Removes the explicit mesos dependency since mesomatic pulls its own in

dposada:


dposada:
Passing build: https://travis-ci.org/twosigma/Cook/builds/231683694

PR #336, opened 2017-05-15
Replace concat with into

pschorf:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -671,7 +678,7 @@
                                               db category->pending-jobs user->quota user->usage num-considerable))
               {:keys [matches failures]} (timers/time!
                                            handle-resource-offer!-match-duration
-                                           (match-offer-to-schedule fenzo (apply concat (vals category->considerable-jobs)) offers))
+                                           (match-offer-to-schedule fenzo (apply concat-vec (vals category->considerable-jobs)) offers))
```
wyegelwel: You can do `(reduce into (vals category->considerable-jobs))` instead of concat-vec

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -381,7 +381,7 @@
         (let [has-gpus? (boolean (or (> (or (.getScalarValue (.getCurrAvailableResources target-vm) "gpus") 0.0) 0)
                                      (some (fn gpu-task? [req]
                                              @(job-needs-gpus? (:job req)))
-                                           (concat (.getRunningTasks target-vm)
+                                           (into (vec (.getRunningTasks target-vm))
```
wyegelwel: I can't put a comment on it, but I think something may have went wrong with a merge at some point. This gpu-host-constraint should be in cook.mesos.constraints.

pschorf:
Build passes: https://travis-ci.org/twosigma/Cook/builds/232530621

scheduler/src/cook/mesos/scheduler.clj
pschorf: I knew there was a better way to do that. Done.

scheduler/src/cook/mesos/scheduler.clj
pschorf: Do you mind if we leave it here for now just to keep this diff small?

pschorf:
https://travis-ci.org/twosigma/Cook/builds/232538564

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: That is fine, just don't want to forget

PR #338, opened 2017-05-15
Replace ad hoc cycle code with trigger chans

wyegelwel:


wyegelwel:
Had to also include the mesos mock changes because other wise some tests are flaky. I tried to minimize the diff

wyegelwel:
Passing build: https://travis-ci.org/twosigma/Cook/builds/232627069

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -730,20 +730,21 @@
...
   (let [chan-length 100
         offers-chan (async/chan (async/buffer chan-length))
         resources-atom (atom (view-incubating-offers fenzo))
-        timer-chan (chime-ch (periodic/periodic-seq (time/now) (time/seconds 1))
-                             {:ch (async/chan (async/sliding-buffer 1))})]
-    (async/thread
-      (loop [num-considerable max-considerable]
-        (reset! fenzo-num-considerable-atom num-considerable)
-
+        num-considerable-atom (reset! fenzo-num-considerable-atom max-considerable)]
```
shamsimam: Should this `reset!` be outside the `let`? Also, `num-considerable-atom` is probably not an atom.

scheduler/src/cook/mesos/util.clj
```diff
@@ -438,3 +439,32 @@
...
+            :or {error-handler identity
+                 on-finished #()}}]]
+  (async/go-loop []
+                 (if-let [x (async/<! ch)]
+                   (do (async/<! (async/thread
+                                   (try
+                                     (f x)
+                                     (catch Exception e
+                                       (error-handler e)))))
+                       (close-when-ch! x)
```
shamsimam: Why does `chime-at-ch` take the responsibility to close `x`? It is possible for `f` to trigger an `async` to populate `x`.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -730,20 +730,21 @@
...
+    (util/chime-at-ch
+      trigger-chan
+      (fn [_]
         ;;TODO make this cancelable (if we want to be able to restart the server w/o restarting the JVM)
```
shamsimam: Can we remove this TODO now as it is cancelable now?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -938,16 +939,14 @@
...
-                       :timeout-hours (* 2 24)}
+  [conn driver config trigger-chan]
+  (let [config (merge {:timeout-hours (* 2 24)}
                       config)]
-    (chime-at (periodic/periodic-seq (time/now) (time/minutes (:timeout-interval-minutes config)))
-              (fn [now] (kill-lingering-tasks now conn driver config))
-              {:error-handler (fn [e]
-                                (log/error e "Failed to reap timeout tasks!"))})))
+    (util/chime-at-ch trigger-chan
+                      (fn [now] (kill-lingering-tasks now conn driver config))
```
shamsimam: nitpick: can we add a name for the `fn`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -973,11 +972,11 @@
...
   "Periodically checks for running jobs that are in groups and runs the associated
    straggler handler."
-  [conn driver {:keys [interval-minutes] :or {interval-minutes 1}}]
-  (chime-at (periodic/periodic-seq (time/now) (time/minutes interval-minutes))
-            (fn [now] (handle-stragglers conn #(mesos/kill-task! driver {:value (:instance/task-id %)})))
-            {:error-handler (fn [e]
-                              (log/error e "Failed to handle stragglers"))}))
+  [conn driver trigger-chan]
+  (util/chime-at-ch trigger-chan
+                    (fn [now] (handle-stragglers conn #(mesos/kill-task! driver {:value (:instance/task-id %)})))
```
shamsimam: nitpick: can we add a name for the `fn`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -438,3 +439,32 @@
...
                                          :instance.status/unknown}))
+
+(defn close-when-ch!
+  "When the value passed in is a channel, close it. Otherwise, do nothing"
+  [maybe-ch]
+  (try
+    (async/close! maybe-ch)
+    (catch Exception e)))
+
+(defn chime-at-ch
```
shamsimam: In many ways, this is just an actor that is reading messages from its queue (`ch`) and processing them (by calling `f`). I'm not sure we want to have `chime` in the name of this function despite its origin.

scheduler/src/cook/mesos/util.clj
```diff
@@ -438,3 +439,32 @@
...
+  [ch f & [{:keys [error-handler on-finished]
+            :or {error-handler identity
+                 on-finished #()}}]]
+  (async/go-loop []
+                 (if-let [x (async/<! ch)]
+                   (do (async/<! (async/thread
+                                   (try
+                                     (f x)
+                                     (catch Exception e
+                                       (error-handler e)))))
```
shamsimam: Should `error-handler` be able to trigger an end for the `go-loop`?

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1012,16 +1012,17 @@
...
         offer-2 {:id {:value "bar"}}
         offer-3 {:id {:value "baz"}}]
     (testing "offer chan overflow"
-      (sched/receive-offers offers-chan mock-driver [offer-1])
-      @(sched/receive-offers offers-chan mock-driver [offer-2])
-      @(sched/receive-offers offers-chan mock-driver [offer-3])
+      (sched/receive-offers offers-chan match-trigger-chan mock-driver [offer-1])
+      @(sched/receive-offers offers-chan match-trigger-chan mock-driver [offer-2])
+      @(sched/receive-offers offers-chan match-trigger-chan mock-driver [offer-3])
       (is (= @declined-offer-ids-atom [(:id offer-2) (:id offer-3)])))))
```
shamsimam: we should also verify the data in `match-trigger-chan`. Possibly make it a larger chan than sliding buffer to match the exact number of trigger calls.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -992,9 +991,9 @@
 (timers/deftimer [cook-mesos scheduler killing-cancelled-tasks-duration])
 
 (defn cancelled-task-killer
```
shamsimam: We should have unit tests for the functions that accept trigger chan arguments?

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
+         (-> state
+             (update :task-id->task #(dissoc % task-id))
+             (update :task-id->completed-task #(assoc % task-id (assoc task :complete-time (t/now))))
+             (update-in [:slave-id->host (:value (:slave-id task)) :available-resources]
+                        #(combine-resources (:resources task) %)))))
+     state)))
+
+(defn complete-tasks!
+  "Removes tasks from state that have completed and calls complete-task!"
+  [state scheduler driver]
```
shamsimam: nitpick: use destructuring on `state`

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
+   (if-let [task (get-in state [:task-id->task task-id])]
+     (do
+       (let [task->complete-status (get-in state [:config :task->complete-status]) 
+             task-state (or task-state (task->complete-status task))]
+         (.statusUpdate scheduler driver (mesos-type/->pb :TaskStatus
+                                                          {:task-id {:value task-id}
+                                                           :state task-state}))
+         (-> state
+             (update :task-id->task #(dissoc % task-id))
+             (update :task-id->completed-task #(assoc % task-id (assoc task :complete-time (t/now))))
```
shamsimam: How accurately does completion time need to be measured? The completion time can now be off by 20 ms.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -287,10 +311,8 @@
       (log/debug "Launching tasks " {:tasks tasks})
       (-> state
           (update :task-id->task #(into % (map-from-vals (comp :value :task-id) tasks)))
-          (update :complete-chan->task-id
-                  #(into % (map (fn [{:keys [task-id] :as task}]
-                                  [(async/timeout (task->runtime-ms task)) (:value task-id)]))
-                         tasks))
+          (update :time-task-id-pairs
+                  #(into % new-time-task-id-pairs))
```
shamsimam: nitpick: can fit in a single line now.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
+             (update :task-id->completed-task #(assoc % task-id (assoc task :complete-time (t/now))))
+             (update-in [:slave-id->host (:value (:slave-id task)) :available-resources]
+                        #(combine-resources (:resources task) %)))))
+     state)))
+
+(defn complete-tasks!
+  "Removes tasks from state that have completed and calls complete-task!"
+  [state scheduler driver]
+  (let [now (:now state)
+        time-task-id-pairs (:time-task-id-pairs state)
```
shamsimam: Should we use `group-by` to create the completed and remaining groups?

scheduler/src/cook/mesos.clj
```diff
@@ -140,6 +143,25 @@
...
+    :or {timeout-interval-minutes 10}
+    :as task-constraints}]
+  (let [match-trigger-chan (async/chan (async/sliding-buffer 1))]
+    ;; Use a normal channel for match-trigger-chan because other events besides chimes will go on channel
+    (async/pipe (chime-ch (periodic/periodic-seq (time/now) (time/seconds 1)))
+                match-trigger-chan)
+    {:rebalancer-trigger-chan (chime-ch (periodic/periodic-seq (time/now)
+                                                               (time/seconds (:interval-seconds rebalancer-config))))
+     :match-trigger-chan match-trigger-chan
+     :rank-trigger-chan (chime-ch (periodic/periodic-seq (time/now) (time/seconds 5)))
```
wyegelwel: Yes, but can it wait for a different PR? I'd like to keep this PR small and.. it is no worse than it was =)

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
-                                                       {:task-id {:value task-id}
-                                                        :state task-state}))
-      (-> state
-          (update :task-id->task #(dissoc % task-id))
-          (update :task-id->completed-task #(assoc % task-id (assoc task :complete-time (t/now))))
-          (update-in [:slave-id->host (:value (:slave-id task)) :available-resources]
-                     #(combine-resources (:resources task) %))))
-    state))
+  ([state task-id scheduler driver]
+   (complete-task! state task-id scheduler driver nil))
```
wyegelwel: I would prefer that but then a lot more logic (getting the task) leaks into this first case. I don't see an obviously nice way to refactor this. I'm going to leave as is for now, let me know if you see a nicer refactor.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
+  ([state task-id scheduler driver task-state]
+   (log/debug "Completing task" {:task-id task-id :state task-state :task (get-in state [:task-id->task task-id])})
+   (if-let [task (get-in state [:task-id->task task-id])]
+     (do
+       (let [task->complete-status (get-in state [:config :task->complete-status]) 
+             task-state (or task-state (task->complete-status task))]
```
wyegelwel: `task->complete-status` is likely a function so this unfortunately won't work.

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: Not particularly accurately. Where are you getting the 20ms from?

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: Ah, that is great idea! I knew there was a good way to do this but couldn't figure it out. Thank you. And thanks for the 2x speed up =)

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -209,19 +211,37 @@
...
+     state)))
+
+(defn complete-tasks!
+  "Removes tasks from state that have completed and calls complete-task!"
+  [state scheduler driver]
+  (let [now (:now state)
+        time-task-id-pairs (:time-task-id-pairs state)
+        complete? (comp (partial t/after? now) first)
+        to-complete-pairs (filter complete? time-task-id-pairs)
+        remaining (remove complete? time-task-id-pairs)
```
wyegelwel: Ideally, yes, but the filter for the kill would be a bit time consuming. I think as is it is fine since eventually the tasks time will be up and when we go to complete it the task id won't be there and it will be a noop. 

I propose leaving it for now since the only use case for the mesos mock, the simulator, doesn't need it and the implementation is correct (though a little wasteful) as is.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -275,8 +295,12 @@
           requested-resources (combine-resources (mapcat :resources tasks))
           resources' (subtract-resources available-resources requested-resources)
           tasks (map #(assoc % :launched-time (t/now)) tasks)
-          task->runtime-ms (-> state :config :task->runtime-ms)]
-      (log/info "Resources requested by tasks: " {:requested-resources requested-resources})
+          task->runtime-ms (-> state :config :task->runtime-ms)
+          new-time-task-id-pairs (map (fn [{:keys [task-id] :as task}]
```
wyegelwel: That would certainly make the filter on kill easier =)

I'm reasonably indifferent though inertia is leading me to lean toward keeping it as is. What is your reasoning why?

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -361,22 +382,25 @@
...
-  ([hosts speed-multiplier task->runtime-ms scheduler]
-   (mesos-mock hosts speed-multiplier task->runtime-ms default-task->complete-status scheduler))
-  ([hosts speed-multiplier task->runtime-ms task->complete-status scheduler]
+  ([hosts offer-trigger-chan scheduler & 
+    {:keys [task->complete-status task->runtime-ms state-atom
+            complete-trigger-chan]
+     :or {task->complete-status default-task->complete-status
+          task->runtime-ms default-task->runtime-ms
+          state-atom (atom nil)
+          complete-trigger-chan (chime-ch (periodic/periodic-seq (t/now) (t/millis 20)))}}]
```
wyegelwel: Sort of, I was observing long delays for tasks to be completed because it would require many cycles for a bunch of tasks to complete around the same time to be completed. Also, moving to this model gives finer control over when completions happen (since we can trigger them).

scheduler/src/cook/mesos/util.clj
wyegelwel: That is a good point. What do you think would be a better name?

scheduler/src/cook/mesos/util.clj
wyegelwel: Perhaps, although it is currently not needed. Are you ok with leaving it as is until it is needed?

scheduler/src/cook/mesos/util.clj
wyegelwel: Currently no, though it is possible. This was a bit of a hack to try to support the old chime logic but get completion notification semantics. I will pass current time instead of x into f since that is closer to the original chime contract anyway.

wyegelwel: Never mind, I will call f with no parameters.

scheduler/src/cook/mesos.clj
shamsimam: Okay, sgtm.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: One option is for `complete-tasks!` to pre-compute `task-state`, then we do not need the overloaded version.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: Oh okay.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: `complete-trigger-chan` chimes every 20ms and calls `complete-tasks!` that eventually calls `complete-task!`

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: Okay.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: Its makes it clear that this structure represent the task completion time. 

Also, when I see a vector I interpret there can be multiple entries for a given task:  [time-1, task-id] and [time-2, task-id].

shamsimam: We can change it in a separate PR.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: Okay, sgtm.

scheduler/src/cook/mesos/util.clj
shamsimam: Okay.

scheduler/src/cook/mesos/util.clj
shamsimam: Naming is more @sradack 's forte.

If we are calling `f` with current time, we could call the function `process-heartbeat`.

wyegelwel:
ok, updated addressing your comments

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: That is in the tests. The simulator will do something different

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: That is true but then it pushes the if task id is found logic into `complete-tasks!`. Although I agree this function is a little gross, I propose we keep it as it seems to be the lesser of a few evils.

scheduler/src/cook/mesos/mesos_mock.clj
wyegelwel: But to your larger question, yea it is fine for it to be inaccurate. It is inaccurate in production, likely much more than 20ms

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: I agree, but don't want to hold up the PR on them. My reasoning is that these functions do very little except call the functions that do real work (and are unit tested). Also, the main ones (rank, rebalance, match) are covered by the mesos mock cook scheduler integration test. Are you ok putting this off?

scheduler/src/cook/mesos/util.clj
wyegelwel: I don't like `heartbeat` since it is so overloaded. 

It is really an in order async trigger processor..

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: ok.

scheduler/src/cook/mesos/mesos_mock.clj
shamsimam: okay

scheduler/src/cook/mesos/scheduler.clj
shamsimam: okay.

scheduler/src/cook/mesos/util.clj
shamsimam: okay, not particularly good with naming :(

shamsimam:
lgtm, let me know if you would like me to merge or have another set of eyes look at it.

scheduler/src/cook/mesos/util.clj
wyegelwel: Wanna leave it for now? Thankfully, the isn't a public api so we can always rename later.

shamsimam: sure.

PR #341, opened 2017-05-16
Initial simulator implementation

wyegelwel:


scheduler/docs/simulator.md
```diff
@@ -0,0 +1,146 @@
+# Cook scheduler faster than real time simulator
```
wyegelwel: I need to make edits to this file still. Jump to the code

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -761,16 +761,10 @@
...
-                               offers-chan ([offers]
-                                            (counters/dec! offer-chan-depth)
-                                            offers)
-                               trigger-chan ([_] [])
-                               :priority true)
                       ;; Try to clear the channel
                       offers (->> (repeatedly chan-length #(async/poll! offers-chan))
-                                  (filter nil?)
-                                  (reduce into offers))
+                                  (remove nil?)
```
wyegelwel: holy shit this was a bad bug..

scheduler/src/cook/components.clj
```diff
@@ -438,7 +438,7 @@
                          (throw (ex-info "You enabled nrepl but didn't configure a port. Please configure a port in your config file." {})))
                        ((lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))}))
 
-(defn- init-logger
+(defn init-logger
```
shamsimam: why does this function need to become public?

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Do we need to reduce the counter?
[`(counters/dec! offer-chan-depth (count offers))`]

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,571 @@
...
+                match-complete-chan (async/chan)
+                rank-complete-chan (async/chan)
+                rebalancer-complete-chan (async/chan)]
+            (log/info "Simulation time: " simulation-time)
+            (log/info "submission-batch size: " (count submission-batch))
+
+            (log/info "Submitting batch")
+            ;; Submit new jobs
+            (doseq [job submission-batch]
+              (org.joda.time.DateTimeUtils/setCurrentMillisFixed (inc (.getTime (tc/to-date (t/now)))))
```
shamsimam: Why does this call have to be inside the loop? Is it a problem if all the jobs have the same submission time?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,571 @@
...
+                                    ;; This relies on
+                                    ;; 1. Time is controlled
+                                    ;; 2. We increment time for each job submitted
+                                    (tc/to-date (t/now)))
+                        50
+                        60000))
+            (log/info "Batch submission complete")
+
+            ;; Rebalance
+            (log/info "Starting rebalance")
```
shamsimam: Why does every simulation cycle trigger rebalancing? Shouldn't it be triggered based on whether enough simulated time has elapsed since last rebalancing?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,571 @@
...
+                                                    (keys (:task-id->task @state-atom))))]
+                          {:running-tasks-ents (count cook-tasks)
+                           :tasks-in-mesos (count mesos-tasks)
+                           :cook-minus-mesos (count (clojure.set/difference cook-tasks mesos-tasks))
+                           :mesos-minus-cook (count (clojure.set/difference mesos-tasks cook-tasks))})
+                       )
+            (log/info "Completion statuses sent")
+
+            ;; Request rank occurs
+            (log/info "Starting rank")
```
shamsimam: Why does every simulation cycle trigger ranking? Shouldn't it be triggered based on whether enough simulated time has elapsed since last ranking?

scheduler/src/cook/components.clj
wyegelwel: I call it in the simulator main.

scheduler/test/cook/test/simulator.clj
wyegelwel: It is just for debugging as a sanity check. It is not strictly necessary.

scheduler/test/cook/test/simulator.clj
wyegelwel: Yea, that would be better but would mean more things to configure for the sim (cycle step time, rebalancer interval, rank interval, match interval...). Essentially as written, all the process intervals equal the cycle step time.

wyegelwel: I will probably end up changing it to your suggestion.

scheduler/test/cook/test/simulator.clj
shamsimam: Okay

scheduler/src/cook/components.clj
shamsimam: Okay.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Can we introduce unit tests that will display the bug fixed by `remove nil?`?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,586 @@
...
+  [mesos-hosts trace cycle-step-ms config]
+  (let [simulation-time (-> trace first :submit-time-ms)
+        mesos-datomic-conn (restore-fresh-database! (str "datomic:mem://mock-mesos"))
+        offer-trigger-chan (async/chan)
+        complete-trigger-chan (async/chan)
+		ranker-trigger-chan (async/chan)
+        matcher-trigger-chan (async/chan)
+        rebalancer-trigger-chan (async/chan) 
+        state-atom (atom {})
+        make-mesos-driver-fn (fn [scheduler _];; _ is framework-id
```
dposada: Why the comment here?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,586 @@
...
+                   job-info)]
+    job-info))
+
+(defn summary-stats [jobs]
+  (->> jobs
+       (map util/job-ent->resources)
+       (map #(assoc % :count 1))
+       (reduce (partial merge-with +))))
+
+(comment
```
dposada: Can we get rid of the comment?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,586 @@
+(ns cook.test.simulator
```
dposada: It doesn't seem like this should live under `scheduler/test`. Maybe it should go in the same directory with the example files?

wyegelwel: I think it is worth keeping it where it is. It is still effectively a test of cook. I will add an actual unit test to highlight that.

scheduler/simulator_files/example-out-trace.csv
```diff
@@ -0,0 +1,1202 @@
...
+58f8f051-7baf-461e-b686-38080a1c6660,33a2d6c6-9b83-4f89-832f-56fda4ccc9a1,1493391232852,1493391472876,1493391472876,1493391862821,22,e29ca590-dce7-43bf-846e-fb41907a87b0,:instance.status/running,,a,1999.0,4.0,dummy_job,502556,finished
+58f8f051-971a-4441-bd9a-7f4db32927c7,2e596bdb-8519-4607-a8dc-de790c517453,1493390212882,1493390212886,1493390212886,1493390842820,12,25c26ed8-515a-4cdb-a2d6-100f862d8df7,:instance.status/success,,d,2447.0,1.0,dummy_job,612498,finished
+58f8f051-c28b-4646-b496-ea9b30cfa4ec,2e185083-8e26-4cd3-9a17-09994dfdded7,1493391472849,1493391472876,1493391472876,1493391862821,22,e29ca590-dce7-43bf-846e-fb41907a87b0,:instance.status/running,,c,2969.0,2.0,dummy_job,553634,finished
+58f8f051-2abc-41d2-b177-b335145cad4c,3f50bbe8-22fc-4a0f-b27f-20b6dd09f3e5,1493390212871,1493390212886,1493390212886,1493390362822,12,25c26ed8-515a-4cdb-a2d6-100f862d8df7,:instance.status/success,,b,1721.0,1.0,dummy_job,146017,finished
+58f8f051-6ede-4d9c-8c75-ead4fa78c3c4,7e7973e8-7e53-4a9e-b336-3c7cd3a3b439,1493391472864,1493391472876,14933914728
```
wyegelwel: good point, will change

scheduler/test/cook/test/simulator.clj
dposada: Makes sense, should we move the example files to the same directory then?

scheduler/simulator_files/example-hosts.json
```diff
@@ -0,0 +1,541 @@
+[ {
```
wyegelwel: I'd prefer to keep these files json so other tools could consume them.

wyegelwel:
I've updated the PR to respond to comments. When you ( @shamsimam and @dposada ) get a chance, please take another pass.

wyegelwel:
@shamsimam @dposada, do you have more comments?

dposada:
@wyegelwel I will take another look today

wyegelwel:
Any more feedback or is this ok to merge?

wyegelwel:
Passing build: https://travis-ci.org/twosigma/Cook/builds/233817393

dposada:
Should we do something to more clearly distinguish this simulator from the existing thing we call simulator (https://github.com/twosigma/Cook/tree/master/simulator)?

wyegelwel:
> Should we do something to more clearly distinguish this simulator from the existing thing we call simulator (https://github.com/twosigma/Cook/tree/master/simulator)?

Yes, though I'm not sure what the best way to do that is. What do you think? Perhaps just add a note in the readme of (https://github.com/twosigma/Cook/tree/master/simulator) and the docs of this simulator distinguishing it from the other?

dposada:
That seems reasonable, unless we can think of a work other than "simulator" for the new thing

wyegelwel:
Ok, I updated the both docs with a message about the other simulator

wyegelwel:
This is approved. Ok to merge?

shamsimam:
Looking at it now.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -590,7 +590,9 @@
   (for [{:keys [tasks leases]} matches
         :let [offers (mapv :offer leases)
               slave-id (-> offers first :slave-id :value)]
-        ^TaskAssignmentResult task tasks
+        ^TaskAssignmentResult task (->> tasks
+                                        ;; sort-by makes match transaction deterministic
+                                        (sort-by (comp :db/id :job #(.getRequest %))))
```
shamsimam: How much does the `sort` cost computationally? Given that we want to optimize schedule times, should we make this configurable?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -761,16 +763,12 @@
...
-                                            (counters/dec! offer-chan-depth)
-                                            offers)
-                               trigger-chan ([_] [])
-                               :priority true)
                       ;; Try to clear the channel
-                      offers (->> (repeatedly chan-length #(async/poll! offers-chan))
-                                  (filter nil?)
-                                  (reduce into offers))
+                      offers (->> (util/read-chan offers-chan chan-length)
+                                  ((fn [offer-lists]
```
shamsimam: nitpick: can we add a name to the `fn`

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,599 @@
...
+                               :label/value)
+     :requested_status  (->> (:job/label job)
+                             (filter #(= (:label/key %) "JOB-STATUS"))
+                             first
+                             :label/value)}))
+
+(defn dump-jobs-to-csv
+  "Given a mesos db, dump a csv with a row per task
+
+   Columns:
```
shamsimam: nitpick: can these be listed and generate the output in sorted order?

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,599 @@
...
+                        :commit-latch/committed? true}
+          txn [runtime-env
+               status-env
+               commit-latch
+               (-> (select-keys job job-keys)
+                   (assoc :db/id (d/tempid :db.part/user))
+                   (update :job/uuid #(java.util.UUID/fromString %))
+                   (update :job/command #(or % ""))
+                   (update :job/name #(or % ""))
+                   (assoc :job/commit-latch commit-latch-id)
```
shamsimam: nitpick: collapse the subsequent `assoc`s into this one.

scheduler/test/cook/test/simulator.clj
```diff
@@ -0,0 +1,599 @@
...
+  (->> (d/q '[:find [?i ...]
+         :where
+         [?i :instance/task-id _]]
+       mesos-db)
+       (map (partial d/entity mesos-db))))
+
+(defn submit-job
+  "Submits a job to datomic"
+  [conn job]
+  (let [job-keys [:job/uuid :job/command :job/user :job/name :job/max-retries
```
shamsimam: can we list the keys in sorted order, it makes it easier to find entries that are being retrieved

scheduler/test/cook/test/simulator.clj
shamsimam: Note to self: answer same as rebalancing one above.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: At most there will be `max-considerable` tasks in the list. We currently run with 200 as `max-considerable` so the sort should be fairly lightweight.

scheduler/test/cook/test/simulator.clj
wyegelwel: They are in a specific order that makes it easy to work with as a csv/dataframe. How strongly do you feel about them being in sorted order?

shamsimam: Okay. Sounds good.

shamsimam: I prefer order unless there is a good reason not to. Looks like there is a good reason here.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Okay. Sounds good.

wyegelwel:
@shamsimam I responded to your comments. Let me know if you have more feedback

scheduler/test/cook/test/simulator.clj
```diff
@@ -220,15 +203,15 @@
                status-env
                commit-latch
                (-> (select-keys job job-keys)
-                   (assoc :db/id (d/tempid :db.part/user))
+                   (assoc :db/id (d/tempid :db.part/user)
```
shamsimam: nitpick: sorted order for keys :(

wyegelwel:
I sorted the keys. If there is no more feedback, please merge.

PR #345, opened 2017-05-17
Allow for at least two runs of the lingering task killer for test_max_runtime_exceeded

shamsimam:


shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/233406805

PR #346, opened 2017-05-17
Adds tests against multiple cook schedulers

dposada:


dposada:
Passing build: https://travis-ci.org/twosigma/Cook/builds/233417110

PR #347, opened 2017-05-18
Optimize init-user->dru-divisors and get-share

shamsimam:


shamsimam:
Local experiments show I can get execution times for `init-user->dru-divisors` down to ~300 ms from 20+ seconds.

shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/233674170

scheduler/src/cook/mesos/share.clj
```diff
@@ -44,30 +44,59 @@
...
+  (defn- get-share-by-type
+    "Query a user's pre-defined share by type (e.g. :cpus, :mem, :gpus).
+
+     If a user's pre-defined share is NOT defined, return either:
+     1. the explicitly provided default value, or
+     2. the share for the \"default\" user. If there is NO \"default\"
+        value for a specific type, return Double.MAX_VALUE."
+    ([db type user]
+     (let [datomic-resource-type (resource-type->datomic-resource-type type)]
+       (or (-> (q query db user datomic-resource-type) ffirst)
```
pschorf: Why do you need `->` here?

scheduler/src/cook/mesos/share.clj
```diff
@@ -44,30 +44,59 @@
...
    \"default\" user. If there is NO \"default\" value for a specific type,
    return Double.MAX_VALUE."
-  [db user]
-  (->> (util/get-all-resource-types db)
-       (map (fn [type] [type (get-share-by-type db type user)]))
-       (into {})))
+  ([db user]
+   (get-share db user (util/get-all-resource-types db)))
+  ([db user all-resource-types]
+   (->> all-resource-types
```
pschorf: Same here, can you get rid of the threading macro?

scheduler/src/cook/mesos/share.clj
shamsimam: Done.

scheduler/src/cook/mesos/share.clj
shamsimam: Done.

scheduler/src/cook/mesos/share.clj
```diff
@@ -146,13 +175,11 @@
...
-                                 (map (fn [user]
-                                        [user (get-share db user)]))
-                                 ;; In case default-user doesn't have an explicit share
-                                 (cons [default-user (get-share db default-user)])
-                                 (into {}))]
+          default-user-share (get-share db default-user)
+          user->share-cache (-> (get-shares db all-share-users)
+                                ;; In case default-user doesn't have an explicit share
+                                (assoc default-user default-user-share))]
       (fn user->share
```
shamsimam: This is indirectly called from `rank-jobs`. So optimizing this will also help speed up the ranking.

shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/233877996

scheduler/src/cook/mesos/share.clj
wyegelwel: @shamsimam, do you have a sense of the improvement this will make to ranking?

shamsimam: Made the change last night, I will measure it later today.

shamsimam: I measured locally, this change doesn't have a noticeable impact `sort-jobs-by-dru` (and hence `rank-jobs`). Runs for the new and old variants of `sort-jobs-by-dru` were running in about 4 seconds.

scheduler/src/cook/mesos/dru.clj
```diff
@@ -31,18 +28,11 @@
...
-                               (into #{}))
-        all-users (clojure.set/union all-running-users all-pending-users)
-        user->dru-divisors (->> all-users
-                                (map (fn [user]
-                                       [user (share/get-share db user)]))
-                                (into {}))]
-    user->dru-divisors)))
+    (let [all-running-users (map util/task-ent->user running-task-ents)
+          all-pending-users (map :job/user pending-job-ents)
+          all-users-set (-> #{} (into all-running-users) (into all-pending-users))
```
sradack: As long as we're optimizing, would it be faster to `(set (concat all-running-users all-pending-users))`?

scheduler/src/cook/mesos/share.clj
sradack: Does that mean we're not merging this?

scheduler/src/cook/mesos/dru.clj
shamsimam: We've run into `concat` issues recently, I prefer avoiding it in favor of `into`.

scheduler/src/cook/mesos/share.clj
shamsimam: Why not? It is still functionally equivalent and an improvement in terms of making DB queries.

shamsimam:
Any more feedback on this?

shamsimam:
Can we get this merged?

scheduler/src/cook/mesos/share.clj
```diff
@@ -44,30 +44,59 @@
...
-                :in $ ?u ?t
-                :where
-                [?e :share/user ?u]
-                [?e :share/resource ?r]
-                [?r :resource/type ?t]
-                [?r :resource/amount ?a]]]
-    (or (-> (q query db user type) ffirst)
-        (-> (q query db default-user type) ffirst)
-        (Double/MAX_VALUE))))
+(let [query '[:find ?a
```
wyegelwel: nitpick: If you change the query to be `:find ?a .` it will return a single element so you don't need to do ffirst down belowhttp://docs.datomic.com/query.html#sec-5-8

scheduler/src/cook/mesos/share.clj
```diff
@@ -44,30 +44,59 @@
...
+     If a user's pre-defined share is NOT defined, return either:
+     1. the explicitly provided default value, or
+     2. the share for the \"default\" user. If there is NO \"default\"
+        value for a specific type, return Double.MAX_VALUE."
+    ([db type user]
+     (let [datomic-resource-type (resource-type->datomic-resource-type type)]
+       (or (ffirst (q query db user datomic-resource-type))
+           (get-share-by-type db type default-user Double/MAX_VALUE))))
+    ([db type user default]
+     (let [datomic-resource-type (resource-type->datomic-resource-type type)]
```
wyegelwel: Don't like the code reuse here. Perhaps use a `delay` so we can have a single implementation?

wyegelwel:
Few small comments otherwise looks good.

scheduler/src/cook/mesos/share.clj
shamsimam: Looks like the `ffirst` is necessary, else the query returns the amounts in a list.

scheduler/src/cook/mesos/share.clj
shamsimam: Thanks for pointing this out. Made two recursive calls and used an `or`. This further allowed privatizing the query. Sweet!

shamsimam:
@wyegelwel addressed your comments

shamsimam:
Any more feedback on this?

PR #348, opened 2017-05-18
Use LRU cache instead of FIFO cache

pschorf:


scheduler/src/cook/components.clj
```diff
@@ -245,7 +245,7 @@
      :mesos-pending-jobs-atom (fnk [] (atom {}))
      :mesos-offer-cache (fnk [[:settings [:offer-cache max-size ttl-ms]]]
                           (-> {}
-                              (cache/fifo-cache-factory :threshold max-size)
+                              (cache/lru-cache-factory :threshold max-size)
```
dposada: Should we add a comment about why we're using lru instead of fifo?

PR #352, opened 2017-05-23
Add binding for disable-mea-culpa-retries

WenboZhao:


jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -795,6 +825,7 @@ public static JSONObject jsonizeJob(Collection<Job> jobs)
             jobBuilder.setCommand(json.getString("command"));
             jobBuilder.setPriority(json.getInt("priority"));
             jobBuilder.setStatus(Status.fromString(json.getString("status")));
+            jobBuilder.disableMeaCulpaRetries(json.getBoolean("disable-mea-culpa-retries"));
```
WenboZhao: @wyegelwel Why this is `disable-mea-culpa-retries` instead of `disable_mea_culpa_retries`. For example, we have `max_retries` instead of `max-retries`

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -358,6 +361,20 @@ public Builder setRetries(Integer retries) {
...
         /**
+         * Whether to disable the "mea-culpa" retries.
+         *
+         * @param disableMeaCulpaRetries {@link Boolean} specifies whether "mea-culpa" retries is disabled.
+         * @see <a href ="https://github.com/twosigma/Cook/blob/master/scheduler/docs/faq.md#how-can-i-configure-my-job-to-run-exactly-once">
+         *      how-can-i-configure-my-job-to-run-exactly-once
+         *      </a>
+         * @return this builder.
+         */
+        public Builder disableMeaCulpaRetries(Boolean disableMeaCulpaRetries) {
```
shamsimam: The method should be named `setMeaCulpaRetries for consistency.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -460,6 +477,7 @@ public Builder setApplication(Application application) {
     final private Long _expectedRuntime;
     final private Integer _priority;
     final private Status _status;
+    final private Boolean _disableMeaCulpaRetries;
```
shamsimam: Maybe name this `_meaCulpaRetries`

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -685,6 +714,7 @@ public static JSONObject jsonizeJob(Job job)
         object.put("cpus", job.getCpus());
         object.put("priority", job.getPriority());
         object.put("max_retries", job.getRetries());
+        object.put("disable-mea-culpa-retries", job.isMeaCulpaRetriesDisabled());
```
shamsimam: The key should use underscores for consistency.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: k, will change

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: will change to `_isMeaCulpaRetriesDisable`

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: From the wiki, it says "-"

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -131,7 +132,8 @@ public Job build() {
                 _name = "cookjob";
             }
```
shamsimam: Do we need to set the default value for `_isMeaCulpaRetriesDisabled` to false?

Alternatively, why are we using `Boolean` instead of `boolean`?

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -358,6 +365,32 @@ public Builder setRetries(Integer retries) {
...
 
         /**
+         * Disable the "mea-culpa" retries.
+         *
+         * @see <a href ="https://github.com/twosigma/Cook/blob/master/scheduler/docs/faq.md#how-can-i-configure-my-job-to-run-exactly-once">
+         *      how-can-i-configure-my-job-to-run-exactly-once
+         *      </a>
+         * @return this builder.
+         */
+        public Builder disableMeaCulpaRetries() {
```
shamsimam: We can have just one setter, why the enable and disable variants?

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -460,6 +493,7 @@ public Builder setApplication(Application application) {
     final private Long _expectedRuntime;
     final private Integer _priority;
     final private Status _status;
+    final private Boolean _isMeaCulpaRetriesDisabled;
```
shamsimam: Same question on `Boolean` vs `boolean`.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: I will set the default. The Boolean is to be consistent with others. I could address those in other CR.

shamsimam: Okay.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: It is more clear. Btw, this is first config for boolean. So prefer to use disable/enableXXX

shamsimam: Ok.

shamsimam:
LGTM

PR #353, opened 2017-05-23
Fix disable-mea-culpa-retries

pschorf:


integration/tests/cook/test_basic.py
```diff
@@ -29,6 +29,12 @@ def test_basic_submit(self):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertEqual('success', job['instances'][0]['status'])
 
+    def test_disable_mea_culpa(self):
+        job_uuid, resp = util.submit_job(self.cook_url, disable_mea_culpa_retries=True)
+        self.assertEqual(201, resp.status_code)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
```
dposada: Do we need to wait for it to complete just to ensure that the field was set?

integration/tests/cook/test_basic.py
```diff
@@ -29,6 +29,12 @@ def test_basic_submit(self):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertEqual('success', job['instances'][0]['status'])
 
+    def test_disable_mea_culpa(self):
```
dposada: Do we also need tests for:

- explicitly setting the field to false
- not setting the field and asserting the default
- checking the actual effect of the field

?

pschorf: Added tests for 1 & 2. Three I think will be difficult to reproduce in an integration test, but there is already a unit test for it.

integration/tests/cook/test_basic.py
pschorf: Done

PR #356, opened 2017-05-24
Add a default value for disable-mea-culpa-retries

pschorf:


wyegelwel:
Lgtm though agree on integration comment

pschorf:
Added a test: also, currently submit_time can be nil, but the schema didn't allow it (which would cause the same error in prod)

I made the schema more inclusive, let me know what you think.

scheduler/src/cook/mesos/api.clj
```diff
@@ -265,7 +265,7 @@
       (merge {:framework-id (s/maybe s/Str)
               :status s/Str
               :state s/Str
-              :submit-time PosInt
+              :submit-time (s/maybe PosInt)
```
wyegelwel: Why wouldn't submit-time be set?

scheduler/src/cook/mesos/api.clj
```diff
@@ -741,7 +741,7 @@
                  :max_retries (:job/max-retries job) ; consistent with input
                  :retries_remaining (- (:job/max-retries job) (util/job-ent->attempts-consumed db job))
                  :max_runtime (:job/max-runtime job Long/MAX_VALUE) ; consistent with input
-                 :framework_id fid
+                 :framework_id (str fid)
```
wyegelwel: why do we need a str here?

scheduler/src/cook/mesos/api.clj
pschorf: There's a comment:

; due to a bug, submit time may not exist for some jobs

In that case, as it stands you can't query those jobs. To me, I'd rather update the schema. We can also populate a default value instead.

scheduler/src/cook/mesos/api.clj
```diff
@@ -729,7 +729,7 @@
         application (:job/application job)
         expected-runtime (:job/expected-runtime job)
         job-map {:command (:job/command job)
-                 :uuid (str (:job/uuid job))
+                 :uuid (:job/uuid job)
```
pschorf: The schema specifies a UUID, not a string. I am guessing there is some magic in compojure that cleans this up but I'd prefer it to conform to the schema here.

scheduler/src/cook/mesos/api.clj
pschorf: The schema specifies a string, not a framework id record

pschorf: Actually, I don't think `str` is the right thing. Let me check this first before we merge.

pschorf: Ok I gave up on this one. I have no idea how this is getting converted, so I made a special case in the test.

PR #363, opened 2017-06-01
Add job constraints

wyegelwel:
This adds the ability to set constraints on individual jobs and provides a way to grow the constraints api in the future. It is inspired by marathon's constraint api: https://mesosphere.github.io/marathon/docs/constraints.html

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -306,6 +306,29 @@ class CookTest(unittest.TestCase):
         resp = util.session.post('%s/rawscheduler' % self.cook_url,
                                  json={'jobs':[job1, job2]})
         self.assertEqual(resp.status_code, 500)
+    
+    def test_constraints(self):
+        # Want to ensure job is put on specific host, but no way to set host names in minimesos
+        job_uuid, resp = util.submit_job(self.cook_url)
+        self.assertEqual(resp.status_code, 201)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        hostname = job['instances'][0]['hostname']
+
+        bad_job_uuid, resp = util.submit_job(self.cook_url, constraints=[["HOSTNAME", 
+                                                                          "EQUALS", 
+                                                                          "lol won't get scheduled"]])
+        self.assertEqual(resp.status_code, 201, resp.text)
+        
+        constraints = [["HOSTNAME", "EQUALS", hostname]]
+        job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints)
+        self.assertEqual(resp.status_code, 201, resp.text)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        hostname_constrained = job['instances'][0]['hostname']
+        self.assertEqual(hostname, hostname_constrained)
+        self.assertEqual(constraints, job['constraints'])
+        # This job should have been scheduled since the job submitted after it has completed
+        # however, its constraint means it won't get scheduled
+        job = util.wait_for_job(self.cook_url, bad_job_uuid, 'waiting', max_delay=3000)
 
     def test_allow_partial(self):
         def absent_uuids(response):
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -17,6 +17,7 @@ class CookTest(unittest.TestCase):
 
     def setUp(self):
         self.cook_url = util.retrieve_cook_url()
+        self.mesos_url = util.retrieve_mesos_url()
         self.logger = logging.getLogger(__name__)
         util.wait_for_cook(self.cook_url)
 
@@ -308,24 +309,26 @@ class CookTest(unittest.TestCase):
         self.assertEqual(resp.status_code, 500)
     
     def test_constraints(self):
-        # Want to ensure job is put on specific host, but no way to set host names in minimesos
-        job_uuid, resp = util.submit_job(self.cook_url)
-        self.assertEqual(resp.status_code, 201)
-        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
-        hostname = job['instances'][0]['hostname']
+        state = util.get_mesos_state(self.mesos_url)
+        hosts = [agent['hostname'] for agent in state['slaves']]
 
         bad_job_uuid, resp = util.submit_job(self.cook_url, constraints=[["HOSTNAME", 
                                                                           "EQUALS", 
                                                                           "lol won't get scheduled"]])
         self.assertEqual(resp.status_code, 201, resp.text)
         
-        constraints = [["HOSTNAME", "EQUALS", hostname]]
-        job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints)
-        self.assertEqual(resp.status_code, 201, resp.text)
-        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
-        hostname_constrained = job['instances'][0]['hostname']
-        self.assertEqual(hostname, hostname_constrained)
-        self.assertEqual(constraints, job['constraints'])
+        host_to_job_uuid = {}
+        for hostname in hosts:
+            constraints = [["HOSTNAME", "EQUALS", hostname]]
+            job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints)
+            self.assertEqual(resp.status_code, 201, resp.text)
+            host_to_job_uuid[hostname] = job_uuid
+
+        for hostname, job_uuid in host_to_job_uuid.items():
+            job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+            hostname_constrained = job['instances'][0]['hostname']
+            self.assertEqual(hostname, hostname_constrained)
+            self.assertEqual([["HOSTNAME", "EQUALS", hostname]], job['constraints'])
         # This job should have been scheduled since the job submitted after it has completed
         # however, its constraint means it won't get scheduled
         job = util.wait_for_job(self.cook_url, bad_job_uuid, 'waiting', max_delay=3000)
--- a/scheduler/docs/scheduler-rest-api.asc
+++ b/scheduler/docs/scheduler-rest-api.asc
@@ -80,6 +80,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |`ports` | integer | Number of ports that should be opened for the job.
 |`uris` | list of URI objects | A list of URIs that will be fetched into the container before launch.
 |`env` | JSON map | A map of environment variables to be provided to the job.
+|`constraints` | list | list of constraints in form [attribute, operator, pattern]
 |`disable_mea_culpa_retries` | boolean | Flag to disable mea culpa retries. If true, mea culpa retries will count against the job's retry count.
 |====
 
@@ -113,12 +114,23 @@ The request body is list of job maps with each entry containing a map of the fol
                  "extract": true
              }
          ],
+         "constraints" : [["instance_type", "EQUALS", "r3.8xlarge"]],
          "command" : "echo hello world"
       }
    ]
 }
 ----
 
+===== Constraints
+
+Constraints provide controls over where a job is placed on the cluster. 
+There are both job and group level constraints.
+This will discuss job constraints, see the docs/groups.md for more details on group constraints.
+
+Constraints are specified as a tuple of attribute, operator, pattern.
+Attribute can be any attribute set on a host in the cluster.
+Cook currently only supports the EQUALS operator. In the future, we will add more operators.
+For the case of the EQUALS, Cook will only schedule a job on a host for which the attribute's value equals pattern.
 
 .Possible response codes
 [options="header"]
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -197,7 +197,7 @@
 (def Constraint
   "Schema for user defined job host constraint"
   [(s/one NonEmptyString "attribute")
-   (s/one (s/pred #(contains? (set (map (comp name util/without-ns) constraint-operators))
+   (s/one (s/pred #(contains? (set (map name constraint-operators))
                               (str/lower-case %))
                   'constraint-operator-exists?)
           "operator")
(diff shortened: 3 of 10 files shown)


wyegelwel:
@shamsimam @sradack , any comments on this?

shamsimam:
Having a look now

integration/tests/cook/test_basic.py
```diff
@@ -306,6 +306,29 @@ def test_error_while_creating_job(self):
         resp = util.session.post('%s/rawscheduler' % self.cook_url,
                                  json={'jobs':[job1, job2]})
         self.assertEqual(resp.status_code, 500)
+    
+    def test_constraints(self):
+        # Want to ensure job is put on specific host, but no way to set host names in minimesos
```
shamsimam: Can we add a test where we use the host name from `http://mesos-master:5050/state.json` user `slaves->*->hostname` ?

Then you can verify whether exactly that host was used. Maybe launch multiple jobs with different host constraints and verify each job only ran on the specified host?

scheduler/docs/scheduler-rest-api.asc
```diff
@@ -113,12 +114,23 @@ The request body is list of job maps with each entry containing a map of the fol
...
 ----
 
+===== Constraints
+
+Constraints provide controls over where a job is placed on the cluster. 
+There are both job and group level constraints.
+This will discuss job constraints, see the docs/groups.md for more details on group constraints.
+
+Constraints are specified as a tuple of attribute, operator, pattern.
+Attribute can be any attribute set on a host in the cluster.
```
shamsimam: Is there a means to query cook to find out available attributes on agents?

scheduler/src/cook/mesos/api.clj
```diff
@@ -28,7 +28,7 @@
             [compojure.core :refer (ANY GET POST routes)]
             [cook.mesos.quota :as quota]
             [cook.mesos.reason :as reason]
-            [cook.mesos.schema :refer (host-placement-types straggler-handling-types)]
+            [cook.mesos.schema :refer (host-placement-types straggler-handling-types constraint-operators)]
```
shamsimam: Nitpick: lexicographic ordering :(

scheduler/src/cook/mesos/api.clj
```diff
@@ -450,6 +461,16 @@
                              :label/key k
                              :label/value v}]))
                        labels)
+        constraints (mapcat (fn [[attribute operator pattern]]
+                              (let [constraint-var-id (d/tempid :db.part/user)]
+                                [[:db/add db-id :job/constraint constraint-var-id]
+                                 (merge {:db/id constraint-var-id
```
shamsimam: Can we use `cond->` and `assoc` instead of `merge`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -609,6 +632,9 @@
                                uris)})
                  (when labels
                    {:labels (walk/stringify-keys labels)})
+                 (when constraints
```
shamsimam: Nitpick: another good spot for `cond->` amd `assoc`

scheduler/src/cook/mesos/api.clj
```diff
@@ -740,6 +766,13 @@
                   "success" "failed")
                 :job.state/running "running"
                 :job.state/waiting "waiting")
+        constraints (->> job
+                         :job/constraint
+                         (map util/remove-datomic-namespacing)
+                         (map (fn [{:keys [attribute operator pattern]}]
+                                (->> [attribute (clojure.string/upper-case (name (util/without-ns operator))) pattern]
+                                     (remove nil?)
```
shamsimam: Why can we have `nil` values in the constraint? Doesn't the schema protect against that?

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -116,7 +117,37 @@
...
+                                      pattern :constraint/pattern
+                                      operator :constraint/operator}]
+            (let [vm-attribute-value (get vm-attributes attribute)]
+              (condp = operator
+                :constraint.operator/equals (= pattern vm-attribute-value)
+                :else (do
+                        (log/error (str "Unknown operator " operator
+                                        " api.clj should have prevented this from happening."))
+                        true))))
+          constraint->passes? (pc/map-from-keys vm-passes-constraint? constraints)
```
shamsimam: Why are we building a map here? Couldn't we directly do: `(every? vm-passes-constraint? constraints)`?

scheduler/test/cook/test/mesos/constraints.clj
```diff
@@ -33,6 +33,15 @@
   (is (= "attribute-equals-host-placement-group-constraint"
          (constraints/group-constraint-name (constraints/->attribute-equals-host-placement-group-constraint nil)))))
 
+(deftest test-user-defined-constraint
+  (let [constraints [{:constraint/attribute "is_spot"
```
shamsimam: We should also add test assertions for multiple constraints scenarios.

scheduler/src/cook/mesos/api.clj
```diff
@@ -450,6 +461,16 @@
                              :label/key k
                              :label/value v}]))
                        labels)
+        constraints (mapcat (fn [[attribute operator pattern]]
+                              (let [constraint-var-id (d/tempid :db.part/user)]
+                                [[:db/add db-id :job/constraint constraint-var-id]
+                                 (merge {:db/id constraint-var-id
+                                         :constraint/attribute attribute
+                                         :constraint/operator (keyword "constraint.operator"
+                                                                       (clojure.string/lower-case operator))}
```
shamsimam: Nitpick: can we require `clojure.string :as str`

scheduler/src/cook/mesos/api.clj
```diff
@@ -740,6 +766,13 @@
                   "success" "failed")
                 :job.state/running "running"
                 :job.state/waiting "waiting")
+        constraints (->> job
+                         :job/constraint
+                         (map util/remove-datomic-namespacing)
+                         (map (fn [{:keys [attribute operator pattern]}]
+                                (->> [attribute (clojure.string/upper-case (name (util/without-ns operator))) pattern]
```
shamsimam: Nitpick: can we require `clojure.string :as str`

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -232,7 +232,8 @@
                              :normal (dru/sorted-task-scored-task-pairs user->dru-divisors user->sorted-running-task-ents)
                              :gpu (dru/sorted-task-cumulative-gpu-score-pairs user->dru-divisors user->sorted-running-task-ents))
          task->scored-task (into (pm/priority-map-keyfn (case category
-                                                          :normal (comp - :dru)
```
shamsimam: Why did this change?

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -116,7 +117,37 @@
   (let [needs-gpus? (job-needs-gpus? job)]
     (->gpu-host-constraint job needs-gpus?)))
 
-(def job-constraint-constructors [build-novel-host-constraint build-gpu-host-constraint])
+(defrecord user-defined-constraint [constraints]
+  JobConstraint
+  (job-constraint-name [this] (get-class-name this))
+  (job-constraint-evaluate
```
shamsimam: Do we also need to implement the overloaded variant?

scheduler/src/cook/mesos/api.clj
wyegelwel: The pattern is optional in the schema. I can make it required for now since the only operator needs it. Would you prefer that?

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Good question. It makes the test deterministic.

scheduler/src/cook/mesos/api.clj
shamsimam: Do you have example constraints where the pattern may be optional. If so, then okay, else I would prefer it to be required.

wyegelwel: I was using the same set up as marathon's constraints, https://mesosphere.github.io/marathon/docs/constraints.html which have an optional pattern, but looking more closely they blend job and group constraints in a way that I'm not sure we want. I will make it required.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -232,7 +232,8 @@
...
+                                                          :normal (juxt (comp - :dru)
+                                                                        (comp util/task-ent->user :task))
                                                           :gpu (fnil - 0)))
```
shamsimam: Does the function for `:gpu` now need to return a vector?

wyegelwel: It isn't necessary, gpu will just not have a tie break when drus are equal. I only added it for normal because there are some tests that would fail when you changed unrelated source.

scheduler/src/cook/mesos/constraints.clj
wyegelwel: I think I may misunderstand the question. I *think* we do.

scheduler/src/cook/mesos/constraints.clj
wyegelwel: Good point, I was going to say which constraints failed but decided against it because it causes a lot of memory allocs.

scheduler/src/cook/mesos/constraints.clj
shamsimam: There is an overloaded version: `[this target-vm-resources target-vm-attrs target-vm-tasks-assigned]` which we are not implementing here. I'm not sure where the overloaded version is implemented?

scheduler/src/cook/mesos/api.clj
wyegelwel: Would rather do that in a separate pr, there are a lot of these conditions that would change

scheduler/docs/scheduler-rest-api.asc
wyegelwel: At the moment, no. That is a good thing to expose. Adding issue for it: #370

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -17,6 +17,7 @@
   (:require [clojure.tools.logging :as log]
             [cook.mesos.group :as group]
             [cook.mesos.util :as util]
+            [plumbing.core :as pc]
```
shamsimam: This import is unused now.

scheduler/src/cook/mesos/constraints.clj
shamsimam: Saw it in line 130, my bad.

wyegelwel:
Ok, removed the import and used mesos state.json to get the list of hosts and submitted a job constrained to each host.

integration/tests/cook/test_basic.py
```diff
@@ -308,24 +309,26 @@ def test_error_while_creating_job(self):
...
-        self.assertEqual(hostname, hostname_constrained)
-        self.assertEqual(constraints, job['constraints'])
+        host_to_job_uuid = {}
+        for hostname in hosts:
+            constraints = [["HOSTNAME", "EQUALS", hostname]]
+            job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints)
+            self.assertEqual(resp.status_code, 201, resp.text)
+            host_to_job_uuid[hostname] = job_uuid
+
+        for hostname, job_uuid in host_to_job_uuid.items():
```
shamsimam: +1

wyegelwel:
Any more feedback on this or can we merge?

shamsimam:
I approved. I saw @sradack as another reviewer, so didn't merge.

travis/minimesosFile
```diff
@@ -2,7 +2,7 @@ minimesos {
     clusterName = "Cook Simulation Cluster"
     loggingLevel = "INFO"
     mapAgentSandboxVolume = true
-    mapPortsToHost = false
+    mapPortsToHost = true
```
sradack: Why is this change being made on this ?

scheduler/src/cook/mesos/api.clj
```diff
@@ -194,6 +194,15 @@
   {:name (s/constrained s/Str non-empty-max-128-characters-and-alphanum?)
    :version (s/constrained s/Str non-empty-max-128-characters-and-alphanum?)})
 
+(def Constraint
+  "Schema for user defined job host constraint"
+  [(s/one NonEmptyString "attribute")
+   (s/one (s/pred #(contains? (set (map (comp name util/without-ns) constraint-operators))
```
sradack: Can `(comp name util/without-ns)` be simply `name`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -740,6 +766,12 @@
                   "success" "failed")
                 :job.state/running "running"
                 :job.state/waiting "waiting")
+        constraints (->> job
+                         :job/constraint
+                         (map util/remove-datomic-namespacing)
+                         (map (fn [{:keys [attribute operator pattern]}]
+                                (->> [attribute (str/upper-case (name (util/without-ns operator))) pattern]
```
sradack: Can `(name (util/without-ns operator))` be simply `(name operator)`?

travis/minimesosFile
sradack: Shams mentioned this is for the integration test.

wyegelwel: I wasn't sure how to find the IP for zookeeper otherwise. Also, it mirrors the dev set up so it is less likely to cause problems where in dev tests pass but in integration they don't

sradack:
Additionally there are some comments from @shamsimam that have gone without a response.

scheduler/src/cook/mesos/api.clj
wyegelwel: Oh wow, yes. I didn't know name would strip the namespace as well. Will change now

scheduler/test/cook/test/mesos/constraints.clj
wyegelwel: Done

scheduler/src/cook/mesos/api.clj
wyegelwel: Since pattern is no longer optional, just going to add :pattern into the map

scheduler/src/cook/mesos/api.clj
wyegelwel: Looking at the implementation of `without-ns` I clearly knew that at some point...

wyegelwel:
Ok, responded to comments. Thanks for pointing out I missed one of Shams!

wyegelwel:
Let me know if there is more feedback. Otherwise, may we merge?

PR #367, opened 2017-06-06
Makes framework id configurable

dposada:


scheduler/src/cook/components.clj
```diff
@@ -231,10 +239,12 @@
...
-     :framework-id (fnk [curator-framework [:settings mesos-leader-path]]
-                     (when-let [bytes (curator/get-or-nil curator-framework
-                                                          (str mesos-leader-path "/framework-id"))]
-                       (String. bytes)))
+     :framework-id-promise (fnk [curator-framework [:settings mesos-leader-path]]
+                             (let [fid-promise (promise)]
+                               (when-let [bytes (curator/get-or-nil curator-framework
+                                                                    (str mesos-leader-path "/framework-id"))]
+                                 (deliver fid-promise (String. bytes)))
+                               fid-promise))
```
shamsimam: Can we instead perform the blocking read on the promise here? That would avoid a lot of the downstream changes.

dposada: Done

scheduler/src/cook/components.clj
```diff
@@ -82,7 +82,14 @@
                    (route/not-found "<h1>Not a valid route</h1>")))})
 
 (def mesos-scheduler
-  {:mesos-scheduler (fnk [[:settings mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout mesos-principal mesos-role mesos-framework-name offer-incubate-time-ms mea-culpa-failure-limit fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset fenzo-fitness-calculator task-constraints riemann mesos-gpu-enabled rebalancer good-enough-fitness] mesos-datomic mesos-datomic-mult curator-framework mesos-pending-jobs-atom mesos-offer-cache]
+  {:mesos-scheduler (fnk [[:settings mesos-master mesos-master-hosts mesos-leader-path mesos-failover-timeout
```
shamsimam: nitpick: can we sort this list?

scheduler/src/cook/mesos.clj
```diff
@@ -214,10 +217,12 @@
                                         (sched/create-datomic-scheduler
                                           mesos-datomic-conn
                                           (fn set-or-create-framework-id [framework-id]
-                                            (curator/set-or-create
-                                              curator-framework
-                                              zk-framework-id
-                                              (.getBytes (-> framework-id mesomatic.types/pb->data :value) "UTF-8")))
+                                            (let [value (-> framework-id mesomatic.types/pb->data :value)]
```
shamsimam: Should this be guarded by a condition that checks that `framework-id-promise` has not been realized and also have an assertion/log if it has been delivered that the `framework-id` is the same value?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1486,7 +1486,7 @@
 (timers/deftimer [cook-scheduler handler list-endpoint])
 
 (defn list-resource
-  [db framework-id is-authorized-fn]
+  [db fid is-authorized-fn]
```
shamsimam: nitpick: I like `framework-id` to `fid`

scheduler/src/cook/mesos.clj
dposada: I added an error log if it's being set to a different value than what `framework-id-promise` already has

scheduler/src/cook/mesos/api.clj
dposada: I created issue #369 to track this

scheduler/src/cook/components.clj
dposada: Done

dposada:
@shamsimam This is now passing on travis and ready for another review

wyegelwel:
This will be a breaking change. When a user upgrades and they don't configure the framework, Cook will connect to mesos with a different framework id and mark all jobs that were running failed because mesos won't have tasks for those jobs under the new framework id. 

Is it necessary to remove the zookeeper framework id store?

dposada:
@wyegelwel True. @shamsimam suggested defaulting the framework id (if not configured) to the value from ZK so that this wouldn't be breaking.

scheduler/src/cook/components.clj
```diff
@@ -399,6 +416,8 @@
                    role)
      :mesos-framework-name (fnk [[:config [:mesos {framework-name "Cook"}]]]
                              framework-name)
+     :mesos-framework-id (fnk [[:config [:mesos {framework-id nil}]]]
```
shamsimam: We do not need this to be separate.

dposada: @shamsimam I don't understand what you mean

shamsimam: Why do we need both `:mesos-framework-id` and `:framework-id`?

dposada: `:mesos-framework-id` resides in the `config-settings` component and represents the value that cook was configured with. `:framework-id` resides in `scheduler-server`, it consumes `:mesos-framework-id`, and it handles the defaulting to the ZK value if nothing was configured.

integration/travis/run_integration.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ev
```
sradack: Why remove `set -e`?

integration/travis/wait-for-cook.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+# Usage: wait-for-cook.sh [PORT]
+#
+# Examples:
+#   timeout 60s wait-for-cook.sh 12321
+#   timeout 60s wait-for-cook.sh
+#
+# Waits for cook to be listening on the provided port (defaults to 12321).
```
sradack: Why create a script instead of a function?  Also, `curl` supports retries.  See `--retry` and `--retry-delay` and `--retry-max-time`.  It seems like we could avoid this entire script and move the curl calls back into `run_integration.sh`.

scheduler/bin/run-docker.sh
```diff
@@ -33,6 +33,7 @@ docker run \
     --publish=${COOK_PORT}:${COOK_PORT} \
     -e "COOK_PORT=${COOK_PORT}" \
     -e "COOK_NREPL_PORT=${COOK_NREPL_PORT}" \
+    -e "COOK_FRAMEWORK_ID=cook-framework-${COOK_PORT}" \
```
sradack: Isn't it Mesos's responsibility to generate the `framework-id`?

integration/travis/run_integration.sh
dposada: If the `timeout` fails, I want to `cat` the cook log

integration/travis/wait-for-cook.sh
dposada: Making it a function sounds good. As far as I can tell, `curl` doesn't support retry on connection refused.

scheduler/bin/run-docker.sh
dposada: Mesos will generate one if you don't provide one. Alternatively, you can provide one when you create your Mesos driver.

integration/travis/wait-for-cook.sh
dposada: Made it a function instead of a script.

dposada:

dposada:
Any additional feedback on this?

dposada:
Can we merge this?

dposada:
@sradack @shamsimam @pschorf Can we merge?

dposada:
@sradack Thoughts / feedback / merge?

integration/travis/run_integration.sh
sradack: It would be better to `timeout ... || curl_error=true` and then handle that specific error instead of turning of `set -e`.  `set -e` helps avoid a lot of issues with bash scripts IMO.

scheduler/bin/run-docker.sh
sradack: I don't follow why we should ever want to set our own `framework-id` instead of letting Mesos set it for us.

dposada: Quoting @wyegelwel from above:

> When a user upgrades and they don't configure the framework, Cook will connect to mesos with a different framework id and mark all jobs that were running failed because mesos won't have tasks for those jobs under the new framework id

So we want to use the same framework-id across restarts

wyegelwel: Reading that now, I realize there are two interpretations.

What I meant is that if we make the framework id configurable and don't read the framework id from zookeeper, then users that upgrade will get bit if they don't know to configure the framework id.

dposada: @wyegelwel I added back in the read from zookeeper such that if it's not configured, then it will attempt to read from zookeeper and use that one.

scheduler/src/cook/components.clj
sradack: I think what @shamsimam is saying, and I agree, is that there isn't a reason to define yet another key related to the framework id when we already have `:framework-id`, inside which we can depend on `:config :mesos :framework-id`.

scheduler/bin/run-docker.sh
sradack: Since we do read from ZK unless framework id is configured, upgrading users should be fine.

scheduler/src/cook/components.clj
```diff
@@ -230,11 +246,12 @@
...
                                    :request-header-size 32768})]
                       (fn [] (.stop jetty))))
-
-     :framework-id (fnk [curator-framework [:settings mesos-leader-path]]
-                     (when-let [bytes (curator/get-or-nil curator-framework
-                                                          (str mesos-leader-path "/framework-id"))]
-                       (String. bytes)))
+     :framework-id (fnk [curator-framework [:settings mesos-leader-path mesos-framework-id]]
+                     (let [framework-id (or mesos-framework-id
+                                            (framework-id-from-zk curator-framework mesos-leader-path)
```
sradack: Please add comment as to why we even bother reading from ZK.

scheduler/src/cook/components.clj
dposada: @sradack @shamsimam The reason to define two keys is that the cook startup code has 2 phases - in the first phase it translates the "literal" config to the config settings used downstream, and in the second phase, it uses these translated settings to actually bring up the components. `:mesos-framework-id` represents the value from phase 1, and `:framework-id` represents the value used in phase 2.

integration/travis/run_integration.sh
dposada: Done

scheduler/src/cook/components.clj
dposada: Done

dposada:
Feedback addressed, ready for another pass

dposada:
Any additional feedback, folks?

integration/tests/cook/test_basic.py
```diff
@@ -24,15 +25,33 @@ def setUp(self):
         util.wait_for_cook(self.cook_url)
 
     def get_job(self, job_uuid):
```
shamsimam: Do we need this function anymore? Maybe we should call `util.get_job` directly?

integration/tests/cook/test_basic.py
```diff
@@ -24,15 +25,33 @@ def setUp(self):
...
-        """Loads a job by UUID using GET /rawscheduler"""
-        return util.query_jobs(self.cook_url, job=[job_uuid]).json()[0]
+        """Loads a job by UUID"""
+        return util.get_job(self.cook_url, job_uuid)
+
+    @retry(stop_max_delay=120000, wait_fixed=5000)
+    def get_output_url(self, job_uuid):
```
shamsimam: Should this belong in `util`?

integration/tests/cook/test_basic.py
```diff
@@ -24,15 +25,33 @@ def setUp(self):
...
+        directories are cached in Cook.
+        """
+        job = self.get_job(job_uuid)
+        instance = job['instances'][0]
+        if 'output_url' in instance:
+            return instance['output_url']
+        else:
+            error_msg = 'Job %s had no output_url' % job['uuid']
+            self.logger.info(error_msg)
+            raise RuntimeError(error_msg)
```
shamsimam: Generic comment: should we use a version of `retry` that retries when `None` (e.g. `@retry(retry_on_result=retry_if_result_none)`) is the result instead of generating and throwing exceptions?

integration/tests/cook/test_basic.py
dposada: The advantage of keeping this function is that each test doesn't have to pass `self.cook_url` as the first argument. I think it's worth it.

integration/tests/cook/test_basic.py
dposada: I don't feel strongly one way or the other

integration/tests/cook/test_basic.py
dposada: Done

integration/tests/cook/test_basic.py
shamsimam: Is this not true for all the `util` functions that take a `cook_url` parameter?

integration/tests/cook/test_basic.py
shamsimam: Okay.

integration/tests/cook/test_basic.py
dposada: Fair enough, removed.

dposada:
Can we merge this?

PR #371, opened 2017-06-08
Make the executor sandbox directory cache configurable

shamsimam:


shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/241056977

scheduler/src/cook/components.clj
```diff
@@ -436,7 +434,20 @@
                      (when enabled?
                        (when (zero? port)
                          (throw (ex-info "You enabled nrepl but didn't configure a port. Please configure a port in your config file." {})))
-                       ((lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))}))
+                       ((lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))
+     :retrieve-url-path-fn (fnk [[:config {agent-query-cache {:threshold 1000, :ttl 6000}}]]
+                             (let [{:keys [threshold ttl]} agent-query-cache
+                                   _ (log/info "Agent query cache will be created using:" {:threshold threshold, :ttl ttl})
```
dposada: Instead of `{:threshold threshold, :ttl ttl}`, can't we just use `agent-query-cache` in the log statement?

scheduler/test/cook/test/testutil.clj
```diff
@@ -28,14 +28,21 @@
 (defn run-test-server-in-thread
   "Runs a minimal cook scheduler server for testing inside a thread. Note that it is not properly kerberized."
   [conn port]
+  (cook.components/init-logger {:file "log/test-server.log"
```
dposada: Is there a more central place we could put this? In other words, instead of just adding logging support to this spot, can we add it for any/all unit tests?

scheduler/src/cook/components.clj
```diff
@@ -436,7 +434,20 @@
                      (when enabled?
                        (when (zero? port)
                          (throw (ex-info "You enabled nrepl but didn't configure a port. Please configure a port in your config file." {})))
-                       ((lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))}))
+                       ((lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))
+     :retrieve-url-path-fn (fnk [[:config {agent-query-cache {:threshold 1000, :ttl 6000}}]]
```
dposada: `config-settings` doesn't seem like the right place for this function -- shouldn't `config-settings` simply propagate the configuration, and this function move to another component?

dposada: Should we add documentation of this new configurability? Also, should we add an entry in the changelog?

scheduler/src/cook/components.clj
shamsimam: We are ignoring any other config present in the `agent-query-cache`, so in the presence of other fields that log statement would be misleading.

scheduler/src/cook/components.clj
shamsimam: I was going off where `is-authorized-fn` is located. Do you have a suggestion of where else to put it?

scheduler/test/cook/test/testutil.clj
shamsimam: Created a new issue for it: https://github.com/twosigma/Cook/issues/373

scheduler/src/cook/components.clj
shamsimam: Updated the configuration doc and the changelog.

dposada: It seems like `scheduler-server` is (somewhat) more appropriate

scheduler/docs/configuration.asc
```diff
@@ -77,6 +77,12 @@ If need it to bind to another port, you can specify that with the `:local-port`
 `:rate-limit`::
   Configure rate limits for the scheduler. `rate-limit` is a map with two possible keys, `user-limit-per-m`. `user-limit-per-m` is the max requests a single user can send in a minute. The default is 600 requests. 
 
+`:agent-query-cache`::
+  Configure the cache used to store sandbox locations of tasks on different mesos agents.
+  `agent-query-cache` is a map with two possible keys, `threshold` and `ttl`.
+  `threshold` is the maximum number of elements in the cache before the FIFO eviction semantics apply. The default is 1000.
+  `ttl` is the default time in milliseconds that entries are allowed to reside in the cache.. The default is 60000, i.e. 1 minute.
```
dposada: Extra `.`

scheduler/src/cook/components.clj
dposada: OK

scheduler/docs/configuration.asc
shamsimam: Removed.

scheduler/src/cook/components.clj
shamsimam: I moved `retrieve-url-path-fn` to `raw-scheduler-routes`, it is now closest to its use.

scheduler/test/cook/test/testutil.clj
shamsimam: Removed the logging config from `run-test-server-in-thread`.

shamsimam:
@wyegelwel do you have any feedback on this?

shamsimam:
Rebased, this branch is ready for another round of review :)

Passing build: https://travis-ci.org/twosigma/Cook/builds/246065501

scheduler/src/cook/components.clj
```diff
@@ -66,11 +64,25 @@
     (resolve var-sym)))
 
 (def raw-scheduler-routes
-  {:scheduler (fnk [mesos-datomic framework-id mesos-pending-jobs-atom settings]
+  {:retrieve-url-path-fn (fnk [[:settings agent-query-cache]]
+                           (log/info "Agent query cache will be created using:" agent-query-cache)
+                           (let [{:keys [threshold ttl]} agent-query-cache
+                                 cache (-> {}
+                                           (cache/fifo-cache-factory :threshold threshold)
```
pschorf: Isn't fifo cache the one with the concat bug? I know it's already in there but might be worth thinking about before we marry ourselves to a config setting.

scheduler/src/cook/mesos/api.clj
```diff
@@ -754,10 +748,10 @@
 
 (defn- instance->instance-map
   "Converts the instance to a map of relevant fields that will be returned to the caller"
-  [db fid instance]
+  [db framework-id retrieve-url-path-fn instance]
```
pschorf: Is there a way to do this without having to pass around `retrieve-url-path-fn`? It can be annoying to navigate around code when you have to keep going back up to components to find the implementation.

scheduler/src/cook/components.clj
shamsimam: I'll switch to `LruCache`, avoids the `concat` hassle.

scheduler/src/cook/mesos/api.clj
shamsimam: We could pass the `cache` reference instead and then invoke `retrieve-url-path` directly. Is that preferred?

dposada:
Needs rebase

shamsimam:
Rebased. Passing build: https://travis-ci.org/twosigma/Cook/builds/248643799

scheduler/src/cook/components.clj
dposada: Are we still planning on making this change?

scheduler/src/cook/mesos/api.clj
dposada: @shamsimam For me, your proposal is preferred

shamsimam: Done, passing around the cache.

shamsimam:
ready for another round, passing build: https://travis-ci.org/twosigma/Cook/builds/249110606

scheduler/docs/configuration.asc
```diff
@@ -77,6 +77,12 @@ If need it to bind to another port, you can specify that with the `:local-port`
 `:rate-limit`::
   Configure rate limits for the scheduler. `rate-limit` is a map with two possible keys, `user-limit-per-m`. `user-limit-per-m` is the max requests a single user can send in a minute. The default is 600 requests. 
 
+`:agent-query-cache`::
+  Configure the cache used to store sandbox locations of tasks on different mesos agents.
+  `agent-query-cache` is a map with two possible keys, `threshold` and `ttl`.
+  `max-size` is the maximum number of elements in the cache before the FIFO eviction semantics apply. The default is 1000.
```
pschorf: fifo->lru

shamsimam: Done.

scheduler/src/cook/components.clj
dposada: Looks like this change was made

shamsimam: Yup, sorry for not responding earlier.

scheduler/src/cook/components.clj
```diff
@@ -325,7 +328,12 @@
 (def config-settings
   "Parses the settings out of a config file"
   (graph/eager-compile
-    {:server-port (fnk [[:config port]]
+    {:agent-query-cache (fnk [[:config {agent-query-cache nil}]]
+                          (merge
+                            {:max-size 1000
```
wyegelwel: Let's make the default larger, maybe 5000 so that most users won't need to set it.

scheduler/test/cook/test/testutil.clj
```diff
@@ -32,13 +32,15 @@
   "Runs a minimal cook scheduler server for testing inside a thread. Note that it is not properly kerberized."
   [conn port]
   (let [authorized-fn (fn [x y z] true)
+        retrieve-url-path-fn (fn [fid hostname executor-id] (str "http://" hostname "/" fid "/" executor-id))
```
wyegelwel: Is this used?

wyegelwel:
Two small comments, otherwise looks fine. 

Larger comment that doesn't need to be resolved in this PR is that adding the cache resulted in a bunch of lines changing just passing the cache through. Can we rework the structure or the way we handle dependencies to remove this need?

scheduler/test/cook/test/testutil.clj
shamsimam: It was not being used, removed.

scheduler/src/cook/components.clj
shamsimam: Done.

shamsimam:
@wyegelwel 

> Can we rework the structure or the way we handle dependencies to remove this need?

Is the plan to define the cache in an atom and then have it initialized by the dependency mechanism?

shamsimam:
Addressed feedback, ready for another round. Passing build: https://travis-ci.org/twosigma/Cook/builds/252259247

pschorf:
I agree that this PR isn't the right place to change it, but I feel the same way about waiter. Maybe there is a better solution in clojure but this really smells like a dependency injection thing to me. I think a larger project would be to try to factor things into logical components that you could wire together via the fnk structure, which would probably also help to address the unit test churn we've talked about a few times,

PR #374, opened 2017-06-09
Update the version strings in the sub-projects

shamsimam:


sradack:
Are we going to do this for every release?

shamsimam:
@sradack that is the intent. Henceforth, we can do this when we prepare the changelog.

pschorf:
I just opened a PR for 1.3.3, I can just do this in that PR

shamsimam:
@pschorf  +1

shamsimam:
You will need to update the versions to 1.3.4* in the files after the 1.3.3 release

PR #376, opened 2017-06-12
Adds instance progress to query response

dposada:


scheduler/src/cook/mesos/api.clj
```diff
@@ -748,7 +749,8 @@
                                mesos-start (:instance/mesos-start-time instance)
                                end (:instance/end-time instance)
                                cancelled (:instance/cancelled instance)
-                               reason (reason/instance-entity->reason-entity db instance)]
+                               reason (reason/instance-entity->reason-entity db instance)
```
shamsimam: Should we extract `fetch-instance-map` from `fetch-job-map`? (https://github.com/twosigma/Cook/pull/308/files#diff-6611b39b65ff49438a4d940b51c85cc0R733) That will simplify the testing.

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1216,4 +1220,43 @@
...
+        fenzo (sched/make-fenzo-scheduler driver-atom 1500 nil 0.8)
+        make-status-update (fn [task-id reason state progress]
+                             (let [task {:task-id {:value task-id}
+                                         :reason reason
+                                         :state state
+                                         :data (ByteString/copyFrom (.getBytes (pr-str {:percent progress}) "UTF-8"))}]
+                               task))
+        job-id (create-dummy-job conn :user "user" :job-state :job.state/running)
+        send-status-update #(async/<!!
+                              (sched/handle-status-update conn driver fenzo
```
shamsimam: In the cook scheduler, the progress will be sent as framework messages. We may need to edit this test for consistency while merging the cook scheduler branch.

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1216,4 +1220,43 @@
                (api/retrieve-url-path target-framework-id agent-hostname "executor-102")))
         (is (nil? (api/retrieve-url-path target-framework-id agent-hostname "executor-103")))
         (is (= (str "http://" agent-hostname ":5051/files/read.json?path=%2Fpath%2Ffor%2Fexecutor-111")
-               (api/retrieve-url-path target-framework-id agent-hostname "executor-111")))))))
\ No newline at end of file
+               (api/retrieve-url-path target-framework-id agent-hostname "executor-111")))))))
+
+(deftest test-instance-progress
```
shamsimam: I feel this test is more for the scheduler (as `handle-status-update` is used) than the fetch job map. Should it be in the scheduler test file?

scheduler/test/cook/test/mesos/api.clj
dposada: You mean cook executor, right?

scheduler/test/cook/test/mesos/api.clj
dposada: I originally had it there, and then moved it here, since the essence of the test is verifying that progress can be retrieved via `fetch-job-map`. In other words, it's not really testing `handle-status-update`, it's just using it as a means to establish a progress value.

scheduler/test/cook/test/mesos/api.clj
shamsimam: Yes, fixed the typo in the previous comment.

scheduler/test/cook/test/mesos/api.clj
shamsimam: Can we then just write to the db directly instead of involving the `handle-status-update`?

scheduler/src/cook/mesos/api.clj
dposada: Done, extracted `instance->instance-map`

scheduler/test/cook/test/mesos/api.clj
dposada: We could, but it makes me have much more confidence in the test when the progress writing is happening in the way it happens in "real life".

shamsimam: How do you feel about splitting the test into two, one for the progress where we write to db directly and then one in the scheduler that verifies that invoking `handle-status-update` actually writes to the db?

dposada: To me, those two tests seem much less valuable than this one test. For example, if we change how we write to the db, only one of the two proposed tests will fail (the one in `scheduler`), while the other one will continue to pass because it's mocking a now-incorrect write to the database.

shamsimam: Both tests rely on reading/writing a common field in the db. I think that if we change how we write, then the unit test that does the assertions on the write should be the one to fail?

Perhaps we need an additional separate integration test for this whole scenario?

dposada: Discussed this in person a bit. @pschorf wants to chime in :)

shamsimam: As per our discussion, the agreed line to differentiate the tests seems to be whether we need to start cook or not. If so, I am okay with this test.

dposada:

PR #378, opened 2017-06-12
Fixes race condition in test-read-chan

dposada:
This test has a race condition, e.g.:

https://travis-ci.org/twosigma/Cook/jobs/242101623#L1147-L1149

I was able to reproduce the race locally. After this fix, I ran this in a `dotimes` 1000 with no fails.


dposada:
@wyegelwel Any feedback on this?

pschorf:
👍 , I've seen this locally a couple of times.

PR #380, opened 2017-06-13
Refactor fenzo constraints to use less memory

pschorf:


scheduler/src/cook/mesos/constraints.clj
```diff
@@ -160,10 +157,13 @@
       (let [vm-resources (.getCurrAvailableResources target-vm)
             vm-attributes (get-vm-lease-attr-map vm-resources)
             [passes? reason] (job-constraint-evaluate constraint vm-resources vm-attributes
-                                                      (into (vec (.getRunningTasks target-vm))
-                                                              (mapv (fn [^com.netflix.fenzo.TaskAssignmentResult result]
-                                                                      (.getRequest result))
-                                                                     (.getTasksCurrentlyAssigned target-vm))))]
+                                                      (loop [tasks-assigned (.getTasksCurrentlyAssigned target-vm)
```
wyegelwel: Could this be achieved with reduce? Could it benefit from transients (https://clojure.org/reference/transients)? 

Please add a comment why this is a loop so it isn't changed back some point later.

pschorf: Much better, thanks

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -160,10 +157,11 @@
       (let [vm-resources (.getCurrAvailableResources target-vm)
             vm-attributes (get-vm-lease-attr-map vm-resources)
             [passes? reason] (job-constraint-evaluate constraint vm-resources vm-attributes
-                                                      (into (vec (.getRunningTasks target-vm))
-                                                              (mapv (fn [^com.netflix.fenzo.TaskAssignmentResult result]
-                                                                      (.getRequest result))
-                                                                     (.getTasksCurrentlyAssigned target-vm))))]
+                                                      ; Use transients to improve speed and reduce allocations
```
wyegelwel: Consider this as a replacement to the current comment:

```
;; This function is called for each job considered, for each host available. It was optimized to reduce memory allocations because it accounted for ___% of all allocs 
```

I prefer comments to explain why something (weird) was done so new eyes (or forgetful eyes) don't need to wonder what compelled us to do the thing.

wyegelwel: Also, consider moving this expression into the let and giving the variable a name

pschorf: 👍

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -160,10 +157,12 @@
       (let [vm-resources (.getCurrAvailableResources target-vm)
             vm-attributes (get-vm-lease-attr-map vm-resources)
             [passes? reason] (job-constraint-evaluate constraint vm-resources vm-attributes
-                                                      (into (vec (.getRunningTasks target-vm))
-                                                              (mapv (fn [^com.netflix.fenzo.TaskAssignmentResult result]
-                                                                      (.getRequest result))
-                                                                     (.getTasksCurrentlyAssigned target-vm))))]
+                                                      ;; This function is called for each job considered, for each host available.
+                                                      ;; It was optimized to reduce memory allocations because it accounted for ~10% of all allocs
+                                                      (persistent! (reduce (fn [tasks ^com.netflix.fenzo.TaskAssignmentResult result]
```
sradack: Do we even need to build a vector at all?

pschorf: This was another improvement. 

@wyegelwel what do you think? I know concat has issues but it seems safe in this case, how many tasks do you think could be assigned to a host?

sradack:
LGTM

scheduler/src/cook/mesos/constraints.clj
wyegelwel: I think it is fine, concat is only dangerous if there is a chain of concats, i.e. `(concat (concat (concat a b) c) d)`

wyegelwel:
Going to wait for travis to complete but looks good

PR #384, opened 2017-06-15
Load mesos master URL from cook and use docker for integration tests

pschorf:


integration/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python
```
shamsimam: Can we use `FROM python:3.5`?

integration/bin/build-docker-image.sh
```diff
@@ -0,0 +1,8 @@
+
```
shamsimam: Please remove the black line.

integration/bin/build-docker-image.sh
```diff
@@ -0,0 +1,8 @@
...
+#!/usr/bin/env bash
+
```
shamsimam: We should add a comment for usage.

integration/bin/run-integration.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
```
shamsimam: We should add a usage comment.

integration/tests/cook/util.py
```diff
@@ -51,8 +51,16 @@ def retrieve_cook_url(varname='COOK_SCHEDULER_URL', value='http://localhost:1232
...
+def retrieve_mesos_url(varname='MESOS_PORT', value='5050'):
+    mesos_port = os.getenv(varname, value)
+    cook_url = retrieve_cook_url()
+    wait_for_cook(cook_url)
+    settings = session.get('%s/settings' % cook_url).json()
+    mesos_master_hosts = settings.get('mesos-master-hosts', ['localhost'])
+    resp = session.get('http://%s:%s/redirect' % (mesos_master_hosts[0], mesos_port), allow_redirects=False)
+    if resp.status_code != 307:
+        raise RuntimeError('Unable to find mesos leader, redirect endpoint returned %d' % resp.status_code)
+    mesos_url = 'http:%s' % resp.headers['Location']
```
shamsimam: are we missing `//` after the `http:`?

integration/Dockerfile
pschorf: I don't think we care about the minor version, why not just take the latest?

integration/bin/build-docker-image.sh
pschorf: Done

integration/bin/build-docker-image.sh
pschorf: Done

integration/bin/run-integration.sh
pschorf: Done

integration/tests/cook/util.py
pschorf: the location header is formatted `//hostname:5050`

integration/tests/cook/util.py
```diff
@@ -51,8 +51,16 @@ def retrieve_cook_url(varname='COOK_SCHEDULER_URL', value='http://localhost:1232
     return cook_url
 
 
-def retrieve_mesos_url(varname='MESOS_URL', value='http://localhost:5050'):
-    mesos_url = os.getenv(varname, value)
+def retrieve_mesos_url(varname='MESOS_PORT', value='5050'):
+    mesos_port = os.getenv(varname, value)
+    cook_url = retrieve_cook_url()
+    wait_for_cook(cook_url)
+    settings = session.get('%s/settings' % cook_url).json()
```
dposada: There is a `settings` function in `test_basic.py`. Might be nice to move it into `util` and use it here.

pschorf: Done

pschorf:
Updated README

PR #385, opened 2017-06-16
Add check to ensure job/group uuids do not exist before creation

pschorf:


wyegelwel:
What error does this return if we catch the conflict at transaction time?

pschorf:
This throws an ExecutionException wrapping an IllegalStateException. That will cause liberator to return a 503

wyegelwel:
> This throws an ExecutionException wrapping an IllegalStateException. That will cause liberator to return a 503

I imagine the answer is no, liberator is too rigid, but is there a way to get it to return a 409 conflict like it would in the case that cook catches the conflict up front?

scheduler/src/cook/mesos/schema.clj
```diff
@@ -945,7 +945,17 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
                          prior-state (:job/state job)]
                      (if (= prior-state :job.state/completed)
                        [[:db/add j :job/state :job.state/waiting]]
-                       []))}}])
+                       []))}}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :ensure-not-exists
+    :db/doc "Ensure that the given entity does not exist"
+    :db/fn #db/fn {:lang "clojure"
+                   :params [db field id]
```
wyegelwel: Change this to take `db` and `id-or-lookup-ref` and then call `(d/entity db id-or-lookup-ref)` that way it is more generic.

scheduler/src/cook/mesos/schema.clj
```diff
@@ -945,7 +945,17 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
                          prior-state (:job/state job)]
                      (if (= prior-state :job.state/completed)
                        [[:db/add j :job/state :job.state/waiting]]
-                       []))}}])
+                       []))}}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :ensure-not-exists
```
wyegelwel: It is datomic best practices to namespace idents. Perhaps `:entity/ensure-not-exists`

scheduler/src/cook/mesos/schema.clj
pschorf: Done

scheduler/src/cook/mesos/schema.clj
pschorf: Done

pschorf:
I don't think there's a way to affect the response code from liberator. Once it decides to create the entity, the response code is always 201 unless you throw in which case it is 500 (not 503.)

PR #386, opened 2017-06-16
Improve the performance of get-jobs-by-user-and-state

wyegelwel:
By chunking by day, we prevent datomic from doing large joins which it apparently is not well optimized for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,12 @@
 All notable changes to this project will be documented in this file
  
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
- 
+
+## [unreleased] 
+### Changed
+- Optimized the query in the list endpoint to avoid an expensive datomic join, from @pschorf and @wyegelwel
+- Change the list endpoint time range to be inclusive on start, from @wyegelwel
+
 ## [1.4.0] - 2017-06-09
 ### Added
 - Added simulator to test scheduler performance, from @wyegelwel
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -16,6 +16,7 @@
 (ns cook.mesos.util
   (:require [clj-time.coerce :as tc]
             [clj-time.core :as t]
+            [clj-time.periodic :refer [periodic-seq]]
             [clojure.core.async :as async]
             [clojure.core.cache :as cache]
             [clojure.walk :as walk]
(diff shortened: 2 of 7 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -198,6 +198,39 @@
...
+      based on the step-fn between start and end
+   step-fn : fn
+      function that returns a clj-time/timedelta given n steps
+
+   Returns:
+   --------
+   list of pairs, [interval-start interval-end]"
+  ([start end]
+   (generate-intervals start end t/in-days t/days))
+  ([start end interval->steps step-fn]
```
pschorf: Do you need `interval->steps`? Can you just iterate `step-fn` until you walk past `end`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -209,16 +242,21 @@
          ;; the total number of jobs for a user should be smaller than
          ;; all completed jobs by any user i.e.
          ;; (waiting_user + running_user < completed - completed_user)
```
pschorf: Update the comment with the new logic?

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,6 +198,39 @@
...
+                  interval-end (t/plus start (step-fn (inc steps-from-start)))
+                  interval-end (if (t/before? interval-end end)
+                                 interval-end
+                                 end)]
+              [interval-start interval-end]))
+          ;; This filter is here because joda time t/days returns the floor of days if there is an
+          ;; a non integer number of days. 
+          (filter (fn [[s e]] (> (t/in-millis (t/interval s e)) 0)))))))
+
 (defn get-jobs-by-user-and-state
```
pschorf: What about passing an optional limit here to stop iterating once you've hit the limit?

wyegelwel: Yea was working on that now. Will push it this weekend or Monday

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,6 +198,39 @@
      get-pending-jobs-duration
      (get-pending-job-ents* unfiltered-db true))))
 
+(defn generate-intervals
```
pschorf: Can this be replaced with http://clj-time.github.io/clj-time/doc/clj-time.periodic.html#var-periodic-seq ?

scheduler/src/cook/mesos/util.clj
```diff
@@ -209,16 +242,21 @@
...
-            db user state start end)
+
+         (mapcat (fn query-in-chunks
+                   [interval]
+                   (let [[interval-start interval-end] (map tc/to-date interval)]
+                     (q '[:find [?j ...]
+                          :in $ ?user ?state ?start ?end
+                          :where
+                          [?j :job/submit-time ?t]
+                          [(< ?start ?t)]
```
sradack: `<` should be `<=`.

scheduler/src/cook/mesos/util.clj
```diff
@@ -209,16 +242,21 @@
...
+                          [(< ?t ?end)]
+                          [?j :job/user ?user]
+                          [?j :job/state ?state]
+                          [?j :job/custom-executor false]]
+                        db user state interval-start interval-end)))
+                 (generate-intervals (tc/from-date start) (tc/from-date end)))
```
shamsimam: Can we make the `step-fn` configurable? 1 day granularity feels too big to me, it'd be great if we can quickly change the config when we come across a user for whom a much lower granularity would do.

shamsimam: Should returning the appropriate date objects be the responsibility of `generate-intervals` so we can avoid the calls to `tc/from-date` and `tc/to-date`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,6 +198,39 @@
...
+
+   Returns:
+   --------
+   list of pairs, [interval-start interval-end]"
+  ([start end]
+   (generate-intervals start end t/in-days t/days))
+  ([start end interval->steps step-fn]
+   (if (t/before? end start)
+     []
+     (->> (for [steps-from-start (range 0 (inc (interval->steps (t/interval start end))))]
```
shamsimam: A `loop` feels a more natural choice to me here as we can quickly build off the next interval from the `end` of the last interval.

scheduler/src/cook/mesos/util.clj
wyegelwel: Yes! Thank you, I knew that function had to exist

scheduler/src/cook/mesos/util.clj
wyegelwel: Switched to using `periodic-seq`

scheduler/src/cook/mesos/util.clj
wyegelwel: done

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,38 +199,73 @@
...
-              [?j :job/state ?state]
-              [?j :job/user ?user]
-              [?j :job/submit-time ?t]
-              [(< ?start ?t)]
-              [(< ?t ?end)]
-              [?j :job/custom-executor false]]
-            db user state start end))
-       (map (partial d/entity db))))
+  [db user state start end limit]
+  (let [intervals (generate-intervals (tc/from-date start) (tc/from-date end) (t/days 1))
```
shamsimam: Can the `(t/days 1)` be made configurable?

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,38 +199,73 @@
...
+                             :where
+                             [?j :job/state ?state]
+                             [?j :job/user ?user]
+                             [?j :job/submit-time ?t]
+                             [(<= ?start ?t)]
+                             [(< ?t ?end)]
+                             [?j :job/custom-executor false]]
+                           db user state start end)])]
+    (->> job-batches
+         (map sort)
```
shamsimam: Why do the batches need to be sorted? Is this intended to be applied on the accumulated results instead?

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,38 +199,73 @@
...
+        (partition 2 1))))
+
+(defn accumulate-n-from-batches
+  "Given a seq of collections, `batches`, take `n` elements from the collections,
+   starting with the first and realizing subsequent batches once previous batches
+   are consumed."
+  [n batches]
+  (loop [coll []
+         [batch & batches] batches]
+    (let [needed-elements (- n (count coll))
```
shamsimam: `needed-elements` can become a loop variable and save the call on `(count coll)`.

scheduler/src/cook/mesos/util.clj
```diff
@@ -198,38 +199,73 @@
...
+   ----------
+   start : clj-time/datetime
+   end : clj-time/datetime
+   period-like : clj-time/period
+
+   Returns:
+   --------
+   list of pairs, [interval-start interval-end]"
+  ([start end period-like]
+   (->> (conj (vec (periodic-seq start end period-like)) end)
```
shamsimam: Should we `conj` only if the last element does not equal `end`?

scheduler/src/cook/mesos/util.clj
wyegelwel: There is no order given by datomic for queried items. This ensures that if you call the function twice with the same inputs you will get the same outputs. If you did it after the accumulate you wouldn't be sure to get the same results.

scheduler/src/cook/mesos/util.clj
wyegelwel: It never will, periodic-seq is exclusive on the right.

scheduler/src/cook/mesos/util.clj
wyegelwel: I'm planning to leave as is. I think it is marginally more readable as is and I don't think the count call will drastically affect performance. Is this ok with you?

sradack: Can we just use `(take n (concat batches))`?

wyegelwel: I'm concerned this will trigger the concat overflow bug because take is lazy

shamsimam: sgtm

pschorf:
LGTM

CHANGELOG.md
```diff
@@ -3,7 +3,12 @@
 All notable changes to this project will be documented in this file
  
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
- 
+
+## [unreleased] 
+### Changed
```
pschorf: 👏 👏 👏

shamsimam:
Approved, will wait for a green build.

wyegelwel:
Build is green. Ok to merge?

PR #392, opened 2017-06-20
Improve the performance of list endpoint

wyegelwel:
Switching to seek-datoms allows that function to be completely lazy, meaning we hold on to a lot less memory. Switching to seek-datoms also means we can seek to exactly the time point we are interested in and has shown query speed ups of >10x

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -228,7 +228,7 @@
 ;; This works because the submission time and job/user field
 ;; are set at the same time, in "real" time. This means that
 ;; jobs submitted after `start` will have been created after
-;; expanded starti
+;; expanded start
 (defn get-jobs-by-user-and-states
   "Returns all job entities for a particular user
    in a particular state, in the specified timeframe,
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -218,6 +218,17 @@
    (->> (conj (vec (periodic-seq start end period-like)) end)
         (partition 2 1))))
 
+;; get-jobs-by-user-and-states is a bit opaque because it is
+;; reaching into datomic internals. Here is a quick explanation.
+;; seek-datoms provides a pointer into the raw datomic indices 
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -217,42 +218,30 @@
...
-         (map sort)
-         (apply concat)
-         (take limit)
-         (map (partial d/entity db)))))
+  [db user states start end limit]
+  (let [states (set states)
+        ;; Expand the time range so that clock skew between cook
+        ;; and datomic doesn't cause us to miss jobs
+        ;; 1 hour was picked because a skew larger than that would be
+        ;; susipious
```
pschorf: suspicious

scheduler/src/cook/mesos/util.clj
```diff
@@ -217,42 +218,30 @@
...
+        expanded-start (Date. (- (.getTime start) 
+                                 (-> 1 t/hours t/in-millis)))
+        expanded-end (Date. (+ (.getTime end)
+                               (-> 1 t/hours t/in-millis)))] 
+    (->> (d/seek-datoms db :avet :job/user user (d/entid-at db :db.part/user expanded-start))
```
pschorf: can you add some comments for how/why this works?

scheduler/src/cook/mesos/util.clj
```diff
@@ -217,42 +218,41 @@
...
+;; reaching into datomic internals. Here is a quick explanation.
+;; seek-datoms provides a pointer into the raw datomic indices 
+;; that we can then seek through. We set the pointer to look
+;; through the avet index, with attribute :job/user, seek to 
+;; user and then seek to the entity id that *would* have been
+;; created at expanded start. 
+;; This works because the submission time and job/user field
+;; are set at the same time, in "real" time. This means that
+;; jobs submitted after `start` will have been created after
+;; expanded starti
```
dposada: Is `starti` a typo?

pschorf: https://github.com/twosigma/Cook/pull/393

PR #396, opened 2017-06-21
Return 503 from /debug endpoint when leader

pschorf:


scheduler/src/cook/mesos.clj
```diff
@@ -207,6 +207,7 @@
                           (reify LeaderSelectorListener
                             (takeLeadership [_ client]
                               (log/warn "Taking mesos leadership")
+                              (swap! mesos-leadership-atom (constantly true))
```
wyegelwel: I think you want just `true` here. I expect this is left over from a refactor where you changed mesos-leadership-atom to hold a value instead of a function

scheduler/src/cook/mesos.clj
```diff
@@ -272,6 +273,7 @@
                               ;; We will give up our leadership whenever it seems that we lost
                               ;; ZK connection
                               (when (#{ConnectionState/LOST ConnectionState/SUSPENDED} newState)
+                                (swap! mesos-leadership-atom (constantly false))
```
wyegelwel: From my comment above, you probably want this to just be `false`, otherwise this will evaluate to truthy

wyegelwel:
Please add a comment in the configuration doc about the new option

wyegelwel:
Bunch of small comments. Once they are fixed, lgtm.

scheduler/src/cook/mesos.clj
pschorf: This was because `swap!` takes a function from old value -> new value, but for this we don't really care about the old value. Maybe there's a better way to express this?

pschorf:
Added a section to the configuration docs.

scheduler/src/cook/mesos.clj
wyegelwel: You can use `reset!` which just sets the value

pschorf: duh, thanks.

scheduler/src/cook/mesos.clj
pschorf: changed to reset!

scheduler/docs/configuration.asc
```diff
@@ -124,6 +124,12 @@ We'll look at the configurable options in turn:
   If you enable this and your cluster doesn't have any GPU machines, Cook will accept GPU jobs, but they'll never be scheduled.
   See https://github.com/apache/mesos/blob/master/docs/gpu-support.md for details on configuring the agents, installing external NVidia dependencies, and configuring Docker/GPU integration.
 
+`:leader-reports-unhealthy`::
+    This configures whether or not the leader reports his status as healthy by returning 200 from the /debug endpoint. This can be used to isolate the leader from
```
shamsimam: Nitpick: newline at `This can be`

scheduler/docs/configuration.asc
```diff
@@ -124,6 +124,12 @@ We'll look at the configurable options in turn:
...
+    query load.
+    If set to true, the leader will return 503 on the /debug endpoint.
+    If set to false, the leader will return 200 on the /debug endpoint.
```
dposada: Can we document the default value as well?

scheduler/docs/configuration.asc
pschorf: Done

scheduler/docs/configuration.asc
```diff
@@ -124,6 +124,12 @@ We'll look at the configurable options in turn:
   If you enable this and your cluster doesn't have any GPU machines, Cook will accept GPU jobs, but they'll never be scheduled.
   See https://github.com/apache/mesos/blob/master/docs/gpu-support.md for details on configuring the agents, installing external NVidia dependencies, and configuring Docker/GPU integration.
 
+`:leader-reports-unhealthy`::
+    This configures whether or not the leader reports his status as healthy by returning 200 from the /debug endpoint.
+    This can be used to isolate the leader from query load.
+    If set to true, the leader will return 503 on the /debug endpoint.
+    If set to false, the leader will return 200 on the /debug endpoint.
```
dposada: Can we also document the default value?

scheduler/test/cook/test/components.clj
```diff
@@ -2,6 +2,35 @@
   (:require [clojure.test :refer :all]
             [cook.components :as components]))
 
+(deftest test-health-check-middleware
+  (let [handler (fn [_] "Called handler!")
+        debug-request {:uri "/debug"
+                       :request-method :get}]
+    (testing "always returns 200 if leader-reports-unhealthy is false"
```
dposada: Nitpick: I would favor a newline before each `testing`

scheduler/docs/configuration.asc
pschorf: done

scheduler/test/cook/test/components.clj
pschorf: Done

pschorf:
Can this be merged?

PR #397, opened 2017-06-21
Use separate query for completed and waiting/running jobs

pschorf:


scheduler/src/cook/mesos/util.clj
```diff
@@ -251,9 +249,41 @@
...
+
+(defn get-jobs-by-user-and-states
+  "Returns all jobs for a particular user in the specified states
+   and timeframe, without a custom executor."
+  [db user states start end limit]
+  (let [get-jobs-by-state (fn get-jobs-by-state [state]
+                            (if (= state :job.state/completed)
+                              (get-completed-jobs-by-user db user start end limit)
+                              (get-active-jobs-by-user-and-state db user start end state)))
+        jobs-by-state (map get-jobs-by-state states)]
```
wyegelwel: Use `mapcat` here so you avoid the `(apply concat ...` below

wyegelwel:
Small comment, otherwise lgtm.

scheduler/src/cook/mesos/util.clj
pschorf: Done

scheduler/src/cook/mesos/util.clj
```diff
@@ -229,13 +229,11 @@
...
-(defn get-jobs-by-user-and-states
-  "Returns all job entities for a particular user
-   in a particular state, in the specified timeframe,
-   without a custom executor."
-  [db user states start end limit]
-  (let [states (set states)
-        ;; Expand the time range so that clock skew between cook
+(defn get-completed-jobs-by-user
+  "Returns all completed job entities for a particular user
+   in the specified timeframe, without a custom executor."
```
wyegelwel: Oh shit, we aren't filtering out custom executor jobs. Can you add another filter for that?

pschorf: Good catch, fixed.

scheduler/src/cook/mesos/util.clj
```diff
@@ -229,13 +229,11 @@
...
 ;; jobs submitted after `start` will have been created after
 ;; expanded start
-(defn get-jobs-by-user-and-states
-  "Returns all job entities for a particular user
-   in a particular state, in the specified timeframe,
-   without a custom executor."
-  [db user states start end limit]
-  (let [states (set states)
-        ;; Expand the time range so that clock skew between cook
+(defn get-completed-jobs-by-user
```
shamsimam: Should we add metrics around these calls?

scheduler/src/cook/mesos/util.clj
dposada: Should we add a test for this fix?

scheduler/src/cook/mesos/util.clj
```diff
@@ -251,9 +249,42 @@
...
+       (map (partial d/entity db))))
+
+(defn get-jobs-by-user-and-states
+  "Returns all jobs for a particular user in the specified states
+   and timeframe, without a custom executor."
+  [db user states start end limit]
+  (let [get-jobs-by-state (fn get-jobs-by-state [state]
+                            (if (= state :job.state/completed)
+                              (get-completed-jobs-by-user db user start end limit)
+                              (get-active-jobs-by-user-and-state db user start end state)))
```
dposada: Why can't we use the completed approach for other statuses?

pschorf: The issue is that the time based scanning is generally worse for running jobs because they'll be at the end of the time range. You can do a different query with a similar principal to look for the first job in a given state, but that performed approximately the same as the simple datomic query for my tests (a few users with a ton of jobs and a few users with few jobs) so I don't think it's worth having the more complicated code.

pschorf:
Added a test

scheduler/src/cook/mesos/util.clj
pschorf: Added a test

scheduler/src/cook/mesos/util.clj
dposada: Ah, makes sense. Would be good to add this in a comment somewhere

scheduler/src/cook/mesos/util.clj
pschorf: Added timers

scheduler/src/cook/mesos/util.clj
pschorf: Added a comment

scheduler/src/cook/mesos/util.clj
```diff
@@ -229,31 +231,72 @@
...
+                            (= (.v %) user)))
+          (map #(.e %))
+          (map (partial d/entity db))
+          (filter #(<= (.getTime start) (.getTime (:job/submit-time %))))
+          (filter #(< (.getTime (:job/submit-time %)) (.getTime end)))
+          (filter #(= :job.state/completed (:job/state %)))
+          (filter #(not (:job/custom-executor %)))
+          (take limit)))))
+
+(timers/deftimer [cook-mesos scheduler get-active-jobs-by-user-duration])
```
dposada: Should we break this out further by state?

pschorf: Definitely.

PR #399, opened 2017-06-22
Parallelize in-order processing of status messages

shamsimam:


shamsimam:
Passing build: https://travis-ci.org/twosigma/Cook/builds/245665805

PR #405, opened 2017-06-23
Do not overwrite end-time on duplicate mesos message

### LINKED ISSUE #403: Don't overwrite end time if we receive duplicate status updates
We currently will overwrite the end time of job if we receive additional terminal status updates for a task. This means that we are reporting the last update received instead of the first which skews metrics.

The change is to check if the end time is set before setting it. https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/scheduler.clj#L263

pschorf:
Closes #403

### DIFF
--- a/scheduler/test/cook/test/mesos/scheduler.clj
+++ b/scheduler/test/cook/test/mesos/scheduler.clj
@@ -906,6 +906,7 @@
                                                [?i :instance/end-time ?end-time]]
                                              (db conn) instance-id)))
               original-end-time (get-end-time)]
+          (Thread/sleep 100)
           (async/<!! (sched/handle-status-update conn driver fenzo
                                                  (make-dummy-status-update task-id :reason-gc-error :task-killed)))
           (is (= original-end-time (get-end-time))))))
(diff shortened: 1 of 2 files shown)


scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -899,7 +899,16 @@
...
+                          (db conn) instance-id))))
+        (let [get-end-time (fn [] (ffirst (q '[:find ?end-time
+                                               :in $ ?i
+                                               :where
+                                               [?i :instance/end-time ?end-time]]
+                                             (db conn) instance-id)))
+              original-end-time (get-end-time)]
+          (async/<!! (sched/handle-status-update conn driver fenzo
+                                                 (make-dummy-status-update task-id :reason-gc-error :task-killed)))
+          (is (= original-end-time (get-end-time))))))
```
shamsimam: Nitpick: It is possible that this test executes so fast that the two `now` return the same value? We should introduce a delay or `with-redef` `now` to ensure that the status message processing is happening at a different time.

PR #408, opened 2017-06-26
Adds version number to /debug and framework name

dposada:


dposada:
Not sure if there's a specific reason. I followed the pattern that was in place. Why don't you like them as variables?

shamsimam:
I prefer running a function to query for the values rather than eagerly computing them.

dposada:
If we put them in a function, then each time someones hits `/debug`, it will look up the resource. We could make them be `delay`s, but that seems like overkill. Thoughts?

shamsimam:
`delay` is not a bad trade-off, middle ground :)

PR #410, opened 2017-07-10
Log what we are preempting and what will run instead

wyegelwel:
This change will make it easier for us to understand the choices the rebalancer is making so we can audit it.

### DIFF
--- a/scheduler/src/cook/mesos/rebalancer.clj
+++ b/scheduler/src/cook/mesos/rebalancer.clj
@@ -391,64 +391,62 @@
     (loop [state init-state
            remaining-preemption max-preemption
            [pending-job-ent & jobs-to-make-room-for] jobs-to-make-room-for
-           pending-job-ents-to-run []
-           task-ents-to-preempt []]
+           preemption-decisions []]
       (if (and pending-job-ent (pos? remaining-preemption))
         (let [[state' preemption-decision] (compute-next-state-and-preemption-decision offer-cache state params pending-job-ent)]
           (if preemption-decision
             (recur state'
                    (dec remaining-preemption)
                    jobs-to-make-room-for
-                   (conj pending-job-ents-to-run pending-job-ent)
-                   (into task-ents-to-preempt
-                         ;; If a task has no id, it must be a synthetic task.
-                         ;; This means that on one iteration of (compute-next-state-and-preemption-decision),
-                         ;; the rebalancer decided to make room for a certain hypothetical task,
-                         ;; but on a subsequent iteration, it became clear that OTHER hypothetical tasks would be an even better outcome.
-                         ;; We shouldn't try to actually preempt tasks that were never scheduled.
-                         (filter :db/id (:task preemption-decision))))
+                   (conj preemption-decisions (assoc preemption-decision
+                                                     :to-make-room-for pending-job-ent)))
             (recur state'
                    remaining-preemption
                    jobs-to-make-room-for
-                   pending-job-ents-to-run
-                   task-ents-to-preempt)))
-
+                   preemption-decisions)))
         (do
           (timers/stop timer)
-          (histograms/update! task-counts-to-preempt (count task-ents-to-preempt))
-          (histograms/update! job-counts-to-run (count pending-job-ents-to-run))
-          ;; pending-job-ents-to-run is only for debugging purpose
-          [pending-job-ents-to-run task-ents-to-preempt]
-          )))))
+          (histograms/update! task-counts-to-preempt (count (mapcat :task preemption-decisions)))
+          (histograms/update! job-counts-to-run (count preemption-decisions))
+
+          preemption-decisions)))))
 
 (defn rebalance!
   [conn driver offer-cache pending-job-ents host->spare-resources params]
   (try
     (log/info "Rebalancing...Params:" params)
     (let [db (mt/db conn)
-          [pending-job-ents-to-run task-ents-to-preempt] (rebalance db offer-cache pending-job-ents host->spare-resources params)]
-      (log/info "Jobs to run:" pending-job-ents-to-run)
-      (log/info "Tasks to preempt:" task-ents-to-preempt)
-      (doseq [task-ent task-ents-to-preempt]
-        (try
-          @(d/transact
-            conn
-            ;; Make :instance/status and :instance/preempted? consistent to simplify the state machine.
-            ;; We don't want to deal with {:instance/status :instance.status/running, :instance/preempted? true}
-            ;; all over the place.
-            (let [job-eid (:db/id (:job/_instance task-ent))
-                  task-eid (:db/id task-ent)]
-              [[:generic/ensure task-eid :instance/status (d/entid db :instance.status/running)]
-               [:generic/atomic-inc job-eid :job/preemptions 1]
-               ;; The database can become inconsistent if we make multiple calls to :instance/update-state in a single
-               ;; transaction; see the comment in the definition of :instance/update-state for more details
-               [:instance/update-state task-eid :instance.status/failed [:reason/name :preempted-by-rebalancer]]
-               [:db/add task-eid :instance/reason [:reason/name :preempted-by-rebalancer]]
-               [:db/add task-eid :instance/preempted? true]]))
-          (catch Throwable e
-            (log/warn e "Failed to transact preemption")))
-        (when-let [task-id (:instance/task-id task-ent)]
-          (mesos/kill-task! driver {:value task-id}))))))
+          preemption-decisons (rebalance db offer-cache pending-job-ents host->spare-resources params)]
+      (doseq [{job-ent-to-make-room-for :to-make-room-for
+               task-ents-to-preempt :task} preemption-decisons]
+        (log/info "Preempting tasks to make room for waiting job"
+                  {:to-make-room-for job-ent-to-make-room-for
+                   :to-preempt task-ents-to-preempt})
+        ;; If a task has no id, it must be a synthetic task.
+        ;; This means that on one iteration of (compute-next-state-and-preemption-decision),
+        ;; the rebalancer decided to make room for a certain hypothetical task,
+        ;; but on a subsequent iteration, it became clear that OTHER hypothetical tasks would be an even better outcome.
+        ;; We shouldn't try to actually preempt tasks that were never scheduled.
+        (doseq [task-ent (filter :db/id task-ents-to-preempt)]
+          (try
+            @(d/transact
+               conn
+               ;; Make :instance/status and :instance/preempted? consistent to simplify the state machine.
+               ;; We don't want to deal with {:instance/status :instance.status/running, :instance/preempted? true}
+               ;; all over the place.
+               (let [job-eid (:db/id (:job/_instance task-ent))
+                     task-eid (:db/id task-ent)]
+                 [[:generic/ensure task-eid :instance/status (d/entid db :instance.status/running)]
+                  [:generic/atomic-inc job-eid :job/preemptions 1]
+                  ;; The database can become inconsistent if we make multiple calls to :instance/update-state in a single
+                  ;; transaction; see the comment in the definition of :instance/update-state for more details
+                  [:instance/update-state task-eid :instance.status/failed [:reason/name :preempted-by-rebalancer]]
+                  [:db/add task-eid :instance/reason [:reason/name :preempted-by-rebalancer]]
+                  [:db/add task-eid :instance/preempted? true]]))
+            (catch Throwable e
+              (log/warn e "Failed to transact preemption")))
+          (when-let [task-id (:instance/task-id task-ent)]
+            (mesos/kill-task! driver {:value task-id})))))))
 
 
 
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
-
+                   preemption-decisions)))
         (do
           (timers/stop timer)
-          (histograms/update! task-counts-to-preempt (count task-ents-to-preempt))
-          (histograms/update! job-counts-to-run (count pending-job-ents-to-run))
-          ;; pending-job-ents-to-run is only for debugging purpose
-          [pending-job-ents-to-run task-ents-to-preempt]
-          )))))
+          (histograms/update! task-counts-to-preempt (count (mapcat :task preemption-decisions)))
```
shamsimam: I ran a small test locally, `reduce` came out to be much faster than `mapcat`.
Can we change this to `(reduce #(+ %1 (-> %2 :task count)) 0 preemption-decisions)`?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
-                   (conj pending-job-ents-to-run pending-job-ent)
-                   (into task-ents-to-preempt
-                         ;; If a task has no id, it must be a synthetic task.
-                         ;; This means that on one iteration of (compute-next-state-and-preemption-decision),
-                         ;; the rebalancer decided to make room for a certain hypothetical task,
-                         ;; but on a subsequent iteration, it became clear that OTHER hypothetical tasks would be an even better outcome.
-                         ;; We shouldn't try to actually preempt tasks that were never scheduled.
-                         (filter :db/id (:task preemption-decision))))
+                   (conj preemption-decisions (assoc preemption-decision
+                                                     :to-make-room-for pending-job-ent)))
```
shamsimam: Can `compute-next-state-and-preemption-decision` prepopulate the `:to-make-room-for` in `preemption-decision`?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
         (do
           (timers/stop timer)
-          (histograms/update! task-counts-to-preempt (count task-ents-to-preempt))
-          (histograms/update! job-counts-to-run (count pending-job-ents-to-run))
-          ;; pending-job-ents-to-run is only for debugging purpose
-          [pending-job-ents-to-run task-ents-to-preempt]
-          )))))
+          (histograms/update! task-counts-to-preempt (count (mapcat :task preemption-decisions)))
+          (histograms/update! job-counts-to-run (count preemption-decisions))
+
```
shamsimam: Nitpick: can we delete this blank line?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
+        (log/info "Preempting tasks to make room for waiting job"
+                  {:to-make-room-for job-ent-to-make-room-for
+                   :to-preempt task-ents-to-preempt})
+        ;; If a task has no id, it must be a synthetic task.
+        ;; This means that on one iteration of (compute-next-state-and-preemption-decision),
+        ;; the rebalancer decided to make room for a certain hypothetical task,
+        ;; but on a subsequent iteration, it became clear that OTHER hypothetical tasks would be an even better outcome.
+        ;; We shouldn't try to actually preempt tasks that were never scheduled.
+        (doseq [task-ent (filter :db/id task-ents-to-preempt)]
```
shamsimam: Should these empty invalid tasks have been filtered in `compute-preemption-decision` before returning the `preemption-decision`?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
-               ;; The database can become inconsistent if we make multiple calls to :instance/update-state in a single
-               ;; transaction; see the comment in the definition of :instance/update-state for more details
-               [:instance/update-state task-eid :instance.status/failed [:reason/name :preempted-by-rebalancer]]
-               [:db/add task-eid :instance/reason [:reason/name :preempted-by-rebalancer]]
-               [:db/add task-eid :instance/preempted? true]]))
-          (catch Throwable e
-            (log/warn e "Failed to transact preemption")))
-        (when-let [task-id (:instance/task-id task-ent)]
-          (mesos/kill-task! driver {:value task-id}))))))
+          preemption-decisons (rebalance db offer-cache pending-job-ents host->spare-resources params)]
```
shamsimam: Nitpick: typo `decisions `

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -391,64 +391,66 @@
...
+      (doseq [{job-ent-to-make-room-for :to-make-room-for
+               task-ents-to-preempt :task} preemption-decisons]
+        ;; Ensure that uuids are loaded in entity
+        (:job/uuid job-ent-to-make-room-for)
+        (doall (map :instance/task-id task-ents-to-preempt))
+
+        (log/info "Preempting tasks to make room for waiting job"
```
shamsimam: Can you add an example of what this log looks like? Should we instead have a `map` to extract only the `:db/id` and `:job/uuid` or `:instance/task-id` in jobs and tasks, respectively?

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: How much faster? 

I think the current version is more readable and since this is only done once per rebalancer I'm hesitant to change it unless it is too slow.

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Thought it was more readable so you can more easily scan to see what is returned. Don't care too much though if you would prefer I remove

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Yes, however, I want to leave it as is for now so that we can see in the logs how frequently this happens

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: done

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: done, thanks

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Here is an example:

```
2017-07-11 13:32:20,009 [async-thread-macro-2] INFO cook.mesos.rebalancer - Preempting tasks to make room for waiting job {:to-make-room-for {:db/id 17592186045652, :job/resour
ce #{{:db/id 17592186045653, :resource/type :resource.type/cpus, :resource/amount 2.0} {:db/id 17592186045654, :resource/type :resource.type/mem, :resource/amount 2000.0}}, :jo
b/priority 50, :job/user b, :job/uuid #uuid "59650b9a-f364-4a13-bb7c-96a6312a2719"}, :to-preempt ({:db/id 17592186045622, :job/_instance {:db/id 17592186045518, :job/resource #
{{:db/id 17592186045520, :resource/type :resource.type/mem, :resource/amount 2000.0} {:db/id 17592186045519, :resource/type :resource.type/cpus, :resource/amount 2.0}}, :job/us
er a, :job/priority 50}, :instance/start-time #inst "2017-07-11T17:32:10.971-00:00", :instance/hostname 156eed6c-fe44-4ed7-971e-e7b4432638b4, :instance/slave-id 156eed6c-fe44-4
ed7-971e-e7b4432638b4, :instance/task-id bfd30118-d811-4807-8e44-f44f3cbe0645})}
```

I went back and forth on this. On the one hand it would be a lot cleaner to reduce to the just the uuids, on the other hand, since this includes the resource info it makes it easier to understand what happened. I'm still largely on the fence so if you have a strong opinion I'm happy to switch it.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: On a vector of size 1000 with nested vector of size `(rand-int 1000)` the timings were around 10 ms vs 120 ms.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Okay, no need to change, was just a nitpick.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: If it's not too expensive to generate, I would prefer a more more readable log like:
`INFO cook.mesos.rebalancer - Preempting tasks to make room for waiting job {:to-make-room-for {:db/id 17592186045652, :cpus 2.0, :mem 2000.0, :job/priority 50, :job/user b, :job/uuid #uuid "59650b9a-f364-4a13-bb7c-96a6312a2719"}, :to-preempt ({:db/id 17592186045622, :job/_instance {:db/id 17592186045518, :cpus 2.0, :mem 2000.0, :job/user a, :job/priority 50}, :instance/task-id bfd30118-d811-4807-8e44-f44f3cbe0645})}`

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Okay. Should we add a todo so that we cleanup it up later?

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: This version quickly tells me what resources were freed and what will be used up. It also has the ids in case I want to use them in grep or something else.

wyegelwel:
Ready for another round of review

shamsimam:
@wyegelwel have some open responses above.

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: I didn't realize reduce would be *that* much faster than mapcat so I reproduced your results:

```
cook.components=> (def a (repeatedly 1000 (fn [] {:tasks (repeat (rand-int 1000) :a)})))
#'cook.components/a
cook.components=> (quick-bench (count (mapcat :tasks a)))
Evaluation count : 18 in 6 samples of 3 calls.
             Execution time mean : 48.031183 ms
    Execution time std-deviation : 981.794831 µs
   Execution time lower quantile : 47.429016 ms ( 2.5%)
   Execution time upper quantile : 49.411953 ms (97.5%)
                   Overhead used : 1.719809 ns
nil
cook.components=> (quick-bench (reduce #(+ %1 (-> %2 :tasks count)) 0 a))
Evaluation count : 42 in 6 samples of 7 calls.
             Execution time mean : 15.862469 ms
    Execution time std-deviation : 172.828084 µs
   Execution time lower quantile : 15.693824 ms ( 2.5%)
   Execution time upper quantile : 16.061297 ms (97.5%)
                   Overhead used : 1.719809 ns
nil
```

However, even 20 tasks preempted for a single job would be quite large. Restricting to a 1000 preemptions, each which preempted 20 tasks, we see even better relative performance by reduce but both are fast enough that I'm leaning toward keeping as is. Is that ok with you?

```
cook.components=> (def b (repeatedly 1000 (fn [] {:tasks (repeat 20 :b)})))
#'cook.components/b
cook.components=> (quick-bench (count (mapcat :tasks b)))
Evaluation count : 294 in 6 samples of 49 calls.
             Execution time mean : 2.086402 ms
    Execution time std-deviation : 13.811516 µs
   Execution time lower quantile : 2.072167 ms ( 2.5%)
   Execution time upper quantile : 2.100173 ms (97.5%)
                   Overhead used : 1.719809 ns
nil
cook.components=> (quick-bench (reduce #(+ %1 (-> %2 :tasks count)) 0 b))
Evaluation count : 912 in 6 samples of 152 calls.
             Execution time mean : 660.952298 µs
    Execution time std-deviation : 4.295272 µs
   Execution time lower quantile : 657.737007 µs ( 2.5%)
   Execution time upper quantile : 667.002116 µs (97.5%)
                   Overhead used : 1.719809 ns
nil
```

shamsimam: Yeah, that is fine.

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: Here is the updated output:

```
2017-07-12 13:26:44,298 [async-thread-macro-8] INFO cook.mesos.rebalancer - Preempting tasks to make room for waiting job {:to-make-room-for {:db/id 17592186045652, :job/uuid #
uuid "59665bca-6ba9-4687-9089-dbc9d5688d7a", :job/user b, :ports 0, :mem 2000.0, :cpus 2.0}, :to-preempt ({:db/id 17592186045622, :instance/task-id afa9c6fe-928e-4689-bd29-0351
269a94d8, :job {:db/id 17592186045518, :job/uuid #uuid "59665bca-0618-4299-8be1-ad2833a409ea", :job/user a, :ports 0, :cpus 2.0, :mem 2000.0}})}
```

scheduler/src/cook/mesos/rebalancer.clj
wyegelwel: done

PR #413, opened 2017-07-13
Return maps instead of entities from rank-jobs

pschorf:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -595,7 +594,7 @@
                                         (sort-by (comp :db/id :job #(.getRequest %))))
         :let [request (.getRequest task)
               task-id (:task-id request)
-              job-id (get-in request [:job :db/id])]]
+              job-id [:job/uuid (get-in request [:job :job/uuid])]]]
```
shamsimam: Should the variable name be updated, it is no longer an `id`, it is now a vector.

scheduler/src/cook/mesos/util.clj
```diff
@@ -89,6 +89,16 @@
                 x))
             m))
 
+(defn entity->map
```
shamsimam: Should we add unit tests with nested entities/components and verify that the returned map is what we expect?

wyegelwel:
Do you have a sense what the performance implications are?

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: I see, you are creating a look up ref instead of using the entity id. I think it is worth re-naming to make that clear.

scheduler/src/cook/mesos/util.clj
```diff
@@ -89,6 +89,16 @@
...
+  "Takes a datomic entity and converts it along with any nested entities
+  into clojure maps"
+  ([entity db]
+   (entity->map (d/entity db (:db/id entity))))
+  ([entity]
+   (->> entity
+        d/touch
```
wyegelwel: I don't think the touch is necessary

pschorf: Removed

scheduler/src/cook/mesos/util.clj
pschorf: I think the existing unit tests cover it

scheduler/src/cook/mesos/scheduler.clj
pschorf: Done

pschorf:
After testing with the simulator, the rank/schedule cycle is about 10% faster with maps. Additionally, the memory usage should be much better.

scheduler/src/cook/mesos/util.clj
shamsimam: I mean an explicit test where an entity with nested components/entities get loaded. Which unit test covers this?

pschorf: Added explicit unit test.

PR #418, opened 2017-07-19
Adds authorization check for job creation

dposada:


scheduler/src/cook/mesos/api.clj
```diff
@@ -1108,7 +1116,7 @@
                        (catch Exception e
                          (log/warn e "Malformed raw api request")
                          [true {::error (.getMessage e)}]))))
-
+     :allowed? (partial job-create-allowed? is-authorized-fn)
```
shamsimam: Is it possible to add an integration test where the creation of a job is disallowed?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1372,4 +1372,9 @@
                                :reason_string "Preempted by rebalancer"
                                :sandbox_directory "/path/to/working/directory"
                                :status "success")]
-            (is (= expected-map (dissoc instance-map :start_time)))))))))
\ No newline at end of file
+            (is (= expected-map (dissoc instance-map :start_time)))))))))
+
+(deftest test-job-create-allowed?
```
shamsimam: Can we also add assertions on the arguments to the `is-authorized-fn` function?

scheduler/src/cook/mesos/api.clj
dposada: Technically, it's possible, but I think it's more effort than it's worth. I think it would require starting up cook configured with a different `:authorization-fn` that would specifically disallow creation based on something unique to the test.

scheduler/test/cook/test/mesos/api.clj
dposada: It seems like that would make the test quite brittle, because it's really close to just re-writing the code in test form.

shamsimam: I care about the action and the entity. Imho, changing them should cause the test to fail.

scheduler/src/cook/mesos/api.clj
shamsimam: okay.

scheduler/test/cook/test/mesos/api.clj
dposada: Compromise! I added a test that invokes the handler with a custom authorization function, so we can actually test the full job-create authorization workflow.

dposada:
@shamsimam Can we merge this?

PR #420, opened 2017-07-19
Throttle rate of publishing progress updates from cook executor

shamsimam:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+          ;; TODO metrics for map size, datomic queries, etc.
+          (log/info "Received" (count instance-id->progress-state) "in progress-update-transactor")
+          (if instance-id->progress-state
+            (do
+              (histograms/update! progress-updater-pending-states (count instance-id->progress-state))
+              (meters/mark! progress-updater-publish-rate)
+              (timers/time!
+                progress-updater-publish-duration
+                (doseq [instance-id->progress-state-partition
+                        (partition batch-size batch-size [] instance-id->progress-state)]
```
wyegelwel: Use `partition-all`

shamsimam: Thanks, done :)

scheduler/docs/configuration.asc
```diff
@@ -340,6 +340,38 @@ The configuration values are defined as follows:
...
+            :pending-threshold 4000
+            :publish-interval-ms 2500}
+ ...}
+```
+
+The configuration values are defined as follows:
+
+`:batch-size`::
+  An integer representing the number of facts that are updated in individual datomic instance progress update transactions.
+  The default value is 100.
```
sradack: It would be nice to determine the optimal value for our configuration.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1316,7 +1432,7 @@
 
 (let [in-order-queue-counter (counters/counter ["cook-mesos" "scheduler" "in-order-queue-size"])
       in-order-queue-timer (timers/timer ["cook-mesos" "scheduler" "in-order-queue-delay-duration"])
-      parallelism 19 ; a prime number to improve distribution after hashing
+      parallelism 19 ; a prime number to potentially help make the distribution uniform
```
sradack: Haha

shamsimam: ;)

scheduler/docs/configuration.asc
shamsimam: Agreed.

shamsimam:
ready for another round of reviews

sradack:
Not sure this needs another round.  Just need @wyegelwel to merge once he approves.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1397,16 +1514,20 @@
...
-                            fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset match-trigger-chan)]
+        (make-offer-handler conn driver-atom fenzo framework-id executor-config pending-jobs-atom offer-cache fenzo-max-jobs-considered
+                            fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset match-trigger-chan)
+        {:keys [batch-size pending-threshold]} progress-config
+        progress-state-chan (progress-update-transactor progress-updater-trigger-chan batch-size conn)
+        {:keys [progress-aggregator-chan]} (progress-update-aggregator pending-threshold progress-aggregator-trigger-chan progress-state-chan)]
     (start-jobs-prioritizer! conn pending-jobs-atom task-constraints rank-trigger-chan)
     {:scheduler (create-mesos-scheduler framework-id gpu-enabled? conn heartbeat-ch
-                                        fenzo offers-chan match-trigger-chan)
+                                        fenzo offers-chan match-trigger-chan progress-aggregator-chan)
```
wyegelwel: Instead of passing the channel, pass a function that is *something* like `handle-progress-update` which just enqueues to update. That way we can better localize the logic for handling progress updates. Also, the code `handle-status-update` code shouldn't need to know about the implementation we are using to handle-progress-updates.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+(defn progress-update-transactor
+  "Launches a long running go block that triggers transacting the latest aggregated instance-id->progress-state to
+   datomic whenever there is a message on the progress-updater-trigger-chan.
+   No more than batch-size facts are updated in individual datomic transactions.
+   It returns a progress-state-chan which can be used to send messages about the latest instance-id->progress-state.
+   If no such message is on the progress-state-chan, then no datomic interactions occur."
+  [progress-updater-trigger-chan batch-size conn]
+  (log/info "Starting progress update transactor")
+  (let [progress-state-chan (async/chan (async/sliding-buffer 1))]
+    (async/go
```
wyegelwel: I think you can replace this go and the loop with https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/util.clj#L523 which handles waiting on the trigger channel and then closing the response channel.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+      (loop []
+        (let [{:keys [response-chan]} (async/<! progress-updater-trigger-chan)
+              instance-id->progress-state (async/<! progress-state-chan)]
+          ;; TODO metrics for map size, datomic queries, etc.
+          (log/info "Received" (count instance-id->progress-state) "in progress-update-transactor")
+          (if instance-id->progress-state
+            (do
```
wyegelwel: Move this code into a function

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+  [progress-updater-trigger-chan batch-size conn]
+  (log/info "Starting progress update transactor")
+  (let [progress-state-chan (async/chan (async/sliding-buffer 1))]
+    (async/go
+      (loop []
+        (let [{:keys [response-chan]} (async/<! progress-updater-trigger-chan)
+              instance-id->progress-state (async/<! progress-state-chan)]
+          ;; TODO metrics for map size, datomic queries, etc.
+          (log/info "Received" (count instance-id->progress-state) "in progress-update-transactor")
+          (if instance-id->progress-state
```
wyegelwel: I'm having a hard to closing my parens, is there the else case for this if?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+            (do
+              (histograms/update! progress-updater-pending-states (count instance-id->progress-state))
+              (meters/mark! progress-updater-publish-rate)
+              (timers/time!
+                progress-updater-publish-duration
+                (doseq [instance-id->progress-state-partition
+                        (partition-all batch-size instance-id->progress-state)]
+                  (try
+                    (let [transactions
+                          (reduce (fn progress-update-transactor-reducer
```
wyegelwel: I'd prefer using `mapcat` here. When I see a `reduce` I need to read it carefully because `reduce` can do anything. Using `mapcat` better signals the intent

wyegelwel:
Why are there

wyegelwel:
(sorry fat fingered the close)

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+  [pending-progress-threshold progress-aggregator-trigger-chan progress-state-chan]
+  (log/info "Starting progress update aggregator")
+  (let [progress-aggregator-chan (async/chan (async/sliding-buffer pending-progress-threshold))
+        query-state-chan (async/chan (async/sliding-buffer 1))]
+    (async/go
+      (loop [instance-id->progress-state {}]
+        (let [[data chan] (async/alts! [progress-aggregator-trigger-chan progress-aggregator-chan query-state-chan]
+                                       :priority true)]
+          (condp = chan
+            progress-aggregator-trigger-chan
```
wyegelwel: Why does this trigger channel need to exist? Could you instead have one of the alts ports be to put the map onto `progress-state-chan` and then clear the map when it is chosen? That would make the intent clearer.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
...
+                (let [progress-state (select-keys data [:progress-message :progress-percent])
+                      instance-id->progress-state' (assoc instance-id->progress-state instance-id progress-state)]
+                  (recur instance-id->progress-state'))
+                (do
+                  (meters/mark! progress-aggregator-drop-rate)
+                  (counters/inc! progress-aggregator-drop-count)
+                  (log/debug "Dropping" data "as threshold has been reached")
+                  (recur instance-id->progress-state))))
+
+            query-state-chan
```
wyegelwel: I don't see where this is used except in tests. This seems like unneeded complexity. If you remove the trigger channel in this loop and allow for pulling you will be able to get the query, though it will wipe the state and then you could close the progress-aggregator-chan to signal exit.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: As an aside, if you remove the trigger and query channels you end up with effectively a single write port (progress-aggregator-chan) and a single read port (progress-state-chan) with some logic in between to transform the inputs to the output. I wonder if you could achieve this as a transducer. Then this code would be a single function and there would be a single channel which would make exiting simpler. 

I haven't played around with transducers enough to know if doing it for this would be a massive pain, but may be worth looking into.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,121 @@
   (-> (d/entity db [:instance/task-id task-id])
       :db/id))
 
+(histograms/defhistogram [cook-mesos scheduler progress-aggregator-pending-states])
+(meters/defmeter [cook-mesos scheduler progress-aggregator-message-rate])
+(meters/defmeter [cook-mesos scheduler progress-aggregator-drop-rate])
+(counters/defcounter [cook-mesos scheduler progress-aggregator-drop-count])
+
+(defn progress-update-aggregator
```
wyegelwel: My own notes:

Chan                                                                Use
progress-aggregator-chan                               Receive progress updates
progress-aggregator-trigger-chan                    "Triggers" sending a progress update on progress-state-chan
progress-state-chan                                         Sends progress-state map when progress-aggregator-trigger-chan is fired
query-state-chan                                              Receive messages (supported [query, exit]), query sends progress map on response, exit sends exited message back

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Passing the function sounds good. To be clear, you intend it to wrap the `async/put!` on `progress-aggregator-chan`?

Did you mean `handle-framework-message` instead of `handle-status-update`?

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Okay, will use `chime-at-ch`

scheduler/src/cook/mesos/scheduler.clj
shamsimam: The else part is: `(log/info "Exiting progress update transactor")`

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Agree, will do.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: The query channel helps simplify the testing as it also acts as a synchronization mechanism guaranteeing the messages that have been processed.

> allow for pulling you will be able to get the query

I'm not sure what pulling refers to here.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: > Could you instead have one of the alts ports be to put the map onto progress-state-chan and then clear the map when it is chosen?

`progress-state-chan` is a channel that holds only the latest value. 
If we remove the `progress-aggregator-trigger-chan` and use `alts!` to put the current progress state `progress-state-chan` we will end up triggering the progress transactor at a much faster rate than what we want - at intervals of `progress-sample-interval-ms`.

> if you remove the trigger and query channels

The trigger channel is being used to control the rate at which we are publishing events to the transactor and eventually the rate at which we are doing datomic writes. More on the query channel below.

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: > If we remove the progress-aggregator-trigger-chan and use alts! to put the current progress state progress-state-chan we will end up triggering the progress transactor at a much faster rate than what we want - at intervals of progress-sample-interval-ms.

I was assuming it would be taken from the progress-update-transactor which is triggered at publish-interval-ms. Where would progress-sample-interval-ms come up?

> In addition, we will end up losing a larger number of the aggregated progress values when the transactor is processing values slowly.

Isn't this true as is since the channel will drop the previous map?

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: > I'm not sure what pulling refers to here.

I meant that if you add putting data onto the progress-state-chan in the alts, you can query the state.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: `progress-aggregator-trigger-chan` is being triggered at `progress-sample-interval-ms` ms. This trigger is causing writes to ` progress-state-chan`. The transactor is being triggered anytime it can read from `progress-state-chan`. 

Using the progress write from `alts!` will trigger writes to `progress-state-chan` more frequently than we currently do. Hence the increased likelihood of dropping progress messages when the transactor falls behind.

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Created `handle-progress-message!`.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,131 @@
...
+  (histograms/update! progress-updater-pending-states (count instance-id->progress-state))
+  (meters/mark! progress-updater-publish-rate)
+  (timers/time!
+    progress-updater-publish-duration
+    (doseq [instance-id->progress-state-partition
+            (partition-all batch-size instance-id->progress-state)]
+      (try
+        (let [transactions
+              (mapcat
+                (fn progress-update-transactor-reducer
```
wyegelwel: Move to:

```clojure
(let [build-progress-txns (fn build-progress-txns 
                              [instance-id {:keys [progress-percent progress-message]}]]
                              (cond-> []
                                      progress-percent (conj [:db/add instance-id :instance/progress (int progress-percent)])
                                      progress-message (conj [:db/add instance-id :instance/progress-message progress-message])))
      txns (mapcat build-progress-txns instance-id->progress-state-partition)]
```

scheduler/src/cook/mesos/scheduler.clj
shamsimam: I like this new code after removing `progress-aggregator-trigger-chan`, thanks!

Let me play around with the transducer stuff now :)

shamsimam: I can see how we would use transducers to update `instance-id->progress-state` as we receive data on `progress-aggregator-chan` by piping the latest value to a sliding buffer chan. However, I'm not sure how I can atomically clear out the portion of the map passed on to `progress-state-chan` read from this new chan as `progress-aggregator-chan` can (1) receive intermediate progress update and (2) it itself has a sliding buffer and can drop messages.

scheduler/src/cook/mesos/scheduler.clj

wyegelwel:
I can't find your comment saying transducers aren't a good fit so will just comment here.

What do you think about making an abstraction like async/pipe but where you specify an initial value,  a reducing function for input elements and a finalizing function for when elements are pulled. Something like:

```clojure
(defn xform-pipe
    "Reads elements from the `in` channel and supplies elements to the `out` channel.
     Maintains a state (initialized to `initial-state`) that is updated by applying the reducing function
     `in-xform` to the current state and the incoming element `(in-xform state element)`.
     `out-finalizer` accepts the current state and produces `[new-state to-return-value]`. 
     When the out channel can receive an element, the `to-return-value` is put on the channel
     and the state is set the `new-state` returned from the `out-finalizer`"
    [in initial-state in-xform out out-finalizer])
 ```

This could be used to implement rolling window aggregations or complex buffering logic without being a mess of async code.

The benefit is that you separate the logic for handling insertion and extraction from the actions you want to take on insertion and extraction. This makes it easier to test both and makes it reusable.

This can't be achieved with transducers because when the state is consumed is not made explicit to the transducer, it must flush the state based on the inputs or some pre-configured value (e.x. `take`). This might be able to be implemented by having the completion arity of the transducer called on each consumption but that is likely a misuse of the transducer api.

What do you think?

**Comment by shamsimam** [comment]
@wyegelwel 

> `defn xform-pipe` What do you think?

Sounds good.

File: scheduler/src/cook/mesos/scheduler.clj [hunk 13]
```diff
@@ -279,9 +279,112 @@
...
+   If no such message is on the progress-state-chan, then no datomic interactions occur."
+  [progress-updater-trigger-chan batch-size conn]
+  (log/info "Starting progress update transactor")
+  (let [progress-state-chan (async/chan)]
+    (letfn [(progress-update-transactor-error-handler [e]
+              (log/error e "Failed to update progress message on tasks!"))
+            (progress-update-transactor-on-finished []
+              (log/info "Exiting progress update transactor"))
+            (process-progress-update-transactor-event []
+              (let [[data _] (async/alts!! [progress-state-chan (async/timeout 100)] :priority true)]
```
wyegelwel: Why the timeout?

shamsimam: To avoid blocking the thread forever.

scheduler/src/cook/mesos/util.clj
```diff
@@ -61,15 +61,15 @@
...
-   (list? form) (outer (apply list (map inner form)))
-   (instance? clojure.lang.IMapEntry form) (outer (vec (map inner form)))
-   (seq? form) (outer (doall (map inner form)))
-   (instance? clojure.lang.IRecord form)
-     (outer (reduce (fn [r x] (conj r (inner x))) form form))
-   ;; Added the line below to work with datomic entities..
-   (instance? datomic.query.EntityMap form) (outer (into {} (map inner) form))
-   (coll? form) (outer (into (empty form) (map inner) form))
-   :else (outer form)))
+    (list? form) (outer (apply list (map inner form)))
```
shamsimam: Sorry for the indentation changes due to reformatting.

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -14,14 +14,14 @@
...
 (ns cook.test.mesos.util
- (:use clojure.test)
- (:require [clj-time.coerce :as tc]
-           [clj-time.core :as t]
-           [clojure.core.async :as async]
-           [cook.mesos.util :as util]
-           [cook.test.testutil :as testutil :refer (create-dummy-instance create-dummy-job restore-fresh-database!)]
-           [datomic.api :as d :refer (q db)])
- (:import [java.util Date]))
+  (:use clojure.test)
```
shamsimam: Sorry for the indentation changes due to reformatting.

scheduler/src/cook/mesos/scheduler.clj
wyegelwel: oh, right because it is an actual thread. Yea, totally makes sense.

shamsimam:
@wyegelwel `xform-pipe` changes ready for review :)

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,26 @@
...
   (->> (repeatedly ch-size #(async/poll! ch))
        (remove nil?)))
+
+(defn xform-pipe
+  "Reads elements from the `in-chan` channel and supplies elements to the `out-chan` channel.
+   Maintains a state (initialized to `initial-state`) that is updated by applying the reducing function
+   `in-xform` to the current state and the incoming element `(in-xform state element)`.
+   `out-finalizer` accepts the current state and produces `[reset-state out-value]`.
+   When the out channel can receive an element, the `value` is put on the channel
+   and the state is set to the `reset-state` returned from the `out-finalizer`.
```
wyegelwel: You should add a note that out-finalizer may be called when the output is not consumed. (i.e. don't make it side effecty)

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,26 @@
...
+
+   Note: This function does not perform error handling, exceptions must be explicitly handled in the
+   provided functions (i.e. in-xform, out-finalizer, on-consumed and on-finished)."
+  [in-chan initial-state in-xform out-chan out-finalizer &
+   {:keys [on-consumed on-finished] :or {on-consumed identity, on-finished #()}}]
```
wyegelwel: These callbacks were a really nice touch

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -279,9 +279,107 @@
...
+  "Launches a long running go block that triggers publishing the latest aggregated instance-id->progress-state
+   whenever there is a read on the progress-state-chan.
+   It drops messages if the progress-aggregator-chan queue is larger than pending-progress-threshold or the aggregated
+   state has more than pending-progress-threshold different entries.
+   It returns the progress-aggregator-chan which can be used to send progress-state messages to the aggregator."
+  [pending-progress-threshold progress-state-chan]
+  (log/info "Starting progress update aggregator")
+  (let [progress-aggregator-chan (async/chan (async/sliding-buffer pending-progress-threshold))
+        progress-aggregator-fn (fn progress-aggregator-fn [instance-id->progress-state data]
+                                 (let [{:keys [instance-id response-chan]} data]
```
wyegelwel: I assume this response-chan is only here for testing. Please remove it, and then move this function into the namespace and test it explicitly. Given the tests on xform-pipe and the tests of this function, you can have simpler tests for the aggregator.

scheduler/src/cook/mesos/util.clj
shamsimam: Will do.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Will do

scheduler/src/cook/mesos/util.clj
shamsimam: Done.

scheduler/src/cook/mesos/scheduler.clj

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
...
+   `out-value` will be consumed, hence it should not have side-effects. The optional parameter `on-consumed`
+   can be used to register notifications of when a value has been supplied to the `out-chan` channel.
+
+   Note: This function does not perform error handling, exceptions must be explicitly handled in the
+   provided functions (i.e. in-xform, out-xform, on-consumed and on-finished)."
+  [in-chan initial-state in-xform out-chan out-xform &
+   {:keys [on-consumed on-finished] :or {on-consumed identity, on-finished #()}}]
+  (async/go-loop [state initial-state]
+    (let [[reset-state out-value] (out-xform state)
+          [data chan] (async/alts! [[out-chan out-value] in-chan] :priority true)]
```
wyegelwel: nitpick: can you use `alt!` instead of `alts!`? I expect it would make things marginally cleaner.

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1642,5 +1683,190 @@
...
+                    "i2" {:progress-message "i2.m2", :progress-percent 25}
+                    "i3" {:progress-message "i3.m1", :progress-percent 10}}
+                   (async/<!! progress-state-chan)))
+            (is (= {} (async/<!! progress-state-chan)))
+
+            (async/close! progress-state-chan)
+            (async/close! progress-aggregator-chan)
+            (poll-until #(nil? (async/<!! progress-state-chan)) 100 1000)))
+
+        (testing "basic progress overflow on channel"
```
wyegelwel: I don't understand how this is testing overflow. This seems like the less message is correctly received.

scheduler/src/cook/mesos/util.clj
shamsimam: I tend to favor `alts!` over `alt!` due to previously having experienced issues getting `alt!` working with cloverage. Some Waiter fragments that use `alt!` don't work with cloverage.

scheduler/test/cook/test/mesos/scheduler.clj
shamsimam: Added an assertion to verify dropped messages by maintaining counter of processed messages

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
...
+   `in-xform` to the current state and the incoming element `(in-xform state element)`.
+   `out-xform` accepts the current state (i.e. `(out-xform state)`) and produces `[reset-state out-value]`.
+   When the `out-chan` channel can receive an element, the `value` is put on the channel and the state is
+   set to the `reset-state` returned from the `out-xform`. Invocation of `out-xform` does not guarantee that
+   `out-value` will be consumed, hence it should not have side-effects. The optional parameter `on-consumed`
+   can be used to register notifications of when a value has been supplied to the `out-chan` channel.
+
+   Note: This function does not perform error handling, exceptions must be explicitly handled in the
+   provided functions (i.e. in-xform, out-xform, on-consumed and on-finished)."
+  [in-chan initial-state in-xform out-chan out-xform &
```
leifwalsh: I might reorder these to be `[to out-xform in-xform from initial-state]`, looking at [`pipeline-async`](https://clojuredocs.org/clojure.core.async/pipeline-async)

scheduler/src/cook/mesos/util.clj
wyegelwel: Makes snese

scheduler/src/cook/mesos/util.clj
wyegelwel: On the other hand, [`pipe`](https://clojuredocs.org/clojure.core.async/pipe) does `[from to]` so they are not consistent. I'm not sure I care enough to argue one direction or the other.

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
...
+   {:keys [on-consumed on-finished] :or {on-consumed identity, on-finished #()}}]
+  (async/go-loop [state initial-state]
+    (let [[reset-state out-value] (out-xform state)
```
leifwalsh: I'm annoyed that you call `out-xform` on every loop even if you can't output yet and your next action is to read another input, but I've been tossing and turning this around in my head for like 20 minutes and can't figure out a way around it.

Nothing to do here just :cry:

wyegelwel: Yea.. I don't think there is a way to avoid it without changing the method signature. I suppose what you could do is put a channel on the out-chan, and then when it is pulled put the value on the returned channel.

Wish alts exposed the ability to call a function to generate the value to put on the channel...

wyegelwel: Holy shit, the [alts code](https://github.com/clojure/core.async/blob/master/src/main/clojure/clojure/core/async.clj#L237) is crazy. I don't understand how they are achieving the promise that it will block until one is available. It looks like they are trying each channel and returning if none are ready.

wyegelwel: One thing you could do is put a promise on the channel and when it is pulled actually call the finalizer and deliver the value. 

You can then provide a transducer to the out channel to deref the promise

shamsimam: I wanted to go down the path of putting a `promise-chan` and letting the consumer of `out-chan` read from the promise channel, but aborted that approach.

Using the promise and transducer saves the consumer from having to change, but now (1) the out has to be created using that transducer and (2) the transducer does a potentially blocking read from the promise.

Not sure which approach is better :(

wyegelwel: I think it is ok as is. Though maybe put a note that it is less than ideal. 

If we wanted to go a little crazy.. One thing we could do is have an intermediate-chan with no buffer that we try to write to. That intermediate-chan is connected to the out-chan using one of the pipeline functions (not sure if we want async vs blocking, probably async) and the function between them is to take from the promise-chan. I *think* this would work, but is definitely worth testing.

wyegelwel: Specifically, one thing I'm unsure of is if the pipeline function will read values from the from channel even if the to channel isn't ready to receive them.

shamsimam: The `pipeline` functions don't work as the `n` argument must be positive which means there are eager reads on the intermediate-chan. This causes the `alts!` to return eagerly with success on `intermediate-chan`.

This is what I tried:
```
TODO
```

shamsimam: I'm not sure the transducer approach will work either as it needs a buffer which needs to be a positive integer or a buffer.

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
   [ch ch-size]
   (->> (repeatedly ch-size #(async/poll! ch))
        (remove nil?)))
+
+(defn xform-pipe
+  "Reads elements from the `in-chan` channel and supplies elements to the `out-chan` channel.
+   Maintains a state (initialized to `initial-state`) that is updated by applying the reducing function
+   `in-xform` to the current state and the incoming element `(in-xform state element)`.
+   `out-xform` accepts the current state (i.e. `(out-xform state)`) and produces `[reset-state out-value]`.
+   When the `out-chan` channel can receive an element, the `value` is put on the channel and the state is
```
wyegelwel: `value` -> `out-value`

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
...
+   set to the `reset-state` returned from the `out-xform`. Invocation of `out-xform` does not guarantee that
```
wyegelwel: You may also want to add that `out-xform` is expected to be a cheap operation

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,28 @@
   [ch ch-size]
   (->> (repeatedly ch-size #(async/poll! ch))
        (remove nil?)))
+
+(defn xform-pipe
```
sradack: I like that we factored this out.  Can it have a better name?  Maybe `reduce-pipe`?  I want to capture that there is state inside this pipe.

scheduler/src/cook/mesos/util.clj
shamsimam: I added the response-chan variant, let me know how you guys feel about the change. The transducer approach did not work as I needed to provide a buffer.

scheduler/src/cook/mesos/util.clj
shamsimam: Done.

scheduler/src/cook/mesos/util.clj
shamsimam: With the response-chan change, this is no longer required.

scheduler/src/cook/mesos/util.clj
shamsimam: Done.

scheduler/src/cook/mesos/util.clj
```diff
@@ -553,27 +553,35 @@
...
+  [in-chan in-xform out-chan out-xform &
+   {:keys [initial-state on-finished] :or {initial-state nil, on-finished #()}}]
+  (async/go-loop [state initial-state
+                  out-data-chan (async/chan 1)]
+    (let [[data chan] (async/alts! [[out-chan out-data-chan] in-chan] :priority true)]
       (condp = chan
-        out-chan (do
-                   (on-consumed out-value)
-                   (recur reset-state))
+        out-chan (if data
```
wyegelwel: What is the data here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -553,27 +553,35 @@
...
+   provided functions (i.e. in-xform, out-xform and on-finished). In particular, `out-xform` must not
+   throw an exception as it will prevent the promise-chan from being fulfilled.
+
+   Note: The roundabout approach of putting a promise-chan on the out-chan before actually placing the
+   value on the promise-chan is necessary as we want to limit calls to `out-xform` to only when the
+   data is ready to be consumed by `out-chan`."
+  [in-chan in-xform out-chan out-xform &
+   {:keys [initial-state on-finished] :or {initial-state nil, on-finished #()}}]
+  (async/go-loop [state initial-state
+                  out-data-chan (async/chan 1)]
```
wyegelwel: Why reuse the chan?

wyegelwel: Use a promise-chan instead of a regular chan. I think you also want a chan with no buffer since I believe this will always be available to write to (since we refresh the chan).

wyegelwel: Actually, I see when you refresh you have the chan with no buffer. I would recommend switching to a promise chan.

scheduler/src/cook/mesos/util.clj
```diff
@@ -553,27 +553,35 @@
...
    Maintains a state (initialized to `initial-state`) that is updated by applying the reducing function
    `in-xform` to the current state and the incoming element `(in-xform state element)`.
    `out-xform` accepts the current state (i.e. `(out-xform state)`) and produces `[reset-state out-value]`.
-   When the `out-chan` channel can receive an element, the `value` is put on the channel and the state is
-   set to the `reset-state` returned from the `out-xform`. Invocation of `out-xform` does not guarantee that
-   `out-value` will be consumed, hence it should not have side-effects. The optional parameter `on-consumed`
-   can be used to register notifications of when a value has been supplied to the `out-chan` channel.
+   When the `out-chan` channel can receive an element, a `out-data-chan` channel is put on the `out-chan`
+   channel. Then the `out-value` is put on this `out-data-chan` and the state is set to the `reset-state`.
+   Invocation of `out-xform` guarantees that `out-value` has been provided, hence it can be used
```
wyegelwel: I actually liked the on-consumed call back so you can separate your stateful actions from the actual processing.

scheduler/src/cook/mesos/util.clj
shamsimam: Okay, started out with a promise-chan and switched it :)

I ran into issues with mults and promise chan, so I try and avoid it when possible. But happy to switch back to it.

scheduler/src/cook/mesos/util.clj
shamsimam: Hmm, okay I'll add it back.

wyegelwel: @sradack what do you think?

scheduler/src/cook/mesos/util.clj
shamsimam: data is true/false depending on whether the put succeeded on out-chan. E.g. it would be false if out-chan was closed.

scheduler/src/cook/mesos/util.clj
shamsimam: Added it back.

scheduler/src/cook/mesos/util.clj
shamsimam: Used `promise-chan`

scheduler/src/cook/mesos/util.clj
wyegelwel: I would also remove the explicit guarantee that when out-xform is called the value is always used. No reason to constrain ourselves.

scheduler/src/cook/mesos/util.clj
```diff
@@ -553,27 +553,35 @@
...
-  (async/go-loop [state initial-state]
-    (let [[reset-state out-value] (out-xform state)
-          [data chan] (async/alts! [[out-chan out-value] in-chan] :priority true)]
+   provided functions (i.e. in-xform, out-xform and on-finished). In particular, `out-xform` must not
+   throw an exception as it will prevent the promise-chan from being fulfilled.
+
+   Note: The roundabout approach of putting a promise-chan on the out-chan before actually placing the
+   value on the promise-chan is necessary as we want to limit calls to `out-xform` to only when the
+   data is ready to be consumed by `out-chan`."
+  [in-chan in-xform out-chan out-xform &
```
wyegelwel: nitpick: out-xform is not a transform or transducer. Call it something like out-finalizer or out-preparer. @sradack , you are good with names, any opinion?

scheduler/src/cook/mesos/util.clj
shamsimam: Done.

scheduler/src/cook/mesos/util.clj
shamsimam: I used `out-preparer` as finalizer reminds me of Java finalizers :(

shamsimam: Should we make out-preparer optional and default it to `[initial-value state]`?

shamsimam: This is how it would look:
```
(defn reducing-pipe
  [in-chan in-xform out-chan &
   {:keys [initial-state on-consumed on-finished out-preparer]
    :or {initial-state nil, on-consumed identity, on-finished #()}}]
  (let [out-preparer (or out-preparer (fn out-preparer [current-state] [initial-state current-state]))]
    (async/go-loop [state initial-state
                    out-data-chan (async/promise-chan)]
      (let [[data chan] (async/alts! [[out-chan out-data-chan] in-chan] :priority true)]
        (condp = chan
          out-chan (if data
                     (let [[reset-state out-value] (out-preparer state)]
                       (async/put! out-data-chan out-value)
                       (async/close! out-data-chan)
                       (on-consumed out-value)
                       (recur reset-state (async/promise-chan)))
                     (recur state out-data-chan))
          in-chan (if (nil? data)
                    (on-finished)
                    (recur (in-xform state data) out-data-chan)))))))
```

scheduler/src/cook/mesos/util.clj
```diff
@@ -552,3 +552,38 @@
...
+   supplied to the `out-chan` channel.
+
+   Note: This function does not perform error handling, exceptions must be explicitly handled in the
+   provided functions (i.e. in-xform, out-preparer and on-finished). In particular, `out-preparer` must
+   not throw an exception as it will prevent the promise-chan from being fulfilled.
+
+   Note: The roundabout approach of putting a promise-chan on the out-chan before actually placing the
+   value on the promise-chan is necessary as we want to limit calls to `out-preparer` to only when the
+   data is ready to be consumed by `out-chan`."
+  [in-chan in-xform out-chan out-preparer &
```
sradack: Why don't we call `in-xform` `reducer` and get rid of `out-preparer` entirely?  The `out-preparer` is not core to what is happening here.

sradack: While we are simplifying, we don't need an `on-finished` either.  You can just track the `chan` returned by the `go-loop`.  The mix of async and handlers bothers me.

shamsimam: Without the callbacks, this is how it looks:
```
(defn reducing-pipe
  "Reads elements from the `in-chan` channel and supplies elements to the `out-chan` channel.
   Maintains a state (initialized to `initial-state`) that is updated by applying the reducing 
   function `reducer` to the current state and the incoming element `(reducer state element)`.
   When the `out-chan` channel can receive an element, the current state is put on the `out-chan`  
   and the state is reset to `initial-state`.

   Note: This function does not perform error handling, exceptions must be explicitly handled in the
   provided functions (i.e. reducer)."
  [in-chan reducer out-chan & {:keys [initial-state]}]
  (async/go-loop [state initial-state]
    (let [[data chan] (async/alts! [[out-chan state] in-chan] :priority true)]
      (condp = chan
        out-chan (if data
                   (recur initial-state)
                   (recur state))
        in-chan (when (not (nil? data))
                  (recur (reducer state data)))))))
```

shamsimam: @wyegelwel @leifwalsh what do you guys think?

wyegelwel: It is definitely simpler. Having the out-preparer allows it to support more use cases but perhaps we just push that off until we need it. 

Seems fine to me. 

Consider closing the out-chan (and/or providing a close? param) when the in-chan is closed.

shamsimam: Done.

scheduler/src/cook/mesos/util.clj
```diff
@@ -451,10 +451,10 @@
...
-     (map (fn [job]
-            (mapv #(vector (:group/uuid %) #{(:job/uuid job)}) (:group/_job job))))
-     (map (partial into {}))
-     (reduce (partial merge-with clojure.set/union))))
+       (map (fn [job]
+              (mapv #(vector (:group/uuid %) #{(:job/uuid job)}) (:group/_job job))))
+       (map (partial into {}))
+       (reduce (partial merge-with clojure.set/union))))
 
 (defn make-guuid->considerable-cotask-ids
```
leifwalsh: This is not part of your change, so sorry, but "guuid" not "guid"?

shamsimam: No worries, there are multiple references to `uuid` in the code: https://github.com/twosigma/Cook/search?utf8=%E2%9C%93&q=uuid&type=

PR #421, opened 2017-07-21
Add support for cook-executor flag and gradual transitioning to Cook executor

shamsimam:


integration/tests/cook/test_basic.py
```diff
@@ -25,7 +25,7 @@ def setUp(self):
         util.wait_for_cook(self.cook_url)
 
     def test_basic_submit(self):
-        job_uuid, resp = util.submit_job(self.cook_url)
+        job_uuid, resp = util.submit_job(self.cook_url, cook_executor=True)
```
dposada: Why does this test need to set the `cook_executor` flag?

shamsimam: As it checks the `exit_code` and `sandbox_directory`

scheduler/docs/configuration.asc
```diff
@@ -327,6 +328,11 @@ The configuration values are defined as follows:
   Care should be taken to avoid setting this value too low as it can end up causing a high rate of message transfer between the executor and the scheduler.
   The default is (1000 * 60 * 5), i.e. 5 minutes.
 
+`:transition-percent`::
+  The percentage of jobs whose first instances that are automatically chosen to run on the Cook executor instead of Mesos' command executor.
+  The flag enables gradual introduction of the COok executor and minimizes the effects of a faulty Cook executor as the job will be reun using the command executor on retry.
+  The default value is 2.
```
dposada: Why wouldn't the default be 0?

scheduler/docs/scheduler-rest-api.asc
```diff
@@ -69,6 +69,7 @@ The request body is list of job maps with each entry containing a map of the fol
 |JSON Key | Type | Description
 |`uuid` | string | a user provided UUID for tracking and referring to the job.
 |`command` | string | the command to run
+|`cook_executor` | boolean | Flag to force use of the Cook executor, defaults to false.
```
dposada: Instead of `cook_executor=true|false`, should we do `executor=cook|mesos`?

scheduler/docs/configuration.asc
```diff
@@ -327,6 +328,11 @@ The configuration values are defined as follows:
   Care should be taken to avoid setting this value too low as it can end up causing a high rate of message transfer between the executor and the scheduler.
   The default is (1000 * 60 * 5), i.e. 5 minutes.
 
+`:transition-percent`::
+  The percentage of jobs whose first instances that are automatically chosen to run on the Cook executor instead of Mesos' command executor.
```
shamsimam: Fixed.

scheduler/docs/configuration.asc
```diff
@@ -327,6 +328,11 @@ The configuration values are defined as follows:
...
+  The flag enables gradual introduction of the COok executor and minimizes the effects of a faulty Cook executor as the job will be reun using the command executor on retry.
```
shamsimam: Fixed. Thanks.

scheduler/docs/scheduler-rest-api.asc
shamsimam: We already have a `custom-executor` flag, following suite. With an `executor` flag the value should be 'cook|custom|mesos', but it is unlikely that we are going to drop `custom-executor` flag.

dposada: Ah, I see. Should we just latch onto that existing flag, and allow it to be `cook|custom|mesos`?

shamsimam: Unfortunately, that flag is configured to store a boolean value.

scheduler/docs/configuration.asc
```diff
@@ -290,6 +290,7 @@ An example configuration looks like:
             :log-level "INFO"
             :max-message-length 512
             :progress-sample-interval-ms 1000
+            :transition-percent 5
```
shamsimam: Done.

scheduler/docs/configuration.asc
shamsimam: Defaulted to zero.

scheduler/docs/scheduler-rest-api.asc
shamsimam: @pschorf @sradack @wyegelwel how do you feel about adding a new field `executor=cook|mesos` and leaving the `custom-executor` flag for backwards compatibility.

sradack: I like it.  I prefer the `executor` solution.

shamsimam: Done.

shamsimam:
ready for another round of reviews

scheduler/docs/configuration.asc
```diff
@@ -322,6 +323,11 @@ The configuration values are defined as follows:
   The maximum length for the unencoded string messages sent from a task via the Mesos executor HTTP API.
   The default is 512.
 
+`:portion`::
```
sradack: This is nice but unnecessary.  We don't need docs because this isn't a permanent feature.  If it is going to be permanent, the user should be able to define what executor makes up the other portion and all portions should add up to 1.

scheduler/example-prod-config.edn
```diff
@@ -20,6 +20,7 @@
                                 :retry-limit 15
                                 :cpus 20}}
  :executor {:command "./cook-executor"
+            :portion 0.25
```
sradack: Even if portion was permanent, having portion be set in our sample config doesn't well convey that it is used only temporarily.

scheduler/src/cook/mesos/api.clj
```diff
@@ -643,6 +649,8 @@
                    {:container container})
                  (when expected-runtime
                    {:expected-runtime expected-runtime})
+                 (when executor
+                   {:executor  executor})
```
sradack: Extra space

scheduler/src/cook/mesos/task.clj
```diff
@@ -25,6 +25,30 @@
...
+          b. This is going to be the first instance of the job, and
+          c. Our random toss yields less than portion percent."
+  [job-ent executor-config]
+  (and (not (use-custom-executor? job-ent))
+       (:command executor-config)
+       (or (= :job.executor/cook (:job/executor job-ent))
+           (and (nil? (:job/executor job-ent))
+                (zero? (count (:job/instance job-ent)))
+                (when-let [portion (:portion executor-config)]
+                  (> (* portion 100) (rand-int 100)))))))
```
sradack: It would be nice if this was deterministic based upon `job-ent`, e.g. `(<= (-> (:id job-ent) .getLeastSignificantBits (mod 100) (/ 100.0)) portion)`

shamsimam: Used `(-> job-ent :job/uuid hash (mod 100))` as it is easier to test and avoid the division

scheduler/src/cook/mesos/api.clj
shamsimam: Thanks, fixed.

scheduler/example-prod-config.edn
shamsimam: Removed portion from the example edn file.

scheduler/docs/configuration.asc
shamsimam: Removed portion from the docs.

shamsimam:
Any additional feedback on this?

integration/tests/cook/test_basic.py
```diff
@@ -25,8 +25,8 @@ def setUp(self):
         util.wait_for_cook(self.cook_url)
 
     def test_basic_submit(self):
-        job_uuid, resp = util.submit_job(self.cook_url)
-        self.assertEqual(resp.status_code, 201)
+        job_uuid, resp = util.submit_job(self.cook_url, executor='cook')
```
dposada: I'm confused about why we're specifying `executor='cook'` here but then below we check for `util.is_using_cook_executor(self.cook_url)`

shamsimam: If the scheduler is configured to not use the Cook executor, it will still accept `cook` as an executor argument, but will end up using the Mesos command executor to execute the job. Do you want the scheduler to reject a `cook` job if the scheduler is not configured to use it?

dposada: Yeah, that seems like a good idea.

shamsimam: Modified the test instead and added a warning log.

Performing the validation requires threading an additional dependency 4-5 levels deep into `api/validate-and-munge-job`, so avoided it.

shamsimam:
@dposada updated the integration tests to avoid confusion.

shamsimam:
Rebased, ready for another review.

scheduler/src/cook/mesos/schema.clj
```diff
@@ -104,7 +104,18 @@
...
-    :db/doc "Determines if this job uses a custom executor (true) or the cook executor (false).
+    :db/doc "Determines if this job will only use the
+             1. cook executor (cook),
+             2. mesos command executor (mesos), or
+             3. custom (custom executor).
+             When missing and :job/custom-executor is true, then uses a custom executor (for legacy compatibility).
+             Else, it may default to any of the Cook executor or Mesos command executor"
+    :db/ident :job/executor
+    :db/valueType :db.type/ref
+    :db/cardinality :db.cardinality/one
```
wyegelwel: If this is a ref, make set component to true.

scheduler/src/cook/mesos/task.clj
```diff
@@ -25,6 +26,31 @@
...
+   2. The Cook executor command has been configured
+   3. Either :job/executor is explicitly enabled
+      Or: a. Cook executor has not been explicitly disabled,
+          b. This is going to be the first instance of the job, and
+          c. The job UUID hash mod 100 yields less than portion percent."
+  [job-ent executor-config]
+  (and (not (use-custom-executor? job-ent))
+       (:command executor-config)
+       (or (= :job.executor/cook (:job/executor job-ent))
+           (and (nil? (:job/executor job-ent))
```
wyegelwel: Make this condition a function and call it. It deserves a name

scheduler/src/cook/mesos/task.clj
```diff
@@ -67,6 +93,9 @@
                    ;;TODO this data is a race-condition
                    {:instance (str (count (:job/instance job-ent)))}))
                "UTF-8")]
+    (when (and (= :job.executor/cook (:job/executor job-ent))
+               (not= executor-key :cook-executor))
+      (log/warn "Task" task-id "requested to use cook executor, but will be executed using" (name executor-key)))
```
wyegelwel: When would this happen?

scheduler/src/cook/mesos/schema.clj
```diff
@@ -737,7 +748,17 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
...
    {:db/id (d/tempid :db.part/user)
     :db/ident :instance/create
-    :db/doc "Creates an instance for a job"}])
+    :db/doc "Creates an instance for a job"}
+   ;; Enum of executor options
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :job.executor/cook
+    :db/doc "Signals intent to use the Cook executor"}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :job.executor/custom
```
wyegelwel: This doesn't appear to be used

scheduler/src/cook/mesos/schema.clj
```diff
@@ -737,7 +748,17 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
...
+    :db/doc "Signals intent to use the custom executor"}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :job.executor/mesos
```
wyegelwel: This also doesn't appear to be used

scheduler/src/cook/mesos/task.clj
```diff
@@ -41,8 +67,8 @@
         container (util/job-ent->container db job-ent)
         ;; If the custom-executor attr isn't set, we default to using a custom
         ;; executor in order to support jobs submitted before we added this field
-        custom-executor? (:job/custom-executor job-ent true)
-        cook-executor? (and (not custom-executor?) (seq executor-config))
+        custom-executor? (use-custom-executor? job-ent)
+        cook-executor? (use-cook-executor? job-ent executor-config)
```
wyegelwel: We should note which executor the instance used if it can be different between instances

scheduler/src/cook/mesos/schema.clj
wyegelwel: I'm generally confused on the use of these.

integration/tests/cook/test_basic.py
```diff
@@ -25,42 +25,59 @@ def setUp(self):
         util.wait_for_cook(self.cook_url)
 
     def test_basic_submit(self):
-        job_uuid, resp = util.submit_job(self.cook_url)
-        self.assertEqual(resp.status_code, 201)
+        is_using_cook_executor = util.is_using_cook_executor(self.cook_url)
```
pschorf: This is repeated a few times: why not have the util method return 'cook' or 'mesos' instead of the boolean?

scheduler/src/cook/mesos/api.clj
```diff
@@ -474,6 +475,10 @@
                                   :constraint/pattern pattern}]))
                             constraints)
         container (if (nil? container) [] (build-container user db-id container))
+        executor (when executor
+                   (case executor ;; will throw an exception if invalid value is provided
```
pschorf: This would be a pretty rare occurrence, right? If the job has a bad value the schema validation should fail.

scheduler/src/cook/mesos/task.clj
pschorf: 👍

scheduler/src/cook/mesos/task.clj
shamsimam: Added the executor to the job instance entity in `cook.mesos.scheduler/matches->task-txns`

scheduler/src/cook/mesos/api.clj
```diff
@@ -485,7 +486,13 @@
                                   :constraint/pattern pattern}]))
                             constraints)
         container (if (nil? container) [] (build-container user db-id container))
-        ;; These are optionally set datoms w/ default values
+        executor (when executor
```
wyegelwel: nitpick: Perhaps make this a function

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -730,41 +730,45 @@
                                 (remove #(contains? matched-job-uuids (:job/uuid %)) existing-jobs)))]
     (pc/map-from-keys remove-matched-jobs (keys category->pending-jobs))))
 
+(defn- update-match-with-task-metadata-seq
+  "Updates the match with an entry for the task metadata for all tasks."
+  [{:keys [tasks] :as match} db framework-id executor-config]
+  (let [task-metadata-seq (map #(task/TaskAssignmentResult->task-metadata db framework-id executor-config %)
```
wyegelwel: There is a lot going on in this expression. Can you switch to use threading?

```clojure
(->> tasks
       (sort-by (comp :db/id :job #(.getRequest ^TaskAssignmentResult %)))
       (map (partial task/TaskAssignmentResult->task-metadata db framework-id executor-config))
```

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -730,41 +730,45 @@
                                 (remove #(contains? matched-job-uuids (:job/uuid %)) existing-jobs)))]
     (pc/map-from-keys remove-matched-jobs (keys category->pending-jobs))))
 
+(defn- update-match-with-task-metadata-seq
+  "Updates the match with an entry for the task metadata for all tasks."
+  [{:keys [tasks] :as match} db framework-id executor-config]
```
wyegelwel: nitpick: s/match/matches/ in this function to be consistent with rest of code

wyegelwel: Never mind, was confused.

scheduler/src/cook/mesos/task.clj
```diff
@@ -60,17 +91,26 @@
...
         data (.getBytes
                (if cook-executor?
                  (json/write-str {"command" (:job/command job-ent)})
                  (pr-str
                    ;;TODO this data is a race-condition
                    {:instance (str (count (:job/instance job-ent)))}))
                "UTF-8")]
+    (when (and (= :executor/cook (:job/executor job-ent))
+               (not= executor-key :cook-executor))
+      (log/warn "Task" task-id "requested to use cook executor, but will be executed using" (name executor-key)))
```
wyegelwel: When would this happen? If they requested to use cook executor but it is not configured?

wyegelwel:
Bunch of small comments otherwise lgtm

scheduler/src/cook/mesos/api.clj
shamsimam: Done, created `str->executor-enum`.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Done, used an anonymous function instead of `partial` as we know exactly how many arguments we will provide.

scheduler/src/cook/mesos/task.clj
shamsimam: Yes, when they requested to use cook executor but it is not configured.

shamsimam:
@sradack @dposada @pschorf any additional feedback?

scheduler/src/cook/mesos/task.clj
wyegelwel: I wish there was a way to inform users why we switched on them. I don't think it is worth holding up this PR, but something like a log on each job of choices made that are "unexpected" so that a user can get some feedback on what happened.

wyegelwel:
> Done, used an anonymous function instead of partial as we know exactly how many arguments we will provide.

Its funny, I use partial not to allow for variable (or unknown) number of arguments but to reduce the cognitive load of reading the code. `(partial f ...)` can only mean that I'm calling with some specified parameters and then potentially some more specified later. `#(...)` can mean anything so I need to read it carefully to see what it is doing. It is the same reason I prefer to avoid reduce when possible.

scheduler/src/cook/mesos/task.clj
shamsimam: Agree.

integration/tests/cook/test_basic.py
```diff
@@ -25,42 +25,56 @@ def setUp(self):
         util.wait_for_cook(self.cook_url)
 
     def test_basic_submit(self):
-        job_uuid, resp = util.submit_job(self.cook_url)
-        self.assertEqual(resp.status_code, 201)
+        job_executor_type = util.get_job_executor_type(self.cook_url)
+        job_uuid, resp = util.submit_job(self.cook_url, executor=job_executor_type)
```
dposada: Why do we need to set the `executor` to the value that's already configured?

shamsimam: To enable the test on line 35 to be triggered.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Changed to use `partial`.

shamsimam:
Ready for more feedback on this :)

integration/tests/cook/test_basic.py
dposada: Should we make `util.submit_job` do this by default so that we don't have to modify so many integration tests?

scheduler/src/cook/mesos/api.clj
```diff
@@ -485,7 +496,9 @@
                                   :constraint/pattern pattern}]))
                             constraints)
         container (if (nil? container) [] (build-container user db-id container))
-        ;; These are optionally set datoms w/ default values
+        executor (str->executor-enum executor)
+        ;; These are optionally set datoms w/ default val
+        ;; ues
```
dposada: Word broken across two lines

integration/travis/scheduler_config.edn
```diff
@@ -19,7 +19,8 @@
          :leader-path "/cook-scheduler"
          :role "cook"
          :framework-id #config/env "COOK_FRAMEWORK_ID"}
- :executor {:command #config/env "COOK_EXECUTOR_COMMAND"}
+ :executor {:command #config/env "COOK_EXECUTOR_COMMAND"
```
dposada: Should this be `:cook-executor` instead of just `:executor`? These configuration options (especially `:portion`) only apply if you're using the cook executor, correct? Maybe that should be more obvious from the name

scheduler/src/cook/mesos/task.clj
dposada: Have we created an issue for @wyegelwel's suggestion? I'm happy to create one if not

scheduler/test/cook/test/simulator.clj
```diff
@@ -105,12 +105,14 @@
...
-                           :progress-sample-interval-ms 1000
-                           :uri {:cache true
+                    :default-progress-output-file "stdout"
+                    :default-progress-regex-string "regex-string"
+                    :log-level "INFO"
+                    :max-message-length 512
+                    :portion 0.25
+                    :progress-sample-interval-ms 1000
+
+                    :uri {:cache true
```
dposada: Indentation seems off on these few lines

integration/tests/cook/test_basic.py
```diff
@@ -25,42 +25,56 @@ def setUp(self):
...
+        job_uuid, resp = util.submit_job(self.cook_url, executor='cook')
+        self.assertEqual(201, resp.status_code, msg=resp.content)
+        job = util.load_job(self.cook_url, job_uuid)
+        self.assertEqual('cook', job['executor'])
+
+        job_uuid, resp = util.submit_job(self.cook_url, executor='mesos')
+        self.assertEqual(201, resp.status_code, msg=resp.content)
+        job = util.load_job(self.cook_url, job_uuid)
+        self.assertEqual('mesos', job['executor'])
+
```
dposada: Can we add integration tests that cover:

- only use cook executor on the first instance of a job
- the portion logic, say portion = 0.5, start several jobs, assert that the executor was chosen as expected

shamsimam: Added:
 - integration test for no cook executor on non-first instances of a job [`test_no_cook_executor_on_subsequent_instances`]
 - unit test to verify no cook executor on non-first instances of a job despite favorable coin toss

integration/tests/cook/test_basic.py
dposada: It looks like 6 tests had to add 

```python
job_executor_type = util.get_job_executor_type(self.cook_url)
..., executor=job_executor_type)
```

Will new tests need this as well? I'm wondering if there is a way to avoid having this in several tests

integration/tests/cook/test_basic.py
dposada: What about integration testing that the coin toss was correct based on the uuid?

integration/tests/cook/test_basic.py
shamsimam: Only tests that care about using the cook executor. All of the tests that specify it, also have a check later for a few assertions based on `if job_executor_type == 'cook':`

integration/tests/cook/test_basic.py
shamsimam: The challenge is that the hash computation is in Clojure and it is hard to mimic using Python. I tried using the Murmur3 python implementation mmh3, but it still produced different results than Clojure's hash function (Murmur3.hashInt). Instead, I have now added assertions in the unit tests that verify that the job uuid is being used in the hash computations.

integration/tests/cook/test_basic.py
dposada: OK

integration/tests/cook/test_basic.py
dposada: OK

shamsimam:
@pschorf @sradack any more feedback?

PR #425, opened 2017-08-01
Adds support for partial results when querying for groups

dposada:


scheduler/src/cook/mesos/api.clj
```diff
@@ -1154,13 +1161,17 @@
...
-                           not-found-guuids (remove #(group-exists? (db conn) %) requested-guuids)]
-                       (if (empty? not-found-guuids)
-                         [false {::guuids requested-guuids}]
+                           allow-partial-results (get-in ctx [:request :query-params :partial])
+                           exists? #(group-exists? (db conn) %)
+                           existing-groups (filter exists? requested-guuids)]
+                       (if (or
+                             (= (count existing-groups) (count requested-guuids))
+                             (and allow-partial-results (pos? (count existing-groups))))
+                         [false {::guuids existing-groups}]
```
shamsimam: Nitpic: the docstring should document the meaning of the returned tuple.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1154,13 +1161,17 @@
                      (let [requested-guuids (->> (get-in ctx [:request :query-params "uuid"])
                                                  vectorize
                                                  (mapv #(UUID/fromString %)))
-                           not-found-guuids (remove #(group-exists? (db conn) %) requested-guuids)]
-                       (if (empty? not-found-guuids)
-                         [false {::guuids requested-guuids}]
+                           allow-partial-results (get-in ctx [:request :query-params :partial])
+                           exists? #(group-exists? (db conn) %)
+                           existing-groups (filter exists? requested-guuids)]
+                       (if (or
```
shamsimam: Nitpick: Can we have the first clause in the same line as the `or`?

scheduler/src/cook/mesos/api.clj
dposada: I moved the code to a new function, `retrieve-groups`, and added a docstring there

scheduler/src/cook/mesos/api.clj
dposada: Done

scheduler/src/cook/mesos/api.clj
```diff
@@ -1151,30 +1151,46 @@
...
+                               (set/difference (set requested-guuids) (set existing-groups)))
+                             " didn't correspond to a group")}]))
+    (catch Exception e
+      [false {::error e}])))
 
 (defn read-groups-handler
   [conn task-constraints is-authorized-fn]
   (base-cook-handler
     {:allowed-methods [:get]
-     :malformed? (fn [ctx]
```
shamsimam: `read-jobs-handler` has a `:malformed?` to validate the parameters.

dposada: The schema is already validating the parameters for us:

```bash
$ curl -v localhost:12321/group
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 12321 (#0)
> GET /group HTTP/1.1
> Host: localhost:12321
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< Cache-Control: max-age=0
< Transfer-Encoding: chunked
< Server: Jetty(9.2.11.v20150529)
< 
* Connection #0 to host localhost left intact
{"errors":{"uuid":"missing-required-key"}}
```

shamsimam: okay

PR #427, opened 2017-08-03
Adds authorization function that supports user whitelisting

dposada:


scheduler/src/cook/authorization.clj
```diff
@@ -144,6 +145,47 @@
                      "is not allowed to perform" verb "on" (str object ",") "denying.")
            false))))
 
+(defn admins-open-gets-auth
```
wyegelwel: Make it clear this also includes checking a whitelist

scheduler/src/cook/authorization.clj
```diff
@@ -144,6 +145,47 @@
...
+  "Allows admins to do anything to any object, and if whitelisted-user is truthy,
+  allows non-admins to manipulate objects that they own and to read (get) any object."
+  [admins user verb {:keys [owner] :as object} whitelisted-user]
```
wyegelwel: `whitelisted-user` implies it contains a user name. Rename it to something like `user-passed-whitelist?`

scheduler/src/cook/authorization.clj
dposada: Done

scheduler/src/cook/authorization.clj
dposada: Renamed to `user-in-whitelist?`

dposada:
@wyegelwel Any other feedback on this?

PR #428, opened 2017-08-04
Use sort instead of sorted-set to speed ranking

wyegelwel:


wyegelwel:
A small test in qa showed this reduced the sort-jobs-hierarchy-duration by ~50%

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1201,7 +1201,7 @@
                sort-jobs-duration
                (->> tasks
                     (group-by util/task-ent->user)
-                    (pc/map-vals (fn [task-ents] (into (sorted-set-by task-comparator) task-ents)))
+                    (pc/map-vals (fn [task-ents] (sort task-comparator task-ents)))
```
shamsimam: Do we need the `set` behavior, i.e. is it possible for there to be duplicates in `task-ents`?
If there can be duplicates, maybe we first create a set and then do the sort?

wyegelwel: It should be impossible to have duplicates since the task ents are coming from the db where they are unique or generated for jobs where there is a one to one mapping from job to task and the job is unique in the database.

wyegelwel:
Using the benchmarking tests from #430, this functions goes from ~900ms to ~450ms

PR #430, opened 2017-08-05
Initial benchmarking tests

wyegelwel:


wyegelwel:
Note, the benchmarking tests currently take around 4 minutes so when running `lein test :all` we should expect to see them take another 4 minutes. There are ways to run [tests in parallel](https://github.com/aredington/parallel-test) if we find this to be too painful.

sradack:
Why are we creating unit tests for benchmarks?  Are we eventually going to put assertions in them?

wyegelwel:
No I just wanted a convenient framework to run them. There doesn't seem to
be any great choices in the Clojure ecosystem* so I went with abusing the
testing tool.

*There is a project perforate that hasn't been updated in 3 years that
kinda does what I want but seems to have a bit of a rigid model

PR #432, opened 2017-08-09
Incremental update functions for ranking

wyegelwel:
This is the start of code to maintain state needed to compute the ranking of jobs. This allows us to incrementally rank jobs instead of re-doing the computation over again. 

This PR provides functions to update a rank-state map and code to compute the ranking of jobs from the rank-state though both need to be cleaned up and tested.

This PR does not include the code to call the functions to update the rank-state. I propose making that a separate PR.

### DIFF
--- a/scheduler/src/cook/mesos/ranker.clj
+++ b/scheduler/src/cook/mesos/ranker.clj
@@ -1,0 +1,75 @@
+;;
+;; Copyright (c) Two Sigma Open Source, LLC
+;;
+;; Licensed under the Apache License, Version 2.0 (the "License");
+;; you may not use this file except in compliance with the License.
+;; You may obtain a copy of the License at
+;;
+;;  http://www.apache.org/licenses/LICENSE-2.0
+;;
+;; Unless required by applicable law or agreed to in writing, software
+;; distributed under the License is distributed on an "AS IS" BASIS,
+;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
+;;
+(ns cook.mesos.ranker
+  (:require [cook.mesos.dru :as dru]
+            [cook.mesos.share :as share]
+            [cook.mesos.util :as util]
+            [plumbing.core :as pc]))
+
+;; TODO: Look into data.avl to store sets and use transients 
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/mesos/ranker.clj
```diff
@@ -0,0 +1,75 @@
...
+                              (= (get-in task [:job/_instance :job/state]) :job.state/waiting)))
+                    (map (fn get-job [[task _]] 
+                           (:job/_instance task))))]))]
+    (into {} (map sort-jobs-by-dru-category-helper) category->user->sorted-tasks)))
+
+(defn rank-jobs
+  "Returns map where keys are categories of jobs and vals are seq of jobs waiting
+   to be scheduled in order of Cook's desire to schedule it"
+  [db rank-state]
+  (let [user->dru-divisors (share/create-user->share-fn db)]
```
sradack: @pschorf recently removed `db` references from entities from Fenzo state, which dramatically reduced the amount of memory needed.  Can we store maps instead and avoid storing entities in `rank-state`?

sradack: Generally, we should be avoiding storing references to `db` or entities in all places that keep state, and keeping references to them only for the amount of time it takes for a unit of work/transaction.

scheduler/test/cook/test/benchmark.clj
```diff
@@ -67,3 +70,25 @@
                                                          sort-task-scored-task-pairs
                                                          sched/sort-jobs-hierarchy-duration))
           nil)))))
+
+(deftest ^:benchmark bench-rank-jobs-state
```
sradack: How did the benchmark perform?

scheduler/src/cook/mesos/ranker.clj
pschorf: `user->dru-divisors` has a cache which returns PersistentArrayMap

scheduler/test/cook/test/benchmark.clj
pschorf: From the travis logs, this was 5 ms vs 2 seconds for the other iteration of rank-jobs

pschorf:
@sradack the inputs to these apis can be generated from jobs using `util/create-task-ent` which is not actually an entity but a map.

PR #433, opened 2017-08-09
Optimize is-offensive?

pschorf:


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1247,10 +1247,18 @@
 
 (defn is-offensive?
   [max-memory-mb max-cpus job]
-  (let [{memory-mb :mem
-         cpus :cpus} (util/job-ent->resources job)]
-    (or (> memory-mb max-memory-mb)
-        (> cpus max-cpus))))
+  (loop [resources (:job/resource job)]
```
sradack: How much of a speed up is this?

PR #438, opened 2017-08-22
Make mesos executor registration timeout a mea-culpa failure

pschorf:


wyegelwel:
Can you take a quick pass and see if there are others that could be set to mea-culpa? Originally we were conservative on which would be mea-culpa because they would be retried indefinitely. We have since added a limit so it is less of a problem.

pschorf:
I made 3 more changes, take a look and see what you think

wyegelwel:
Test failed. I'm assuming it is flaky. Restarted it, hopefully it will pass this time.

PR #439, opened 2017-08-22
Comments out two flaky assertions

dposada:


dposada:
@pschorf This is good to go now

PR #443, opened 2017-08-25
Adds support for filtering by name in /list

dposada:


integration/tests/cook/test_basic.py
```diff
@@ -59,7 +59,7 @@ def test_failing_submit(self):
 
     def test_progress_update_submit(self):
         command = 'echo "progress: 25 Twenty-five percent"; sleep 1; exit 0'
-        job_uuid, resp = util.submit_job(self.cook_url, command=command, max_runtime=5000)
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, max_runtime=60000)
```
pschorf: Why did this need to change? 12x seems like a big bump

scheduler/src/cook/mesos/api.clj
```diff
@@ -1585,25 +1609,35 @@
...
-                                    ::user user
-                                    ::since-hours-ago (parse-int-default since-hours-ago 24)
-                                    ::start-ms (parse-long-default start-ms nil)
-                                    ::limit (parse-int-default limit Integer/MAX_VALUE)
-                                    ::end-ms (parse-long-default end-ms (System/currentTimeMillis))}]
-                            (catch NumberFormatException e
-                              [true {::error (.toString e)}]))
-                          [true {::error (str "unsupported state in " state ", must be one of: " allowed-list-states)}]))
-                      [true {::error "must supply the state and the user name"}])))
+                                             (get-in ctx [:request :body-params])))
```
pschorf: 👏

scheduler/src/cook/mesos/api.clj
```diff
@@ -1576,6 +1580,26 @@
...
+  [name]
+  (re-pattern (-> name
+                  (str/replace #"\." "\\\\.")
+                  (str/replace #"\*+" ".*"))))
+
+(defn name-filter-str->name-filter-fn
+  "Returns a name-filtering function (or nil) given a user-provided name filter string"
+  [name]
+  (if name
+    #(re-matches (name-filter-str->name-filter-pattern name) %)
```
pschorf: Not sure if this actually matters but would it make a difference to compile the regex once and re-use it?

integration/tests/cook/test_basic.py
dposada: On Travis, the test was failing due to "task max runtime exceeded"

scheduler/src/cook/mesos/api.clj
dposada: Good idea, done

dposada:
Build is green now

PR #444, opened 2017-08-25
A job can only be a straggler once

wyegelwel:
Currently straggler handling will kill an instance if it passes the straggler test regardless of if the job has been considered a straggler before. This is problematic for jobs that actually *do* run much longer than all the others in the batch. In order to let the job progress, a job will now only be able to have a single straggler.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -497,24 +497,32 @@ class CookTest(unittest.TestCase):
                 'multiplier': 2.0
             }
         }
-        slow_job_wait = 1200
+        settings_timeout_interval_minutes = util.get_in(util.settings(self.cook_url), 'task-constraints',
+                                                        'timeout-interval-minutes')
+        # the value needs to be a little more than 2 times settings_timeout_interval_minutes to allow
+        # at least two runs of the straggler handler loop
+        slow_job_wait = (2 * settings_timeout_interval_minutes * 60) + 30
+
         group_spec = self.minimal_group(straggler_handling=straggler_handling)
         job_fast = util.minimal_job(group=group_spec["uuid"])
         job_slow = util.minimal_job(group=group_spec["uuid"],
-                                    command='sleep %d' % slow_job_wait)
+                                    command='sleep %d' % slow_job_wait,
+                                    max_retries=2)
         data = {'jobs': [job_fast, job_slow], 'groups': [group_spec]}
         resp = util.session.post('%s/rawscheduler' % self.cook_url, json=data)
         self.assertEqual(resp.status_code, 201)
         util.wait_for_job(self.cook_url, job_fast['uuid'], 'completed')
+        # Need to wait for initial job to be killed, then restarted and run to completion
         util.wait_for_job(self.cook_url, job_slow['uuid'], 'completed',
-                          slow_job_wait * 1000)
+                          3*slow_job_wait * 1000)
         jobs = util.session.get('%s/rawscheduler?job=%s&job=%s' %
                                 (self.cook_url, job_fast['uuid'], job_slow['uuid']))
         self.assertEqual(200, jobs.status_code)
         jobs = jobs.json()
         self.logger.debug('Loaded jobs %s', jobs)
         self.assertEqual('success', jobs[0]['state'], 'Job details: %s' % (json.dumps(jobs[0], sort_keys=True)))
-        self.assertEqual('failed', jobs[1]['state'])
+        self.assertEqual('success', jobs[1]['state'])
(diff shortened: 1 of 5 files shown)


scheduler/docs/groups.md
```diff
@@ -42,13 +42,15 @@ When running a lot of jobs, occasionally it happens that one or two of them run
...
 these problem cases, Cook can be configured to kill and restart jobs in a group that have been
-running much longer then the other jobs in the group. 
+running much longer then the other jobs in the group.
 
 To configure this, set the straggler handling type to `quantile-deviation`. This will have Cook look
 for jobs that have been running multiple times longer than a particular quantile of the group. The
 parameters, `quantile` and `multiplier` decide when a job is considered a straggler. The
 `quantile` decides the target runtime of the group and the `multipier` decides how many times
 longer than the run time of the quantile job before a job is marked as a straggler.
+If a job has already had an instance killed for being as straggler subsequent instances will not be
```
dposada: "a straggler" instead of "as straggler", and I think we should add a comma after "straggler" as well

scheduler/docs/groups.md
```diff
@@ -42,13 +42,15 @@ When running a lot of jobs, occasionally it happens that one or two of them run
...
+considered a straggler regardless of how long it has run for.
```
dposada: you're using singular here, but plural in "subsequent instances"

scheduler/test/cook/test/mesos/group.clj
```diff
@@ -90,6 +90,60 @@
             straggler-task-a (d/entity db straggler-a)
             straggler-task-b (d/entity db straggler-b)]
         (is (= #{straggler-task-a straggler-task-b} (set (group/find-stragglers group-ent))))))
+    (testing "quantile deviation straggler-handling, stragglers found but was already a straggler"
```
dposada: Nitpick: I would favor a newline before the `testing`

wyegelwel:
@dposada I responded to your comments on the code. I can add an integration test but it would likely mean it would be the slowest integration test or flaky. The reason is that we would need to wait for the first job in the group to complete, the other job to run longer enough to be treated as a straggler and then complete because it isn't killed for being a straggler. The problem is we need to set the run time of the straggling job to be long enough that straggler handling has a chance to kill it but short enough that we are willing to wait for it to complete after being killed the first time.

What do you think? I can try to find reasonable settings here but they may not work well for a different environment.

dposada:
Is the straggler handling loop interval configured in settings? If so, would it make sense to write the test such that it fetches that value dynamically and adjusts the job running times accordingly? I think we do something similar to that in one of the other integration tests

wyegelwel:
Ok, I updated the straggler handling integration test to include the new behavior.

integration/tests/cook/test_basic.py
```diff
@@ -390,24 +390,32 @@ def test_straggler_handling(self):
...
+                                    command='sleep %d' % slow_job_wait,
+                                    max_retries=2)
         data = {'jobs': [job_fast, job_slow], 'groups': [group_spec]}
         resp = util.session.post('%s/rawscheduler' % self.cook_url, json=data)
         self.assertEqual(resp.status_code, 201)
         util.wait_for_job(self.cook_url, job_fast['uuid'], 'completed')
+        # Need to wait for initial job to be killed, then restarted and run to completion
         util.wait_for_job(self.cook_url, job_slow['uuid'], 'completed',
-                          slow_job_wait * 1000)
+                          3*slow_job_wait * 1000)
```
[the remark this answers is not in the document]
wyegelwel: Unfortunately not; that setting is also the interval at which stragglers are detected and restarted in cook so the test needs the job to live long enough for straggler detection to occur.

scheduler/test/cook/test/mesos/group.clj
```diff
@@ -85,6 +85,14 @@
             job-straggler-b (create-dummy-job conn :group group-ent-id)
             straggler-b (create-dummy-instance conn job-straggler-b :instance-status :instance.status/running
                                              :start-time (tc/to-date (t/ago (t/minutes 121))))
+            ;; Not treated as a straggler because instance already failed as a straggler
```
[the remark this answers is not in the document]
wyegelwel: I think that is effectively what is happening because job-straggler-b is detected as a straggler while job-straggler-c is configured to already have a straggler instance so the current instance is not killed.

shamsimam:
Needs rebase.

shamsimam:
@wyegelwel can you please rebase

dposada:
@wyegelwel Should we close this?

dposada:
@wyegelwel Can we close this?

PR #445, opened 2017-08-29
Load job UUIDs before checking auth

pschorf:


scheduler/src/cook/mesos/api.clj
```diff
@@ -870,6 +870,12 @@
        :job/_instance
        :job/uuid))
 
+(defn- wrap-seq [v]
```
dposada: Can we add a docstring?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1145,6 +1145,35 @@
...
+          is-authorized-fn (partial auth/is-authorized? {:authorization-fn 'cook.authorization/configfile-admins-auth-open-gets})
+          agent-query-cache (Object.)
+
+          handler (api/destroy-jobs-handler conn framework-id is-authorized-fn agent-query-cache)
+          {:keys [uuid] :as job} (assoc (minimal-job) :user "creator")
+          _ (api/create-jobs! conn {::api/jobs [job]})
+          resp (handler {:request-method :delete
+                         :authorization/user "destroyer"
+                         :query-params {:job uuid}})]
+      (is (= 403 (:status resp))))))
```
dposada: Almost all the code is duplicated between the two `testing` clauses, should we extract some of it?

scheduler/src/cook/mesos/api.clj
pschorf: Done

scheduler/test/cook/test/mesos/api.clj
pschorf: Done

pschorf:
Currently, any user can delete any other user's jobs (since ::jobs is nil, no jobs are checked for auth)

dposada:
Should we add an integration test? It would require the multiple user support we discussed a few days ago, right? Is that overkill?

pschorf:
I've tested it out internally and it works, I'd vote to punt on the integration test for now and add it once we add multi-user support.

dposada:
SGTM, let's create an issue for it then

PR #446, opened 2017-09-07
Create the stdout and stderr files in the sandbox directory

shamsimam:


wyegelwel:
This seems fine to me

PR #447, opened 2017-09-07
Simplify Python3 support in tests

shamsimam:


.travis.yml
```diff
@@ -19,7 +20,7 @@ env:
     - MESOS_NATIVE_JAVA_LIBRARY=/usr/lib/libmesos.so
     - CLJ_HTTP_ASYNC_POOL_TEST_DURATION_MULTIPLIER=5
   matrix:
-    - COOK_EXECUTOR='1' TEST_DIR=executor DEPS_CMD='python3 setup.py install' TEST_CMD='python3 setup.py nosetests'
+    - COOK_EXECUTOR='1' TEST_DIR=executor DEPS_CMD='pip --version && pip install virtualenv && virtualenv --version && virtualenv py35 && ls -la py35/bin && source py35/bin/activate py35 && python setup.py install' TEST_CMD='python setup.py nosetests'
```
DaoWen: Does the `py35` argument in `source py35/bin/activate py35` do something? I thought that the `activate` script didn't accept any options.

executor/tests/utils.py
```diff
@@ -10,6 +10,13 @@ def get_random_task_id():
     return str(random.randint(100000, 999999))
 
 
+def ensure_directory(output_filename):
```
dposada: Missing docstring

executor/travis/setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Sets up the travis worker to be able to run executor tests.
+
+export PROJECT_DIR=`pwd`
+cd ${PROJECT_DIR}
+
+pip install virtualenv
+virtualenv -p python3 py35
```
dposada: Why do we need to use virtualenv for this?

shamsimam: `python setup.py install` will attempt to download and install modules, we dont have write access to the standard python installation.

.travis.yml
shamsimam: You're right, removed the unnecessary argument.

executor/tests/utils.py
shamsimam: Added docstring.

shamsimam:
Travis recently updated their build environments (https://docs.travis-ci.com/user/build-environment-updates/2017-09-06/) as a result we were having issues with the use of Python 3 in our build scripts. This PR explicitly installs the Python3 dependencies and ensures the executor build runs successfully.

executor/travis/setup.sh
dposada: 👍. Nitpick: I think I would favor `venv` for the virtual env directory

dposada:
@DaoWen do you have any more feedback on this? Would be nice to get this merged so other PRs can take advantage of it

DaoWen:
I'm pretty sure there's a simpler way to fix this, but I need to double-check the documentation. Please hold off on merging this for now.

DaoWen:
OK, the change with `pyenv` looks much cleaner.

PR #449, opened 2017-09-11
Fix javac issue with Cook scheduler docker container

DaoWen:
Avoid running `javac` to compile `SimpleAssignmentResult.java`, which is only used for unit tests.

Also added the `--rm` option to our `docker run` scripts to auto-remove the instances on exit. (That way you shouldn't have to manually run `docker rm` to clean up the clutter quite so often.)

Also bumped the `mesosphere/mesos` docker image used as the base for our Cook Scheduler container up to a more recent stable version (we were using a release candidate image before).

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -162,7 +162,12 @@
     :source-paths []}
 
    :test-console
-   {:jvm-opts ["-Dcook.test.logging.console"]}}
+   {:jvm-opts ["-Dcook.test.logging.console"]}
+
+   :docker
+   ; avoid calling javac in docker
+   ; (.java sources are only used for unit test support)
+   {:java-source-paths ^:replace []}}
 
   :plugins [[lein-print "0.1.0"]]
 
--- a/integration/bin/run-integration.sh
+++ b/integration/bin/run-integration.sh
@@ -34,6 +34,7 @@ then
 fi
 
 docker run \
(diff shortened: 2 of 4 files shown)


scheduler/project.clj
```diff
@@ -162,7 +162,10 @@
     :source-paths []}
 
    :test-console
-   {:jvm-opts ["-Dcook.test.logging.console"]}}
+   {:jvm-opts ["-Dcook.test.logging.console"]}
+   :docker
```
dposada: Nitpick: I would favor a newline before the new profile

scheduler/project.clj
```diff
@@ -162,7 +162,10 @@
...
+   ; avoid calling javac in docker
```
dposada: Can we add onto this comment that the reason this is OK is that the Java file is only used by unit tests?

DaoWen:
@dposada - I added that missing newline and clarified the comment.

dposada:
LGTM. @pschorf?

wyegelwel:
Could we solve this by adding javac to the container as well or is there a
deeper problem?

dposada:
We could solve it by adding javac to the container (or by requiring that the java file gets compiled before building the docker image), but it seemed cleaner to exclude the java file in the container since it's only used by unit tests.

Issue #226, opened 2017-01-18
Provide a way to kill and restart a group of jobs


wyegelwel:
Given that we have a way to group jobs, provide retry and kill semantics on the group.

dposada:
@DaoWen isn't this complete?

PR #452, opened 2017-09-15
Allow killing a group of jobs

DaoWen:
See issue #226

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -10,9 +10,10 @@ from retrying import retry
 
 from tests.cook import util
 
-
 class CookTest(unittest.TestCase):
     _multiprocess_can_split_ = True
+
+    CMD_NON_ZERO_EXIT_REASON = 99003
 
     @staticmethod
     def minimal_group(**kwargs):
@@ -568,17 +569,17 @@ class CookTest(unittest.TestCase):
                 'multiplier': 2.0
             }
         }
-        slow_job_wait = 1200
+        slow_job_wait_seconds = 1200
         group_spec = self.minimal_group(straggler_handling=straggler_handling)
         job_fast = util.minimal_job(group=group_spec["uuid"])
         job_slow = util.minimal_job(group=group_spec["uuid"],
-                                    command='sleep %d' % slow_job_wait)
+                                    command='sleep %d' % slow_job_wait_seconds)
         data = {'jobs': [job_fast, job_slow], 'groups': [group_spec]}
         resp = util.session.post('%s/rawscheduler' % self.cook_url, json=data)
         self.assertEqual(resp.status_code, 201)
         util.wait_for_job(self.cook_url, job_fast['uuid'], 'completed')
         util.wait_for_job(self.cook_url, job_slow['uuid'], 'completed',
-                          slow_job_wait * 1000)
+                          slow_job_wait_seconds * 1000)
         jobs = util.query_jobs(self.cook_url, True, job=[job_fast, job_slow]).json()
         self.logger.debug('Loaded jobs %s', jobs)
         self.assertEqual('success', jobs[0]['state'], 'Job details: %s' % (json.dumps(jobs[0], sort_keys=True)))
@@ -768,6 +769,60 @@ class CookTest(unittest.TestCase):
         group_info = resp.json()[0]
         for k in detail_keys:
             self.assertIn(k, group_info)
+
+    def test_group_kill_simple(self):
+        # Create and submit jobs in group
+        slow_job_wait_seconds = 1200
+        group_spec = self.minimal_group()
+        group_uuid = group_spec['uuid']
+        job_fast = util.minimal_job(group=group_uuid)
+        job_slow = util.minimal_job(group=group_uuid, command=f'sleep {slow_job_wait_seconds}')
+        _, resp = util.submit_jobs(self.cook_url, [job_fast, job_slow], groups=[group_spec])
+        self.assertEqual(resp.status_code, 201)
+        # Wait for the fast job to finish
+        util.wait_for_job(self.cook_url, job_fast, 'completed')
+        # Now try to cancel the group (just the long job)
+        util.kill_groups(self.cook_url, [group_uuid])
+        # Wait for the slow job (and its instance) to die
+        query = lambda: util.query_jobs(self.cook_url, True, job=[job_slow])
+        util.wait_until(query, util.all_instances_killed)
+        # The fast job should have Success, slow job Failed (because we killed it)
+        jobs = util.query_jobs(self.cook_url, True, job=[job_fast, job_slow]).json()
+        self.assertEqual('success', jobs[0]['state'], f"Job details: {json.dumps(jobs[0], sort_keys=True)}")
+        self.assertEqual('failed',  jobs[1]['state'], f"Job details: {json.dumps(jobs[1], sort_keys=True)}")
+        self.assertEqual(CookTest.CMD_NON_ZERO_EXIT_REASON, jobs[1]['instances'][0]['reason_code'])
+        # Now try to kill the group again
+        # (ensure it still works when there are no live jobs)
+        util.kill_groups(self.cook_url, [group_uuid])
+
+    def test_group_kill_multi(self):
+        # Create and submit jobs in group
+        slow_job_wait_seconds = 1200
+        group_spec = self.minimal_group()
+        group_uuid = group_spec['uuid']
+        job_spec = {'group': group_uuid, 'command': f'sleep {slow_job_wait_seconds}'}
+        jobs, resp = util.submit_jobs(self.cook_url, job_spec, 10, groups=[group_spec])
+        self.assertEqual(resp.status_code, 201)
+        # Wait for some job to start
+        def some_job_started(group_response):
+            group = group_response.json()[0]
+            running_count = group['running']
+            self.logger.info(f"Currently {running_count} jobs running in group {group_uuid}")
+            return running_count > 0
+        def group_detail_query():
+            response = util.query_groups(self.cook_url, uuid=[group_uuid], detailed='true')
+            self.assertEqual(200, response.status_code)
+            return response
+        util.wait_until(group_detail_query, some_job_started)
+        # Now try to kill the whole group
+        util.kill_groups(self.cook_url, [group_uuid])
+        # Wait for all the jobs to die
+        # Ensure that each job Failed (because we killed it)
+        query = lambda: util.query_jobs(self.cook_url, True, job=jobs)
+        util.wait_until(query, util.all_instances_killed)
+        # Now try to kill the group again
+        # (ensure it still works when there are no live jobs)
+        util.kill_groups(self.cook_url, [group_uuid])
 
     def test_400_on_group_query_without_uuid(self):
         resp = util.query_groups(self.cook_url)
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -118,7 +118,12 @@ def submit_jobs(cook_url, job_specs, clones=1, **kwargs):
     """
     if isinstance(job_specs, dict):
         job_specs = [job_specs] * clones
-    jobs = [ minimal_job(**spec) for spec in job_specs ]
+    def full_spec(spec):
+        if 'uuid' not in spec:
+            return minimal_job(**spec)
+        else:
+            return spec
+    jobs = [ full_spec(j) for j in job_specs ]
     request_body = {'jobs': jobs}
     request_body.update(kwargs)
     logger.info(request_body)
@@ -130,6 +135,15 @@ def kill_jobs(cook_url, jobs, assert_response=True):
     """Kill one or more jobs"""
     params = {'job': [unpack_uuid(j) for j in jobs]}
     response = session.delete(f'{cook_url}/rawscheduler', params=params)
+    if assert_response:
+        assert 204 == response.status_code, response.content
+    return response
+
+
+def kill_groups(cook_url, groups, assert_response=True):
+    """Kill one or more groups of jobs"""
+    params = {'uuid': [unpack_uuid(g) for g in groups]}
+    response = session.delete(f'{cook_url}/group', params=params)
     if assert_response:
         assert 204 == response.status_code, response.content
     return response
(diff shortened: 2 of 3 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -700,6 +700,41 @@ def test_detailed_for_groups(self):
...
 
+    def test_group_kill(self):
+        # Create and submit jobs in group
+        slow_job_wait = 1200
+        group_spec = self.minimal_group()
+        job_fast = util.minimal_job(group=group_spec['uuid'])
+        job_slow = util.minimal_job(group=group_spec['uuid'],
+                                    command='sleep %d' % slow_job_wait)
+        data = {'jobs': [job_fast, job_slow], 'groups': [group_spec]}
+        resp = util.session.post('%s/rawscheduler' % self.cook_url, json=data)
```
dposada: We have `util.submit_job` for a single job, maybe we should create `util.submit_jobs` instead of repeating this pattern in multiple places

integration/tests/cook/test_basic.py
```diff
@@ -700,6 +700,41 @@ def test_detailed_for_groups(self):
...
+        util.wait_for_exit_code(self.cook_url, job_slow['uuid'])
+        # The fast job should have Success, slow job Failed (because we killed it)
+        jobs = util.session.get('%s/rawscheduler?job=%s&job=%s' %
+                                (self.cook_url, job_fast['uuid'], job_slow['uuid']))
+        self.assertEqual(200, jobs.status_code)
+        jobs = jobs.json()
+        self.logger.debug("JOBS:\n" + (json.dumps(jobs, sort_keys=True)))
+        self.assertEqual('success', jobs[0]['state'], 'Job details: %s' % (json.dumps(jobs[0], sort_keys=True)))
+        self.assertEqual('failed', jobs[1]['state'], 'Job details: %s' % (json.dumps(jobs[1], sort_keys=True)))
+        self.assertEqual(99003, jobs[1]['instances'][0]['reason_code'])
```
dposada: What is 99003?

integration/tests/cook/test_basic.py
```diff
@@ -700,6 +700,41 @@ def test_detailed_for_groups(self):
...
+        group_spec = self.minimal_group()
+        job_fast = util.minimal_job(group=group_spec['uuid'])
+        job_slow = util.minimal_job(group=group_spec['uuid'],
+                                    command='sleep %d' % slow_job_wait)
+        data = {'jobs': [job_fast, job_slow], 'groups': [group_spec]}
+        resp = util.session.post('%s/rawscheduler' % self.cook_url, json=data)
+        self.assertEqual(resp.status_code, 201)
+        self.logger.debug("group " + group_spec['uuid'])
+        self.logger.debug("fast  " + job_fast['uuid'])
+        self.logger.debug("slow  " + job_slow['uuid'])
```
dposada: Do we still need these 3 log lines?

integration/tests/cook/test_basic.py
```diff
@@ -700,6 +700,41 @@ def test_detailed_for_groups(self):
...
+        self.assertEqual(resp.status_code, 204)
+        # Wait for the slow job (and its instance) to die
+        util.wait_for_job(self.cook_url, job_slow['uuid'], 'completed')
+        util.wait_for_exit_code(self.cook_url, job_slow['uuid'])
+        # The fast job should have Success, slow job Failed (because we killed it)
+        jobs = util.session.get('%s/rawscheduler?job=%s&job=%s' %
+                                (self.cook_url, job_fast['uuid'], job_slow['uuid']))
+        self.assertEqual(200, jobs.status_code)
+        jobs = jobs.json()
+        self.logger.debug("JOBS:\n" + (json.dumps(jobs, sort_keys=True)))
```
dposada: Do we still need this log line?

scheduler/src/cook/mesos/api.clj
```diff
@@ -844,11 +844,23 @@
             application (assoc :application (util/remove-datomic-namespacing application))
             expected-runtime (assoc :expected-runtime expected-runtime))))
 
-(defn fetch-group-job-details
+(defn fetch-group-jobs-by-state
```
dposada: Missing docstring

scheduler/src/cook/mesos/api.clj
```diff
@@ -844,11 +844,23 @@
...
   [db guuid]
   (let [group (d/entity db [:group/uuid guuid])
-        jobs (:group/job group)
-        jobs-by-state (group-by :job/state jobs)]
+        jobs (:group/job group)]
+    (group-by :job/state jobs)))
+
+(defn fetch-group-live-jobs
```
dposada: Missing docstring

scheduler/src/cook/mesos/api.clj
```diff
@@ -844,11 +844,23 @@
...
+  [db guuid]
+  (->> guuid
+       (fetch-group-jobs-by-state db)
+       ((juxt :job.state/running :job.state/waiting))
+       (apply concat)
+       (vec)))
+
+(defn fetch-group-job-details
```
dposada: Missing docstring

scheduler/src/cook/mesos/api.clj
```diff
@@ -844,11 +844,23 @@
...
+        jobs (:group/job group)]
+    (group-by :job/state jobs)))
+
+(defn fetch-group-live-jobs
+  [db guuid]
+  (->> guuid
+       (fetch-group-jobs-by-state db)
+       ((juxt :job.state/running :job.state/waiting))
+       (apply concat)
+       (vec)))
```
dposada: Could we just `filter` the jobs by state rather than `group-by`, `juxt`, `concat`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1871,7 +1892,14 @@
                                :description "The groups were returned."}
                           400 {:description "Non-UUID values were passed."}
                           403 {:description "The supplied UUIDs don't correspond to valid groups."}}
-              :handler (read-groups-handler conn task-constraints is-authorized-fn)}}))
+              :handler (groups-action-handler conn task-constraints is-authorized-fn)}
+        :delete
+        {:summary "Kill all jobs within a set of groups"
+         :parameters {:query-params QueryGroupsParams}
```
dposada: Should we name this something other than `QueryGroupsParams`?

integration/tests/cook/test_basic.py
```diff
@@ -700,6 +700,41 @@ def test_detailed_for_groups(self):
...
+        self.assertEqual('success', jobs[0]['state'], 'Job details: %s' % (json.dumps(jobs[0], sort_keys=True)))
+        self.assertEqual('failed', jobs[1]['state'], 'Job details: %s' % (json.dumps(jobs[1], sort_keys=True)))
+        self.assertEqual(99003, jobs[1]['instances'][0]['reason_code'])
+        # Now try to kill the group again
+        # (ensure it still works when there are no live jobs)
+        resp = util.session.delete('%s/group?uuid=%s' % (self.cook_url, group_spec['uuid']))
+        self.assertEqual(resp.status_code, 204)
+
```
dposada: Should we add a test that multiple jobs get cancelled? Maybe some waiting and some running?

integration/tests/cook/test_basic.py
DaoWen: 99003 is "Command exited non-zero". I'll promote this to a named constant.

integration/tests/cook/test_basic.py
DaoWen: deleted

integration/tests/cook/test_basic.py
DaoWen: deleted

integration/tests/cook/test_basic.py
DaoWen: I added both `util.minimal_jobs` and `util.submit_jobs`. It actually didn't help with this specific case, but I was able to use `util.minimal_jobs` in my new test since the jobs being created were homogeneous.

integration/tests/cook/test_basic.py
DaoWen: done

scheduler/src/cook/mesos/api.clj
DaoWen: fixed

scheduler/src/cook/mesos/api.clj
DaoWen: fixed

scheduler/src/cook/mesos/api.clj
DaoWen: fixed

scheduler/src/cook/mesos/api.clj
DaoWen: Broke this out into a separate schema that doesn't include `detailed`.

scheduler/src/cook/mesos/api.clj
DaoWen: I removed this function when I refactored to use `filter` based on your other suggestion.

scheduler/src/cook/mesos/api.clj
DaoWen: Reverted this change when I refactored to use `filter` like you suggested.

DaoWen:
@dposada - I've added another test for when some jobs are running, some are waiting. I believe your other comments are addressed as well.

integration/tests/cook/test_basic.py
dposada: I was thinking `submit_jobs` could take in the `job_specs` as an argument, so that it would work even for heterogeneous jobs

dposada: And maybe it could also take in the `group_specs` (optionally)

integration/tests/cook/util.py
```diff
@@ -184,6 +196,54 @@ def wait_for_exit_code_inner():
         raise
 
 
+def wait_until(query, predicate, max_wait_ms=30000):
```
dposada: Missing docstring

dposada: Should `wait_for_job` and `wait_for_exit_code` switch to using this?

integration/tests/cook/util.py
```diff
@@ -184,6 +196,54 @@ def wait_for_exit_code_inner():
...
+
+    try:
+        return wait_until_inner()
+    except:
+        final_response = query()
+        logger.info("Timeout exceeded waiting for condition. Details: %s" % final_response)
+        raise
+
+
+def jobs_list_query(url, jobs, assert_response=False):
```
dposada: Should this just replace `query_jobs` altogether?

scheduler/src/cook/mesos/api.clj
```diff
@@ -864,6 +870,22 @@
...
+        completed? #(-> % :job/state (= :job.state/completed))]
+    (remove completed? jobs)))
+
+(defn fetch-group-live-jobs
+  "Get all jobs from a group that are currently running or waiting (not complete)"
+  [db guuid]
+  (let [group (d/entity db [:group/uuid guuid])
+        jobs (:group/job group)
+        completed? #(-> % :job/state (= :job.state/completed))]
+    (remove completed? jobs)))
```
dposada: Duplicated `defn`

scheduler/src/cook/mesos/api.clj
```diff
@@ -1215,30 +1237,39 @@
...
   (base-cook-handler
-    {:allowed-methods [:get]
+    {:allowed-methods [:get :delete]
      :allowed? (fn [ctx]
                  (let [user (get-in ctx [:request :authorization/user])
                        guuids (::guuids ctx)
                        group-user (fn [guuid] (-> (d/entity (db conn) [:group/uuid guuid])
                                                   :group/job first :job/user))
+                       request-method (get-in ctx [:request :request-method])
+                       action-name ({:get "view" :delete "kill"} request-method)
```
dposada: I think this would be clearer:

```clj
(request-method {:get "view" :delete "kill"})
```

sradack:
Where are we at on this?

DaoWen:
I just pushed #494 to address the points that @dposada pointed out with the integration tests during the review. I'll rebase this PR on top of those changes once they get merged.

scheduler/src/cook/mesos/api.clj
DaoWen: deleted

DaoWen:
@dposada - Now that #494 is merged, I think this patch is also ready.

integration/tests/cook/test_basic.py
```diff
@@ -769,6 +770,60 @@ def test_detailed_for_groups(self):
         for k in detail_keys:
             self.assertIn(k, group_info)
 
+    def test_group_kill_simple(self):
+        # Create and submit jobs in group
+        slow_job_wait = 1200
```
dposada: Nitpick: I like having the units in the name, e.g. `slow_job_wait_seconds`

integration/tests/cook/test_basic.py
```diff
@@ -769,6 +770,60 @@ def test_detailed_for_groups(self):
...
+        group_spec = self.minimal_group()
+        group_uuid = group_spec['uuid']
+        job_fast = util.minimal_job(group=group_uuid)
+        job_slow = util.minimal_job(group=group_uuid, command=f'sleep {slow_job_wait}')
+        _, resp = util.submit_jobs(self.cook_url, [job_fast, job_slow], groups=[group_spec])
+        self.assertEqual(resp.status_code, 201)
+        # Wait for the short job to finish
```
dposada: Nitpick: variable is `fast`, comment says "short"

scheduler/src/cook/mesos/api.clj
```diff
@@ -354,6 +354,12 @@
    (s/optional-key :detailed) s/Bool
    (s/optional-key :partial) s/Bool})
 
+(def KillGroupsParams
+  "Schema for killing groups, allowing optionally for 'partial'
+  results, meaning that some uuids can be valid and others not"
+  {:uuid [s/Uuid]
+   (s/optional-key :partial) s/Bool})
```
dposada: Am I missing something, or are we not using the `partial` flag? 

I'm not sure if we need it for kill, what do you think? 

In the CLI, when we implement `kill`, we will probably need to first query for the UUIDs, to make sure they don't exist on multiple clusters, since we don't want to kill on multiple clusters (the user will need to tell us which cluster). Since we've already queried, once we actually go to do the kill, we will know the UUID(s) exist(s) on the cluster we're doing the kill against.

Thoughts?

DaoWen: The partial flag is checked in the `:exists?` method, but after discussing this offline with you and Paul, I went ahead and deleted this. I agree with both of your reasoning that users should be precise about what they want to delete.

DaoWen:
@dposada - I made those last few small changes, and it looks like Travis is happy too.

PR #460, opened 2017-09-21
dockerignore files to reduce context copied over to Docker

shamsimam:


scheduler/.dockerignore
```diff
@@ -0,0 +1,15 @@
+.idea
+.lein-failures
+.minimesos
+.nrepl-port
+bin
+classes
+docs
+gclog.*
+log
+resources
```
pschorf: I don't think this is a good idea: maybe we can ignore it now and things will work, but eventually someone's going to add something here and it's going to be a pain to debug.

shamsimam: Removed `resources` from the dockerignore file

pschorf:
I'll merge this once the build passes

PR #464, opened 2017-09-25
update .gitignore to exclude cook-executor from the resources directory

shamsimam:


DaoWen:
I just verified that this gets the `resources` directory out of my untracked files list. :+1:

PR #465, opened 2017-09-26
fix test_no_cook_executor_on_subsequent_instances

shamsimam:


integration/tests/cook/test_basic.py
```diff
@@ -42,12 +42,14 @@ def test_no_cook_executor_on_subsequent_instances(self):
         job_uuid, resp = util.submit_job(self.cook_url, command='exit 1', max_retries=10)
         self.assertEqual(resp.status_code, 201, msg=resp.content)
         try:
-            util.wait_for_job(self.cook_url, job_uuid, 'completed', max_delay=60000)
+            job = util.wait_for_job(self.cook_url, job_uuid, 'completed', max_delay=60000)
+            message = json.dumps(job, sort_keys=True)
+            self.assertEqual('failed', job['state'], message)
+            self.assertEqual('completed', job['status'], message)
```
dposada: Doesn't the `wait_for_job` call take care of making sure the `status` is `completed`?

shamsimam: Yes, adding the assertion doesn't hurt though. Happy to remove it if you prefer.

PR #467, opened 2017-09-26
WIP: Use non-standard Mesos files for stdout and stderr

shamsimam:
Older versions of Mesos (e.g. 0.27.1) create the `stdout` and `stderr` with only read permissions for the user running the task. As a result, the Cook executor receives a permission error when it tries to open these files in append mode. As a workaround, we can instead use different files (`executor.stdout` and `executor.stderr`) which get created with read/write permission for the user running the task.

Note: This issue does not occur in later versions of Mesos (e.g. 1.2) where the `stdout` and `stderr` files are created with read/write permissions of the user running the task.


dposada:
Can we delete the branch?

PR #468, opened 2017-09-26
Configurable cook executor fields job

shamsimam:


scheduler/src/cook/mesos/api.clj
```diff
@@ -240,6 +240,11 @@
    (s/optional-key :constraints) [Constraint]
    (s/optional-key :container) Container
    (s/optional-key :executor) (s/enum "cook" "mesos")
+   (s/optional-key :executor-log-level) (s/enum "DEBUG" "ERROR" "INFO" "WARNING")
```
dposada: Do we really want to let users modify this?

scheduler/src/cook/mesos/api.clj
```diff
@@ -240,6 +240,11 @@
...
+   (s/optional-key :executor-max-message-length) PosInt
```
dposada: Do we really want to let users modify this?

scheduler/src/cook/mesos/api.clj
```diff
@@ -240,6 +240,11 @@
...
+   (s/optional-key :progress-output-file) NonEmptyString
+   (s/optional-key :progress-regex-string) NonEmptyString
+   (s/optional-key :progress-sample-interval-ms) PosInt
```
dposada: What's the use case for letting users modify the sample interval?

scheduler/src/cook/mesos/api.clj
shamsimam: @dposada @wyegelwel 

I introduced `executor-log-level` and `executor-max-message-length` as developer configs that can help us with debuggability. Both these values are sanitized and restricted to their allowed values in `task/build-executor-environment`.

scheduler/src/cook/mesos/api.clj
shamsimam: If users know their rate of output is going to be really slow, they can set `progress-sample-interval-ms` to a higher value than the default.

scheduler/src/cook/mesos/api.clj
shamsimam: Please see https://github.com/twosigma/Cook/pull/468#discussion_r141377993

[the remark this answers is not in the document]
shamsimam: Please see https://github.com/twosigma/Cook/pull/468#discussion_r141377993

scheduler/src/cook/mesos/api.clj
wyegelwel: Id prefer dropping them but don't care that much. If you are going to keep them, please move the value checks to submission time and fail the submission if they are invalid so the user doesn't get any surprises.

dposada: I'd prefer dropping them as well. If we find in the future that we're really missing them for debuggability, we can consider adding them

shamsimam: Removed configurability of `executor-log-level`, `executor-max-message-length`, and `progress-sample-interval-ms`.

shamsimam:
RTM

shamsimam:
@dposada any more feedback on this?

integration/tests/cook/test_basic.py
```diff
@@ -115,6 +115,35 @@ def test_progress_update_submit(self):
...
+        self.assertEqual('message: (\d*) (.*)', job['progress_regex_string'], message)
+        self.assertEqual(1, len(job['instances']))
+        message = json.dumps(job['instances'][0], sort_keys=True)
+        self.assertEqual('success', job['instances'][0]['status'], message)
+        self.assertEqual('success', job['instances'][0]['status'], message)
+
+        # allow enough time for progress updates to be submitted
+        publish_interval_ms = util.get_in(util.settings(self.cook_url), 'progress', 'publish-interval-ms')
+        wait_publish_interval_secs = min(2 * publish_interval_ms / 1000, 20)
+        time.sleep(wait_publish_interval_secs)
```
dposada: Can we accompish this with "wait for" instead of sleeping?

shamsimam: The issue is that the progress field can change after the `exit_code` or `end_time` have been provided. So we can't wait for those fields to be set or for the progress field to be set.

I can move the sleep inside a `if job_executor_type == 'cook':` if you like, that way tests not using the cook executor will not need to sleep. In any case, the sleep is capped at 20 seconds.

shamsimam: Done.

shamsimam:
RTM

PR #470, opened 2017-09-28
Makes the CLI http modules configurable

dposada:


cli/cook/colors.py
```diff
@@ -8,7 +8,7 @@ def failed(s):
 
 
 def success(s):
-    return term.blue(s)
+    return term.green(s)
```
dposada: Green looks much better as an indication of success

cli/cook/cli.py
```diff
@@ -90,9 +92,10 @@ def run(args):
 
     log_format = '%(asctime)s [%(levelname)s] %(message)s'
     if verbose:
+        logging.getLogger('').handlers = []
```
dposada: Without this, we are not guaranteed that our `basicConfig` call will have any effect

cli/cook/subcommands/show.py
```diff
@@ -284,7 +284,7 @@ def query_across_clusters(clusters, query_fn):
     """
     count = 0
     all_entities = {'clusters': {}}
-    max_workers = len(os.sched_getaffinity(0))
+    max_workers = os.cpu_count()
```
dposada: `sched_getaffinity` is not available on some platforms

cli/cook/__main__.py
```diff
@@ -14,11 +14,11 @@ def main(args=None):
 
     try:
         result = run(args)
-        exit(result)
+        sys.exit(result)
```
dposada: `pyinstaller` was unhappy without the `sys.`

Issue #479, opened 2017-10-05
Force docker container user


m4ce:
Hi,

I noticed that following logic in the build-container definition, line 416

```
    (if (= (clojure.string/lower-case ctype) "docker")
      (let [docker (:docker container)
            params (or (:parameters docker) [])
            port-mappings (or (:port-mapping docker) [])
            user-params (filter #(= (:key %) "user") params)
            expected-user-param (str (uid user) ":" (gid user))]
        (when (some #(not= expected-user-param (:value %)) user-params)
          (throw (ex-info "user parameter must match uid and gid of user submitting."
                          {:expected-user-param expected-user-param
                           :user-params-submitted user-params})))
```

The only problem is that if the user doesn't specify the user in the container parameters, it will run as root by default and no check will obviously be performed.

It might be good to force the user-params to be the uid and gid of the user submitting if it exists, otherwise error out?

dposada:
@m4ce Just so I understand, is the point that you want to avoid ever running as root inside the container?

m4ce:
@dposada, yes. It's problematic in terms of security, especially when you bind volumes from the host into the container.. :/

DaoWen:
I did some reading into this issue, and it looks like the standard solution for docker is to use *user namespaces*. If I understand the feature correctly, it allows the docker container to think that it's running as root, when really it's running as a less-privileged user on the host machine.

Here are a few of the references I was reading:

* https://docs.docker.com/engine/security/userns-remap/
* https://success.docker.com/KBase/Introduction_to_User_Namespaces_in_Docker_Engine
* https://www.jujens.eu/posts/en/2017/Jul/02/docker-userns-remap/

Based on that, I think this security issue might better addressed in the docker deployment by ensuring this uid-remapping feature is correctly configured.

@m4ce - What are your thoughts on this?

m4ce:
@DaoWen - I already new about user namespaces. However, it is my understanding that it a one-to-one mapping. It only works if you always want to re-map for instance uid 0 to a specific uid on the host. However, this won't work if you want to map uid 0 to many uids, which is the case here since we have multi-tenancy.

DaoWen:
OK, that makes sense. It looks like you can only specify a single remapping, and it must be set statically when the docker daemon is launched. That's probably not ideal for your use case.

I'd prefer not having this user-check enforced by default. See my latest comment on #480 for more details.

PR #480, opened 2017-10-05
Force docker container user parameters

m4ce:
See https://github.com/twosigma/Cook/issues/479 for more information.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -407,21 +407,24 @@
 
 (defn- build-container
   "Helper for submit-jobs, deal with container structure."
-  [user id container]
+  [user id container force-container-user?]
   (let [container-id (d/tempid :db.part/user)
         docker-id (d/tempid :db.part/user)
         ctype (:type container)
         volumes (or (:volumes container) [])]
     (if (= (clojure.string/lower-case ctype) "docker")
       (let [docker (:docker container)
-            params (or (:parameters docker) [])
+            container-params (or (:parameters docker) [])
             port-mappings (or (:port-mapping docker) [])
-            user-params (filter #(= (:key %) "user") params)
-            expected-user-param (str (uid user) ":" (gid user))]
-        (when (some #(not= expected-user-param (:value %)) user-params)
+            expected-user-params {:key "user" :value (str (uid user) ":" (gid user))}
+            defined-user-params (or (first (filter #(= (:key %) "user") container-params)) expected-user-params)
+            params (if (nil? defined-user-params) container-params (concat container-params [expected-user-params]))]
+        (if (and (force-container-user? (or (= (uid user) "") (= (gid user) ""))))
+          (throw (ex-info "failed to look up uid/gid of user submitting." {:user user})))
+        (if (not= (expected-user-params :value) (defined-user-params :value))
           (throw (ex-info "user parameter must match uid and gid of user submitting."
-                          {:expected-user-param expected-user-param
-                           :user-params-submitted user-params})))
+                          {:expected-user-params expected-user-params
+                           :defined-user-params defined-user-params})))
         [[:db/add id :job/container container-id]
          (merge {:db/id container-id
                  :container/type "DOCKER"}
@@ -447,7 +450,7 @@
 
 (s/defn make-job-txn
   "Creates the necessary txn data to insert a job into the database"
-  [job :- Job]
+  [job :- Job force-container-user?]
   (let [{:keys [uuid command max-retries max-runtime expected-runtime priority cpus mem gpus
                 user name ports uris env labels container group application disable-mea-culpa-retries
                 constraints executor]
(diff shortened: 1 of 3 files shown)


DaoWen:
Thanks for the patch! Your input on docker container jobs has been very helpful. I'm going to look into why this isn't passing the Travis tests a little later today.

Also, have you signed the [Cook project CLA](https://github.com/twosigma/Cook/tree/master/cla/)? Two Sigma Open Source needs to have a copy of the CLA on file from you before we can merge any changes. There are instructions in the document about how to send it. Scanning and emailing the signed document is probably the easiest option (that's what I did).

m4ce:
@DaoWen, I have not. I will sign it and email it to you guys.

m4ce:
CLA signed.

wyegelwel:
Sorry for the short reply I'm on my phone.

There is a danger in this check that the uid and gid on the host that is
running cook for the user doesn't match the uid and gid of the user on the
box the task is run on. I'm not sure the best way to address that.

m4ce:
It's definitely a possibility :/ The current check also implies that the user and gid must be the same on the box where Cook runs though. The alternative would be to do away with uid/gid and rely on the actual user name and group name. However, the user will have to make sure that the user/group is present in the running container.

DaoWen:
It looks like this patch isn't passing our Travis test suite because there are cases where the new checks fail. I think I would prefer making this check conditional, with an optional configuration setting that needs to be explicitly set to enable this feature. That would allow the tests to pass, and not break any of our production deployments, while giving you the option to enforce matching user IDs in your environment.

I think we also need some unit tests that check proper behavior with and without the setting enabled in the configuration. I'd lean towards unit tests rather than integration tests here, since integration tests would require adding another dimension to the Travis test matrix using an alternate configuration file. Adding some unit tests that ensure an exception is thrown when the check is enabled and violated should be straightforward though.

m4ce:
@DaoWen, before I make any more modifications, do my code changes look okay? I've propagated the configuration option all the way to `build-container`.

Let me know and I will then proceed adjusting the tests.

DaoWen:
I think the design looks fine. Ideally, I'd like to see *all* of our configuration-settings code get refactored so that we don't have to chop it up and pipe pieces of it through multiple layers of functions like this—but that's not happening today.

I noticed some syntactic issues while I was looking at your change. You can view the diff here:
https://github.com/DaoWen/Cook/commit/d830c8fc8dcb3f7a06ac386db5342033d934c127

But yeah, overall your approach looks good to me.

DaoWen:
@m4ce - Do you need help debugging this? It looks like something is getting messed up with the job submissions, and the job aren't actually being created, but the request return success anyway. Since jobs aren't actually being created, almost all of the tests end up failing.

m4ce:
@DaoWen - I was away for a week and only could come back to you now. Would you have any advice for me here?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1172,9 +1177,9 @@
      ;; Implementing put! doesn't mean that PUT requests are actually supported
      ;; (see :allowed-methods above). It is simply what liberator eventually calls to persist
      ;; resource changes when conflict? has returned false.
-     :put! (partial create-jobs! conn)
+     :put! (partial create-jobs! conn force-container-user?)
```
DaoWen: `force-container-user?` is passed as the *second* argument to `create-jobs!` here.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1059,7 +1064,7 @@
   to Datomic.
   Preconditions:  The context must already have been populated with both
   ::jobs and ::groups, which specify the jobs and job groups."
-  [conn {:keys [::groups ::jobs] :as ctx}]
+  [conn {:keys [::groups ::jobs] :as ctx} force-container-user?]
```
DaoWen: `force-container-user?` is declared as the *third* argument to `create-jobs!` here. I think it should be the *second* argument; i.e., it should be moved right after `conn`. The context needs to be the last argument because all of the other arguments are passed via a partial function application, and then the context is passed implicitly by the Liberator framework. (See my other comment below for the partial function application.)

DaoWen:
Now it's failing because `api/create-jobs!` is also called several times in `scheduler/test/cook/test/mesos/api.clj`. (There may be other reasons too—but this was the obvious reason looking at the new Travis failure log.)

DaoWen:
OK, hopefully Travis will be happy with ea8d9f2...

Assuming that it's all working now when the flag is `false`, we still need a couple of new tests to check that the option works as intended when it's enabled. Maybe add a couple more test cases at the end of `test-create-jobs!` (that's the same test you just updated in `scheduler/test/cook/test/mesos/api.clj`) to test the option with `true`; e.g., testing that it accepts jobs with matching credentials, and rejects jobs if one or more has mismatched credentials.

DaoWen:
@m4ce - Based on the inactivity, I'm guessing that you don't have time to work on this feature at the moment. I'm going close the PR for now. We can always reopen it later when you have time to work on it again.

PR #481, opened 2017-10-06
Order in priority message handling

shamsimam:
Cook executor: Add support for sequence numbers while sending progress status messages.

Cook scheduler: `progress-aggregator` respects the sequence number in-memory while performing local aggregation of progress messages as it is possible for them to arrive out-of-order.

### DIFF
--- a/scheduler/docs/configuration.adoc
+++ b/scheduler/docs/configuration.adoc
@@ -289,7 +289,7 @@ An example configuration looks like:
 {...
  :executor {:command "cook-executor"
             :default-progress-output-file "stdout"
-            :default-progress-regex-string "progress: (\d*)(?: )?(.*)"
+            :default-progress-regex-string "progress: (\d+)(?: )?(.*)"
             :log-level "INFO"
             :max-message-length 512
             :progress-sample-interval-ms 1000
--- a/scheduler/docs/configuration.adoc
+++ b/scheduler/docs/configuration.adoc
@@ -377,8 +377,10 @@ The configuration values are defined as follows:
   The default value is 2500.
 
 `:sequence-cache-threshold`::
-  An integer representing the size of the task id to progress sequence cache to maintain.
+  An integer representing the max number of items in the task sequence cache.
   This cache is used to track the latest sequence number of progress message processed for a given task.
+  In order to avoid the potential for out-of-order progress updates,
+  this cache should be sized to handle the maximum number of active tasks that are reporting progress.
   The default value is 1000.
 
 === Debugging Facilities
--- a/scheduler/docs/configuration.adoc
+++ b/scheduler/docs/configuration.adoc
@@ -356,7 +356,8 @@ An example configuration looks like:
 {...
  :progress {:batch-size 100
             :pending-threshold 4000
-            :publish-interval-ms 2500}
+            :publish-interval-ms 2500
+            :sequence-cache-threshold 1000}
  ...}
 ```
 
@@ -374,6 +375,11 @@ The configuration values are defined as follows:
 `:publish-interval-ms`::
   An integer representing the number of millisecond intervals at which progress updates will be published to datomic.
   The default value is 2500.
+
+`:sequence-cache-threshold`::
+  An integer representing the size of the task id to progress sequence cache to maintain.
+  This cache is used to track the latest sequence number of progress message processed for a given task.
+  The default value is 1000.
 
 === Debugging Facilities
 
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -292,7 +292,7 @@
    stale data based on the value of progress-sequence in the message.
    We avoid this stale check in the datomic transaction as it is relatively expensive to perform the checks on a per
    instance basis in the query."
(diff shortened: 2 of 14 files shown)


shamsimam:
RTM

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -287,19 +287,34 @@
...
-        (counters/inc! progress-aggregator-pending-states-count (- new-count old-count)))
-      instance-id->progress-state')
+    (if (integer? (:progress-sequence data))
+      (let [progress-state (select-keys data [:progress-message :progress-percent :progress-sequence])
+            instance-id->progress-state' (update instance-id->progress-state instance-id
+                                                 (fn [current-state]
+                                                   (let [new-progress-sequence (:progress-sequence progress-state)
+                                                         old-progress-sequence (:progress-sequence current-state)]
+                                                     (if (or (nil? current-state)
+                                                             (< old-progress-sequence new-progress-sequence))
```
sradack: Unclear how effective this is giving that the latest sequence is being flushed out with every write to Datomic.

shamsimam: I wasn't sure how much we want to protect against out-of-order messages. If we want to preserve across write, probably need to introduce an LRU cache to store the task to progress sequence mapping. Do you want me to go down that path?

shamsimam: We now preserve the `instance-id->sequence` values i  the `sequence-cache-store` that persist across datomic writes.

shamsimam:
RTM

scheduler/docs/configuration.adoc
```diff
@@ -539,4 +545,4 @@ Here's a table with some options that should work for a cluster with thousands o
 
 ### License
 
-© Two Sigma Open Source, LLC
+?? Two Sigma Open Source, LLC
```
sradack: What happened here?

scheduler/docs/configuration.adoc
```diff
@@ -375,6 +376,11 @@ The configuration values are defined as follows:
   An integer representing the number of millisecond intervals at which progress updates will be published to datomic.
   The default value is 2500.
 
+`:sequence-cache-threshold`::
+  An integer representing the size of the task id to progress sequence cache to maintain.
```
sradack: The max number of items in the task sequence cache.

scheduler/docs/configuration.adoc
```diff
@@ -375,6 +376,11 @@ The configuration values are defined as follows:
...
+  This cache is used to track the latest sequence number of progress message processed for a given task.
+  The default value is 1000.
+
```
sradack: In order to avoid the potential for out-of-order progress updates, this cache should be sized to handle the maximum number of active tasks that are reporting progress.

scheduler/src/cook/mesos/util.clj
```diff
@@ -598,3 +598,21 @@
                (when close?
                  (async/close! to))
                (recur (reducer state data)))))))
+
+(defn cache-lookup!
```
sradack: Please make unit tests.

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1683,32 +1695,83 @@
...
-                                      {:instance-id "i2" :progress-message "i2.m1" :progress-percent 20}))))
-
-  (testing "handle threshold limit reached"
-    (is (= {"i1" {:progress-message "i1.m1" :progress-percent 10}
-            "i2" {:progress-message "i2.m1" :progress-percent 20}}
-           (sched/progress-aggregator 2 {"i1" {:progress-message "i1.m1" :progress-percent 10}}
-                                      {:instance-id "i2" :progress-message "i2.m1" :progress-percent 20})))))
+  (let [pending-progress-threshold 10
+        sequence-cache-store-fn (fn [& {:keys [data threshold] :or {data {} threshold 100}}]
+                                  (atom (cache/lru-cache-factory data :threshold threshold)))]
```
sradack: Is a TTL-based cache more appropriate?

shamsimam: I think it would still need to be backed by a cache of bounded size, e.g. LRU. So I chose to just use LRU directly and bypass the TTL cache requirement. Since the data is just a map and an integer, I was okay skipping the TTL eviction.

scheduler/docs/configuration.adoc
shamsimam: Done.

scheduler/docs/configuration.adoc
shamsimam: Done.

scheduler/docs/configuration.adoc
shamsimam: Used `(C)`

scheduler/src/cook/mesos/util.clj
shamsimam: Added unit tests.

scheduler/test/cook/test/mesos/scheduler.clj
shamsimam: Added ttl cache.

shamsimam:
RTM

PR #484, opened 2017-10-13
Adds ssh command to the CLI

dposada:


cli/cook/version.py
```diff
@@ -1 +1 @@
-VERSION = '0.1.2a2'
+VERSION = '0.1.2a3'
```
sradack: This is a new feature.  Typically we should be bumping the minor version, although maybe we had something else in mind.  What does the "a" stand for?

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+    resp_json = resp.json()
+    flags = resp_json['flags']
+    if 'work_dir' not in flags:
+        logging.error(f'work_dir was not present in mesos agent state flags: {flags}')
+        raise Exception('Mesos agent state did not include its work directory.')
+
+    print_info(f'Executing ssh to {colors.bold(hostname)}.')
+    agent_work_dir = flags['work_dir']
+    directory = sandbox_directory(agent_work_dir, instance, job)
+    file = os.environ.get('SSH', 'ssh')
```
sradack: Using `CS_SSH` instead of `SSH` in case someone has set `SSH` for some other reason?

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+            print_info(str(e))
+    else:
+        raise Exception(f'Unable to ssh for any instance of this job.')
+
+
+def ssh(clusters, args):
+    """Attempts to ssh (using os.execlp) to the Mesos agent corresponding to the given job or instance uuid."""
+    uuids = strip_all(args.get('uuid'))
+    if len(uuids) > 1:
+        # argparse should prevent this, but we'll be defensive anyway
```
sradack: These comments are excellent.

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
+import logging
+import os
+from operator import itemgetter
+from urllib.parse import urlparse
+
+from cook import colors, http
+from cook.subcommands.show import query, print_no_data
```
sradack: Should we factor out functions shared by subcommands so that they don't depend on each other?  `query` and `print_no_data` seem general enough to not be specifically in the `show` command.

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+    flags = resp_json['flags']
+    if 'work_dir' not in flags:
+        logging.error(f'work_dir was not present in mesos agent state flags: {flags}')
+        raise Exception('Mesos agent state did not include its work directory.')
+
+    print_info(f'Executing ssh to {colors.bold(hostname)}.')
+    agent_work_dir = flags['work_dir']
+    directory = sandbox_directory(agent_work_dir, instance, job)
+    file = os.environ.get('SSH', 'ssh')
+    logging.info(f'using ssh file: {file}')
```
sradack: What do you think about using `using ssh command` instead of `using ssh file`?

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+    os.execlp(file, 'ssh', '-t', hostname, f'cd {directory} ; bash')
```
sradack: If the directory has a space, we might run into an issue here.

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+    if len(instances) == 0:
+        raise Exception('This job currently has no instances.')
+
+    for instance in sorted(instances, key=itemgetter('start_time'), reverse=True):
+        try:
+            print_info(f'Attempting ssh for job instance {colors.bold(instance["task_id"])}...')
+            ssh_to_instance(instance, job)
+        except Exception as e:
+            print_info(str(e))
+    else:
```
sradack: Falling back to `ssh`ing into the non-latest instance might be confusion to users.

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if num_results > 1:
+        # This is unlikely to happen in the wild, but it could.
+        # A couple of examples of how this could happen:
+        # - same uuid on an instance and a job (not necessarily the parent job)
+        # - same uuid on a job in cluster x as another job in cluster y
+        print_info('There is more than one match for the given uuid.')
+        return 1
+
+    for cluster_name, entities in query_result['clusters'].items():
```
sradack: Is it possible for there to be multiple clusters?  Using `for` implies that it's possible.

sradack: This command wants to resolve a uuid to a unique something.  If future commands wants the same semantics, it might be useful to have a function that works something like this:

<pre>
query_unique(uuid) -> {'type': 'job', 'data': {jobdata}}
                    | {'type': 'instance', 'data': {instance-data}}
                    | {'type': 'group', 'data': {group-data}}</pre>

That way this and other functions that want one thing don't have to do the same processing.

cli/cook/subcommands/ssh.py
dposada: Done (created `querying.py`)

cli/cook/subcommands/ssh.py
dposada: Good idea, done

cli/cook/subcommands/ssh.py
dposada: Done

cli/cook/subcommands/ssh.py
dposada: Added quotes, thanks

cli/cook/subcommands/ssh.py
dposada: Ditched this, and now we only try the most recently started instance

cli/cook/subcommands/ssh.py
dposada: Good call. Extracted a function `query_unique`. For now, it doesn't support a group uuid, because that's the same pattern we want for ssh, fuse, ls, tail

cli/cook/version.py
dposada: Changed to `0.2.0`. The "a" was for alpha, but I ditched it

dposada:
@sradack I addressed your feedback, thanks. Anything more?

cli/cook/subcommands/ssh.py
```diff
@@ -0,0 +1,119 @@
...
+    instances = job['instances']
+    if len(instances) == 0:
+        raise Exception('This job currently has no instances.')
+
+    instance = max(instances, key=itemgetter('start_time'))
+    print_info(f'Attempting ssh for job instance {colors.bold(instance["task_id"])}...')
+    ssh_to_instance(instance, job)
+
+
+def query_unique(clusters, uuid):
```
sradack: What do you think about having this in the querying package?

dposada: I was thinking about that on the drive in this morning. Done.

PR #487, opened 2017-10-17
Keep cook-executor binary updated

DaoWen:
Ensure cook-executor binary is rebuilt when sources change, and always
copy the newest version when building the scheduler docker image.

### DIFF
--- a/executor/bin/check-version.sh
+++ b/executor/bin/check-version.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# USAGE: ./bin/check-version.sh [-q|--quiet]
+#
+# Returns 0 (true) if the currently-installed cook executor version
+# matches the source version, and 1 (false) otherwise.
+#
+# When using the quiet flag (-q|--quiet), no additional output is printed.
+#
+# Note that versions containing the case-insensitive string "SNAPSHOT" never match
+# (i.e., they're considered "unstable", and should always require a rebuild).
+
+case "$1" in
+    -q|--quiet)
+        shift
+        verbose=false
+        ;;
+    '')
+        verbose=true
+        ;;
+    *)
+        echo "USAGE: $0 [-q|--quiet]"
+        exit -1
+esac
+
+source_version=`python3 -c 'from cook._version import __version__; print(__version__)'`
+
+shopt -s nocasematch
+
+if [[ "$source_version" == *SNAPSHOT* ]]; then
+    if $verbose; then
+        echo "Detected snapshot (unstable) version: $source_version"
+    fi
+    exit 1
+fi
+
+binary_app=./dist/cook-executor
+
+if [ -e $binary_app ]; then
+    installed_version="$($binary_app --version)"
+else
+    installed_version=none
+fi
+
+if [ "$installed_version" == "$source_version" ]; then
+    if $verbose; then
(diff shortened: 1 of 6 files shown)


executor/bin/check-version.sh
```diff
@@ -0,0 +1,55 @@
...
+        ;;
+    '')
+        verbose=true
+        ;;
+    *)
+        echo "USAGE: $0 [-q|--quiet]"
+        exit -1
+esac
+
+source_version=`python3 -c 'from cook._version import __version__; print(__version__)'`
```
shamsimam: Can we ensure that this command works when the script is invoked from any directory?

executor/bin/check-version.sh
```diff
@@ -0,0 +1,55 @@
...
+    fi
+    exit 1
+fi
+
+binary_app=./dist/cook-executor
+
+if [ -e $binary_app ]; then
+    installed_version="$($binary_app --version)"
+else
+    installed_version=none
```
shamsimam: Should we eagerly return 1 here?

executor/cook/__main__.py
```diff
@@ -21,6 +21,11 @@
 
 
 def main(args=None):
+    if len(sys.argv) == 2 and sys.argv[1] == "--version":
+        from _version import __version__
+        print(__version__)
+        sys.exit(0)
+
     print('Starting cook executor...')
```
shamsimam: Can we include the version here?

scheduler/bin/build-docker-image.sh
```diff
@@ -3,5 +3,29 @@
...
+if $DO_EXECUTOR_REBUILD; then
+    echo "Triggering build of cook-executor before proceeding."
+    ${EXECUTOR_DIR}/bin/build-cook-executor.sh
+fi
+
+echo "Copying cook-executor from ${COOK_EXECUTOR_FILE} to ${SCHEDULER_EXECUTOR_DIR}"
+mkdir -p ${SCHEDULER_EXECUTOR_DIR}
+cp -f ${COOK_EXECUTOR_FILE} ${SCHEDULER_EXECUTOR_DIR}
+
+
```
shamsimam: Nitpick: delete this blank line

executor/bin/check-version.sh
DaoWen: I wanted to let this fall through to one of the `echo`s below.

executor/cook/__main__.py
DaoWen: How do you want it formatted?

executor/bin/check-version.sh
DaoWen: Done

executor/cook/__main__.py
shamsimam: I like: `print('Starting cook executor {}...'.format(version))`

executor/bin/check-version.sh
```diff
@@ -23,8 +27,24 @@ case "$1" in
         exit -1
 esac
 
+#
+# Switch to executor directory
+#
+
+executor_bin_dir="$(dirname ${BASH_SOURCE[0]})"
+
+cd "$executor_bin_dir/.."
```
shamsimam: Does this need to be a `pushd` followed by a `popd` later?

PR #488, opened 2017-10-17
Fixes support for host placement + straggler handling in Java client

dposada:


wyegelwel:
This code all looks good, thought there are a lot of changes. What was the specific thing that needed to be fixed?

dposada:
Without these changes, attempting to submit a job group with an `attribute-equals` host placement or a `quantile-deviation` straggler handling through the job client doesn't work because it's sending underscores in the JSON posted to the API instead of dashes. I fixed that and also added some assertions to the test that was attempting to submit these.

wyegelwel:
Ok, makes sense, though I was under the impression that the api would accept either underscore or dash. Glad to know the answer.

dposada:
@wyegelwel I believe the API will accept either underscore or dash for keys, but not (necessarily) for values

PR #490, opened 2017-10-18
Use uuid lookup ref in record-placement-failures

pschorf:


scheduler/src/cook/mesos/fenzo_utils.clj
```diff
@@ -71,7 +71,7 @@
                   (reduce summarize-placement-failure {} failures))]
     (log/debug "Job" (:job/uuid job) " placement failure summary:" summary)
     (when (and under-investigation? (seq summary))
-      [(:db/id job) summary])))
+      [(:job/uuid job) summary])))
```
dposada: Why is this changing?

scheduler/src/cook/mesos/fenzo_utils.clj
```diff
@@ -81,11 +81,11 @@
   (log/debug "Fenzo placement failure information:" failure-results)
   (let [failures-by-job (into {} (map accumulate-placement-failures
                                       failure-results))
-        transactions (mapv (fn [[job-db-id failures-description]]
+        transactions (mapv (fn [[job-uuid failures-description]]
                              (let [failure-summary (str failures-description)]
                                (when (> (count failure-summary) 25000)
                                  (log/warn "Fenzo placement Failure summary is too long!" failure-summary))
-                               {:db/id job-db-id
+                               {:db/id [:job/uuid job-uuid]
```
dposada: What effect does `[:job/uuid job-uuid]` have in the transaction?

scheduler/test/cook/test/mesos/fenzo_utils.clj
```diff
@@ -85,3 +89,16 @@
...
+  (let [uri "datomic:mem://test-record-placement-failures"
+        conn (restore-fresh-database! uri)
+        job-id (create-dummy-job conn :under-investigation true)
+        job (->> job-id
+                 (d/entity (d/db conn))
+                 util/entity->map)
+        ^TaskRequest task-request (scheduler/make-task-request job)
+        failure (assignment-failure :mem)
+        assignment-result (SimpleAssignmentResult. [failure] nil task-request)]
+    (is (fenzo/record-placement-failures! conn [[assignment-result]]))))
```
dposada: Is this just asserting that transacting the placement failures didn't throw? Should we also assert on what got transacted?

scheduler/src/cook/mesos/fenzo_utils.clj
pschorf: Jobs don't have a `:db/id` after the refactoring I did a few months ago, so this will always be nil

scheduler/src/cook/mesos/fenzo_utils.clj
pschorf: Datomic will look up the job with that UUID and replace the ref with that ID.

scheduler/test/cook/test/mesos/fenzo_utils.clj
pschorf: This is just asserting that it returned non-nil (a transaction was performed). I will add an additional query afterwards to check the results.

wyegelwel:
Code looks good to me. Would you add an integration test for the unscheduled endpoint so we can have more confidence it will work? If you set a task constraint that will ensure it won't get shceudled you can check that is returned from the request

pschorf:
Good point, added an integration test.

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+
+
+    def test_unscheduled_jobs(self):
+        job_uuid, resp = util.submit_job(self.cook_url,
+                                         command='ls',
+                                         constraints=[['HOSTNAME', 'EQUALS', 'fakehost']])
+        self.assertEqual(resp.status_code, 201, resp.content)
+        try:
+            unscheduled_resp = util.session.get('%s/unscheduled_jobs?job=%s' % (self.cook_url,
+                                                                                job_uuid))
```
dposada: Nitpick: can we keep this on the previous line?

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+            self.assertEqual(200, unscheduled_resp.status_code, unscheduled_resp)
+            unscheduled_resp = unscheduled_resp.json()[0]
+            self.assertTrue(
+                any(['The job is now under investigation. Check back in a minute for more details!' == reason['reason'] for reason in unscheduled_resp['reasons']]),
```
dposada: Would there ever be more than 1 reason here? Should we assert `len(reasons) == 1`?

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+                unscheduled_resp)
+            self.assertEqual(job_uuid, unscheduled_resp['uuid'])
+
+            @retry(stop_max_delay=60000, wait_fixed=1000)
+            def check_unscheduled_reason():
+                unscheduled_resp_2 = util.session.get('%s/unscheduled_jobs?job=%s' % (self.cook_url,
```
dposada: Why `_2`?

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+                                                                                      job_uuid))
```
dposada: Nitpick: previous line

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+                self.assertEqual(200, unscheduled_resp_2.status_code, unscheduled_resp_2)
+                unscheduled_resp_2 = unscheduled_resp_2.json()[0]
+                self.assertTrue(
+                    any(['The job couldn\'t be placed on any available hosts.' == reason['reason'] for reason in unscheduled_resp_2['reasons']]),
```
dposada: Should we assert there are no other reasons?

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
+
+
+    def test_unscheduled_jobs(self):
+        job_uuid, resp = util.submit_job(self.cook_url,
+                                         command='ls',
+                                         constraints=[['HOSTNAME', 'EQUALS', 'fakehost']])
+        self.assertEqual(resp.status_code, 201, resp.content)
+        try:
+            unscheduled_resp = util.session.get('%s/unscheduled_jobs?job=%s' % (self.cook_url,
```
dposada: Should we add `util` function for this endpoint?

integration/tests/cook/test_basic.py
```diff
@@ -845,3 +845,33 @@ def get_docker_info():
...
+                                                                                      job_uuid))
+                self.assertEqual(200, unscheduled_resp_2.status_code, unscheduled_resp_2)
+                unscheduled_resp_2 = unscheduled_resp_2.json()[0]
+                self.assertTrue(
+                    any(['The job couldn\'t be placed on any available hosts.' == reason['reason'] for reason in unscheduled_resp_2['reasons']]),
+                    unscheduled_resp_2)
+                self.assertEqual(job_uuid, unscheduled_resp_2['uuid'])
+            check_unscheduled_reason()
+        finally:
+            util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
```
dposada: Should we add a `util` function for deleting a job?

integration/tests/cook/test_basic.py
pschorf: Done

integration/tests/cook/test_basic.py
pschorf: Done

integration/tests/cook/test_basic.py
pschorf: If multiple tests run at once, there can be a separate reason that "You have jobs ahead of this in the queue" but that is nondeterministic so I opted for the `any` approach

integration/tests/cook/test_basic.py
pschorf: done

integration/tests/cook/test_basic.py
pschorf: done

integration/tests/cook/test_basic.py
pschorf: Same issue as above

integration/tests/cook/test_basic.py
pschorf: Done

integration/tests/cook/test_basic.py
dposada: Can we add a comment explaining this?

integration/tests/cook/test_basic.py
dposada: Comment?

integration/tests/cook/util.py
```diff
@@ -240,3 +240,9 @@ def get_executor(agent_state, executor_id):
 def get_user(cook_url, job_uuid):
     """Retrieves the job corresponding to the given job_uuid and returns the user"""
     return load_job(cook_url, job_uuid)['user']
+
+def unscheduled_jobs(cook_url, job_uuid):
+    return session.get('%s/unscheduled_jobs?job=%s' % (cook_url, job_uuid)).json()
+
+def kill_job(cook_url, job_uuid):
```
dposada: docstrings on these two

integration/tests/cook/test_basic.py
dposada: I think this comment may have been hidden due to other commits :point_up:

integration/tests/cook/test_basic.py
dposada: I think this comment may have been hidden due to other commits :point_up:

pschorf: done

integration/tests/cook/test_basic.py
pschorf: done

dposada:
Must be a flaky test because the comments made the build fail

pschorf:
Weird, one of the stages timed out in Travis. I reran that step and it succeeded.

PR #491, opened 2017-10-18
Delete docker containers automatically with --rm flag

DaoWen:


executor/bin/build-cook-executor.sh
```diff
@@ -5,18 +5,13 @@ set -e
...
-    # Cleanup
-    docker rm ${NAME}
-fi
-
 EXECUTOR_DIR="$(dirname ${DIR})"
 
 mkdir -p ${EXECUTOR_DIR}/dist
 
 # build cook-executor inside docker image to avoid local python environment and architecture hassles
-pushd ${EXECUTOR_DIR}
```
shamsimam: Why remove the `pushd` and `popd`?

DaoWen: Since the `popd` was the last thing in the script, it seemed superfluous (this doesn't seem like a script you'd want to source as opposed to running in a subshell)—but if I remove the `popd`, then the `pushd` looks weird since we've lost the symmetry, so I ended up changing it to a `cd`. Also, `pushd` and `popd` print messages that I think just add more noise to the output.

PR #496, opened 2017-10-20
Use pre-installed Python 3.6 on Travis

DaoWen:
Since pyenv has been failing to download and install python3.6.2 all morning, I think making this switch is worthwhile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,11 @@ jdk:
 services:
   - docker
 before_install:
+  - source ~/virtualenv/python3.6.2/bin/activate
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv E56151BF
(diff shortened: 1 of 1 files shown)


DaoWen:
Hmmm... I tested this using the latest [travisci/ci-garnet image from DockerHub](https://hub.docker.com/r/travisci/ci-garnet/), but apparently that's not the image they actually use in production. I found a closer match based on [the language→image mapping in their documentation](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally), and it doesn't have python3.6 pre-installed anywhere in the image as far as I can tell.

Anyway, from what I can tell, our call to pyenv should automatically pick up on a local install of python3.6 whenever Travis decides to update to the newer garnet images, so there's no reason to change our Travis config for this.

PR #497, opened 2017-10-24
Fix flaky integration test for max_runtime_exceeded

### LINKED ISSUE #299: test_max_runtime_exceeded sometimes fails the reason code assertion
Example here:

https://travis-ci.org/twosigma/Cook/jobs/223179765#L1393-L1398

I can't seem to find any usages of reason `:mesos-command-executor-failed` (or 99003) in the scheduler. @wyegelwel, do you have any ideas as to what might cause this?

DaoWen:
Closes #299

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -223,22 +223,24 @@ class CookTest(unittest.TestCase):
             util.wait_for_job(self.cook_url, job_uuid, 'running')
             util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_ms)
             job = util.wait_for_end_time(self.cook_url, job_uuid)
-            self.assertEqual(1, len(job['instances']))
+            job_details = f"Job details: {json.dumps(job, sort_keys=True)}"
+            self.assertEqual(1, len(job['instances']), job_details)
             instance = job['instances'][0]
-            self.assertEqual('failed', instance['status'])
+            self.assertEqual('failed', instance['status'], job_details)
             valid_reasons = [# we killed the job because it took too long
                              CookTest.REASON_MAX_RUNTIME_EXCEEDED,
                              # the job was killed before the executor was properly registered
                              CookTest.REASON_EXECUTOR_UNREGISTERED,
                              # the job didn't die gracefully when we terminated it
                              CookTest.REASON_CMD_NON_ZERO_EXIT]
-            self.assertIn(instance['reason_code'], valid_reasons)
+            self.assertIn(instance['reason_code'], valid_reasons, job_details)
             actual_running_time_ms = instance['end_time'] - instance['start_time']
-            self.assertGreater(actual_running_time_ms, max_runtime_ms)
+            self.assertGreater(actual_running_time_ms, max_runtime_ms, job_details)
+            self.assertGreater(job_timeout_interval_ms, actual_running_time_ms, job_details)
             if job_executor_type == 'cook':
                 job = util.wait_for_exit_code(self.cook_url, job_uuid)
-                self.assertNotEqual(0, job['instances'][0]['exit_code'])
-                self.assertTrue(bool(job['instances'][0]['sandbox_directory']))
+                self.assertNotEqual(0, job['instances'][0]['exit_code'], job_details)
+                self.assertTrue(bool(job['instances'][0]['sandbox_directory']), job_details)
         finally:
             util.kill_jobs(self.cook_url, [job_uuid])
 
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -208,26 +208,35 @@ class CookTest(unittest.TestCase):
                                                         'timeout-interval-minutes')
         # the value needs to be a little more than 2 times settings_timeout_interval_minutes to allow
         # at least two runs of the lingering task killer
-        job_timeout_interval_seconds = (2 * settings_timeout_interval_minutes * 60) + 15
-        job_timeout_interval_ms = job_timeout_interval_seconds * 1000
+        job_sleep_seconds = (2 * settings_timeout_interval_minutes * 60) + 15
+        job_sleep_ms = job_sleep_seconds * 1000
         max_runtime_ms = 5000
-        assert max_runtime_ms < job_timeout_interval_ms
+        assert max_runtime_ms < job_sleep_ms
         try:
+            # schedule a job that will go over its max_runtime time limit
             job_uuid, resp = util.submit_job(self.cook_url,
-                                             command=f'sleep {job_timeout_interval_seconds}',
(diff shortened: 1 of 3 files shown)


DaoWen:
@dposada and I were discussing what to do about this test this morning. We concluded that making the test a little more robust/forgiving would be better than marking it as explicit.

integration/tests/cook/util.py
```diff
@@ -278,6 +278,25 @@ def predicate(response):
     return response.json()[0]
 
 
+def wait_for_end_time(cook_url, job_id):
+    """Wait for the given job's end_time field to appear in instance 0."""
```
shamsimam: Please update the docstring to clarify that it returns the job when successful.

integration/tests/cook/test_basic.py
```diff
@@ -213,19 +215,32 @@ def test_max_runtime_exceeded(self):
...
-            self.assertTrue(bool(job['instances'][0]['sandbox_directory']))
+        try:
+            job_uuid, resp = util.submit_job(self.cook_url,
+                                             command=f'sleep {job_timeout_interval_seconds}',
+                                             executor=job_executor_type, max_runtime=max_runtime_ms)
+            self.assertEqual(201, resp.status_code, msg=resp.content)
+            util.wait_for_job(self.cook_url, job_uuid, 'running')
+            util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_ms)
+            job = util.wait_for_end_time(self.cook_url, job_uuid)
+            self.assertEqual(1, len(job['instances']))
```
shamsimam: Can we include the job details in the assertion message. `f"Job details: {json.dumps(job, sort_keys=True)}"` ?

integration/tests/cook/test_basic.py
```diff
@@ -213,19 +215,32 @@ def test_max_runtime_exceeded(self):
...
+            self.assertEqual('failed', instance['status'])
+            valid_reasons = [# we killed the job because it took too long
+                             CookTest.REASON_MAX_RUNTIME_EXCEEDED,
+                             # the job was killed before the executor was properly registered
+                             CookTest.REASON_EXECUTOR_UNREGISTERED,
+                             # the job didn't die gracefully when we terminated it
+                             CookTest.REASON_CMD_NON_ZERO_EXIT]
+            self.assertIn(instance['reason_code'], valid_reasons)
+            actual_running_time_ms = instance['end_time'] - instance['start_time']
+            self.assertGreater(actual_running_time_ms, max_runtime_ms)
```
shamsimam: Shouldn't there be some kind of bound on how much larger the actual running time is?

DaoWen: I can add another explicit assertion for that, but it's already checked by this call a few lines above: `util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_ms)`

shamsimam: I'd prefer another assertion on `actual_running_time_ms`

DaoWen:
@shamsimam - Feedback addressed.

integration/tests/cook/test_basic.py
```diff
@@ -213,19 +215,34 @@ def test_max_runtime_exceeded(self):
...
+            job = util.wait_for_end_time(self.cook_url, job_uuid)
+            job_details = f"Job details: {json.dumps(job, sort_keys=True)}"
+            self.assertEqual(1, len(job['instances']), job_details)
+            instance = job['instances'][0]
+            self.assertEqual('failed', instance['status'], job_details)
+            valid_reasons = [# we killed the job because it took too long
+                             CookTest.REASON_MAX_RUNTIME_EXCEEDED,
+                             # the job was killed before the executor was properly registered
+                             CookTest.REASON_EXECUTOR_UNREGISTERED,
+                             # the job didn't die gracefully when we terminated it
```
dposada: Why would this happen?

integration/tests/cook/test_basic.py
```diff
@@ -213,19 +215,34 @@ def test_max_runtime_exceeded(self):
         job_timeout_interval_ms = job_timeout_interval_seconds * 1000
```
dposada: Should these just be `job_sleep_seconds` and `job_sleep_ms`? I kept on having to scroll up to check what these were actually doing

integration/tests/cook/test_basic.py
```diff
@@ -213,19 +215,34 @@ def test_max_runtime_exceeded(self):
...
-            self.assertEqual(-15, job['instances'][0]['exit_code'])
-            self.assertTrue(bool(job['instances'][0]['sandbox_directory']))
+        try:
+            job_uuid, resp = util.submit_job(self.cook_url,
+                                             command=f'sleep {job_timeout_interval_seconds}',
+                                             executor=job_executor_type, max_runtime=max_runtime_ms)
+            self.assertEqual(201, resp.status_code, msg=resp.content)
+            util.wait_for_job(self.cook_url, job_uuid, 'running')
+            util.wait_for_job(self.cook_url, job_uuid, 'completed', job_timeout_interval_ms)
+            job = util.wait_for_end_time(self.cook_url, job_uuid)
```
dposada: Do we need to wait for running, then wait for completed, then wait for end time? Seems like we could just have the wait for end time, no?

DaoWen: The wait-for-completed has a timeout. My worry is that if there's a delay scheduling the job, then the timeout could fail and give us a false-negative. There's also a weird delay waiting for the `end_time` to show up... Not sure what causes that. This is more verbose, but it seemed more robust. If we just had a single timeout, it wouldn't be clear if it failed because 1) the job was delayed being scheduled, 2) the job wasn't cancelled when we expected, or 3) the `end_time` didn't show up in a timely manner.

integration/tests/cook/test_basic.py
DaoWen: This is one of the exit codes reported in #299. I'm actually thinking now that we shouldn't accept either of these other reasons until we figure out why they happen. Since the assertions now print the full job details string, we should have some extra info for debugging this if they come up again. However, if these were happening because of a scheduling delay issue (which is the only reason that I can think of where we'd be getting the `REASON_EXECUTOR_UNREGISTERED` error), then the more robust wait call chain we're using now may resolve this.

dposada: :+1:

DaoWen:
@dposada - Feedback addressed.

integration/tests/cook/test_basic.py
dposada: What do you think about adding some of the explanation above into comments for the wait for running, wait for completed, wait for end time lines?

DaoWen:
@dposada - I added the comment explaining why we have three consecutive wait calls.

PR #500, opened 2017-10-25
Add single-test.sh script for integration testing

DaoWen:
I find this utility script useful for debugging failed tests.

### DIFF
--- a/integration/bin/single-test.sh
+++ b/integration/bin/single-test.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# Usage: ./bin/single-test.sh TEST
+# Locates the specified TEST method name,
+# then builds, runs and times the nosetests command for that single test.
+# Sample usage: ./bin/single-test.sh test_basic_submit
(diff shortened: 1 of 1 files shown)


integration/bin/single-test.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "Test not found: $test_name" >&2
+    exit 1
+elif [ -z "$test_class" ]; then
+    echo "Test class not found for $test_name in $test_file" >&2
+    exit 1
+fi
+
+echo "Found $test_name in $test_file:$test_class"
+
+time nosetests --verbosity 3 --tests "$test_file:$test_class.$test_name"
```
dposada: I am also a fan of using `--nologcapture` when running a single test, what do you think?

DaoWen: I wasn't aware of that option! I'll add it.

DaoWen: Done!

PR #501, opened 2017-10-26
Configurable progress file in cook executor

shamsimam:


executor/cook/config.py
```diff
@@ -49,12 +49,28 @@ def initialize_config(environment):
...
+    sandbox_directory = environment.get('MESOS_SANDBOX', '')
+    default_progress_output_name = '{}.progress'.format(executor_id)
+    if sandbox_directory:
+        default_progress_output_file = '{}/{}'.format(sandbox_directory, default_progress_output_name)
+    else:
+        default_progress_output_file = default_progress_output_name
+
+    if 'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV' in environment:
+        progress_output_file_env = environment.get('EXECUTOR_PROGRESS_OUTPUT_FILE_ENV')
+        logging.info('Progress location environment variable is {}'.format(progress_output_file_env))
```
sradack: We can remove the duplication and move this after the `if`

executor/cook/config.py
```diff
@@ -49,12 +49,28 @@ def initialize_config(environment):
...
+        if progress_output_file_env not in environment:
```
sradack: Can we make this warning apply to both cases?

executor/cook/executor.py
```diff
@@ -218,7 +222,46 @@ def get_task_state(exit_code):
...
+    Nothing.
+    """
+    if key not in environment or environment[key] != value:
+        logging.info('Setting process environment[{}]={}'.format(key, value))
+        environment[key] = value
+
+
+def retrieve_process_environment(config, os_environ):
+    """Prepares the environment for the subprocess.
+    The function also ensures that env[EXECUTOR_PROGRESS_OUTPUT_FILE_ENV]
```
sradack: Can we make this comment more clear?

executor/cook/config.py
```diff
@@ -49,12 +49,28 @@ def initialize_config(environment):
     """Initializes the config using the environment.
     Populates the default values for missing environment variables.
     """
+    executor_id = environment.get('MESOS_EXECUTOR_ID', 'executor')
+    sandbox_directory = environment.get('MESOS_SANDBOX', '')
+    default_progress_output_name = '{}.progress'.format(executor_id)
+    if sandbox_directory:
+        default_progress_output_file = '{}/{}'.format(sandbox_directory, default_progress_output_name)
```
sradack: Use `os.path.join`

executor/cook/config.py
```diff
@@ -49,12 +49,28 @@ def initialize_config(environment):
     """Initializes the config using the environment.
     Populates the default values for missing environment variables.
     """
+    executor_id = environment.get('MESOS_EXECUTOR_ID', 'executor')
+    sandbox_directory = environment.get('MESOS_SANDBOX', '')
+    default_progress_output_name = '{}.progress'.format(executor_id)
+    if sandbox_directory:
```
sradack: When is this not true?

executor/cook/executor.py
```diff
@@ -218,7 +222,46 @@ def get_task_state(exit_code):
...
+    and PROGRESS_OUTPUT_FILE is set to config.progress_output_name in the
```
sradack: What is `PROGRESS_OUTPUT_FILE`?  It's not set in this function.

executor/cook/executor.py
```diff
@@ -218,7 +222,46 @@ def get_task_state(exit_code):
...
+        A dictionary representing the current environment.
+
+    Returns
+    -------
+    The environment dictionary for the subprocess.
+    """
+    environment = dict(os_environ)
+
+    progress_output_name = config.progress_output_name
+    custom_progress_output_file_env = environment.get('EXECUTOR_PROGRESS_OUTPUT_FILE_ENV', '')
```
sradack: We're already checking for `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV` in `config.py`.  The config could know the name of the env var to set with the progress output file, whether it's customized or not.  There need not be branching in this function.  It's responsibility can be to simply set the value of  configured environment variable to the correct value for the progress output file.

sradack: Do we set an environment variable to the progress output file if `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV` isn't set?

executor/cook/config.py
shamsimam: In the tests.

executor/cook/config.py
shamsimam: Done.

executor/cook/config.py
shamsimam: Changed. I previously wanted to output additional log only when `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV` is specified.

executor/cook/config.py
shamsimam: Done.

executor/cook/executor.py
shamsimam: Added documentation.

executor/cook/executor.py
shamsimam: Removed it, I hadn't updated the documentation after cleaning up the code.

executor/cook/executor.py
shamsimam: > Do we set an environment variable to the progress output file if EXECUTOR_PROGRESS_OUTPUT_FILE_ENV isn't set?

No. That is what I was doing with `PROGRESS_OUTPUT_FILE` previously.

shamsimam: > The config could know the name of the env var to set with the progress output file, whether it's customized or not. There need not be branching in this function. It's responsibility can be to simply set the value of configured environment variable to the correct value for the progress output file.

Even if config knows the name, the branching would then be if the name is a non-empty string before setting the configured environment. Is your goal to limit the exposure of the `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV` string?

integration/tests/cook/test_cli.py
```diff
@@ -872,7 +872,9 @@ def entry(name):
     def test_show_progress_message(self):
         executor = util.get_job_executor_type(self.cook_url)
         cp, uuids = cli.submit('echo "progress: 99 We\'re so close!"', self.cook_url,
-                               submit_flags=f'--executor {executor}')
+                               submit_flags=f'--executor {executor} '
+                                            f'--env EXECUTOR_PROGRESS_OUTPUT_FILE_ENV=PROGRESS_OUTPUT_FILE '
+                                            f'--env PROGRESS_OUTPUT_FILE=stdout')
```
dposada: The second and third lines don't need to be f-strings

executor/cook/executor.py
shamsimam: @sradack any thoughts on this?

integration/tests/cook/test_cli.py
shamsimam: Done.

sradack:
Rebase

shamsimam:
Rebased.

executor/cook/config.py
```diff
@@ -31,13 +32,15 @@ def parse_time_ms(time_string):
     def __init__(self,
                  max_bytes_read_per_line=1024,
                  max_message_length=512,
+                 progress_output_env_variable='EXECUTOR_PROGRESS_OUTPUT_FILE',
```
sradack: This default value is never used.

executor/cook/config.py
```diff
@@ -49,12 +52,24 @@ def initialize_config(environment):
...
+    default_progress_output_name = '{}.progress'.format(executor_id)
+    if sandbox_directory:
+        default_progress_output_file = os.path.join(sandbox_directory, default_progress_output_name)
+    else:
+        default_progress_output_file = default_progress_output_name
+
+    progress_output_env_variable = environment.get('EXECUTOR_PROGRESS_OUTPUT_FILE_ENV', 'EXECUTOR_PROGRESS_OUTPUT_FILE')
+    logging.info('Progress location environment variable is {}'.format(progress_output_env_variable))
+    if progress_output_env_variable not in environment:
+        logging.warning('No entry found for {} in the environment'.format(progress_output_env_variable))
```
sradack: Do we need to warn if this isn't set?  Most, if not all, of the time the user won't be setting this.

executor/cook/config.py
```diff
@@ -49,12 +52,24 @@ def initialize_config(environment):
...
+
     max_bytes_read_per_line = int(environment.get('EXECUTOR_MAX_BYTES_READ_PER_LINE', 4 * 1024))
     max_message_length = int(environment.get('EXECUTOR_MAX_MESSAGE_LENGTH', 512))
-    progress_output_name = environment.get('PROGRESS_OUTPUT_FILE', 'stdout')
+    progress_output_name = environment.get(progress_output_env_variable, default_progress_output_file)
```
sradack: Do we need to allow the user to set this?  What's the use case?

executor/cook/executor.py
```diff
@@ -271,14 +307,15 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
...
+        environment = retrieve_process_environment(config, os.environ)
+        process = launch_task(task, environment)
         if process:
             # task has begun running successfully
             update_status(driver, task_id, cook.TASK_RUNNING)
             cio.print_and_log('Forked command at {}'.format(process.pid))
         else:
             # task launch failed, report an error
-            logging.info('Error in launching task')
+            logging.warning('Error in launching task')
```
sradack: Why not `logging.error`?

executor/cook/progress.py
```diff
@@ -142,8 +142,12 @@ def tail(self, sleep_time_ms):
                 logging.debug('{} has not yet been created, sleeping {} ms'.format(self.target_file, sleep_time_ms))
                 time.sleep(sleep_param)
 
+            if not os.path.isfile(self.target_file):
```
sradack: Are we expecting the task to create this file?  Should we consider having the executor create it?

shamsimam:
Rebased. Awaiting green build.

executor/cook/config.py
shamsimam: The user will set this when they are not happy with the default. E.g. if the progress is being written to `stdout`.

executor/cook/progress.py
shamsimam: I think the task should be responsible for creating the file. E.g. when the file is `stdout` on Mesos 0.27, the executor doesn't have enough privileges to create it.

executor/cook/config.py
shamsimam: Changed it to `info`.

executor/cook/executor.py
shamsimam: Changed to `error`

executor/cook/config.py
shamsimam: It is used in `test_retrieve_process_environment`

executor/cook/executor.py
```diff
@@ -252,6 +254,40 @@ def get_task_state(exit_code):
...
+    Nothing.
+    """
+    if key not in environment or environment[key] != value:
+        logging.info('Setting process environment[{}]={}'.format(key, value))
+        environment[key] = value
+
+
+def retrieve_process_environment(config, os_environ):
+    """Prepares the environment for the subprocess.
+    The function also ensures that env[EXECUTOR_PROGRESS_OUTPUT_FILE_ENV] is set to config.progress_output_name.
```
pschorf: Should this comment refer to `config.progress_output_env_variable` instead of `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV`?

integration/tests/cook/test_basic.py
```diff
@@ -96,8 +96,11 @@ def test_failing_submit(self):
 
     def test_progress_update_submit(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
-        command = 'echo "progress: 25 Twenty-five percent"; sleep 1; exit 0'
-        job_uuid, resp = util.submit_job(self.cook_url, command=command, executor=job_executor_type, max_runtime=60000)
+        command = 'echo "progress: 25 Twenty-five percent in ${CUSTOM_PROGRESS_OUTPUT_FILE}"; sleep 1; exit 0'
+        job_uuid, resp = util.submit_job(self.cook_url, command=command,
+                                         env={'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV': 'CUSTOM_PROGRESS_OUTPUT_FILE',
```
pschorf: Should we just add these to the default job env?

executor/cook/executor.py
shamsimam: Done.

integration/tests/cook/test_basic.py
shamsimam: I'd rather the test configure it explicitly.

pschorf: Sounds good, just wanted to check

executor/cook/config.py
```diff
@@ -49,9 +52,22 @@ def initialize_config(environment):
     """Initializes the config using the environment.
     Populates the default values for missing environment variables.
     """
+    executor_id = environment.get('MESOS_EXECUTOR_ID', 'executor')
+    sandbox_directory = environment.get('MESOS_SANDBOX', '')
+    default_progress_output_name = '{}.progress'.format(executor_id)
```
dposada: Nitpick: `f'{executor_id}.progress'`?

executor/cook/config.py
dposada: Should we create a constant for `'EXECUTOR_PROGRESS_OUTPUT_FILE'` since we use it here and below?

executor/cook/config.py
dposada: Should we use f-strings throughout all the changes?

integration/tests/cook/test_cli.py
```diff
@@ -871,7 +871,9 @@ def entry(name):
     def test_show_progress_message(self):
         executor = util.get_job_executor_type(self.cook_url)
         cp, uuids = cli.submit('echo "progress: 99 We\'re so close!"', self.cook_url,
-                               submit_flags=f'--executor {executor}')
+                               submit_flags=f'--executor {executor} '
+                                            '--env EXECUTOR_PROGRESS_OUTPUT_FILE_ENV=PROGRESS_OUTPUT_FILE '
+                                            '--env PROGRESS_OUTPUT_FILE=stdout')
```
dposada: It's unfortunate that this test (and other similar ones) are getting more complex simply to post progress. We discussed offline the possibility of making the default progress output file configurable, so that in our test and local environments we can configure it to `stdout`.

executor/tests/test_config.py
```diff
@@ -57,17 +60,77 @@ def test_initialize_config_defaults(self):
...
         self.assertEqual(2500, config.progress_sample_interval_ms)
         self.assertEqual('/sandbox/location', config.sandbox_directory)
         self.assertEqual(4000, config.shutdown_grace_period_ms)
+
+    def test_initialize_config_custom_progress_file_without_sandbox(self):
+        environment = {'EXECUTOR_MAX_BYTES_READ_PER_LINE': '1234',
+                       'EXECUTOR_MAX_MESSAGE_LENGTH': '1024',
+                       'MESOS_EXECUTOR_SHUTDOWN_GRACE_PERIOD': '4secs',
+                       'OUTPUT_TARGET_FILE': 'progress.out',
+                       'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV': 'OUTPUT_TARGET_FILE',
```
dposada: For me, it's more clear to have the `_ENV` entry listed first, since it is kind of like a "pointer" to the actual file entry (occurs in multiple places)

scheduler/src/cook/mesos/task.clj
```diff
@@ -59,13 +59,17 @@
...
-   "PROGRESS_SAMPLE_INTERVAL_MS" progress-sample-interval-ms})
+  (let [progress-output-file (:job/progress-output-file job-ent)]
+    (cond-> (assoc environment
+              "EXECUTOR_LOG_LEVEL" log-level
+              "EXECUTOR_MAX_MESSAGE_LENGTH" max-message-length
+              "PROGRESS_REGEX_STRING" (:job/progress-regex-string job-ent default-progress-regex-string)
+              "PROGRESS_SAMPLE_INTERVAL_MS" progress-sample-interval-ms)
+            progress-output-file
+            (assoc "EXECUTOR_PROGRESS_OUTPUT_FILE" progress-output-file
+                   "EXECUTOR_PROGRESS_OUTPUT_FILE_ENV" "EXECUTOR_PROGRESS_OUTPUT_FILE"))))
```
dposada: For me, it's more clear to have the `_ENV` entry listed first, since it is kind of like a "pointer" to the actual file entry (occurs in multiple places)

executor/cook/config.py
shamsimam: We're using python 3.5 in the executor.

executor/tests/test_config.py
shamsimam: Sorted properly, which caused it to move above.

executor/cook/config.py
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 
 import logging
+import os
 
 from pymesos.utils import parse_duration
 
+EXECUTOR_PROGRESS_OUTPUT_FILE = 'EXECUTOR_PROGRESS_OUTPUT_FILE'
```
dposada: Maybe `DEFAULT_PROGRESS_FILE_ENV_VARIABLE` for the name of the constant? "executor" and "output" seem superfluous to me, while it might be nice to clarify that this is the default and that it's the name of an env variable

integration/tests/cook/test_basic.py
```diff
@@ -96,8 +96,11 @@ def test_failing_submit(self):
 
     def test_progress_update_submit(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
-        command = 'echo "progress: 25 Twenty-five percent"; sleep 1; exit 0'
-        job_uuid, resp = util.submit_job(self.cook_url, command=command, executor=job_executor_type, max_runtime=60000)
+        command = 'echo "progress: 25 Twenty-five percent in ${PROGRESS_FILE}" >> ${PROGRESS_FILE}; sleep 1; exit 0'
+        job_uuid, resp = util.submit_job(self.cook_url, command=command,
+                                         env={'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV': 'PROGRESS_FILE',
+                                              'PROGRESS_FILE': 'progress.txt'},
+                                         executor=job_executor_type, max_runtime=60000)
```
dposada: I thought we were going with the default staying as `stdout` in our test environments so these tests wouldn't need to become more complex

integration/tests/cook/test_cli.py
```diff
@@ -892,10 +892,43 @@ def test_show_progress_message(self):
...
+            self.assertIn("99 We are so close!", cli.decode(cp.stdout))
+            self.assertIn('Command exited with status 0', cli.decode(cp.stdout))
+            self.assertIn('Executor completed execution', cli.decode(cp.stdout))
+
+    def test_show_progress_message_custom_progress_file(self):
+        executor = util.get_job_executor_type(self.cook_url)
+        cp, uuids = cli.submit('\'touch progress.txt && '
+                               'echo "Hello World" >> progress.txt && '
+                               'echo "progress: 99 We are so close!" >> progress.txt && '
+                               'echo "Done" >> progress.txt\' ',
```
dposada: Nitpick: unneeded space at the end of the command?

integration/tests/cook/test_cli.py
```diff
@@ -892,10 +892,43 @@ def test_show_progress_message(self):
...
+                                            '--env PROGRESS_OUTPUT_FILE=progress.txt')
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        util.wait_for_job(self.cook_url, uuids[0], 'completed')
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        cp, jobs = cli.show_json(uuids, self.cook_url)
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        self.assertEqual(executor, jobs[0]['instances'][0]['executor'])
+        if executor == 'cook':
+            instance = util.wait_until(lambda: cli.show_json(uuids, self.cook_url)[1][0]['instances'][0],
+                                       lambda i: 'progress' in i and 'progress_message' in i)
```
dposada: Would it make sense to create a helper function for this since it's duplicated above?

scheduler/docs/configuration.adoc
```diff
@@ -288,8 +288,8 @@ An example configuration looks like:
 ```clojure
 {...
  :executor {:command "cook-executor"
-            :default-progress-output-file "stdout"
             :default-progress-regex-string "progress: (\d+)(?: )?(.*)"
+            :environment {"COOK_EXECUTOR" "1"}
```
dposada: What is this for?

scheduler/src/cook/components.clj
```diff
@@ -375,12 +375,17 @@
                                {:executor-config executor})
                      {})
                    (do
-                     (when (and (:uri executor) (nil? (get-in executor [:uri :value])))
-                       (throw (ex-info "Executor uri value is missing!" {:executor executor})))
+                     (when-let [environment (:environment executor)]
+                       (when-not (and (map? environment)
+                                      (every? string? (keys environment))
+                                      (every? string? (vals environment)))
+                         (throw (ex-info "Executor environment map a string to a string!" {:executor executor}))))
```
dposada: "must be a map from string to string"?

executor/cook/config.py
shamsimam: Renamed to `DEFAULT_PROGRESS_FILE_ENV_VARIABLE `

integration/tests/cook/test_basic.py
shamsimam: Some of the other tests (e.g. `test_multiple_progress_updates_submit` and `test_multiple_rapid_progress_updates_submit`) are relying on `stdout` being the default. These tests are checking if writing progress to another file also works.

integration/tests/cook/test_cli.py
shamsimam: Removed trailing space at the end of the command.

integration/tests/cook/test_cli.py
shamsimam: Created `__wait_for_progress_message`

scheduler/docs/configuration.adoc
shamsimam: It was an example. I have replaced it with `{"EXECUTOR_DEFAULT_PROGRESS_OUTPUT_NAME" "stdout"}` as this one we actually use now.

scheduler/src/cook/components.clj
shamsimam: Done.

shamsimam:
@dposada  build is green now.

PR #503, opened 2017-10-26
Adds the instance progress message to 'show'

dposada:


cli/cook/format.py
```diff
@@ -36,13 +36,16 @@ def format_instance_status(instance):
     """Formats the instance status field"""
     status_text = format_state(instance['status'])
     if 'reason_string' in instance:
```
shamsimam: It is possible for a task can have both `reason_string` and `progress` values? Do we want to avoid showing the progress in this scneario?

cli/cook/subcommands/submit.py
```diff
@@ -210,6 +210,8 @@ def register(add_parser, add_defaults):
     submit_parser.add_argument('--ports', help='number of ports to reserve for job', type=int)
     submit_parser.add_argument('--application-name', '-a', help='name of application submitting the job')
     submit_parser.add_argument('--application-version', '-v', help='version of application submitting the job')
+    submit_parser.add_argument('--executor', '-E', help='executor to use to run the job on the Mesos agent',
+                               choices=('cook', 'mesos'))
```
shamsimam: Should we also add the `progress-output-file` and `progress-regex-string` flags? I'm okay if you feel it needs to be a separate PR.

cli/cook/format.py
dposada: Good call. I made it so that this field can accommodate either/both:

Both:
image

Just `reason_string`:
image

Just progress:
image

dposada: Done

cli/cook/subcommands/submit.py
dposada: Created issue #508 for this

PR #507, opened 2017-10-26
Add sysouts to get output closer to Mesos' CommandExecutor

shamsimam:
The main change is that we now pipe the output from the subprocess that is launched.
We also wrap the stdout/stderr calls to print in a lock to avoid data races.

Sample output in the `stdout` file:
```
Cook Executor version 0.1.2.dev0

Starting task c8df9002-9a30-4a11-b484-1cc42b83ec1b

progress: 25 Twenty-five percent
progress: 50 Fifty percent
progress: Sixty percent invalid format
progress: 75 Seventy-five percent
progress: Eighty percent invalid format

Command exited with status 0 (pid: 6984)

Executor completed execution of c8df9002-9a30-4a11-b484-1cc42b83ec1b
```

### DIFF
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -192,16 +192,14 @@ def kill_task(process, shutdown_grace_period_ms):
         for i in range(loop_limit):
             time.sleep(0.01)
             if not is_process_running(process):
-                message = 'Command terminated with signal Terminated (pid: {})'.format(process.pid)
-                cio.printline_out(message, flush=True)
-                logging.info(message)
+                cio.print_and_log('Command terminated with signal Terminated (pid: {})'.format(process.pid),
+                                  flush=True)
                 break
         if is_process_running(process):
             logging.info('Process did not terminate, forcefully killing it')
             process.kill()
-            message = 'Command terminated with signal Killed (pid: {})'.format(process.pid)
-            cio.printline_out(message, flush=True)
-            logging.info(message)
+            cio.print_and_log('Command terminated with signal Killed (pid: {})'.format(process.pid),
+                              flush=True)
 
 
 def await_process_completion(stop_signal, process_info, shutdown_grace_period_ms):
@@ -267,7 +265,7 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
     Nothing
     """
     task_id = get_task_id(task)
-    cio.printline_out('Starting task {}'.format(task_id))
+    cio.print_and_log('Starting task {}'.format(task_id))
     try:
         # not yet started to run the task
         update_status(driver, task_id, cook.TASK_STARTING)
@@ -300,9 +298,8 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
         # propagate the exit code
         process, _, _ = process_info
         exit_code = process.returncode
-        message = 'Command exited with status {} (pid: {})'.format(exit_code, process.pid)
-        cio.printline_out(message, flush=True)
-        logging.info(message)
+        cio.print_and_log('Command exited with status {} (pid: {})'.format(exit_code, process.pid),
+                          flush=True)
 
         # await progress updater termination if executor is terminating normally
         if not stop_signal.isSet():
@@ -329,9 +326,7 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
     finally:
         # ensure completed_signal is set so driver can stop
         completed_signal.set()
-        message = 'Executor completed execution of {}'.format(task_id)
-        cio.printline_out(message, flush=True)
-        logging.info(message)
+        cio.print_and_log('Executor completed execution of {}'.format(task_id), flush=True)
 
 
 def run_mesos_driver(stop_signal, config):
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -301,18 +301,20 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
         cio.print_and_log('Command exited with status {} (pid: {})'.format(exit_code, process.pid),
                           flush=True)
 
+        # force send any available latest progress state
+        cp.force_send_progress_update(progress_watcher, progress_updater)
+
+        exit_message = json.dumps({'exit-code': exit_code, 'task-id': task_id})
+        send_message(driver, exit_message, config.max_message_length)
+
         # await progress updater termination if executor is terminating normally
         if not stop_signal.isSet():
             logging.info('Awaiting progress updater completion')
             progress_complete_event.wait()
             logging.info('Progress updater completed')
 
-        # force send the latest progress state if available
+        # force send the latest progress state after progress tracking completion
         cp.force_send_progress_update(progress_watcher, progress_updater)
-
-        # send the exit code after the last progress message has been sent
-        exit_message = json.dumps({'exit-code': exit_code, 'task-id': task_id})
-        send_message(driver, exit_message, config.max_message_length)
 
         # task either completed successfully or aborted with an error
         task_state = get_task_state(exit_code)
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -106,8 +106,7 @@ def launch_task(task, max_bytes_read_per_line):
 
     Returns
     -------
-    When command is provided and a process can be started, the process launched along
-    with the two threads that are piping the stdout and stderr.
+    When command is provided and a process can be started, the process launched.
     Else it logs the reason and returns None.
     """
     try:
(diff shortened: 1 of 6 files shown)


sradack:
A general impression I have is that we are dumping too much data into stdout.  We should only lightly be touching `stdout`.  `stdout` should mainly be the output of the program, not the executor.

executor/cook/progress.py
```diff
@@ -150,16 +150,19 @@ def tail(self, sleep_time_ms):
...
                     # exit if program has completed and there are no more lines to read
                     if self.task_completed_signal.isSet():
-                        logging.info('Done processing file for progress messages')
+                        cio.print_and_log('Done processing progress messages, {} lines read'.format(
+                            line_index))
                         break
                     # no new line available, sleep before trying again
                     time.sleep(sleep_param)
                     continue
+                line_index += 1
```
sradack: We're assuming that `readline` returns a whole line, which may not be true.

executor/tests/test_executor.py
```diff
@@ -463,3 +482,35 @@ def sleep_and_set_stop_signal():
...
+            self.assertEqual(2, len(driver.messages))
+
+            actual_encoded_message_0 = driver.messages[0]
+            expected_message_0 = {'sandbox-directory': sandbox_directory, 'task-id': task_id}
+            assert_message(self, expected_message_0, actual_encoded_message_0)
+
+            actual_encoded_message_1 = driver.messages[1]
+            expected_message_1 = {'exit-code': 0, 'task-id': task_id}
+            assert_message(self, expected_message_1, actual_encoded_message_1)
+
```
sradack: We're missing the part where we check if the stdout file has the expected string.

shamsimam:
@sradack reduced the outputs to stdout

executor/cook/executor.py
```diff
@@ -95,17 +94,20 @@ def send_message(driver, message, max_message_length):
...
+    task: dictionary
         The task to execute.
+    max_bytes_read_per_line: int
+        The maximum number of bytes to read per call to readline().
 
     Returns
     -------
-    When command is provided and a process can be started, the process launched.
+    When command is provided and a process can be started, the process launched along
+    with the two threads that are piping the stdout and stderr.
```
sradack: This function doesn't launch two threads.

executor/cook/executor.py
```diff
@@ -229,36 +265,43 @@ def manage_task(driver, task, stop_signal, completed_signal, config, stdout_name
...
                                               config.progress_sample_interval_ms, send_message)
         progress_complete_event = cp.launch_progress_tracker(progress_watcher, progress_updater)
 
-        await_process_completion(stop_signal, process, config.shutdown_grace_period_ms)
+        process_info = process, stdout_thread, stderr_thread
+        await_process_completion(stop_signal, process_info, config.shutdown_grace_period_ms)
         task_completed_signal.set()
 
         # propagate the exit code
+        process, _, _ = process_info
```
sradack: Why overwrite the existing `process`?

executor/cook/io_helper.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Returns
+    -------
+    Nothing.
+    """
+    message = '{}{}'.format(string_data, os.linesep) if newline else string_data
+    with __stdout_lock__:
+        sys.stdout.write(message)
+    if flush:
+        sys.stdout.flush()
```
sradack: Why shouldn't flush be protected by the the lock?

executor/cook/io_helper.py
```diff
@@ -0,0 +1,143 @@
...
+    flush: boolean
+        Flag determining whether to trigger a sys.stderr.flush()
+    newline: boolean
+        Flag determining whether to output a newline at the end
+
+    Returns
+    -------
+    Nothing.
+    """
+    message = '{}{}'.format(string_data, os.linesep) if newline else string_data
```
sradack: Would it be better to `sys.stderr.write(os.linesep) if newline`, instead of building a string every time?

executor/cook/io_helper.py
```diff
@@ -0,0 +1,143 @@
...
+    -------
+    Nothing.
+    """
+    print_and_log('Starting to pipe {}'.format(label))
+    try:
+        while True:
+            line = out_file.readline(max_bytes_read_per_line)
+            if not line:
+                break
+            out_fn(line.decode('utf-8'), newline=False)
```
sradack: Do we have to assume an encoding?

executor/cook/io_helper.py
```diff
@@ -0,0 +1,143 @@
...
+        flush_fn()
+        print_and_log('Done piping {}'.format(label))
+    except Exception:
+        logging.exception('Error in process_output of {}'.format(label))
+        print_out('Error in process_output of {}{}'.format(label, os.linesep), flush=True)
```
sradack: Why doesn't this use the `newline` option?

executor/cook/executor.py
```diff
@@ -117,20 +119,24 @@ def launch_task(task):
             logging.warning('No command provided!')
             return None
 
-        process = subprocess.Popen(command, shell=True, stdout=sys.stdout.fileno(), stderr=sys.stderr.fileno())
+        process = subprocess.Popen(command,
+                                   bufsize=2*max_bytes_read_per_line,
```
sradack: Could we be slowing down processes that block waiting for the executor to take data off the buffer?  Should we consider a larger buffer?

shamsimam: Does `10* max_bytes_read_per_line` sound reasonable? The default is `io.DEFAULT_BUFFER_SIZE` which on my system is `8192`. `max_bytes_read_per_line` defaults to `4096`.

executor/cook/executor.py
shamsimam: Removed.

executor/cook/io_helper.py
shamsimam: I want to reduce the work done in the critical section. Do you see any harm in having it outside and potentially running flush in parallel?

executor/cook/io_helper.py
shamsimam: Yes. Changed.

executor/cook/io_helper.py
shamsimam: Removed the `encoding` argument. It will however default to `utf-8` (https://docs.python.org/3/library/stdtypes.html#bytes.decode)

executor/cook/progress.py
shamsimam: Added a check for `line.endswith(os.linesep)`

executor/tests/test_executor.py
shamsimam: Added assertion on the string being there in the output.

executor/cook/io_helper.py
shamsimam: The print statement has been removed.

executor/cook/executor.py
sradack: I played around with `bufsize` locally.  When the subprocess is writing sufficiently large data (at least 64k, which is the default pipe buffer on my Linux system), the subprocess blocked no matter how big this buffer was, even if I set it to something huge like 1MB.  When the data is small (less than 64k), no blocking occurs, no matter the size of bufsize.

My conclusion is that `bufsize` should be set sufficiently large enough to transfer data from the pipe buffer to python's buffer without degrading performance.  We probably can avoid setting `bufsize` entirely, since `4096` seems like a reasonable default.

executor/cook/io_helper.py
sradack: I couldn't get two `sys.stdout.write` calls to interleave.  If the newlines were inside the buffer being written, we might be able to avoid these locks entirely.

I can't find any docs indicating whether `flush()` is thread-safe or not.  Since we're perhaps being overprotective with the write locks, we could also be overprotective on the `flush()` calls.  It still feels odd to me that one is protected and one isn't.

sradack: Maybe `sys.stdout` calls aren't thread-safe after all:

<pre>
>>> import sys
>>> sys.stdout
<_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
>>>
</pre>

From https://docs.python.org/3/library/io.html:

> 16.2.4.3. Multi-threading
FileIO objects are thread-safe to the extent that the operating system calls (such as read(2) under Unix) they wrap are thread-safe too.

> Binary buffered objects (instances of BufferedReader, BufferedWriter, BufferedRandom and BufferedRWPair) protect their internal structures using a lock; it is therefore safe to call them from multiple threads at once.

> TextIOWrapper objects are not thread-safe.

executor/cook/executor.py
sradack: What do you think about using a fixed size `Queue` and having one dedicated writer for `stdout` and `stderr`?

https://docs.python.org/3/library/queue.html

shamsimam: We discussed offline and will not use the queue. I'll remove the default buffer size.

PR #509, opened 2017-10-27
Adds kill command to the CLI

dposada:


cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+            duplicate_uuids.append(uuid)
+        else:
+            uuid_to_entries[uuid] = [entry_map]
+
+    for cluster_name, entities in query_result['clusters'].items():
+        jobs = entities['jobs']
+        for job in jobs:
+            add(job['uuid'], 'job', cluster_name)
+
+        instances = [i for j in entities['instances'] for i in j['instances'] if i['task_id'] in uuids]
```
DaoWen: This looks like a typo: `for j in entities['instances']`
... but apparently it isn't? It would be nice if `entities['instances']` was actually a sequence of instances, and there could be a more clearly named member for the instance-jobs sequence being used here.

Also, I'd prefer having this expanded into an imperative nested for-loop.

This is irrelevant if you expand this into nested loops, but on a site note: I think it's usually more efficient to use a generator than a list comprehension if you don't actually need a list (i.e., swapping the `[ ... ]` with `( ... )` should result in slightly faster code).

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+    if len(duplicate_uuids) > 0:
+        messages = []
+        for duplicate_uuid in duplicate_uuids:
+            bullets = []
+            for entry in uuid_to_entries[duplicate_uuid]:
+                bullets.append(f'- as a {entry["type"]} on {entry["cluster_name"]}')
+            message = f'{duplicate_uuid} is duplicated:\n' + '\n'.join(bullets)
+            messages.append(message)
+        details = '\n\n'.join(messages)
+        message = f'Unable to kill due to duplicate UUIDs.\n\n{details}'
```
DaoWen: "Refusing to kill" might be a more accurate description.

It would also be good to tell the user how to get around this issue (i.e., tell them to explicitly set the cluster where they want to kill the duplicated entity).

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+        raise Exception(message)
+
+
+def kill_entities(uuids, query_result, clusters):
+    """Attempts to kill the jobs / instances / groups with the given UUIDs"""
+    exit_code = 0
+    for cluster_name, entities in query_result['clusters'].items():
+        cluster = next(c for c in clusters if c['name'] == cluster_name)
```
DaoWen: You could build a dict for this outside the loop:

```python
clusters_by_name = { c['name']: c for c in clusters }
for cluster_name, entities in query_result['clusters'].items():
    cluster = clusters_by_name[cluster_name]
```

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+
+        jobs = entities['jobs']
+        instances = [i for j in entities['instances'] for i in j['instances'] if i['task_id'] in uuids]
```
DaoWen: The confusing `for j in entities['instances']` is repeated here.

Even if you don't refactor the `entities` data structure, I'd recommend at least pulling this out into a separate function to avoid repeating this complicated comprehension twice.

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+            resp = http.delete(cluster, 'group', params={'uuid': group_uuids})
+            if resp.status_code == 204:
+                for group in groups:
+                    print_info(f'Killed job group {colors.bold(group["uuid"])} on {colors.bold(cluster_name)}.')
+            else:
+                exit_code = 1
+                for group in groups:
+                    print(colors.failed(f'Failed to kill job group {group["uuid"]} on {cluster_name}.'))
+
+    return exit_code
```
DaoWen: If you killed a lot of jobs across several clusters, some early "Failed" messages might not be visible by the time the job ends. It might be good to add another `print` here when `exit_code` is non-zero to remind the user that there were failures.

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+    """Attempts to kill the jobs / instances / groups with the given UUIDs."""
+    uuids = strip_all(args.get('uuid'))
+    query_result = query(clusters, uuids)
+    if query_result['count'] == 0:
+        print_no_data(clusters)
+        return 1
+
+    # If the user provides UUIDs that map to more than one entity,
+    # we will raise an Exception that contains the details
+    guard_against_duplicates(uuids, query_result)
```
DaoWen: I think it would be better to catch the exception thrown in this method, print the error message, and return a non-zero exit code. I know our users are probably going to be disappointed they didn't get to see the stack trace, but they'll get over it eventually...

integration/tests/cook/test_cli.py
```diff
@@ -79,9 +79,8 @@ def test_wait_for_multiple_jobs(self):
         cp = cli.wait(uuids, self.cook_url)
         self.assertEqual(0, cp.returncode, cp.stderr)
         cp, jobs = cli.show_json(uuids, self.cook_url)
-        self.assertEqual(uuids[0], jobs[0]['uuid'])
-        self.assertEqual(uuids[1], jobs[1]['uuid'])
-        self.assertEqual(uuids[2], jobs[2]['uuid'])
+        returned_uuids = [j['uuid'] for j in jobs]
+        self.assertEqual(sorted(uuids), sorted(returned_uuids))
```
DaoWen: I think this sorting can go away if you maintain the order of the uuids internally.

integration/tests/cook/test_cli.py
```diff
@@ -896,3 +895,105 @@ def test_submit_with_executor(self):
...
+        cp, jobs = cli.show_json(uuids, self.cook_url)
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        duplicate_uuid = jobs[0]['instances'][0]['task_id']
+        cp, uuids = cli.submit('ls', self.cook_url, submit_flags=f'--group {duplicate_uuid}')
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        cp = cli.kill([duplicate_uuid], self.cook_url)
+        self.assertEqual(1, cp.returncode, cp.stderr)
+        self.assertIn('Unable to kill due to duplicate UUIDs', cli.decode(cp.stderr))
+        self.assertIn('as a job group', cli.decode(cp.stderr))
+        self.assertIn('as a job instance', cli.decode(cp.stderr))
```
DaoWen: What happens when you have a duplicate of the same type (e.g., two jobs) on two different clusters?

Is there a simple way of testing 3 clusters? I'm thinking not, since that would require bringing up a third cluster for all of these tests, which is probably not worth it... If that's the case, I'd suggest adding some unit tests for handling a response with duplicate entries on 3 or more clusters. A unit test for this case would have caught the repeated entries in the `duplicate_uuids` that I flagged above.

cli/cook/http.py
```diff
@@ -64,6 +70,14 @@ def get(cluster, endpoint, params):
     return resp
 
 
+def delete(cluster, endpoint, params):
+    """DELETEs data corresponding to the given params on cluster at /endpoint"""
+    url = __make_url(cluster, endpoint)
+    resp = __delete(url, params)
+    logging.info('response from cook: %s' % resp.text)
```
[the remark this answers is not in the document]
dposada: Good idea, done

cli/cook/querying.py
```diff
@@ -102,6 +102,10 @@ def query(clusters, uuids, pred_jobs=None, pred_instances=None, pred_groups=None
     Uses query_across_clusters to make the /rawscheduler
     requests in parallel across the given clusters
     """
+    # Cook will give us back two copies if the user asks for the same UUID twice, e.g.
+    # $ cs show d38ea6bd-8a26-4ddf-8a93-5926fa2991ce d38ea6bd-8a26-4ddf-8a93-5926fa2991ce
+    # Prevent this by creating a set:
+    uuids = set(uuids)
```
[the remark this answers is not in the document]

integration/tests/cook/test_cli.py
dposada: Yep, done

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+        return
+
+    uuid_to_entries = {}
+    duplicate_uuids = []
+
+    def add(uuid, entity_type, cluster):
+        entry_map = {'type': entity_type, 'cluster_name': cluster}
+        if uuid in uuid_to_entries:
+            uuid_to_entries[uuid].append(entry_map)
+            duplicate_uuids.append(uuid)
```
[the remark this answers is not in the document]
dposada: Fixed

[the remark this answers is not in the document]
dposada: Fixed

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def guard_against_duplicates(uuids, query_result):
+    """
+    Checks for UUIDs that are duplicated and throws if any are found (unlikely to happen in the wild, but it could)
+    """
+    if query_result['count'] == 1:
+        return
+
+    uuid_to_entries = {}
```
pschorf: You can use a `defaultdict` here to avoid the if/else logic in `add`

cli/cook/subcommands/kill.py
pschorf: I also prefer the nested loops to the list comprehension here.

cli/cook/subcommands/kill.py
```diff
@@ -0,0 +1,109 @@
...
+            if resp.status_code == 204:
+                for job in jobs:
+                    print_info(f'Killed job {colors.bold(job["uuid"])} on {colors.bold(cluster_name)}.')
+                for instance in instances:
+                    instance_uuid = instance['task_id']
+                    print_info(f'Killed job instance {colors.bold(instance_uuid)} on {colors.bold(cluster_name)}.')
+            else:
+                exit_code = 1
+                for job in jobs:
+                    print(colors.failed(f'Failed to kill job {job["uuid"]} on {cluster_name}.'))
```
pschorf: Should we do this on stderr instead of stdout?

cli/cook/subcommands/kill.py
dposada: :+1: I moved this code to the `query_entities` function, because it was being used by other subcommands too. I also changed it to loops.

cli/cook/subcommands/kill.py
dposada: Done and done

cli/cook/subcommands/kill.py
dposada: Done

cli/cook/subcommands/kill.py
dposada: See other comment

cli/cook/subcommands/kill.py
dposada: The exception gets caught in the `main` function, which just prints the error message and gives a non-zero exit code:

```bash
$ cs kill $uuid
Refusing to kill due to duplicate UUIDs.

c3ac6e2d-dc2f-41ed-a3bd-4c2215459869 is duplicated:
- as a job on dev0
- as a job group on dev0

You might need to explicitly set the cluster where you want to kill by using the --cluster flag.
```

integration/tests/cook/test_cli.py
dposada: This is what a duplicate of the same type on two clusters looks like:

```bash
$ uuid=$(uuidgen)
$ cs --cluster dev0 submit --uuid $uuid ls
Attempting to submit on dev0 cluster...
Job submission succeeded on dev0. Your job UUID is:
3c9ea901-664e-47ec-b00f-5cf3ded3b20d
$ cs --cluster dev1 submit --uuid $uuid ls
Attempting to submit on dev1 cluster...
Job submission succeeded on dev1. Your job UUID is:
3c9ea901-664e-47ec-b00f-5cf3ded3b20d
$ cs kill $uuid
Refusing to kill due to duplicate UUIDs.

3c9ea901-664e-47ec-b00f-5cf3ded3b20d is duplicated:
- as a job on dev0
- as a job on dev1

You might need to explicitly set the cluster where you want to kill by using the --cluster flag.
```

dposada: I also added another test case which duplicates a UUID 3 times - for a job, an instance, and a group. This new test case would have exposed the bug you identified.

cli/cook/subcommands/kill.py
dposada: We could, but it seems like the failed lines and success lines should go to the same place (either both to stdout or both to stderr)

pschorf: Makes sense to me

cli/cook/subcommands/kill.py
dposada: Nice, done

PR #512, opened 2017-10-27
Add support for groups on retry endpoint

DaoWen:


scheduler/src/cook/mesos/api.clj
```diff
@@ -1369,27 +1370,51 @@
       (when-let [uncoerced-query-params (get-in ctx [:request :query-params "job"])]
         (mapv #(UUID/fromString %) (vectorize uncoerced-query-params)))
       (get-in ctx [:request :body-params :jobs])
-      [(get-in ctx [:request :body-params :job])]))
+      (vectorize (get-in ctx [:request :body-params :job]))))
```
dposada: Was this broken without the `vectorize`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1369,27 +1370,51 @@
...
         jobs-not-existing (remove (partial job-exists? (d/db conn)) jobs)]
-    (if (seq jobs-not-existing)
+    (cond
+      ; error case
+      (seq jobs-not-existing)
       [false {::error (->> jobs-not-existing
                            (map #(str "UUID " % " does not correspond to a job."))
                            (str/join \space))}]
-      [true {::jobs jobs}])))
+      ; if the context doesn't already have ::jobs, add it
```
dposada: Nitpick: I would favor a newline separating the condition/action pairs

scheduler/src/cook/mesos/api.clj
```diff
@@ -1369,27 +1370,51 @@
       (when-let [uncoerced-query-params (get-in ctx [:request :query-params "job"])]
         (mapv #(UUID/fromString %) (vectorize uncoerced-query-params)))
       (get-in ctx [:request :body-params :jobs])
-      [(get-in ctx [:request :body-params :job])]))
+      (vectorize (get-in ctx [:request :body-params :job]))))
 
 (defn check-jobs-exist
   [conn ctx]
-  (let [jobs (jobs-from-request ctx)
+  (let [jobs (or (::jobs ctx) (jobs-from-request ctx))
```
dposada: Should we instead modify `jobs-from-request`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1350,6 +1350,7 @@
 (def UpdateRetriesRequest
   {(s/optional-key :job) s/Uuid
    (s/optional-key :jobs) [s/Uuid]
+   (s/optional-key :groups) [s/Uuid]
```
dposada: I thought we decided to add a flag for only retrying non-completed jobs in the group

DaoWen: I'm going to do that as a follow-on. Thanks for the reminder though!

scheduler/src/cook/mesos/api.clj
DaoWen: It wasn't broken before the addition of the `:groups` parameter, but now it's possible for both `:job` and `:jobs` to be empty, in which case we want an empty vector to return here, not `[nil]`.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1350,6 +1350,7 @@
...
    (s/optional-key :retries) PosNum
    (s/optional-key :increment) PosNum})
```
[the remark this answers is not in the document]
DaoWen: We talked about this option offline, but I want to note that we don't actually have any integration tests that use `:increment`. (We have some unit tests, but not integration tests.)

scheduler/src/cook/mesos/api.clj
DaoWen: I like that a lot better. Done.

scheduler/src/cook/mesos/api.clj
DaoWen: Done.

integration/tests/cook/test_basic.py
```diff
@@ -818,28 +818,57 @@ def test_group_kill_multi(self):
...
+            job_data = util.query_jobs(self.cook_url, job=jobs)
+            self.assertEqual(200, job_data.status_code)
+            for job in job_data.json():
+                if job['status'] not in ['waiting', 'running']:
+                    job_details = f'Job details: {json.dumps(job, sort_keys=True)}'
+                    self.assertEqual('completed', job['status'], job_details)
+                    self.assertEqual('success', job['state'], job_details)
+        finally:
+            # ensure that we don't leave a bunch of jobs running/waiting
+            util.kill_groups(self.cook_url, [group_uuid])
```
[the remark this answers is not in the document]
DaoWen: I broke this test into two different cases: `test_group_change_killed_retries` and `test_group_change_completed_retries`. The latter tests a mixed case, where some are running, some are completed, and some are still waiting. I'll add a couple more variants when I implement the "failed-only" flag for retries.

DaoWen:
@dposada - I addressed your feedback, and Travis finally went green!

integration/tests/cook/test_basic.py
```diff
@@ -869,6 +869,94 @@ def query():
...
             util.kill_groups(self.cook_url, [group_uuid])
 
+    def test_group_change_killed_retries(self):
+        group_spec = self.minimal_group()
+        group_uuid = group_spec['uuid']
+        job_spec = {'group': group_uuid, 'command': f'sleep 1'}
+        try:
+            jobs, resp = util.submit_jobs(self.cook_url, job_spec, 10, groups=[group_spec])
+            self.assertEqual(resp.status_code, 201)
+            # wait for some job to start
```
dposada: Nitpick: misplaced comment?

integration/tests/cook/test_basic.py
```diff
@@ -869,6 +869,94 @@ def query():
...
+            util.wait_until(group_query, util.group_some_job_started)
+            # ensure none of the jobs are still in a failed state
+            job_data = util.query_jobs(self.cook_url, job=jobs)
+            self.assertEqual(200, job_data.status_code)
+            for job in job_data.json():
+                if job['status'] not in ['waiting', 'running']:
+                    job_details = f'Job details: {json.dumps(job, sort_keys=True)}'
+                    self.assertEqual('completed', job['status'], job_details)
+                    self.assertEqual('success', job['state'], job_details)
```
dposada: Nitpick: I think this would be more clear if we asserted that `job['status']` is either waiting, running, or completed, instead of only having the `assertEquals` if the status is not waiting or running

integration/tests/cook/test_basic.py
```diff
@@ -869,6 +869,94 @@ def query():
...
+            util.kill_groups(self.cook_url, [group_uuid])
+
+    def test_group_change_completed_retries(self):
+        group_spec = self.minimal_group()
+        group_uuid = group_spec['uuid']
+        job_spec = {'group': group_uuid, 'command': f'sleep 1'}
+        try:
+            jobs, resp = util.submit_jobs(self.cook_url, job_spec, 10, groups=[group_spec])
+            self.assertEqual(resp.status_code, 201)
+            # wait for some job to start
```
dposada: Nitpick: misplaced comment?

integration/tests/cook/test_basic.py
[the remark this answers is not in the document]
dposada: Does `test_group_change_completed_retries` actually ensure that some jobs are running and some are still waiting when it does the group retry? It seems like it only ensures that at least one job has completed (but it could be that all are completed, or that the non-completed ones are all waiting or all running). I'm not necessarily advocating for making the test more complex, but maybe we should make its intent more clear with a comment. Also, the name is a little confusing.

scheduler/src/cook/mesos/api.clj
dposada: Cool, do we have an issue for that?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1433,14 +1471,17 @@
...
-      [false {::error (->> unauthorized-jobs
-                           (map #(str "You are not authorized to retry job " % "."))
-                           (str/join \space))}]
-      true)))
+        unauthorized-job? #(not (user-can-retry-job? conn is-authorized-fn request-user %))]
+    (or (first (for [guuid (::groups ctx)
+                     :let [group (d/entity (db conn) [:group/uuid guuid])
+                           job (first (:group/job group))]
+                     :when (some-> job unauthorized-job?)]
+                 [false {::error (str "You are not authorized to retry jobs from group " guuid ".")}]))
```
dposada: Should we add some tests for attempting to retry a group that the user is not authorized to retry?

scheduler/src/cook/mesos/api.clj
DaoWen: It's in JIRA.

integration/tests/cook/test_basic.py
DaoWen: This definitely looks out of place after adding the blank lines around he def. I moved it down to just before the corresponding `wait_until` call.

integration/tests/cook/test_basic.py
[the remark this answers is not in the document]
DaoWen: I'll change the name to `test_group_change_retries`, and add a comment to clarify what the expected state is after the `wait_until` call with the some-job-completed predicate.

DaoWen:
@dposada - I added the unit test cases for `:retry`. I think this is ready to merge.

scheduler/test/cook/test/authorization.clj
```diff
@@ -33,6 +33,8 @@
     (is (true? (auth/open-auth {} "foo" :create test-job)))
     (is (true? (auth/open-auth {} "bar" :read test-job)))
     (is (true? (auth/open-auth {} "baz" :update test-job)))
+    (is (true? (auth/open-auth {} "bazz" :get test-job)))
```
dposada: Do we need both `:get` and `:read` in all these tests?

DaoWen: Honestly, I have no idea what `:read` is... It looks like it's only used on `/queue` and `/running`, so it probably doesn't even make sense to test these verbs on jobs. I think the same applies for most of the other verbs here. But I'd prefer changing that in a separate PR since it's not retry-related (both the `:get` and `:retry` verbs are used on the `/retry` endpoint, which is why I added those).

PR #516, opened 2017-10-27
Do not trigger heartbeats for non-heartbeat messages

shamsimam:
Since the Cook Executor doesn't send heartbeat messages to the scheduler, the scheduler treats long running tasks as unreachable. 

The fix is to send heartbeat messages from the executor to the scheduler.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -279,19 +279,6 @@
   (-> (d/entity db [:instance/task-id task-id])
       :db/id))
 
-(defn- task-id->executor
-  "Retrieves the executor given a task-id"
-  [db task-id]
-  (when task-id
-    (-> (d/q '[:find ?executor
-               :in $ ?task-id
-               :where
-               [?i :instance/task-id ?task-id]
-               [?i :instance/executor ?e]
-               [?e :db/ident ?executor]]
-             db task-id)
-        ffirst)))
-
 (meters/defmeter [cook-mesos scheduler progress-aggregator-message-rate])
 (meters/defmeter [cook-mesos scheduler progress-aggregator-drop-rate])
 (counters/defcounter [cook-mesos scheduler progress-aggregator-drop-count])
@@ -1531,13 +1518,11 @@
                      )
     (framework-message [this driver executor-id slave-id message]
                        (try
-                         (let [{:strs [task-id] :as parsed-message} (json/read-str (String. ^bytes message "UTF-8"))
-                               executor (task-id->executor (d/db conn) task-id)]
-                           (when (some #(not (contains? #{"task-id" "timestamp"} %)) (keys parsed-message))
+                         (let [{:strs [task-id type] :as parsed-message} (json/read-str (String. ^bytes message "UTF-8"))]
+                           (if (= "heartbeat" type)
+                             (heartbeat/notify-heartbeat heartbeat-ch executor-id slave-id parsed-message)
                              (async-in-order-processing
-                               task-id #(handle-framework-message conn handle-progress-message parsed-message)))
-                           (when (not= :executor/cook executor)
-                             (heartbeat/notify-heartbeat heartbeat-ch executor-id slave-id parsed-message)))
+                               task-id #(handle-framework-message conn handle-progress-message parsed-message))))
                          (catch Exception e
                            (log/error e "Unable to process framework message"
                                       {:executor-id executor-id, :message message, :slave-id slave-id}))))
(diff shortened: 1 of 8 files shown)


wyegelwel:
The heartbeat namespace (and code) was for our old executor before reconciliation existed in mesos. Instead of this pr, you can have the heart beat namespace check that the executor is the old executor (or not the new cook executor). That way you don't need add more functionality. 

Once we move away from the old executor we can remove the heartbeat namespace and rely on reconciliation with mesos

shamsimam:
@wyegelwel thanks, will do.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1519,10 +1530,17 @@
...
                      )
     (framework-message [this driver executor-id slave-id message]
-                       ;; the executor-id is the same as the task-id (see cook.mesos.task/task-info->mesos-message)
-                       (async-in-order-processing (:value executor-id)
-                                                  #(handle-framework-message conn handle-progress-message message))
-                       (heartbeat/notify-heartbeat heartbeat-ch executor-id slave-id message))
+                       (try
+                         (let [{:strs [task-id] :as parsed-message} (json/read-str (String. ^bytes message "UTF-8"))
+                               executor (task-id->executor (d/db conn) task-id)]
+                           (when (some #(not (contains? #{"task-id" "timestamp"} %)) (keys parsed-message))
```
sradack: Can we be more explicit about what we're handling here?  Perhaps we need a `:message-type` field.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1519,10 +1530,17 @@
                      ;; TODO: Rescind the offer in fenzo
                      )
     (framework-message [this driver executor-id slave-id message]
-                       ;; the executor-id is the same as the task-id (see cook.mesos.task/task-info->mesos-message)
-                       (async-in-order-processing (:value executor-id)
-                                                  #(handle-framework-message conn handle-progress-message message))
-                       (heartbeat/notify-heartbeat heartbeat-ch executor-id slave-id message))
+                       (try
+                         (let [{:strs [task-id] :as parsed-message} (json/read-str (String. ^bytes message "UTF-8"))
+                               executor (task-id->executor (d/db conn) task-id)]
```
sradack: We query the database without using the result when the message is a Cook Executor message.

sradack:
Why did we once implement heartbeats for custom executor, but leave alone jobs that used the standard executor?  Was it that we could take advantage of the custom executor's ability to send heartbeats to improve Cook Scheduler's awareness of problems with executors?

scheduler/src/cook/mesos/scheduler.clj
shamsimam: Isn't the result from the database being used to avoid forwarding the message to the `heartbeat` mechanism.

scheduler/src/cook/mesos/scheduler.clj
shamsimam: The executor is free to merge multiple values into a message. We are optimizing for filtering out heartbeat messages that only provide the timestamp. Would you be happy with a comment explicitly mentioning that we are filtering out messages with only heartbeat values?

shamsimam: Also, with the `message-type` field we will then need all executors (i.e. custom and mesoso along with cook executor) to send this field.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1519,10 +1517,15 @@
...
     (framework-message [this driver executor-id slave-id message]
-                       ;; the executor-id is the same as the task-id (see cook.mesos.task/task-info->mesos-message)
-                       (async-in-order-processing (:value executor-id)
-                                                  #(handle-framework-message conn handle-progress-message message))
-                       (heartbeat/notify-heartbeat heartbeat-ch executor-id slave-id message))
+                       (try
+                         (let [{:strs [task-id] :as parsed-message} (json/read-str (String. ^bytes message "UTF-8"))]
+                           (async-in-order-processing
+                             task-id #(handle-framework-message conn handle-progress-message parsed-message))
+                           (when (contains? parsed-message "timestamp")
```
sradack: Should we add some comments talking about the where we expect these messages to come from, why we have the condition, etc?

sradack:
We should update the name of this PR.

shamsimam:
> Did you get a chance to look at the custom executor?

The custom executor sets a `type` attribute on the message which is what I have ended up using to decide which function to call.

shamsimam:
> We should update the name of this PR.

Renamed.

shamsimam:
Rebased.

PR #523, opened 2017-10-31
Fixes bug with defaults for application name, version

dposada:


integration/tests/cook/test_cli.py
```diff
@@ -502,6 +502,15 @@ def test_submit_with_application(self):
         self.assertEqual(0, cp.returncode, cp.stderr)
         cp, jobs = cli.show_json(uuids, self.cook_url)
         self.assertEqual('cook-scheduler-cli', jobs[0]['application']['name'])
```
shamsimam: Can we add a test/assertion that checks the default application version?

dposada: Done

PR #525, opened 2017-11-01
Add logging to unscheduled_jobs query

pschorf:


pschorf:
Fixed a `logger` instead of `self.logger` bug

PR #526, opened 2017-11-01
Adds command-prefix to cs submit

dposada:


dposada:
@DaoWen any feedback on this?

PR #529, opened 2017-11-03
Adds cs config command

dposada:


cli/cook/subcommands/config.py
```diff
@@ -0,0 +1,118 @@
...
+
+def get_in(dct, keys):
+    """Gets the value in dct at the nested path indicated by keys"""
+    for key in keys:
+        if not isinstance(dct, dict):
+            return None
+        try:
+            dct = dct[key]
+        except KeyError:
+            return None
```
DaoWen: Is there a reason to prefer the `try`/`except` here rather than using `dct.get(key)`?

cli/cook/subcommands/config.py
```diff
@@ -0,0 +1,118 @@
...
+    return dct
+
+
+def set_in(dct, keys, value):
+    """Sets the value in dct at the nested path indicated by keys"""
+    for key in keys[:-1]:
+        if key not in dct:
+            dct[key] = {}
+        dct = dct[key]
```
DaoWen: `dct = dct.get(key, {})`

dposada: I actually need to set `dct[key]` to `{}` to create the nested map if it's not present

cli/cook/subcommands/config.py
dposada: Switched to checking if `key in dct`

PR #531, opened 2017-11-03
Add failed-only option for retry endpoint

DaoWen:


scheduler/src/cook/mesos/api.clj
```diff
@@ -1957,7 +1984,7 @@
         {:summary "Change a job's retry count"
          :parameters {:body-params UpdateRetriesRequest}
          :handler (put-retries-handler conn is-authorized-fn task-constraints)
-         :responses {201 {:schema PosInt
+         :responses {201 {:schema NonNegInt
```
DaoWen: I had to change this since `display-retries` now returns `0` if no jobs had failed, but the `failed_only` flag was set. I felt like this made more sense than returning an error status in the case that there were no failed jobs.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1408,15 +1425,23 @@
...
-       (check-groups-exist conn ctx)))
+  (let [[jobs-ok? jobs-ctx' :as jobs-result] (vectorize (check-jobs-exist conn ctx))]
+    (if-not jobs-ok?
+      jobs-result
+      (let [[groups-ok? groups-ctx' :as groups-result] (vectorize (check-groups-exist conn ctx))]
+        (cond
+          (not groups-ok?) groups-result
+          (not (or jobs-ctx' groups-ctx')) true
+          (and jobs-ctx' groups-ctx') [true (combine jobs-ctx' groups-ctx')]
+          :else [true (or jobs-ctx' groups-ctx')])))))
```
DaoWen: The logic here became a lot more complex because I didn't want any potential future changes to `check-jobs-exist` or `check-groups-exist` to unexpectedly break this function.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1408,15 +1425,23 @@
...
 
 (defn validate-retries
   [conn task-constraints ctx]
   (let [retries (get-in ctx [:request :body-params :retries])
         increment (get-in ctx [:request :body-params :increment])
         jobs (jobs-from-request ctx)
         groups (get-in ctx [:request :body-params :groups])
+        failed-only? (get-in ctx [:request :body-params :failed_only])
```
dposada: Can we make the param use dash instead of underscore? We use dash for other params like `start-ms`, `end-ms` on `/list`

DaoWen: I originally had it as `:failed-only?`, but that screwed up my ability to set it via named arguments in our Python integration tests. If you really prefer the hyphen, maybe I can do something in the `retry_jobs` method in Python to convert from snake_case to kabob-case.

dposada: I thought `cook-coercer` took care of that for us (https://github.com/twosigma/Cook/blob/3e1d88b7747a3c5bf95a2afbd62dec2c269e2c6b/scheduler/src/cook/mesos/api.clj#L1036), but maybe I'm mis-remembering

DaoWen: I thought we had something doing that too, but it didn't work when I tried it before. That was probably due to a different schema issue though, so I'll try it again and push the change if it works. Thanks!

DaoWen: I didn't realize that I had to update the `cook-coercer` to make this happen. Now it works.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1858,6 +1851,40 @@
...
+              (fn [their-matchers]
+                (fn [s]
+                  (or (their-matchers s)
+                      (get {;; can't use form->kebab-case because env and label
+                            ;; accept arbitrary kvs
+                            JobRequestMap (partial map-keys ->kebab-case)
+                            Group (partial map-keys ->kebab-case)
+                            HostPlacement (fn [hp]
+                                            (update hp :type keyword))
+                            UpdateRetriesRequest (partial map-keys ->kebab-case)
```
DaoWen: The only thing I changed in this function was the addition of this line. I moved the whole function down here because it was being defined before `UpdateRetriesRequest`. Since it's not used earlier in the file, I moved it as far down as possible so that we can avoid moving it again in the future.

integration/tests/cook/test_basic.py
```diff
@@ -892,45 +921,46 @@ def jobs_query():
 
             util.wait_until(jobs_query, util.all_instances_done)
             # retry all jobs in the group
-            util.retry_jobs(self.cook_url, retries=2, groups=[group_uuid])
+            util.retry_jobs(self.cook_url, retries=2, groups=[group_uuid], **extra_retry_args)
```
pschorf: Can we just inline the args here instead of creating the separate `extra_retry_args` variable?

integration/tests/cook/test_basic.py
```diff
@@ -892,45 +921,46 @@ def jobs_query():
...
             for job in job_data.json():
-                job_details = f'Job details: {json.dumps(job, sort_keys=True)}'
+                job_details = f'failed_only={failed_only}; Job details: {json.dumps(job, sort_keys=True)}'
                 self.assertNotEqual('failed', job['state'], job_details)
         finally:
             # ensure that we don't leave a bunch of jobs running/waiting
             util.kill_groups(self.cook_url, [group_uuid])
 
+    def test_group_change_killed_retries(self):
+        self._help_test_group_change_killed_retries(failed_only=True)
```
pschorf: It doesn't seem like this test actually asserts anything differently. Is it worth the overhead of running it twice with a different parameter?

integration/tests/cook/test_basic.py
```diff
@@ -965,6 +995,66 @@ def all_completed_restarted(response):
             # ensure that we don't leave a bunch of jobs running/waiting
             util.kill_groups(self.cook_url, [group_uuid])
 
+    def _help_test_group_failed_only_change_retries(self, config):
```
pschorf: To me, this is hard to understand without reading the corresponding tests and jumping back and forth. How long would it be if you just had 3 separate tests (assuming there's some generic code in here that not all tests need) You could probably kill a lot of the comments in that case as well, because the code would be more straightforward.

integration/tests/cook/util.py
```diff
@@ -228,7 +229,7 @@ def wait_until_inner():
         return wait_until_inner()
     except:
         final_response = query()
-        logger.info(f"Timeout exceeded waiting for condition. Details: {final_response}")
+        logger.info(f"Timeout exceeded waiting for condition. Details: {final_response.content}")
```
pschorf: Sometimes `final_response` isn't actually a response. Can you add a guard here and only log `.content` if it actually exists?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1378,15 +1354,22 @@
...
     (cond
       ; error case
       (seq jobs-not-existing)
       [false {::error (->> jobs-not-existing
                            (map #(str "UUID " % " does not correspond to a job."))
                            (str/join \space))}]
 
+      ; remove non-failed jobs if necessary
+      (::failed-only? ctx)
+      [true {::jobs (with-meta
```
pschorf: I'm guessing the `with-meta` thing is to deal with liberator bs?

integration/tests/cook/test_basic.py
DaoWen: Yes. I think this is an artifact of refactoring an earlier version of the test. This should definitely be inlined.

integration/tests/cook/test_basic.py
DaoWen: I think it's worthwhile to ensure that the `failed_only` flag isn't completely broken.

integration/tests/cook/util.py
DaoWen: Done.

scheduler/src/cook/mesos/api.clj
DaoWen: Yup! I'll add a comment. This prevents Liberator from *concatenating* the vectors (which is the default behavior) because we only want to keep the failed jobs in this case.

integration/tests/cook/test_basic.py
DaoWen: I agree that the generic `config` arg makes this harder to understand, but I also don't want to duplicate all of this code three times. I don't see a winning solution here... We'll have to defer to @dposada to decide which is the lesser evil in this case.

integration/tests/cook/test_basic.py
pschorf: I thought about this some more and I talked myself out of it again. I think the test above checks that `failed_only` isn't completely broken, and if there's other behavior of the flag that's worth testing we should do that explicitly.

integration/tests/cook/test_basic.py
dposada: I am generally against having test assertions not be in the test proper (i.e. in a helper function). How about keeping the meat of this function, but moving the assertions back out to each individual test?

scheduler/src/cook/mesos/api.clj
DaoWen: Done. (The comment is just above the line with `:replace`.)

integration/tests/cook/test_basic.py
DaoWen: That seems like a good compromise.

DaoWen: Done! Lifting the test assertions out of the helper method also simplified the config map I was passing to the helper. I think this change was definitely an improvement. Thanks for the feedback!

DaoWen:
@pschorf - Good?

integration/tests/cook/test_basic.py
```diff
@@ -551,6 +551,34 @@ def test_change_retries(self):
...
+        job_specs = util.minimal_jobs(2, max_retries=1, command='sleep 5')
+        try:
+            jobs, resp = util.submit_jobs(self.cook_url, job_specs)
+            self.assertEqual(resp.status_code, 201)
+            # wait for first job to start running, and kill it
+            job_uuid = jobs[0]
+            util.wait_for_job(self.cook_url, job_uuid, 'running')
+            util.kill_jobs(self.cook_url, [job_uuid])
+            job = util.load_job(self.cook_url, job_uuid)
+            self.assertEqual('failed', job['state'])
```
dposada: I think we have a race between the kill and the job completing. If the job completes before the kill, the job state will not be "failed". Maybe we should make it sleep longer than 5 seconds?

integration/tests/cook/test_basic.py
```diff
@@ -551,6 +551,34 @@ def test_change_retries(self):
...
+            # retry both jobs, but with the failed_only=true flag
+            resp = util.retry_jobs(self.cook_url, retries=4, failed_only=True, jobs=jobs)
+            self.assertEqual(201, resp.status_code, resp.text)
+            job = util.load_job(self.cook_url, job_uuid)
+            self.assertIn(job['status'], ['waiting', 'running'])
+            jobs = util.wait_for_jobs(self.cook_url, jobs, 'completed')
+            # We expect both jobs to be completed successfully now.
+            # The first job (which we killed and retried) should have 2 retries remaining.
```
dposada: I don't understand why the first job would have 2 retries instead of 3

integration/tests/cook/test_basic.py
```diff
@@ -551,6 +551,34 @@ def test_change_retries(self):
...
+            # The second job (which started with the default 1 retries)
+            # should have 0 remaining since the failed_only flag was set.
+            for job, expected_retries in zip(jobs, [2, 0]):
+                job_details = f"Job details: {json.dumps(job, sort_keys=True)}"
+                self.assertEqual('success', job['state'], job_details)
+                self.assertEqual(expected_retries, job['retries_remaining'], job_details)
```
dposada: For only 2 jobs, I would favor not having the `for` loop

integration/tests/cook/test_basic.py
```diff
@@ -892,45 +920,48 @@ def jobs_query():
...
             # ensure that we don't leave a bunch of jobs running/waiting
             util.kill_groups(self.cook_url, [group_uuid])
 
+    def test_group_change_killed_retries(self):
+        jobs = self._help_test_group_change_killed_retries(failed_only=True)
+        # ensure none of the jobs are still in a failed state
+        for job in jobs:
+            self.assertNotEqual('failed', job['state'], f'Job details: {json.dumps(job, sort_keys=True)}')
+
+    def test_group_change_killed_retries_failed_only(self):
```
dposada: How come the test that sets `failed_only` to `True` is called `test_group_change_killed_retries` and the test that sets it to `False` is called `test_group_change_killed_retries_failed_only`?

integration/tests/cook/test_basic.py
```diff
@@ -870,7 +898,7 @@ def query():
             # (ensure it still works when there are no live jobs)
             util.kill_groups(self.cook_url, [group_uuid])
 
-    def test_group_change_killed_retries(self):
+    def _help_test_group_change_killed_retries(self, failed_only):
```
dposada: Maybe we could name this something like `group_submit_kill_retry`? Should it be in `util`?

integration/tests/cook/test_basic.py
```diff
@@ -965,6 +996,66 @@ def all_completed_restarted(response):
...
+            # wait again for the parameterized condition from the config
+            util.wait_until(group_query, config_condition)
+            # return final job details to caller for assertion checks
+            return util.query_jobs(self.cook_url, assert_response=True, job=jobs).json()
+        finally:
+            # ensure that we don't leave a bunch of jobs running/waiting
+            util.kill_groups(self.cook_url, [group_uuid])
+
+    def test_group_failed_only_change_retries_all_active(self):
+        config = {'command': 'sleep 10', 'predicate_statuses': ['running', 'waiting']}
```
dposada: Why not just have `command` and `statuses` be named arguments to the helper function?

integration/tests/cook/test_basic.py
```diff
@@ -965,6 +996,66 @@ def all_completed_restarted(response):
             # ensure that we don't leave a bunch of jobs running/waiting
             util.kill_groups(self.cook_url, [group_uuid])
 
+    def _help_test_group_failed_only_change_retries(self, config):
```
dposada: Maybe we should just call this `group_submit_retry`? Should it be in `util`?

integration/tests/cook/util.py
```diff
@@ -228,7 +229,11 @@ def wait_until_inner():
         return wait_until_inner()
     except:
         final_response = query()
-        logger.info(f"Timeout exceeded waiting for condition. Details: {final_response}")
+        try:
+            details = final_response.content
+        except AttributeError:
```
dposada: Is there a better way to do this? Maybe using `isintance`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1352,12 +1318,22 @@
...
    (s/optional-key :groups) [s/Uuid]
+   (s/optional-key :failed-only) s/Bool
    (s/optional-key :retries) PosNum
    (s/optional-key :increment) PosNum})
 
 (defn display-retries
   [conn ctx]
-  (:job/max-retries (d/entity (db conn) [:job/uuid (-> ctx ::jobs first)])))
+  (if-let [first-job (-> ctx ::jobs first)]
+    (:job/max-retries (d/entity (db conn) [:job/uuid first-job]))
```
dposada: Maybe `(-> conn db (d/entity [:job/uuid first-job]) :job/max-retries)`

scheduler/src/cook/mesos/api.clj
dposada: Is it misleading to respond 201 in this case?

integration/tests/cook/util.py
DaoWen: I think this is one of those places where [EAFP](https://docs.python.org/3.6/glossary.html#term-eafp) applies.

scheduler/src/cook/mesos/api.clj
DaoWen: Now using 202 Accepted for the empty case

integration/tests/cook/test_basic.py
DaoWen: Bumped up to 10 seconds.

integration/tests/cook/test_basic.py
DaoWen: Each run counts against the retries-remaining count. Since it ran twice (run, kill, restart, complete), it's left with 2/4.

integration/tests/cook/test_basic.py
DaoWen: That was backwards! I fixed it. Thanks!

DaoWen:
@dposada - This is ready for another review. I finished addressing your comments. I also updated the default behavior for requests including one or more group uuids to `failed_only=true` based on our discussion on Wednesday.

dposada:
Looks like this needs a merge/rebase

DaoWen:
@dposada - Rebased (and merge conflict resolved).

scheduler/src/cook/mesos/api.clj
```diff
@@ -1302,17 +1276,28 @@
...
 
-(def UpdateRetriesRequest
+(def UpdateRetriesRequestBase
   {(s/optional-key :job) s/Uuid
    (s/optional-key :jobs) [s/Uuid]
-   (s/optional-key :groups) [s/Uuid]
    (s/optional-key :retries) PosNum
    (s/optional-key :increment) PosNum})
 
+(def UpdateRetriesRequestDeprecated UpdateRetriesRequestBase)
```
dposada: Would be nice to have a comment explaining what's going on here

scheduler/src/cook/mesos/api.clj
```diff
@@ -1479,10 +1488,15 @@
...
     conn is-authorized-fn
     {:allowed-methods [:put]
-     :exists? (partial check-jobs-and-groups-exist conn)
      :malformed? (partial validate-retries conn task-constraints)
-     :handle-created (partial display-retries conn)
-     :put! (partial retry-jobs! conn)}))
+     :exists? (partial check-jobs-and-groups-exist conn)
+     :put! (partial retry-jobs! conn)
+     ;; :put-enacted? decides whether to respond with Accepted (false) or Created (true).
+     :put-enacted? (comp seq ::jobs)
```
dposada: https://httpstatuses.com/202 says:

> The request has been accepted for processing, but the processing has not been completed.

It's not really true in this case that processing has not completed, right?

integration/tests/cook/test_basic.py
```diff
@@ -1030,6 +1039,33 @@ def all_completed_restarted(response):
...
+            job_details = f'Job details: {json.dumps(job, sort_keys=True)}'
+            self.assertIn(job['status'], statuses, job_details)
+            self.assertEqual(job['retries_remaining'], 0, job_details)
+            self.assertEqual(len(job['instances']), 1, job_details)
+
+    def test_group_failed_only_change_retries_all_failed(self):
+        statuses = ['completed']
+        jobs = util.group_submit_retry(self.cook_url, command='exit 1', predicate_statuses=statuses)
+        for job in jobs:
+            job_details = f'Job details: {json.dumps(job, sort_keys=True)}'
```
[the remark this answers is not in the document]
DaoWen: That's a good idea, but this PR is already pretty big, so I think I'll do that as a follow-on.

scheduler/src/cook/mesos/api.clj
DaoWen: Changed to 200 OK.

integration/tests/cook/test_basic.py
DaoWen: Done. (We're killing the jobs now instead of waiting for completion, so the longer sleep won't hurt us.)

DaoWen:
@dposada - Updated

PR #532, opened 2017-11-03
adds logs before the sleep + ensures terminating message is printed

shamsimam:


integration/tests/cook/test_cli.py
```diff
@@ -900,7 +900,9 @@ def test_show_progress_message(self):
             self.assertEqual(0, cp.returncode, cp.stderr)
             self.assertIn("99% We are so close!", cli.stdout(cp))
 
-            time.sleep(1.0) # allow the executor to wrap up
+            self.logger.info('Sleeping 5 seconds to allow cook executor to complete execution')
+            time.sleep(5.0) # allow the executor to wrap up
+
```
dposada: Can't we use the `wait_until` approach instead of sleeping?

integration/tests/cook/test_cli.py
```diff
@@ -882,6 +882,10 @@ def __wait_for_progress_message(self, uuids):
         return util.wait_until(lambda: cli.show_json(uuids, self.cook_url)[1][0]['instances'][0],
                                lambda i: 'progress' in i and 'progress_message' in i)
 
+    def __wait_for_executor_completion_message(self, uuids):
```
dposada: Should we put this in `util`?

integration/tests/cook/test_cli.py
shamsimam: Done.

integration/tests/cook/test_cli.py
shamsimam: This function relies on the use of the `cli`. I didn't want to introduce dependency on the `cli` in the `util.py`.

dposada: Got it

PR #533, opened 2017-11-06
Wrap agent query in wait_until

pschorf:


shamsimam:
Awaiting green build.

PR #535, opened 2017-11-06
upgrades pymesos version to 0.2.15

shamsimam:
logs error from the executor


executor/cook/executor.py
```diff
@@ -419,6 +417,10 @@ def shutdown(self, driver):
         logging.info('Mesos requested executor to shutdown!')
         self.stop_signal.set()
 
+    def error(self, driver, message):
```
dposada: Docstring?

shamsimam: This is an overridden method like the others above, e.g. `shutdown`, `killTask`, `launchTask`, etc.

PR #536, opened 2017-11-06
Fixes flakiness in test_ls

dposada:


shamsimam:
Good to merge?

integration/tests/cook/test_cli.py
```diff
@@ -771,7 +771,7 @@ def entry(name):
         self.logger.debug(entries)
         foo = entry('foo')
         self.assertEqual('drwxr-xr-x', foo['mode'])
-        self.assertEqual(2, foo['nlink'])
+        self.assertGreaterEqual(2, foo['nlink'])
```
DaoWen: This should be `assertLessEqual` if `foo['nlink']` is sometimes greater than 2.

dposada: Fixed, thanks

dposada:
@DaoWen any other feedback on this?

PR #539, opened 2017-11-07
supports running cook scheduler on a local mesos

shamsimam:


executor/bin/prepare-executor.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# USAGE: ./bin/prepare-executor.sh MODE TARGET_DIR
+# Builds the cook executor and then copies it to TARGET_DIR
+# Examples:
+#   ./bin/prepare-executor.sh docker /target/directory
+#   ./bin/prepare-executor.sh local /target/directory
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+EXECUTOR_DIR="$(dirname ${DIR})"
+MODE=${1}
```
DaoWen: Rather than specifying a *mode* (`docker` vs `local`), I think it would be better to have the different versions based on your target *platform*. We could generate the `PLATFORM` string with something like `$(uname -s)-$(uname -i)`, which should yield `Linux-x86_64` or `Darwin-x86_64` for 64-bit Linux or OS X, respectively.

I think it's still fine to have `docker` vs `local` for the build script names (since that's describing where the build is done, not the target platform), but I'd prefer having the generated binary use the target platform in its name.

shamsimam: I've undone changes that added the `-docker` suffix, the locally created executable will have the `-local` suffix.

executor/Dockerfile.build
```diff
@@ -10,4 +10,4 @@ RUN pip install -r requirements.txt
 
 COPY . /opt/cook
 
-CMD ["pyinstaller", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
+CMD ["pyinstaller", "-F", "-n", "cook-executor-docker", "-p", "cook", "cook/__main__.py"]
```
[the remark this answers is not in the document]
shamsimam: Undid the change to `cook-executor-binary`.

dposada:
@shamsimam Is using a local mesos now the recommended workflow? The quickstart still references minimesos https://github.com/twosigma/Cook#quickstart

DaoWen:
@dposada - I still like minimesos on Linux. I think it's just a bit more of a pain on mac due to the headaches around docker...

dposada:
Should we clarify the quickstart to reflect that?

shamsimam:
I don't know if I want to recommend it, but I prefer using local Mesos over Minimesos/Docker. I get to run my test commands and also don't need to mess with Docker.

I'll make a pass on the quickstart later today.

scheduler/bin/run-local.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Usage: ./bin/run-local.sh
+# Runs the cook scheduler locally.
+
+# Defaults (overridable via environment)
+: ${COOK_DATOMIC_URI="datomic:mem://cook-jobs"}
+: ${COOK_FRAMEWORK_ID:=cook-framework-$(date +%s)}
+: ${COOK_NREPL_PORT:=${2:-8888}}
+: ${COOK_PORT:=${1:-12321}}
+: ${MASTER_IP:="127.0.0.2"}
```
dposada: @shamsimam why does this default to 127.0.0.2?

shamsimam: I have ZK running on 127.0.0.1, master on 127.0.0.2 and the workers on 3/4. I used Waiter's add loopback script to get all these IPs working locally.

PR #540, opened 2017-11-07
Remove jobs from group entity in job-ent->map

pschorf:


scheduler/src/cook/mesos/util.clj
```diff
@@ -91,18 +91,19 @@
...
+     (deep-transduce-kv (filter (fn [[k _]] (not (contains? skip-keys k)))) entity'))))
 
 (defn job-ent->map
   "Convert a job entity to a map. This also loads the associated group and converts it to a map."
   ([job db]
    (job-ent->map (d/entity db (:db/id job))))
   ([job]
    (-> (entity->map job)
-       (assoc :group/_job (entity->map (:group/_job job))))))
+       (assoc :group/_job (entity->map (:group/_job job) :skip-keys #{:group/job})))))
```
dposada: Might be nice to add a comment explaining why we're skipping that key

dposada: This line is pretty confusing in general, given the `:group/_job` and `:group/job`, so some extra explanation may not hurt

integration/tests/cook/test_basic.py
```diff
@@ -999,13 +999,26 @@ def test_400_on_group_query_without_uuid(self):
...
         try:
             self.assertEqual(201, resp.status_code, resp.content)
-            time.sleep(30)  # Need to wait for a rank cycle
-            queue = util.session.get('%s/queue' % self.cook_url)
-            self.assertEqual(200, queue.status_code, queue.content)
-            self.assertTrue(any([job['job/uuid'] == job_uuid for job in queue.json()['normal']]))
+            def query_queue():
+                return util.session.get('%s/queue' % self.cook_url)
+            def queue_predicate(resp):
+                return any([job['job/uuid'] == job_uuid for job in resp.json()['normal']])
```
dposada: Nitpick: I would favor a blank line before and after each `def`

integration/tests/cook/test_basic.py
```diff
@@ -999,13 +999,26 @@ def test_400_on_group_query_without_uuid(self):
...
+            resp = util.wait_until(query_queue, queue_predicate)
+            self.assertEqual(200, resp.status_code, resp.content)
+            job = [job for job in resp.json()['normal']
+                   if job['job/uuid'] == job_uuid][0]
+            self.assertTrue('group/_job' in job.keys())
```
dposada: Should we use a key other than `group/_job`? It's a little odd explaining to consumers that they should get the group data from `group/_job`. I know that is the Datomic key, but how much do we want those internals to be exposed in the API?

integration/tests/cook/test_basic.py
pschorf: If only they were lambdas....

Done

scheduler/src/cook/mesos/util.clj
pschorf: Done

integration/tests/cook/test_basic.py
dposada: Makes sense

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,18 +91,22 @@
...
 (defn job-ent->map
   "Convert a job entity to a map. This also loads the associated group and converts it to a map."
   ([job db]
    (job-ent->map (d/entity db (:db/id job))))
   ([job]
    (-> (entity->map job)
-       (assoc :group/_job (entity->map (:group/_job job))))))
+       ; Load the group related to the job (if any) but without loading the job entities
+       ; inside the group entity. If someone needs to access them, it's better to query
+       ; at that point instead of relying on the cached data.
```
dposada: 👍

PR #541, opened 2017-11-07
Marks flaky ssh tests as explicit

dposada:


DaoWen:
The Travis check failed on def0545 because `test_ssh_duplicate_uuid` flaked (no output URL).

dposada:
@pschorf RTM

PR #543, opened 2017-11-07
adds COOK_JOB_UUID and COOK_JOB_GROUP_UUID to the job environment

shamsimam:


scheduler/src/cook/mesos/task.clj
```diff
@@ -81,7 +81,11 @@
         custom-executor? (use-custom-executor? job-ent)
         cook-executor? (and (not container) ;;TODO support cook-executor in containers
                             (use-cook-executor? job-ent executor-config))
+        job-uuid (:job/uuid job-ent)
+        group-uuid (util/job-ent->group-uuid job-ent)
         environment (cond-> (util/job-ent->env job-ent)
+                            group-uuid (assoc "COOK_JOB_GROUP_ID" (str group-uuid))
+                            job-uuid (assoc "COOK_JOB_ID" (str job-uuid))
```
dposada: When would `job-uuid` be falsy?

dposada: Should these be `_UUID` instead of `_ID`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -128,6 +128,11 @@
   [db job-ent]
   (some-> job-ent :job/container remove-datomic-namespacing))
 
+(defn job-ent->group-uuid
+  "Take a job entity and return its container"
```
dposada: Why "its container"?

integration/tests/cook/test_cli.py
```diff
@@ -963,6 +963,18 @@ def test_submit_with_executor(self):
         self.assertEqual(2, cp.returncode, cp.stderr)
         self.assertIn("invalid choice: 'bogus'", cli.decode(cp.stderr))
 
+    def test_job_environment(self):
+        """Note: This test doesn't directly test the cli. It relies on stdout to test the job environment."""
```
dposada: It's a little odd that this lives in `test_cli` but isn't meant to test the CLI

shamsimam: I know :( 
Any suggestions on how to better test the feature?

scheduler/src/cook/mesos/task.clj
shamsimam: I like `_ID` as Mesos names its variables `MESOS_EXECUTOR_ID` and `MESOS_SLAVE_ID`.

integration/tests/cook/test_cli.py
dposada: Can we fetch job output without using `cs tail`?

scheduler/src/cook/mesos/task.clj
dposada: Sure, but in Cook we already refer to job uuid and group uuid in other places

scheduler/src/cook/mesos/util.clj
shamsimam: Fixed.

scheduler/src/cook/mesos/task.clj
shamsimam: > When would job-uuid be falsy?

Avoided the condition check.

integration/tests/cook/test_cli.py
shamsimam: None of the other non-cli tests do it and figured it would pretty much repeat what `cs tail` does.
I considered progress messages, but didn't want to tie this test to the cook executor.

dposada: Even though none of the other tests do it currently, it does seem like a generally useful util function for tests to get the stdout, stderr from a job.

As for the duplication, we could consider making the util function just shell out to `cs tail`

shamsimam: I think I'll rewrite the command to check and env and exit `1` if it doesn't find those entries in the environment. That way we can avoid the dependency on the cli.

shamsimam: Rewrote the test.

scheduler/src/cook/mesos/task.clj
shamsimam: Added the `_UUID` suffix to the variable names.

scheduler/src/cook/mesos/util.clj
```diff
@@ -132,6 +132,11 @@
   [db job-ent]
   (some-> job-ent :job/container remove-datomic-namespacing))
 
+(defn job-ent->group-uuid
+  "Take a job entity and return its group UUID"
+  [job-ent]
+  (some-> job-ent :group/_job first :group/uuid))
```
dposada: Can a job be in more than 1 group? Asking because of the `first`

shamsimam: I already spoke to @pschorf about it since the groups are returned as a set. A job can only belong to one group.

wyegelwel:
When we were adding groups we imagined a job belonging to multiple groups.
For example, hierarchy. However, we realized this complicated a lot of the
constraint logic so we exposed only setting a single group in the API until
we figured out how to make multiple groups work. We kept it as a list
internally because we thought we would figure it out quickly. Unfortunately
that didn't pan out.

PR #546, opened 2017-11-09
Sandbox sync for pending hosts

shamsimam:
- retries sandbox syncing of hosts when the cache entry expires
- removes bad hosts from the pending sandbox sync process
- asserts for `sandbox_directory` and `output_url` in integration tests

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -333,13 +333,18 @@ def wait_for_exit_code(cook_url, job_id, max_delay_ms=2000):
     return response.json()[0]
 
 
-def wait_for_sandbox_directory(cook_url, job_id, max_delay_ms=4000):
+def wait_for_sandbox_directory(cook_url, job_id):
     """
     Wait for the given job's sandbox_directory field to appear.
     Returns an up-to-date job description object on success,
     and raises an exception if the max_delay_ms wait time is exceeded.
     """
     job_id = unpack_uuid(job_id)
+
+    cook_settings = settings(cook_url)
+    cache_ttl_ms = get_in(cook_settings, 'agent-query-cache', 'ttl-ms')
+    sync_interval_ms = get_in(cook_settings, 'sandbox-syncer', 'sync-interval-ms')
+    max_delay_ms = min(4 * max(cache_ttl_ms, sync_interval_ms), 4 * 60 * 1000)
 
     def query():
         response = query_jobs(cook_url, True, job=[job_id])
--- a/scheduler/src/cook/mesos/sandbox.clj
+++ b/scheduler/src/cook/mesos/sandbox.clj
@@ -137,16 +137,16 @@
 (defn- aggregate-pending-sync-hostname
   "Aggregates the hostname as a pending sync item into pending-sync-state."
   [pending-sync-state agent-hostname reason]
-  ;; TODO Shams use reason to track consecutive failures
-  (-> pending-sync-state
-      (update :pending-sync-hosts conj agent-hostname)))
+  (cond-> (update pending-sync-state :pending-sync-hosts conj agent-hostname)
+          (= :error reason)
+          (update-in [:host->consecutive-failures agent-hostname] (fnil inc 0))))
 
 (defn- clear-pending-sync-hostname
   "Clears the hostname as a pending sync item from pending-sync-state."
   [pending-sync-state agent-hostname reason]
-  ;; TODO Shams use reason to track consecutive failures
-  (-> pending-sync-state
-      (update :pending-sync-hosts disj agent-hostname)))
+  (cond-> (update pending-sync-state :pending-sync-hosts disj agent-hostname)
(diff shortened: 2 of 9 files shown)


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -182,11 +187,51 @@
...
+            (let [consecutive-failures (get host->consecutive-failures agent-hostname 0)]
+              (cond
+                (>= consecutive-failures max-consecutive-sync-failure)
+                (do
+                  (log/info "Removing entry for" agent-hostname "with" consecutive-failures "consecutive failures")
+                  (send pending-sync-agent clear-pending-sync-hostname agent-hostname :threshold))
+                (not (cache/has? @mesos-agent-query-cache agent-hostname))
+                (do
+                  (log/info "Triggering pending sandbox sync of" agent-hostname)
+                  (send pending-sync-agent clear-pending-sync-hostname agent-hostname :sync)
```
pschorf: Is this call necessary? Should we just let `sync-agent-sandboxes` clean this up?

shamsimam: We need this to cleanup pending hosts. I restructured the code a bit to ensure the removal actually happens before the next round of `sync`s are tried.

shamsimam:
@pschorf request for feedback

integration/tests/cook/util.py
```diff
@@ -333,14 +333,19 @@ def predicate(resp):
...
     Wait for the given job's sandbox_directory field to appear.
     Returns an up-to-date job description object on success,
     and raises an exception if the max_delay_ms wait time is exceeded.
     """
     job_id = unpack_uuid(job_id)
 
+    cook_settings = settings(cook_url)
+    cache_ttl_ms = get_in(cook_settings, 'agent-query-cache', 'ttl-ms')
+    sync_interval_ms = get_in(cook_settings, 'sandbox-syncer', 'sync-interval-ms')
+    max_delay_ms = min(4 * max(cache_ttl_ms, sync_interval_ms), 4 * 60 * 1000)
```
pschorf: One last question...is it possible to write a non-flaky integration test that would have failed before but pass now? I'm not sure it is, but it would be great to have.

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -182,11 +189,64 @@
...
+  (let [syncer-state {:mesos-agent-query-cache mesos-agent-query-cache
+                      :pending-sync-agent pending-sync-agent
+                      :task-id->sandbox-agent task-id->sandbox-agent}]
+    (chime/chime-at
+      (periodic/periodic-seq (time/now) (time/millis sync-interval-ms))
+      (fn host-sandbox-syncer-task [_]
+        (let [{:keys [framework-id host->consecutive-failures pending-sync-hosts]} @pending-sync-agent
+              num-pending-sync-hosts (count pending-sync-hosts)]
+          (log/info num-pending-sync-hosts "hosts have pending syncs")
+          (loop [[hostname & remaining-hostnames] (vec pending-sync-hosts)
```
pschorf: Why is the `vec` necessary?

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -182,11 +189,64 @@
...
+                 pending-sync-agent-send-performed false]
+            (let [consecutive-failures (get host->consecutive-failures hostname 0)]
+              (cond
+                (nil? hostname)
+                (when pending-sync-agent-send-performed
```
pschorf: Would it be that bad to just have the `await` regardless of whether we sent anything? I don't think it's costing us too much, and it would be nice to clean up the logic a bit.

shamsimam: I'd rather only trigger the wait if this iteration performed any agent operations.

integration/tests/cook/util.py
shamsimam: Let me see if I can make use of `mesos_mock`. An integration test might be hard as we use so few hosts that any of the other finishing jobs may trigger a sync.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Clojure's sequential destructuring does not work on a set :(

pschorf: What about `seq`? :)

shamsimam: `seq` worked.

integration/tests/cook/util.py
shamsimam: Added `test-sandbox-directory-population` in `scheduler.clj`.

shamsimam:
@pschorf ready for another round :)

scheduler/src/cook/mesos/sandbox.clj
pschorf: 👍

PR #549, opened 2017-11-09
Defer calculation of group components until constraint construction

pschorf:
Please hold off on this PR until I have time to do a simulator run to make sure it actually helps. Just wanted to get the travis build going.


pschorf:
The simulator match cycles for 1.8.1 were 12% slower than 1.7.1, running a simulation with this patch now to see if it improves things.

pschorf:
This branch is 1% slower than v1.7.1 in the simulator, so let's merge it.

PR #551, opened 2017-11-09
Corrects condition on if check in prepare-executor.sh

shamsimam:


executor/bin/prepare-executor.sh
```diff
@@ -29,7 +29,7 @@ if [ -z "${TARGET_DIR}" ]; then
 fi
 
 COOK_EXECUTOR_NAME="cook-executor-${MODE}"
-if [[ ! "${MODE}" =~ ^(docker)$ ]]; then
+if [[ "${MODE}" =~ ^(docker)$ ]]; then
```
DaoWen: Why are you using a regular expression here? Using the basic test command seems simpler and clearer:
```bash
if [ "${MODE}" == docker ]; then
```

DaoWen:
Can we get Travis to invoke this script too? That would help us catch issues like this.

shamsimam:
@DaoWen I'd rather not mess with Travis. Travis is testing whether we can run the command directly without having to download the file which is the mechanism we intend to use.

PR #552, opened 2017-11-10
Add user agent to requests in python scripts

DaoWen:
Clarifies the origin of some requests in the cook scheduler access logs.

### DIFF
--- a/cli/cook/http.py
+++ b/cli/cook/http.py
@@ -5,6 +5,8 @@ from urllib.parse import urljoin
 
 import requests
 
+import cook
+
 session = None
 timeouts = None
 
@@ -29,6 +31,7 @@ def configure(config):
     http_adapter = adapters_module.HTTPAdapter(max_retries=retries)
     session = session_module.Session()
     session.mount('http://', http_adapter)
+    session.headers['User-Agent'] = f"cs/{cook.version.VERSION} ({session.headers['User-Agent']})"
 
 
 def __post(url, json_body):
--- a/cli/cook/version.py
+++ b/cli/cook/version.py
@@ -1,1 +1,1 @@
-VERSION = '0.7.1'
+VERSION = '0.7.2'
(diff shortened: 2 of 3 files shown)


cli/cook/http.py
```diff
@@ -29,6 +31,7 @@ def configure(config):
     http_adapter = adapters_module.HTTPAdapter(max_retries=retries)
     session = session_module.Session()
     session.mount('http://', http_adapter)
+    session.headers['User-Agent'] = f'CookCLI/{cook.version.VERSION}'
```
dposada: I would favor `cs/...`

dposada: Can we bump the version of the CLI for this change?

cli/cook/version.py
```diff
@@ -1 +1 @@
-VERSION = '0.7.0'
+VERSION = '0.7.1'
```
DaoWen: Bumped Cook CLI version number.

dposada: We're already at 0.7.1 on master:

https://github.com/twosigma/Cook/blob/master/cli/cook/version.py

cli/cook/http.py
```diff
@@ -29,6 +31,7 @@ def configure(config):
     http_adapter = adapters_module.HTTPAdapter(max_retries=retries)
     session = session_module.Session()
     session.mount('http://', http_adapter)
+    session.headers['User-Agent'] += f' (cs/{cook.version.VERSION})'
```
DaoWen: Name changed: `CookCLI` → `cs`

I also decided that it made more sense to append the extra user agent info to the default user-agent header of the response library. Now it looks like this in the logs: `"python-requests/2.18.4 (cs/0.7.1)"`

shamsimam: Can we change the order to `cs/0.7.1 (python-requests/2.18.4)`? This makes more sense to me as `cs` is what is more important, that it uses `python-request` is an implementation detail.

shamsimam: Based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent , `cs` is our product, `0.7.1` is the version and `(python-requests/2.18.4)` is system information.

DaoWen: OK, you convinced me. I'll reverse it.

cli/cook/version.py
DaoWen: Rebased (and bumped to 0.7.2).

PR #556, opened 2017-11-10
Adds --group-name to cs submit

dposada:


cli/cook/subcommands/submit.py
```diff
@@ -156,6 +161,16 @@ def submit(clusters, args, _):
     application_version = job.pop('application-version', version.VERSION)
     job['application'] = {'name': application_name, 'version': application_version}
 
+    group = None
+    if 'group-name' in job:
```
DaoWen: I thought this code might generate a unique group UUID for each submitted job rather than using the same one for all of them. Your (offline) explanation made sense, but renaming `job` to `job_template` here might make this more clear.

integration/tests/cook/cli.py
```diff
@@ -82,15 +82,28 @@ def show(uuids=None, cook_url=None, flags=None, show_flags=None):
...
-    jobs = [job for entities in response['clusters'].values() for job in entities['jobs']]
+    cp, data = __show_json(uuids, cook_url, flags)
+    jobs = [job for entities in data['clusters'].values() for job in entities['jobs']]
     return cp, jobs
 
 
+def show_groups(uuids, cook_url=None, flags=None):
+    """Shows the group JSON corresponding to the given UUID(s)"""
+    cp, data = __show_json(uuids, cook_url, flags)
+    groups = [job for entities in data['clusters'].values() for job in entities['groups']]
```
DaoWen: `job` → `group`

integration/tests/cook/test_cli.py
```diff
@@ -1200,3 +1200,26 @@ def test_config_command_advanced(self):
...
+
+        # Group name and group uuid
+        group_uuid = str(uuid.uuid4())
+        cp, uuids = cli.submit_stdin(['ls', 'ls'], self.cook_url, submit_flags=f'--group-name bar --group {group_uuid}')
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        _, groups = cli.show_groups([group_uuid], self.cook_url)
+        self.assertEqual('bar', groups[0]['name'])
+        self.assertEqual(group_uuid, groups[0]['uuid'])
+        self.assertIn(uuids[0], groups[0]['jobs'])
+        self.assertIn(uuids[1], groups[0]['jobs'])
```
DaoWen: Maybe instead we should test `assertEqual(sorted(uuids), sorted(groups[0]['jobs'])`? Or we could test that in addition. It seems like a good idea to check if there are duplicates or other extra entries.

cli/cook/subcommands/submit.py
dposada: Good call, renamed

integration/tests/cook/cli.py
dposada: Thanks, fixed

integration/tests/cook/test_cli.py
dposada: Changed the assertion to equals on the sorted lists

PR #557, opened 2017-11-10
Adds support for piping from cs jobs to cs kill, show, and wait

dposada:


cli/cook/querying.py
```diff
@@ -206,3 +216,57 @@ def query_unique_and_run(clusters, job_or_instance_uuid, command_fn):
...
+
+    if stdin_from_pipe:
+        s = sys.stdin.read()
+        if not s:
+            raise Exception('No data received.')
+
+        try:
+            piped_data = json.loads(s)
+        except:
+            raise Exception('Unable to parse data.')
```
DaoWen: Unable to parse <em>JSON</em> data from stdin?

cli/cook/querying.py
```diff
@@ -206,3 +216,57 @@ def query_unique_and_run(clusters, job_or_instance_uuid, command_fn):
...
+
+        piped_cluster_names = set()
+        piped_types = set()
+        uuids = []
+        for datum in piped_data:
```
DaoWen: Should we assert that `piped_data` is a `list` (or isn't a `dict`)?

cli/cook/querying.py
```diff
@@ -206,3 +216,57 @@ def query_unique_and_run(clusters, job_or_instance_uuid, command_fn):
...
+        raise Exception(f'When piping to {command}, you cannot also supply UUIDs as arguments.')
+
+    include_jobs = True
+    include_instances = True
+    include_groups = True
+
+    if stdin_from_pipe:
+        s = sys.stdin.read()
+        if not s:
+            raise Exception('No data received.')
```
DaoWen: Expected job data (JSON) on stdin?

cli/cook/querying.py
DaoWen: These comments about JSON might be irrelevant if you decide to switch to the format we were discussing with Steven yesterday afternoon, e.g.:
<i>public-cloud:group/13d2e8da-83aa-4b0c-a2be-6c86116db923</i>

dposada: This now uses the URL format, e.g. `http://cook.example.com/jobs/13d2e8da-83aa-4b0c-a2be-6c86116db923`

dposada:
@DaoWen This is ready for another pass

cli/cook/subcommands/jobs.py
```diff
@@ -158,7 +185,10 @@ def register(add_parser, add_defaults):
                                              "alphanumeric characters, '.', '-', '_', and '*' as a wildcard)")
     parser.add_argument('--limit', '-l', help=f'limit the number of results (default = {DEFAULT_LIMIT})',
                         type=check_positive)
-    parser.add_argument('--json', help='show the data in JSON format', dest='json', action='store_true')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--json', help='show the data in JSON format', dest='json', action='store_true')
+    group.add_argument('-1', help='list one job per line, without table formatting',
```
DaoWen: Weren't we going to have a long form of this option too? Maybe `--entity-urls`?

DaoWen:
Once this is merged, I think it also makes sense to return a URL for an entity rather than a UUID in a lot of cases. For example, it would be nice to get a clickable URL rather than a UUID after a `cs submit`. It would also probably be better to get a URL when using the `-s` flag.

cli/cook/subcommands/jobs.py
dposada: I added a long form: `--urls`

dposada:
@DaoWen RTM

PR #559, opened 2017-11-13
fixes formatting, typos, messaging

shamsimam:
removes backslashes from progress regex
adds assertion messages on couple of the tests


integration/tests/cook/util.py
```diff
@@ -483,4 +483,5 @@ def progress_line(cook_url, percent, message):
         raise Exception('{} not present in {} regex string'.format('(\d+)', regex_string))
     if not '(.*)' in regex_string:
         raise Exception('{} not present in {} regex string'.format('(.*)', regex_string))
-    return regex_string.replace('(\d+)', str(percent)).replace('(.*)', str(message)).replace('(?: )?', ' ')
+    simplified_regex_string = regex_string.replace('(?: )?', ' ').replace('\\', '')
+    return simplified_regex_string.replace('(\d+)', str(percent)).replace('(.*)', str(message))
```
DaoWen: I think you need to replace the backslash *last*, otherwise it's going to clobber the backslashes in your other patterns (e.g., `(\d+)` → `(d+)`, which will probably cause the test to fail).

Nitpick: I'd prefer this all as one expression. You can force Python to treat multiple lines as a single expression if you wrap it all in parentheses:
```python
return (regex_string
        .replace('(\d+)', str(percent))
        .replace('(.*)', str(message))
        .replace('(?: )?', ' ')
        .replace('\\', ''))

File: scheduler/src/cook/mesos/api.clj [hunk 2]
```diff
@@ -1829,158 +1829,158 @@
...
+                               404 {:description "The UUID doesn't correspond to a job."}}}})))
+      (ANY "/queue" []
+        (waiting-jobs mesos-pending-jobs-fn is-authorized-fn mesos-leadership-atom leader-selector))
+      (ANY "/running" []
+        (running-jobs conn is-authorized-fn))
+      (ANY "/list" []
+        (list-resource (db conn) framework-id is-authorized-fn)))
+    (format-params/wrap-restful-params {:formats [:json-kw]
+                                        :handle-error c-mw/handle-req-error})
+    (streaming-json-middleware)))
```
DaoWen: Nitpick: It would be nice if we could prevent improperly-intended code from being checked in. We should all settle on a standard formatting that we're happy with, then make a [cljfmt](https://github.com/weavejester/cljfmt) configuration that we can run locally and in Travis.

shamsimam:
RTM

PR #563, opened 2017-11-14
moves task filtering logic to refresh-agent-cache-entry

shamsimam:


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -80,30 +80,16 @@
...
-                (when (seq txns)
-                  (log/info "Inserting" (count txns) "facts in sandbox state update")
-                  (meters/mark! sandbox-updater-tx-rate)
-                  (timers/time!
-                    sandbox-updater-tx-duration
-                    @(d/transact datomic-conn txns)))))
-            (send task-id->sandbox-agent clear-agent-state task-id->sandbox-partition))
+          (letfn [(build-sandbox-txns [[task-id sandbox]]
+                    [[:db/add [:instance/task-id task-id] :instance/sandbox-directory sandbox]])]
+            (let [txns (mapcat build-sandbox-txns task-id->sandbox-partition)]
```
wyegelwel: Nitpick: you can use `map` and then only one set of brackets in build-sandbox-txns

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -150,19 +136,38 @@
...
-  [framework-id mesos-agent-query-cache pending-sync-agent task-id->sandbox-agent agent-hostname]
+  [{:keys [datomic-conn mesos-agent-query-cache pending-sync-agent task-id->sandbox-agent]} framework-id agent-hostname]
   (try
     (let [run (delay
                 (try
-                  (let [task-id->sandbox-directory
-                        (retrieve-sandbox-directories-on-agent framework-id agent-hostname)]
-                    (log/info "Found sandbox directories for" (count task-id->sandbox-directory) "tasks on" agent-hostname)
-                    (send task-id->sandbox-agent aggregate-sandbox task-id->sandbox-directory)
+                  (let [task-id->sandbox-directory (retrieve-sandbox-directories-on-agent framework-id agent-hostname)
```
wyegelwel: nitpick: Consider this rewrite:

```
(let [task-id->sandbox-directory (->> (retrieve-sandbox-directories-on-agent framework-id agent-hostname)
                                      (filter-task-ids-with-sandbox datomic-conn))] 
   ...)

File: scheduler/src/cook/mesos/sandbox.clj [hunk 3]
```diff
@@ -150,19 +136,38 @@
           (contains? #{:success :threshold} reason)
           (update :host->consecutive-failures dissoc hostname)))
 
+(defn- filter-task-ids-with-sandbox
```
wyegelwel: Nitpick: Consider renaming to `remove-task-ids-with-sandbox`

scheduler/src/cook/mesos/sandbox.clj
shamsimam: I use `task-id->sandbox-directory` in the logs below to get a sense of items that were dropped.

wyegelwel:
Bunch of nitpicking comments.

Not sure how easy this would be to add (if it is not easy, just create a to-do or issue) to track a metric of how long it takes from when the sandbox directory is available on the agent to when we transact it. We *want* that to be small so it would be nice to track it.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Done.

shamsimam:
@wyegelwel addressed the nitpicks. Let's do that metric in a separate PR. We can track it by stroing the entry time in the `task-id->sandbox-directory` map (i.e. it would now become a `task-id->sandbox-state` data structure. Then in the publish we can maintain a metric that stores the time difference as a histogram.

wyegelwel:
That's fine with me

shamsimam:
Created https://github.com/twosigma/Cook/issues/564

shamsimam:
@wyegelwel build is green

PR #566, opened 2017-11-14
fixes the partial call to render-jobs-for-response from destroy-jobs-handler

shamsimam:


pschorf:
Can you add an integration test for this?

shamsimam:
@pschorf okay

shamsimam:
Turns out `handle-ok` is unused in our deletes. I strengthened the assertions on the unit tests. We already have integration tests for delete.

pschorf:
Yeah I was honestly surprised the tests were passing as it was. Thanks for cleaning up the dead code.

PR #568, opened 2017-11-14
Optimize job-ent->map

pschorf:


scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+   (deep-transduce-kv (map identity) entity)))
+
+; During testing, we found that calling (keys) on a group entity with a large number of jobs was
+; quite slow. As an optimization, load the group keys ahead of time from the schema.
+(let [group-keys (->> schema/schema-attributes
+                      (map :db/ident)
+                      (filter #(= "group" (namespace %)))
+                      ; Do not load the nested job entities for a group. These should be queried on-demand
+                      ; as required.
+                      (filter #(not= :group/job %)))]
```
shamsimam: Nitpick: this is easier to read as a `remove`

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+  (defn job-ent->map
+    "Convert a job entity to a map. This also loads the associated group and converts it to a map."
+    ([job db]
+     (job-ent->map (d/entity db (:db/id job))))
+    ([job]
+     (let [group-ent (first (:group/_job job))
+           job (entity->map job)
+           group (when group-ent
+                   (->> group-keys
+                        (map (fn [k] [k (entity->map (get group-ent k))]))
```
shamsimam: nitpick: `pc/map-from-keys` might be a better fit here

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+                        (filter (fn [[k v]] (not (nil? v))))
```
shamsimam: nitpick: reads better as a `remove`.

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
-       ; inside the group entity. If someone needs to access them, it's better to query
-       ; at that point instead of relying on the cached data.
-       (assoc :group/_job (entity->map (:group/_job job) :skip-keys #{:group/job})))))
+  ([entity db]
+   (entity->map (d/entity db (:db/id entity))))
+  ([entity]
+   (deep-transduce-kv (map identity) entity)))
+
+; During testing, we found that calling (keys) on a group entity with a large number of jobs was
+; quite slow. As an optimization, load the group keys ahead of time from the schema.
```
shamsimam: question: Do we know why that call is slow? What is it iterating over to load those keys?

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+(let [group-keys (->> schema/schema-attributes
+                      (map :db/ident)
+                      (filter #(= "group" (namespace %)))
+                      ; Do not load the nested job entities for a group. These should be queried on-demand
+                      ; as required.
+                      (filter #(not= :group/job %)))]
+  (defn job-ent->map
```
shamsimam: Should we unit test this?

scheduler/src/cook/mesos/util.clj
pschorf: I don't know, without digging into datomic I'm not really sure. I wanted to get this wrapped up but I can spend some more time digging into it.

scheduler/src/cook/mesos/util.clj
pschorf: done

scheduler/src/cook/mesos/util.clj
pschorf: We already have unit tests, this change doesn't affect those (the returned object is the same)

scheduler/src/cook/mesos/util.clj
pschorf: IIRC it was faster this way (with the nil filtering before making the map) but I don't think it was really compelling either way. I kind of like this style but I am also open to changing it.

scheduler/src/cook/mesos/util.clj
pschorf: Done

scheduler/src/cook/mesos/util.clj
shamsimam: We can do it outside this PR and update the comment later.

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+                      ; Do not load the nested job entities for a group. These should be queried on-demand
+                      ; as required.
+                      (remove (partial = :group/job)))]
+  (defn job-ent->map
+    "Convert a job entity to a map. This also loads the associated group and converts it to a map."
+    ([job db]
+     (job-ent->map (d/entity db (:db/id job))))
+    ([job]
+     (let [group-ent (first (:group/_job job))
+           job (entity->map job)
```
shamsimam: question: what prevents `entity->map` from loading `:group/_job` (i.e. the group)?

pschorf: When you iterate over the datomic entities, it only loads the "forward" mappings (group->job, job->instance, etc) and not the "reverse" mappings (job->group, instance->job, etc). You can only access those by directly querying on the key.

scheduler/src/cook/mesos/util.clj
wyegelwel: Can we quantify "quite slow"?

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+; During testing, we found that calling (keys) on a group entity with a large number of jobs was
+; quite slow. As an optimization, load the group keys ahead of time from the schema.
+(let [group-keys (->> schema/schema-attributes
+                      (map :db/ident)
+                      (filter #(= "group" (namespace %)))
+                      ; Do not load the nested job entities for a group. These should be queried on-demand
+                      ; as required.
+                      (remove (partial = :group/job)))]
+  (defn job-ent->map
+    "Convert a job entity to a map. This also loads the associated group and converts it to a map."
```
wyegelwel: Please comment that it removes the jobs in the group

scheduler/src/cook/mesos/util.clj
```diff
@@ -91,22 +92,35 @@
...
+     (job-ent->map (d/entity db (:db/id job))))
+    ([job]
+     (let [group-ent (first (:group/_job job))
+           job (entity->map job)
+           group (when group-ent
+                   (->> group-keys
+                        (map (fn [k] [k (entity->map (get group-ent k))]))
+                        (remove (fn [[k v]] (nil? v)))
+                        (into {})
+                        hash-set))]
```
wyegelwel: why is hash-set here?

scheduler/src/cook/mesos/util.clj
wyegelwel: I prefer map-from-keys too. I think it is also ok to leave the nil values

scheduler/src/cook/mesos/util.clj
pschorf: It's what gets returned normally by datomic. I want to just refactor this whole thing into a proper record or something with explicit components but that's a bigger change, so to get the release out I just wanted to keep things compatible for now.

wyegelwel: oh because if you normally check that key it returns a set. I think that deserves a comment.

scheduler/src/cook/mesos/util.clj
pschorf: Done

scheduler/src/cook/mesos/util.clj
pschorf: Agreed, added

scheduler/src/cook/mesos/util.clj
pschorf: Added a comment for posterity

scheduler/src/cook/mesos/util.clj
pschorf: Done

wyegelwel: I don't see the change to `map-from-keys`. Did you forget to push that commit?

wyegelwel:
Code approved, but build failed

pschorf:
Build is passing now

PR #569, opened 2017-11-15
Adds support for querying using Cook Scheduler resource URL

dposada:
This PR introduces support in the CLI for querying for an entity using an "entity ref string". An entity ref string can either be simply a UUID, in which case all configured clusters and all three types (job, instance, and group) will be queried, or it can be a Cook Scheduler URL that represents the way to retrieve that specific entity.

Some examples of valid entity ref strings follow:

- 5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster2.example.com/jobs/5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster2.example.com/jobs?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster3.example.com/instances/5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster3.example.com/instances?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster4.example.com/groups/5ab383b1-5b8b-4483-b2b2-45126923f4df
- http://cluster4.example.com/groups?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df


DaoWen:
I like the idea of using `cluster:type/uuid` as the format, since the colon makes it look more like a resource access on a remote host. It would also let us allow (and clearly distinguish) other combinations, like `cluster:uuid` or `type/uuid`.

dposada:
With `cluster/type/uuid`, I was going for something that looks like a git ref, e.g. `refs/remotes/origin/master`. As far as allowing other combinations, you can do that with the current scheme using e.g. `cluster/*/uuid` or `*/type/uuid`. 

I worry a little about the Uncanny Valley effect with `cluster:type/uuid`, because it looks very close to a resource access on a remote host, but we're not really using it as such (i.e. we're not treating the cluster name as a host, instead we're using it as a name lookup in the configured clusters).

cli/cook/querying.py
```diff
@@ -59,31 +85,55 @@ def make_group_request(cluster, uuids):
...
+
+    # Query for instances
+    instance_uuids = uuids_by_type[Types.INSTANCE]
+    instance_parent_job_pairs = []
+    parent_jobs = query_cluster(cluster, instance_uuids, pred_instances, timeout,
+                                interval, make_instance_request, Types.INSTANCE)
+    for job in parent_jobs:
+        for instance in job['instances']:
+            instance_uuid = instance['task_id']
+            if instance_uuid in instance_uuids:
```
DaoWen: Nitpick: I don't think introducing `instance_uuid` actually gains us anything.

cli/cook/querying.py
```diff
@@ -190,19 +230,60 @@ def __get_latest_instance(job):
...
+    """
+    entity_refs = []
+    for ref_string in ref_strings:
+        ref_string = ref_string.strip()
+        ref_parts = ref_string.split('/')
+        num_parts = len(ref_parts)
+        if num_parts == 1:
+            entity_refs.append({'cluster': Clusters.ALL, 'type': Types.ALL, 'uuid': ref_parts[0]})
+        elif num_parts == 3:
+            entity_refs.append({'cluster': unquote_plus(ref_parts[0]), 'type': ref_parts[1], 'uuid': ref_parts[2]})
```
DaoWen: Why is the call to `unquote_plus` needed here?

cli/cook/subcommands/kill.py
```diff
@@ -91,7 +91,7 @@ def kill_entities(query_result, clusters):
 def kill(clusters, args, _):
     """Attempts to kill the jobs / instances / groups with the given UUIDs."""
     guard_no_cluster(clusters)
-    uuids = strip_all(args.get('uuid'))
+    uuids = parse_entity_refs(args.get('uuid'))
```
DaoWen: Should we have an alias for the `uuid` option? Maybe `entity`?

cli/cook/querying.py
dposada: The idea is that the cluster name may have a `/` in it, in which case commands which output entity refs (in a future PR) will escape them in the cluster portion of the ref

cli/cook/subcommands/kill.py
dposada: What do you mean by an alias?

cli/cook/querying.py
dposada: Agreed, removed

cli/cook/subcommands/kill.py
DaoWen: I forgot that these are positional arguments. I was thinking you had to do something like `cs kill --uuid $x`, in which case supporting `cs kill --entity $x` might be good too.

cli/cook/querying.py
```diff
@@ -190,19 +229,60 @@ def __get_latest_instance(job):
...
+      */group/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      cluster2/job/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      cluster3/instance/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      cluster4/group/5ab383b1-5b8b-4483-b2b2-45126923f4df
+
+    Throws if an invalid entity ref string is encountered.
+    """
+    entity_refs = []
+    for ref_string in ref_strings:
+        ref_string = ref_string.strip()
```
DaoWen: Remove this strip?

cli/cook/querying.py
DaoWen: Based on our conversation just now, it sounds like we want to use `unquote` to avoid having any `+` characters in the cluster name changed to spaces.

cli/cook/subcommands/kill.py
dposada: Got it, yeah, they are positional

cli/cook/querying.py

dposada:
@DaoWen Any more feedback on this?

DaoWen:
Is this WIP? If I remember correctly, you wanted to merge a change to create those new endpoints before merging PR. Right?

dposada:
Correct, I think it makes sense to add the new endpoints before merging this.

That being said, this change is complete (except for feedback).

cli/cook/querying.py
```diff
@@ -190,19 +227,91 @@ def __get_latest_instance(job):
...
+                        raise Exception(f'{ref_string} refers to an unsupported resource.')
+
+                    cluster_name = cluster_names[0]
+                    if num_path_parts > 2:
+                        entity_refs.append({'cluster': cluster_name, 'type': entity_type, 'uuid': path_parts[2]})
+                    elif result.query:
+                        query_args = parse_qs(result.query)
+                        if 'uuid' in query_args:
+                            for uuid in query_args['uuid']:
+                                entity_refs.append({'cluster': cluster_name, 'type': entity_type, 'uuid': uuid})
```
DaoWen: It's not clear to me if we actually want this branch. (Same concern as with the sample URLs in the docstring.)

cli/cook/querying.py
```diff
@@ -190,19 +227,91 @@ def __get_latest_instance(job):
...
+
+    Some examples of valid entity ref strings follow:
+
+      5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster2.example.com/jobs/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster2.example.com/jobs?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster3.example.com/instances/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster3.example.com/instances?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster4.example.com/groups/5ab383b1-5b8b-4483-b2b2-45126923f4df
+      http://cluster4.example.com/groups?uuid=5ab383b1-5b8b-4483-b2b2-45126923f4df
```
DaoWen: Why do we want to support both a `/«uuid»` and `?uuid=«uuid»`? Since all of these endpoints are new anyway, why not just do away with the non-REST-like use of query parameters to identify the resource? If we want to be able to identify more than one group in the same URL, maybe we should have a special identifier in place of the UUID, e.g.: `http://cluster2.example.com/jobs/multiple?uuid=...&uuid=...&uuid=...`

integration/tests/cook/cli.py
```diff
@@ -90,20 +90,36 @@ def __show_json(uuids, cook_url=None, flags=None):
...
     cp, data = __show_json(uuids, cook_url, flags)
     jobs = [job for entities in data['clusters'].values() for job in entities['jobs']]
     return cp, jobs
 
 
+def show_instances(uuids, cook_url=None, flags=None):
+    """Shows the instance JSON corresponding to the given UUID(s)"""
+    cp, data = __show_json(uuids, cook_url, flags)
+    instance_job_pairs = [pair for entities in data['clusters'].values() for pair in entities['instances']]
+    return cp, instance_job_pairs
```
DaoWen: Again, are we changing the way instances are displayed? Assuming so, does this need to change?

cli/cook/querying.py
```diff
@@ -190,19 +227,91 @@ def __get_latest_instance(job):
...
+                if len(cluster_names) >= 1:
+                    resource = path_parts[1].lower()
+                    if resource == 'jobs':
+                        entity_type = Types.JOB
+                    elif resource == 'instances':
+                        entity_type = Types.INSTANCE
+                    elif resource == 'groups':
+                        entity_type = Types.GROUP
+                    else:
+                        raise Exception(f'{ref_string} refers to an unsupported resource.')
```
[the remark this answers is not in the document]
DaoWen: You've addressed this in #582.

cli/cook/querying.py
```diff
@@ -59,31 +83,54 @@ def make_group_request(cluster, uuids):
...
+
+    # Query for instances
+    instance_uuids = uuids_by_type[Types.INSTANCE]
+    instance_parent_job_pairs = []
+    parent_jobs = query_cluster(cluster, instance_uuids, pred_instances, timeout,
+                                interval, make_instance_request, Types.INSTANCE)
+    for job in parent_jobs:
+        for instance in job['instances']:
+            if instance['task_id'] in instance_uuids:
+                instance_parent_job_pairs.append((instance, job))
```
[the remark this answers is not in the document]
DaoWen: :ok: I think it makes sense to do this in a separate PR. I've created an issue to track this: #589.

cli/cook/querying.py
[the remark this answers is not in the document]
DaoWen: :ok: I think it makes sense to do this in a separate PR. I've created an issue to track this: #589.

integration/tests/cook/cli.py
DaoWen: :ok: I think it makes sense to do this in a separate PR. I've created an issue to track this: #589.

cli/cook/querying.py
```diff
@@ -190,19 +227,91 @@ def __get_latest_instance(job):
...
+                        else:
+                            raise Exception(f'Unable to determine UUID from {ref_string}.')
+                    else:
+                        raise Exception(f'Unable to determine UUID from {ref_string}.')
+                else:
+                    raise Exception(f'There is no configured cluster that matches {ref_string}.')
+            else:
+                raise Exception(f'Unable to determine entity type and UUID from {ref_string}.')
+        else:
+            raise Exception(f'{ref_string} is not a valid entity reference.')
```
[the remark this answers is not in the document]
DaoWen: :x: I think it would be worth making this change before merging. (The other things I can comment on can wait for a different PR.)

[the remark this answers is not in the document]
dposada: Good call, done

cli/cook/querying.py
```diff
@@ -274,44 +289,42 @@ def parse_entity_refs(clusters, ref_strings):
     entity_refs = []
     for ref_string in ref_strings:
         result = urlparse(ref_string)
-        if result.path and not result.netloc:
-            if is_valid_uuid(result.path):
-                entity_refs.append({'cluster': Clusters.ALL, 'type': Types.ALL, 'uuid': result.path})
-            else:
+
+        if not result.path:
+            raise Exception(f'{ref_string} is not a valid entity reference.')
```
DaoWen: :+1: I find this is *much* easier to read now.

PR #571, opened 2017-11-15
WIP: Fixes bug with cs kill-ing multiple UUIDs

dposada:


cli/cook/subcommands/kill.py
```diff
@@ -18,7 +18,8 @@ def guard_against_duplicates(query_result):
     def add(uuid, entity_type, cluster):
         entry_map = {'type': entity_type, 'cluster_name': cluster}
         uuid_to_entries[uuid].append(entry_map)
-        duplicate_uuids.add(uuid)
+        if len(uuid_to_entries[uuid]) > 1:
```
DaoWen: Nitpick: Should this be `== 2` rather than `> 1`? (Does it make sense to add it again for 3, 4, ...?) It might actually be more clear to move this check up 2 lines (before doing the `append`), and change it to `if uuid in uuid_to_entries`.

dposada: Done, thanks

PR #573, opened 2017-11-16
Fixes bug with cs ls on empty sandbox directory

dposada:


cli/cook/subcommands/ls.py
```diff
@@ -76,7 +76,7 @@ def browse_files(instance, sandbox_dir, path):
 def ls_for_instance(instance, sandbox_dir, path, long_format, as_json):
     """Lists contents of the Mesos sandbox path for the given instance"""
     entries = browse_files(instance, sandbox_dir, path)
-    if len(entries) == 0:
+    if len(entries) == 0 and path:
```
DaoWen: Why would the path be falsy? Is it `""`, or `None`, or either?

I'm not clear why that's a problem, and what it has to do with the sandbox directory being empty. Adding a comment to clarify what's going on here might be good.

dposada:
Nope, this is RTM

PR #574, opened 2017-11-16
disable triggering syncing agent sandbox entries into datomic

shamsimam:


scheduler/docs/configuration.adoc
```diff
@@ -89,13 +89,12 @@ If need it to bind to another port, you can specify that with the `:local-port`
 `:sandbox-syncer`::
```
pschorf: I think the name `sandbox-syncer` isn't appropriate now that it's just a publisher.

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -120,77 +120,32 @@
...
-                 datomic-db task-ids))
-          (remove-task-id-with-sandbox [task-id->sandbox task-id]
-            (dissoc! task-id->sandbox task-id))]
-    (->> (keys task-id->sandbox)
-         (retrieve-task-ids-with-sandbox (d/db datomic-conn))
-         (reduce remove-task-id-with-sandbox (transient task-id->sandbox))
-         persistent!)))
-
 (defn refresh-agent-cache-entry
   "If the entry for the specified agent is not cached:
```
pschorf: update comment

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -212,63 +178,23 @@
...
         publisher-cancel-fn (start-sandbox-publisher
                               task-id->sandbox-agent datomic-conn publish-batch-size publish-interval-ms)
-        syncer-cancel-fn (start-host-sandbox-syncer publisher-state sync-interval-ms max-consecutive-sync-failure)]
+        hostname->task-id->sandbox-directory-fn (fn hostname->task-id->sandbox-directory-fn [agent-hostname task-id]
+                                                  (hostname->task-id->sandbox-directory
+                                                    publisher-state framework-id agent-hostname task-id))]
     (assoc publisher-state
-      :publisher-cancel-fn publisher-cancel-fn
-      :syncer-cancel-fn syncer-cancel-fn)))
+      :hostname->task-id->sandbox-directory-fn hostname->task-id->sandbox-directory-fn
```
pschorf: I really don't like the name of this function, is there something shorter that makes sense?

scheduler/docs/configuration.adoc
shamsimam: Renamed to `sandbox-publisher`

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Updated.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Renamed to `retrieve-sandbox-directory-from-agent`

PR #576, opened 2017-11-17
WIP: does not trigger sandbox syncing for cook executor tasks

shamsimam:


shamsimam:
We've disabled the sandbox syncer on task start.

dposada:
Can we delete the branch?

PR #577, opened 2017-11-17
Disable fenzo group constraints

pschorf:


wyegelwel:
What is the context for this?

shamsimam:
We did some performance benchmarking and this seems to slow down the handle resource offer rate when using group constraints. We need more time to develop and fine tune this feature, so we are removing it from branch to avoid any potential performance issues meanwhile.

wyegelwel:
> We did some performance benchmarking and this seems to slow down the handle resource offer rate when using group constraints. We need more time to develop and fine tune this feature, so we are removing it from branch to avoid any potential performance issues meanwhile.

Isn't that expected? How much slower was it?

pschorf:
Given the performance impact, I think it makes more sense to explicitly disable these constraints while. I do a deeper dive into what is going on but keep the bug fix for loading the groups with the jobs since that doesn't impact the performance. I am planning on reverting this in a branch and spending some time looking into what's actually going on, then roll forward.

DaoWen:
We should re-enable `test_balanced_host_constraint_cannot_place` (see #579) after this is merged.

pschorf:
I don't think reverting it will be too bad because the actual scheduler change is only a few lines, almost all the changes are to the test code which should be relatively easy to add back in.

integration/tests/cook/test_basic.py
```diff
@@ -1241,208 +1241,3 @@ def test_unscheduled_jobs_partial(self):
...
-                            return sub_reason
-                return None
-
-            reason = util.wait_until(query, lambda r: r is not None)
-            self.assertEqual(reason['reason'],
-                             "Host had a different attribute than other jobs in the group.")
-        finally:
-            util.kill_jobs(self.cook_url, uuids)
-
-    def test_retrieve_jobs_with_deprecated_api(self):
```
dposada: I'm guessing deleting this test was unintentional?

pschorf: PR incoming

PR #578, opened 2017-11-17
specifies the temp directory used by pyinstaller …

shamsimam:
…to unpackage the cook executor


executor/Dockerfile.build
```diff
@@ -10,4 +10,4 @@ RUN pip install -r requirements.txt
 
 COPY . /opt/cook
 
-CMD ["pyinstaller", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
+CMD ["pyinstaller", "--runtime-tmpdir", ".cook-executor", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
```
dposada: What is this relative to? Should we make it absolute?

dposada: (same question for the other places)

shamsimam: We want it to be relative to the sandbox directory, we don't have access to that path while running this command. I can add a comment in the Cook executor README if that helps.

shamsimam:
@dposada build is green now

Issue #579, opened 2017-11-20
test_balanced_host_constraint_cannot_place is flaky


DaoWen:
We've been seeing repeated failure of this test. I think Paul has already determined that this is due to some broken logic in the constraints handling code.

```
======================================================================
ERROR: test_balanced_host_constraint_cannot_place (tests.cook.test_basic.CookTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/twosigma/Cook/integration/tests/cook/test_basic.py", line 1338, in test_balanced_host_constraint_cannot_place
    placement_reasons = util.wait_until(query_unscheduled, lambda r: len(r) > 0)
  File "/home/travis/build/twosigma/Cook/integration/tests/cook/util.py", line 242, in wait_until
    return wait_until_inner()
  File "/home/travis/build/twosigma/Cook/integration/.eggs/retrying-1.3.3-py3.6.egg/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/home/travis/build/twosigma/Cook/integration/.eggs/retrying-1.3.3-py3.6.egg/retrying.py", line 212, in call
    raise attempt.get()
  File "/home/travis/build/twosigma/Cook/integration/.eggs/retrying-1.3.3-py3.6.egg/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/home/travis/.pyenv/versions/3.6.2/lib/python3.6/site-packages/six-1.11.0-py3.6.egg/six.py", line 693, in reraise
    raise value
  File "/home/travis/build/twosigma/Cook/integration/.eggs/retrying-1.3.3-py3.6.egg/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/home/travis/build/twosigma/Cook/integration/tests/cook/util.py", line 236, in wait_until_inner
    raise RuntimeError(error_msg)
RuntimeError: wait_until condition not yet met, retrying...
-------------------- >> begin captured logging << --------------------
tests.cook.util: INFO: Using cook url http://localhost:12321
tests.cook.util: INFO: Using cook url http://localhost:12321
tests.cook.util: DEBUG: Waiting for connection to cook...
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): localhost
[...]

PR #580, opened 2017-11-20
Mark test_balanced_host_constraint_cannot_place as explicit

DaoWen:
See #579.


DaoWen:
@pschorf - We should re-enable this test once the host constraint logic is fixed. In the meantime, I want to disable it to avoid the constant Travis failures.

shamsimam:
Awaiting green build

PR #581, opened 2017-11-20
uses one-folder cook executor (instead of one-file) built using…

shamsimam:
…pyinstaller


scheduler/bin/run-docker.sh
```diff
@@ -31,7 +31,7 @@ fi
 SCHEDULER_DIR="$( dirname ${DIR} )"
 SCHEDULER_EXECUTOR_DIR=${SCHEDULER_DIR}/resources/public
 EXECUTOR_NAME=cook-executor
-COOK_EXECUTOR_COMMAND="mkdir .cook-executor && ./${EXECUTOR_NAME}"
+COOK_EXECUTOR_COMMAND="./${EXECUTOR_NAME}/${EXECUTOR_NAME}"
```
DaoWen: Maybe we should still make this a hidden file? (It seems like the consensus last week was that we don't want to clutter up the user's sandbox directory with files that they're not expecting.)
```
"./.${EXECUTOR_NAME}/${EXECUTOR_NAME}"
~~~^
```
I think that would require changing the build script to rename the folder. But if you guys think users won't be bothered by the presence of a `cook-executor` directory in their sandbox, then I'm happy leaving it like this.

shamsimam: This path is used mainly for local development where the tar file is served by the scheduler and unzipped by the Mesos fetcher. I'd vote for avoiding the extra work for this scenario which is mainly used in development.

executor/Dockerfile.build
```diff
@@ -10,4 +10,4 @@ RUN pip install -r requirements.txt
 
 COPY . /opt/cook
 
-CMD ["pyinstaller", "--runtime-tmpdir", ".cook-executor", "-F", "-n", "cook-executor", "-p", "cook", "cook/__main__.py"]
+CMD ["pyinstaller", "--onedir", "--name", "cook-executor", "--paths", "cook", "cook/__main__.py"]
```
dposada: What does `--onedir` accomplish? Should we add a comment explaining?

executor/bin/build-local.sh
```diff
@@ -11,6 +11,7 @@ NAME=cook-executor-build
 EXECUTOR_DIR="$(dirname ${DIR})"
 
 mkdir -p ${EXECUTOR_DIR}/dist
+rm -rf ${EXECUTOR_DIR}/dist/cook-executor-local
 
 cd ${EXECUTOR_DIR}
-pyinstaller --runtime-tmpdir .cook-executor -F -n cook-executor-local -p cook cook/__main__.py
+pyinstaller --onedir --name cook-executor-local --paths cook cook/__main__.py
```
dposada: Same question as above

scheduler/bin/run-docker.sh
DaoWen: Ah, right—this is in the run-docker script. That makes sense. No point in changing this for docker.

executor/Dockerfile.build
shamsimam: Done.

executor/bin/build-local.sh
shamsimam: Done.

shamsimam:
@dposada  I also added a modification to support running locally from the one-folder version directly.

scheduler/src/cook/components.clj
```diff
@@ -404,10 +402,15 @@
                                                     :progress-sample-interval-ms (* 1000 60 5)}
                            default-uri-config {:cache true
                                                :executable true
-                                               :extract false}]
+                                               :extract false}
+                           executor-uri-configured (and (:uri executor) (not (str/blank? (get-in executor [:uri :value]))))]
+                       (when executor-uri-configured
```
dposada: `when-not`?

scheduler/src/cook/components.clj
```diff
@@ -404,10 +402,15 @@
...
+                         (log/warn "Executor uri value is missing, the uri config will be disabled" {:executor executor}))
```
dposada: Should this be `info` instead of `warn`?

scheduler/src/cook/components.clj
```diff
@@ -404,10 +402,15 @@
...
                        (cond-> (merge default-executor-config executor)
-                               (:uri executor)
-                               (update :uri #(merge default-uri-config %1)))))))
+                               executor-uri-configured
+                               (update :uri #(merge default-uri-config %1))
+                               (not executor-uri-configured)
+                               (dissoc :uri))))))
```
dposada: Seems like this would be cleaner as an `if executor-uri-configured`

scheduler/src/cook/components.clj
shamsimam: Done.

scheduler/src/cook/components.clj
shamsimam: Done.

scheduler/src/cook/components.clj
shamsimam: Done.

shamsimam:
@dposada build is now green

PR #586, opened 2017-11-21
Renames cs list to cs jobs

dposada:


cli/cook/cli.py
```diff
@@ -3,7 +3,7 @@
 from urllib.parse import urlparse
 
 from cook import util, http, metrics, version, configuration
-from cook.subcommands import submit, show, wait, list, ssh, ls, tail, kill, config
+from cook.subcommands import submit, show, wait, jobs, ssh, ls, tail, kill, config
```
shamsimam: Nitpick: we could sort these too :)

PR #590, opened 2017-11-22
Fix check-version script for onedir-mode executor

DaoWen:


shamsimam:
Awaiting green build.

PR #591, opened 2017-11-22
Support task constraint in job client

WenboZhao:
To create a constraint and supply it to `Job.Builder`,  one could do
```java
Constraint c = Constraints.buildEqualsConstraint("host", "foo.bar.com");
...
jobBuilder.addConstraint(c);
```
The implementation is based on the following docs:
 - https://github.com/twosigma/Cook/blob/master/scheduler/docs/scheduler-rest-api.adoc#constraints
 - In api code: https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/api.clj#L208
 - Example usage in integration tests: https://github.com/twosigma/Cook/blob/master/integration/tests/cook/test_basic.py#L782


jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -93,10 +93,11 @@ public static Status fromString(String name) {
         private Boolean _isMeaCulpaRetriesDisabled;
         private List<Instance> _instances = Collections.synchronizedList(new ArrayList<Instance>());
         private List<FetchableURI> _uris = new ArrayList<>();
-        private Map<String,String> _env = new HashMap<>();
+        private Map<String, String> _env = new HashMap<>();
         private JSONObject _container;
-        private Map<String,String> _labels = new HashMap<>();
+        private Map<String, String> _labels = new HashMap<>();
         private List<UUID> _groups = new ArrayList<>();
+        private List<TaskConstraint.Constraint> _constraints = new ArrayList<>();
```
shamsimam: Should this be a set instead of a list?

jobclient/src/test/java/com/twosigma/cook/jobclient/JobTest.java
```diff
@@ -38,6 +39,11 @@
      */
     private Job _initializedJob;
 
+    /**
+     * A constraint will be applied to the job.
+     */
+    private TaskConstraint.Constraint _JobConstraint;
```
shamsimam: Should we also add tests in `cook.test.jobclient.jobclient`?

WenboZhao:
I changed the API which I believe it make more sense now. Also add more tests.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: I prefer `List` here. Is there any reason we use `Set`.

shamsimam: `Set` will mean that we cannot repeat the same constraint. With lists, it will be possible to repeat a constraint and will unnecessarily be transmitted in the JSON.

WenboZhao: That make sense. Will change that accordingly

jobclient/src/test/java/com/twosigma/cook/jobclient/JobTest.java
WenboZhao: Added.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,170 @@
...
+ * A simple class to represent a constraint.
+ *
+ * A constraint has 3 fields: an attribute, an operator, and a value.
+ * An example of constraints could be: "host","EQUALS", "foo.bar.com".
+ *
+ * For now, only "EQUALS" operator is supported by Cook. One could construct
+ * a constraint with an "EQUALS" operator by
+ *
+ * <pre>
+ *  Constraint c = Constraint.EQUALS.apply("host", "foo.bar.com");
```
wyegelwel: Nitpick: Can you add the other way that a constraint can be constructed : 

```java
Constraint c = new Constraint("host", Constraint.EQUALS, "foo.bar.com")
```

I find this version easier to read.

Also, I haven't coded java recently so I expect I am wrong but should this be `Constraint.Operator.EQUALS`?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,170 @@
...
+    Constraint(String attribute, Operator operator, String value) {
+        Preconditions.checkNotNull(attribute);
+        Preconditions.checkNotNull(operator);
+        Preconditions.checkNotNull(value);
+        _attribute = attribute;
+        _operator = operator;
+        _value = value;
+    }
+
+    Constraint(String attribute, String operator, String value) {
```
wyegelwel: This looks like a duplicate of the constructor on line 69

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: I explicitly added the `EQUALS` in the constraint level.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        Constraint that = (Constraint) o;
+
+        if (_attribute != null ? !_attribute.equals(that._attribute) : that._attribute != null) return false;
```
wyegelwel: It took me a really long time to grok this, would you mind changing to strict boolean logic. I *think* this is the correct switch:

```java
if (_attribute != null && that._attribute == null) return false;
if (_attribute == null && that._attribute != null) return false;
if (_attribute != null && !_attribute.equals(that._attribute)) return false;
```

I know it is less concise but this makes it much easier to verify that the checks are correct.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: Yes, all package level constructors. I should call the other one internally

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
wyegelwel: Never mind, I see you are binding it at the bottom. Adding a comment to that line

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,170 @@
...
+        JSONArray jsonArray = new JSONArray();
+        jsonArray.put(0, _attribute);
+        jsonArray.put(1, _operator.toString());
+        jsonArray.put(2, _value);
+        return jsonArray;
+    }
+
+    // Supported operators
+
+    final public static Operator EQUALS = Operator.EQUALS;
```
wyegelwel: Can you move this binding as close to the operator class definition as possible?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,162 @@
...
```
shamsimam: Nitpick: This is redundant. Can we just refer `Operator.EQUALS` directly?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: This is a generated code. Not written by me.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: sure thing

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -276,26 +281,26 @@ public Builder addLabel(String key, String value) {
...
-         *
+         * <p>
          * This resets the labels of the job; it will remove labels that were previously set.
          *
          * @param labels specifies the labels to set for this job.
          * @return this builder
          */
-        public Builder setLabels(Map<String,String> labels) {
+        public Builder setLabels(Map<String, String> labels) {
             _labels = ImmutableMap.copyOf(labels);
```
shamsimam: I know this isn't part of this PR, but a line in this method changed. But, since `_labels` is now an immuatble map, doesn't this mean that calls to `addLabels` will now fail? Do we have a unit test that covers this case? I'm okay if this is an issue and covered in a separate PR.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -238,34 +243,34 @@ public Builder addEnv(String name, String value) {
...
         /**
          * Adds a collection of env vars to the job.
-         *
+         * <p>
          * This resets the enviroment to the job; it will remove variables that were previously set.
          *
          * @param environment specifies the environment to set for this job.
          * @return this builder
          */
-        public Builder setEnv(Map<String,String> environment) {
```
shamsimam: Same `ImmutableMap`  issue as `_labels`. Please see below.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        /**
+         * Returns a constraint by "applying" this operator to the given attribute and value.
+         *
+         * @param attribute The constraint attribute
+         * @param value     The constraint value
+         * @return a constraint.
+         */
+        public Constraint apply(String attribute, String value) {
```
shamsimam: Do we expect all operators we add in the future to live up to this specification?

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
WenboZhao: I will address it in different PR

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: I am ok either way. I just thought ```Constraint.EQUALS``` is more user friendly.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: That is good question but I guess I am not the right person to answer it :-)

shamsimam: Since we are not sure, can we avoid this restriction then?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
wyegelwel: I agree with @WenboZhao. A middle ground would be Constraint.Operator.EQUALS but then there are more characters.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
wyegelwel: You could imagine an operator "IN" which takes a list of values. @shamsimam, @WenboZhao  what api would you imagine in that case?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
wyegelwel: intellij is smarter than me. I'd prefer switching it to the boolean logic but knowing it is generated is comforting. @shamsimam what do you think?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: I will refactor those ```Operator``` to e.g. 
```
OneToZeroOperator;   \\ E.g. UNIQUE
OneToOneOperator;    \\ E.g. EQUALS
OneToManyOperator;  \\ E.g. IN
```

shamsimam: My vote would be to do away with the `apply` method and have an external factory method that builds the constraint for each specific operator as the Java generics support will not be able to help us here.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
shamsimam: I like that @wyegelwel version is more readable, but I'm good either way.

WenboZhao: Maybe in the future, machine/AI generated codes will be completely human unreadable and we have to just trust them. I will just keep as it is.

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: hmm, actually, the `Constraint.XXX.apply` way is doing a factory pattern here. Although I admit that `build` instead `apply` is a better for java style naming convention. `apply()` is too scala-ish.

shamsimam: I am not against factory pattern that the `apply`/`build` methods offer. My issue is with the signature of those methods. If we keep the operators as enums (I like that they are enums currently), we do not have a good way to deal with individual signatures for the different operators. So, we may need to introduce explicit factory methods in `Constraint` like `buildEqualsConstraint(String, String)` and `buildInConstraint(String, Collection<String>)`

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
DaoWen: This logic looks like the inverse of [Objects.equals](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-); i.e., it would be much simpler to write this:
```java
if (!java.util.Objects.equals(_attribute, that._attribute)) return false;
```
I'm guessing that the code generator was written before Java 7, which is why they generate this complicated ternary construct instead of using the library. It's too bad that they haven't updated the templates. If you want to make this method more readable, I recommend delegating to the `Objects` static helpers. However, I'm OK with leaving the auto-generated code too—I totally understand being hesitant to "improve" working code when really we're better off just re-generating the methods if we ever update the class again.

jobclient/src/main/java/com/twosigma/cook/jobclient/constraint/api/Constraint.java
```diff
@@ -0,0 +1,60 @@
...
+ * <li>attribute, operator, list of values
+ * </ul>
+ * For examples,
+ * <ul>
+ * <li>"host", UNIQUE
+ * <li>"host", EQUALS, foo.bar.com
+ * <li>"host", IN, [foo1.bar.com,foo2.bar.com]
+ * </ul>
+ */
+public interface Constraint {
```
shamsimam: I think we can use generics here and have `Constraint<V>` with `V getValue()` and change Constraint to a final class.

jobclient/src/main/java/com/twosigma/cook/jobclient/constraint/api/ConstraintBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.twosigma.cook.jobclient.constraint.api;
+
+import org.json.JSONArray;
+
+public interface ConstraintBuilder {
```
shamsimam: I imagined the constraint builder to have factory methods like:
`Constraint<String> buildEqualsConstraint(String, String)` and `Constraint<String[]> buildInConstraint(String, String[])`

jobclient/src/main/java/com/twosigma/cook/jobclient/constraint/Constraints.java
```diff
@@ -0,0 +1,70 @@
...
+    public static ConstraintBuilder get(Operator operator) {
+        switch (operator) {
+            case EQUALS:
+                return new OneToOneConstraintBuilder(operator);
+            default:
+                throw new UnsupportedOperationException(operator + " is not supported.");
+        }
+    }
+
+    public static Constraint parseFrom(JSONArray constraint) {
```
shamsimam: Once we have the operator, we can call `operator.parseFrom(constraint)` with the necessary casting, where `parseFrom` is an abstract method on the `enum Operator` and avoid the need for `ConstraintBuilder.parseFrom`.

jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -883,6 +903,12 @@ public static JSONObject jsonizeJob(Collection<Job> jobs)
                     jobBuilder.addUri(FetchableURI.parseFromJSON(urisJson.getJSONObject(j)));
                 }
             }
+            if (json.has("constraints")) {
+                JSONArray constraintsJson = json.getJSONArray("constraints");
+                for (int j = 0; j < constraintsJson.length(); j++) {
```
shamsimam: We can parse out the operators and call the `operator.parseFrom()` here and avoid the need for the `Constraints` class.

jobclient/src/main/java/com/twosigma/cook/jobclient/constraint/api/ConstraintBuilder.java
WenboZhao: How about the UNIQUE operator?

jobclient/src/main/java/com/twosigma/cook/jobclient/Constraint.java
WenboZhao: updated.

wyegelwel:
This update looks good to me. 

@shamsimam @DaoWen what do you think?

PR #592, opened 2017-11-24
reads progress messages from stdout if the progress file is set to /dev/null

shamsimam:


executor/cook/config.py
```diff
@@ -48,21 +49,25 @@ def __init__(self,
...
 
 def initialize_config(environment):
     """Initializes the config using the environment.
     Populates the default values for missing environment variables.
     """
     executor_id = environment.get('MESOS_EXECUTOR_ID', 'executor')
     sandbox_directory = environment.get('MESOS_SANDBOX', '')
-    default_progress_output_name = environment.get('EXECUTOR_DEFAULT_PROGRESS_OUTPUT_NAME',
-                                                   '{}.progress'.format(executor_id))
+    default_progress_output_name = \
```
pschorf: Why did this change? I'm not a fan of the trailing slash syntax.

shamsimam: I want the `get` arguments to be on the same line. I undid the change by introducing a shorter name variable.

executor/cook/progress.py
```diff
@@ -7,12 +7,25 @@
...
+
 class ProgressUpdater(object):
     """This class is responsible for sending progress updates to the scheduler.
     It throttles the rate at which progress updates are sent.
     """
 
-    def __init__(self, driver, task_id, max_message_length, poll_interval_ms, send_message_fn):
+    def __init__(self, task_id, max_message_length, poll_interval_ms, send_progress_message_fn):
         """
         driver: MesosExecutorDriver
```
sradack: `driver` is no longer an parameter.

executor/cook/progress.py
```diff
@@ -138,6 +156,10 @@ def tail(self, sleep_time_ms):
         try:
             sleep_param = sleep_time_ms / 1000
 
+            if os.path.exists(self.target_file) and not os.path.isfile(self.target_file):
```
sradack: Do we need this?  Why wouldn't we change for `!isfile` after the file is created in the loop below?

executor/cook/progress.py
```diff
@@ -231,41 +252,50 @@ def track_progress(progress_watcher, progress_complete_event, send_progress_upda
...
-                                    args=(progress_watcher, progress_complete_event,
-                                          progress_updater.send_progress_update))
-    progress_update_thread.start()
-    return progress_complete_event
+    def __init__(self, task_id, config, stop_signal, task_completed_signal, send_progress_message, location, counter):
+        """Launches the threads that track progress and send progress updates to the driver.
 
+        Parameters
+        ----------
+        driver: MesosExecutorDriver
```
sradack: `driver` is not a parameter

executor/cook/progress.py
```diff
@@ -231,41 +252,50 @@ def track_progress(progress_watcher, progress_complete_event, send_progress_upda
...
-    -------
-    Nothing.
-    """
-    latest_progress = progress_watcher.current_progress()
-    progress_updater.send_progress_update(latest_progress, force_send=True)
+    def wait(self, timeout=None):
+        """Waits for the progress tracker thread to run to completion."""
+        self.progress_complete_event.wait(timeout=timeout)
+        if self.progress_complete_event.isSet():
+            logging.info('Progress monitoring from {} complete'.format(self.watcher.progress_target_file()))
```
sradack: A log message would also be useful if the event *wasn't* set.

executor/cook/progress.py
```diff
@@ -231,41 +252,50 @@ def track_progress(progress_watcher, progress_complete_event, send_progress_upda
...
-        The progress watcher which maintains the current progress state.
-    progress_updater: ProgressUpdater
-        The progress updater which sends progress updates to the scheduler.
-        
-    Returns
-    -------
-    The progress complete threading.Event.
-    """
-    progress_complete_event = Event()
-    progress_update_thread = Thread(target=track_progress,
```
sradack: Why isn't `track_progress` part of `ProgressTracker`?

executor/cook/progress.py
shamsimam: Okay, will add

executor/cook/progress.py
shamsimam: Will remove.

executor/cook/progress.py
shamsimam: I think the two `if` clauses address different scenarios. 
- This `if` specifies that the location specified already exists and is not a file.

The `while` loop will keep running for inputs like `/dev/null` which aren`t files and which will never be created. We would only stop looping once `task_completed_signal` is set. The `if` condition below is `true` only after the task completed and is a notification that we were expecting to read from a file which was never created.

executor/cook/progress.py
shamsimam: Will move.

executor/cook/progress.py
shamsimam: Will remove.

shamsimam:
> I'm not sure how much we gained with the creation of the `ProgressTracker` class

It simplifies the `manage_task` method, else we would have to return all the created instances and pass them along to subsequence method calls.

executor/cook/progress.py
shamsimam: removed.

executor/cook/progress.py
shamsimam: removed.

executor/cook/progress.py
shamsimam: added log message

executor/cook/progress.py
shamsimam: Moved as a static method inside `ProgressTracker` to keep testing simple.

executor/cook/progress.py
sradack: A "non-file" may be created after this `if` statement.  I may not be following.

executor/cook/progress.py
sradack: It still feels awkward to me.  We seem to be mixing OO and procedural style.

executor/cook/progress.py
shamsimam: Let's talk offline

executor/cook/progress.py
shamsimam: Made `track_progress` non-static and removed the unit test. We have other tests that cover this function.

shamsimam:
Build is green

PR #596, opened 2017-11-27
adds logic to skip faulty progress entries in the cook executor

shamsimam:


executor/cook/progress.py
```diff
@@ -192,17 +192,22 @@ def retrieve_progress_states(self):
...
-                                     'progress-percent': int(percent),
-                                     'progress-sequence': sequence}
-                    yield self.progress
+                if progress_report is None:
+                    continue
+                percent_str, message = progress_report
+                if not percent_str or not percent_str.isdigit():
+                    logging.info('Skipping "{}" as the percent entry is not an int'.format(progress_report))
+                    continue
+                progress_percent_int = int(percent_str, 10)
```
shamsimam: The `10` is not strictly necessary, just being explicit here about the base.

PR #597, opened 2017-11-27
User resource usage endpoint

DaoWen:


dposada:
@DaoWen I merged from master and resolved a conflict at the end of `test_basic.py` due to a test that I had added at the end colliding with your new tests here. Sorry for the close and re-open, it's the only way I know of to make GitHub realize that the conflict has been resolved when resolving through their UI.

dposada:
Can we add documentation to the API doc for this endpoint?

DaoWen:
> Can we add documentation to the API doc for this endpoint?

I think we should scrap our current API document in favor of an auto-generated document from our swagger annotations—but that would require beefing up all of our swagger annotations, which is going to take some work. I'll update the asciidoc file for now.

DaoWen:
Sample basic output:
```json
{
  "total_usage": {
    "cpus": 10,
    "mem": 2560,
    "gpus": 0,
    "jobs": 10
  }
}
```
Sample breakdown output:
```json
{
  "total_usage": {
    "cpus": 2,
    "mem": 512,
    "gpus": 0,
    "jobs": 2
  },
  "grouped": [
    {
      "group": {
        "uuid": "d250d3ed-f397-47b6-9a71-173871e81c63",
        "name": "cookgroup",
        "running_jobs": [
          "ba9b69a7-5fce-4047-a484-167290134322"
        ]
      },
      "usage": {
        "cpus": 1,
        "mem": 256,
        "gpus": 0,
        "jobs": 1
      }
    }
  ],
  "ungrouped": {
    "running_jobs": [
      "81a23761-25c8-4714-a72f-3e0057ecbab6"
    ],
    "usage": {
      "cpus": 1,
      "mem": 256,
      "gpus": 0,
      "jobs": 1
    }
  }
}
```

DaoWen:
@dposada - I added some basic documentation to the current REST API adoc file: https://github.com/DaoWen/Cook/blob/d9811198054638c3760551c824c968175f982904/scheduler/docs/scheduler-rest-api.adoc#user-resource-usage

Several of the other endpoint descriptions in that file are either out of date or otherwise inaccurate. I want to revisit the swagger annotations sometime soon—but we'll see if I have time. :persevere:

scheduler/src/cook/mesos/util.clj
```diff
@@ -375,16 +389,32 @@
 (timers/deftimer [cook-mesos scheduler get-running-tasks-duration])
 
 (defn get-running-task-ents
-  "Returns all running task entities"
+  "Returns all running task entities."
   [db]
   (timers/time!
     get-running-tasks-duration
-    (->> (q '[:find ?i
+    (->> (q '[:find [?i ...]
```
dposada: Why was this change needed?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1679,6 +1679,65 @@
...
+
+(def zero-usage
+  "Resource usage map 'zero' value"
+  (util/total-resources-of-jobs nil))
+
+(defn get-user-usage
+  "Query a user's current resource usage based on running jobs."
+  [db ctx]
+  (let [user (get-in ctx [:request :query-params :user])
+        running-jobs (mapv :job/_instance (util/get-user-running-task-ents db user))
```
dposada: Does it make sense to add support for waiting instead of running jobs?

scheduler/src/cook/mesos/util.clj
DaoWen: I think this is what we meant to have originally. This lets us avoid unpacking the single-element tuples that were being returned by the original query.
http://docs.datomic.com/query.html#find-specifications

scheduler/src/cook/mesos/util.clj
```diff
@@ -375,16 +389,32 @@
...
               :in $ [?status ...]
               :where
               [?i :instance/status ?status]]
             db [:instance.status/running :instance.status/unknown])
-         (map (fn [[x]] (d/entity db x))))))
+         (map (partial d/entity db)))))
```
DaoWen: (In response to your question below:)
The `find ?i` → `:find [?i ...]` change bought us this cleaner mapping operation here.

scheduler/docs/scheduler-rest-api.adoc
```diff
@@ -515,6 +515,42 @@ To set a user's quota back to the default, an operator can retract the quota the
...
+|Code | HTTP Meaning | Possible reason
+|200 | OK | Returns the user's current resource usage (JSON).
+|400 | Malformed | Returned if a user is not specified
+|403 | Forbidden | This is returned if the user issuing the request is not authorized
+|====
+
+.Example Usage
+[source,bash]
+----
+$ curl -u: --negotiate "$cook_uri/usage?user=$user"
```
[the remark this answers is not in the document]
DaoWen: Again, copy/pasted from query (so yes).

scheduler/docs/scheduler-rest-api.adoc
```diff
@@ -515,6 +515,42 @@ To set a user's quota back to the default, an operator can retract the quota the
...
+
+`GET /usage?user=:user`
+
+.Possible response codes
+[options="header"]
+|====
+|Code | HTTP Meaning | Possible reason
+|200 | OK | Returns the user's current resource usage (JSON).
+|400 | Malformed | Returned if a user is not specified
+|403 | Forbidden | This is returned if the user issuing the request is not authorized
```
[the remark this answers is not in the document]
DaoWen: Copy/pasted from query. I really want to just scrap this document in favor of better swagger.

integration/tests/cook/test_basic.py
```diff
@@ -1265,3 +1265,105 @@ def test_retrieve_jobs_with_deprecated_api(self):
...
+            util.wait_for_jobs(self.cook_url, job_uuids, 'running')
+            resp = util.user_current_usage(self.cook_url, user=user, group_breakdown='true')
+            self.assertEqual(resp.status_code, 200, resp.content)
+            usage_data = resp.json()
+            # Check that the response structure looks as expected
+            self.assertEqual(set(usage_data.keys()), {'total_usage', 'grouped', 'ungrouped'}, usage_data)
+            self.assertEqual(set(usage_data['ungrouped'].keys()), {'running_jobs', 'usage'}, usage_data)
+            my_group_usage = next(x for x in usage_data['grouped'] if x['group']['uuid'] == group_uuid)
+            self.assertEqual(set(my_group_usage.keys()), {'group', 'usage'}, my_group_usage)
+            # Our jobs should be in the breakdown for our job group
```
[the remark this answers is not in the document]
DaoWen: Yeah, I should either do that or compare the whole list.

scheduler/src/cook/mesos/api.clj
DaoWen: Do you want to open a ticket for that? I can't think of a compelling use case, but you know our users better than I do.

dposada: I can't either, so let's wait and see if anyone asks for it

scheduler/src/cook/mesos/util.clj
```diff
@@ -375,16 +389,32 @@
...
+
+(timers/deftimer [cook-mesos scheduler get-user-running-tasks-duration])
+
+(defn get-user-running-task-ents
+  "Returns all running task entities for a specific user."
+  [db user]
+  (timers/time!
+    get-user-running-tasks-duration
```
dposada: Have we confirmed that this timer works and has reasonable values?

DaoWen: Not yet.

DaoWen:
@pschorf - If you approve, I think this is ready to merge.

scheduler/src/cook/mesos/util.clj
```diff
@@ -375,16 +389,32 @@
...
+    (->> (q '[:find [?i ...]
+              :in $ [?status ...] ?user
+              :where
+              [?j :job/user ?user]
+              [?j :job/instance ?i]
+              [?i :instance/status ?status]]
```
pschorf: Can this be based off of `:job/status` instead of `:instance/status`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -375,16 +389,32 @@
...
+
+(defn get-user-running-task-ents
+  "Returns all running task entities for a specific user."
+  [db user]
+  (timers/time!
+    get-user-running-tasks-duration
+    (->> (q '[:find [?i ...]
+              :in $ [?status ...] ?user
+              :where
+              [?j :job/user ?user]
```
pschorf: I think this query will be more performant if we put the `:job/status` condition before the `:job/user` condition. Some users will have a couple of orders of magnitude more completed jobs than currently running jobs. In any case it's worth testing the query out first.

scheduler/docs/scheduler-rest-api.adoc
```diff
@@ -515,6 +515,42 @@ To set a user's quota back to the default, an operator can retract the quota the
 $ curl -X DELETE -u: --negotiate "$cook_uri/quota?user=$user"
 ----
 
+==== User resource usage
+
+Each user's currently-running jobs have a resource footprint.
```
pschorf: The dash in currently-running seems weird to me.

scheduler/docs/scheduler-rest-api.adoc
```diff
@@ -515,6 +515,42 @@ To set a user's quota back to the default, an operator can retract the quota the
...
+===== Get user's current resource usage
+
+`GET /usage?user=:user`
+
+.Possible response codes
+[options="header"]
+|====
+|Code | HTTP Meaning | Possible reason
+|200 | OK | Returns the user's current resource usage (JSON)
+|400 | Malformed | Returned if a user is not specified
```
pschorf: In the liberator resource, we have docs for codes (200, 401, 403) as opposed to (200, 400, 403) here. These should be consistent

DaoWen: But it's consistent with the same wrong error codes being used for the `/quota` endpoint docs!
Like I said, we should switch to just using swagger. (But I fixed these codes here anyway.)

scheduler/src/cook/mesos/api.clj
```diff
@@ -1679,6 +1679,65 @@
...
+      ; basic user usage response
+      {:total-usage (util/total-resources-of-jobs running-jobs)}
+      ; (optional) user's total usage with breakdown by job groups
+      (when with-group-breakdown?
+        (let [breakdowns (->> running-jobs
+                              (group-by util/job-ent->group-uuid)
+                              (map-vals (juxt #(mapv :job/uuid %)
+                                              util/total-resources-of-jobs
+                                              #(-> % first :group/_job first))))]
+          {:grouped (vec (for [[guuid [job-uuids usage group]] breakdowns
```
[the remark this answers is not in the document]
DaoWen: I agree that the `:when` here is a little weird, but I think the destructuring is a win.

I'll try removing the `vec`. I wasn't sure if the schema was OK with any sequence, or if specifically wanted a vector.

scheduler/src/cook/mesos/util.clj
DaoWen: Using `:job/state` here actually simplified a lot of other stuff too. Thanks for the suggestion!

DaoWen:
@pschorf - It's green again.

PR #598, opened 2017-11-28
sorts instances by start_time (field expected to be always present on…

shamsimam:
… an instance) in test_no_cook_executor_on_subsequent_instances

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -54,7 +54,7 @@ class CookTest(unittest.TestCase):
             pass
         job = util.load_job(self.cook_url, job_uuid)
         message = json.dumps(job, sort_keys=True)
-        job_instances = sorted(job['instances'], key=lambda instance: instance['end_time'])
+        job_instances = sorted(job['instances'], key=lambda instance: instance['start_time'])
         self.assertTrue(len(job_instances) >= 2, message)  # sort job instances
         for i in range(1, len(job_instances)):
             message = 'Index ' + str(i) + json.dumps(job_instances[i], sort_keys=True)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -61,7 +61,7 @@ class CookTest(unittest.TestCase):
                 self.assertEqual('failed', job_instance['status'], message)
                 self.assertEqual('Command exited non-zero', job_instance['reason_string'], message)
             else:
-                self.assertEqual('unknown', job_instance['status'], message)
+                self.assertIn(job_instance['status'], ['running', 'unknown'], message)
             self.assertEqual('mesos', job_instance['executor'], message)
 
     def test_disable_mea_culpa(self):
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2,6 +2,7 @@ from collections import Counter
 import json
 import logging
 import math
+import operator
 import subprocess
 import time
 import unittest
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -54,7 +54,7 @@ def test_no_cook_executor_on_subsequent_instances(self):
             pass
         job = util.load_job(self.cook_url, job_uuid)
         message = json.dumps(job, sort_keys=True)
-        job_instances = sorted(job['instances'], key=lambda instance: instance['end_time'])
```
pschorf: Could you figure out why `end_time` was missing? If the job was completed I'd expect it to be on the instance.

integration/tests/cook/test_basic.py
```diff
@@ -54,7 +54,7 @@ def test_no_cook_executor_on_subsequent_instances(self):
             pass
```
DaoWen: The `self.assertEqual` will be caught and ignored by this try/accept block if they fail.

integration/tests/cook/test_basic.py
shamsimam: As we discussed offline, the job wasn't completed.

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
-            job = util.wait_for_job(self.cook_url, job_uuid, 'completed', max_wait_ms=60000)
-            message = json.dumps(job, sort_keys=True)
-            self.assertEqual('failed', job['state'], message)
-            self.assertEqual('completed', job['status'], message)
-        except:
-            pass
+        util.wait_until(lambda: util.load_job(self.cook_url, job_uuid, assert_response=False),
+                        lambda job: 'instances' in job and len(job['instances']) > 4, # try and get at least 5 instances
+                        max_wait_ms=60000)
         job = util.load_job(self.cook_url, job_uuid)
```
pschorf: You can just set `job = util.wait_until(...)` instead of calling `load_job` again

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
         message = json.dumps(job, sort_keys=True)
-        job_instances = sorted(job['instances'], key=lambda instance: instance['end_time'])
-        self.assertTrue(len(job_instances) >= 2, message)  # sort job instances
+        job_instances = sorted(job['instances'], key=lambda instance: instance['start_time'])
+        self.assertTrue(len(job_instances) >= 2, message) # happy with at least 2 in case the scheduler is slow
```
pschorf: You can remove this assertion with the `wait_until` call. It will throw if there are fewer than 5 instances.

shamsimam: The intent is to try and run it to quite a few (5?) instances. But the assertion only needs more than 1 instance.

integration/tests/cook/test_basic.py
shamsimam: O do not want to throw an exception if the predicate fails, the intent is for it to run and generate as many instances as possible with 5 being deemed enough. But only 2 are needed for the assertions.

DaoWen: > I do not want to throw an exception if the predicate fails

In that case, don't we need a try/catch around the `wait_until` call?

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
-        except:
-            pass
+        util.wait_until(lambda: util.load_job(self.cook_url, job_uuid, assert_response=False),
+                        lambda job: 'instances' in job and len(job['instances']) > 4, # try and get at least 5 instances
+                        max_wait_ms=60000)
         job = util.load_job(self.cook_url, job_uuid)
         message = json.dumps(job, sort_keys=True)
-        job_instances = sorted(job['instances'], key=lambda instance: instance['end_time'])
-        self.assertTrue(len(job_instances) >= 2, message)  # sort job instances
+        job_instances = sorted(job['instances'], key=lambda instance: instance['start_time'])
```
DaoWen: Nitpick: Using [*itemgetter*](https://docs.python.org/3.6/library/operator.html#operator.itemgetter) would be more idiomatic:
```python
key=operator.itemgetter('start_time')
```

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
         self.assertEqual(resp.status_code, 201, msg=resp.content)
-        try:
-            job = util.wait_for_job(self.cook_url, job_uuid, 'completed', max_wait_ms=60000)
-            message = json.dumps(job, sort_keys=True)
-            self.assertEqual('failed', job['state'], message)
-            self.assertEqual('completed', job['status'], message)
-        except:
-            pass
+        util.wait_until(lambda: util.load_job(self.cook_url, job_uuid, assert_response=False),
+                        lambda job: 'instances' in job and len(job['instances']) > 4, # try and get at least 5 instances
```
DaoWen: I'm pretty sure that a job is never without an `instances` field in the response here. Furthermore, even if it is missing, the KeyError is just going to trigger another retry.

If you still want the predicate to be "safe", I think using `get` with a default is simpler than the `in` check:
```python
lambda job: len(job.get('instances', [])) > 4
```

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
+        job_uuid, resp = util.submit_job(self.cook_url, command='exit 1', max_retries=10) # should launch many instances
         self.assertEqual(resp.status_code, 201, msg=resp.content)
-        try:
-            job = util.wait_for_job(self.cook_url, job_uuid, 'completed', max_wait_ms=60000)
-            message = json.dumps(job, sort_keys=True)
-            self.assertEqual('failed', job['state'], message)
-            self.assertEqual('completed', job['status'], message)
-        except:
-            pass
+        util.wait_until(lambda: util.load_job(self.cook_url, job_uuid, assert_response=False),
```
DaoWen: Why explicitly set `assert_response=False`? If the response code is an error code and the assertion fails, that will just trigger another retry.

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
-        job_instances = sorted(job['instances'], key=lambda instance: instance['end_time'])
-        self.assertTrue(len(job_instances) >= 2, message)  # sort job instances
+        job_instances = sorted(job['instances'], key=lambda instance: instance['start_time'])
+        self.assertTrue(len(job_instances) >= 2, message) # happy with at least 2 in case the scheduler is slow
         for i in range(1, len(job_instances)):
-            message = 'Index ' + str(i) + json.dumps(job_instances[i], sort_keys=True)
-            self.assertEqual('failed', job_instances[i]['status'], message)
-            self.assertEqual('Command exited non-zero', job_instances[i]['reason_string'], message)
-            self.assertEqual('mesos', job_instances[i]['executor'], message)
+            job_instance = job_instances[i]
```
DaoWen: Why iterate over indices instead of instances?
```python
for job_instance in job_instances[1:]:
```

integration/tests/cook/test_basic.py
```diff
@@ -43,24 +43,24 @@ def test_basic_submit(self):
...
+            message = 'Index ' + str(i) + json.dumps(job_instance, sort_keys=True)
+            if 'reason_string' in job_instance:
+                self.assertEqual('failed', job_instances[i]['status'], message)
+                self.assertEqual('Command exited non-zero', job_instance['reason_string'], message)
+            else:
+                self.assertEqual('unknown', job_instances[i]['status'], message)
```
DaoWen: For consistency, I think you should replace `job_instances[i]` with `job_instance` in these assertions.

PR #601, opened 2017-11-28
adds interactive output support by adding flush timer task

shamsimam:


executor/cook/io_helper.py
```diff
@@ -91,31 +91,49 @@ def process_output(label, out_file, max_bytes_read_per_line, out_fn, flush_fn):
...
     logging.info('Starting to pipe {}'.format(label))
+
+    io_lock = Lock()
+    lines_buffered_event = Event()
+
+    def trigger_flush():
+        if lines_buffered_event.is_set():
+            with io_lock:
+                logging.info('Flushing contents {}'.format(label))
+                flush_fn()
```
pschorf: Is it possible for `flush_fn` to throw an error? If so we should handle it so the thread doesn't die.

shamsimam:
@DaoWen the PR now uses non-blocking reads as per your snippet :)

executor/cook/io_helper.py
```diff
@@ -73,49 +78,80 @@ def print_err(string_data, flush=False, newline=True):
...
+
+    bytes_in_buffer = 0
     try:
         while True:
-            line = out_file.readline(max_bytes_read_per_line)
-            if not line:
+            data = nb_read(out_fd)
+            reads_since_flush += 1
+            if data is not None:
+                logging.info('Read: {} [label={}]'.format(data, label))
```
pschorf: Do we want to log every read at info?

executor/cook/io_helper.py
```diff
@@ -73,49 +78,80 @@ def print_err(string_data, flush=False, newline=True):
...
+                out_fn(data, newline=False)
+                bytes_in_buffer += len(data)
+            if ((reads_since_flush >= min_reads_before_flush and bytes_in_buffer > 0) or
+                        bytes_in_buffer > max_read_nb_chunk_size):
```
pschorf: Shouldn't the size-based buffering be handled by the output stream?

executor/cook/io_helper.py
```diff
@@ -91,31 +91,58 @@ def process_output(label, out_file, max_bytes_read_per_line, out_fn, flush_fn):
...
     logging.info('Starting to pipe {}'.format(label))
+
+    io_lock = Lock()
+    lines_buffered_event = Event()
+
+    def safe_flush():
+        try:
+            flush_fn()
+        except:
+            logging.exception('Error while flushing contents {}'.format(label))
```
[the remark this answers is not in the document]
shamsimam: `logging.exception` logs the exception.

executor/cook/io_helper.py
shamsimam: Reverted the commit, so not relevant anymore.

executor/cook/io_helper.py
shamsimam: Reverted the commit, so not relevant anymore.

shamsimam:
@pschorf @DaoWen ready for another round of review

PR #602, opened 2017-11-28
handles binary data in stdout and stderr

shamsimam:


executor/cook/io_helper.py
```diff
@@ -12,13 +12,15 @@
...
+def print_out(data, bytes=False, flush=False, newline=True):
     """Wrapper function that prints to stdout in a thread-safe manner using the __stdout_lock__ lock.
 
     Parameters
     ----------
-    string_data: string
-        The string to output
+    data: string or bytes
+        The data to output
+    bytes: boolean
```
pschorf: When you would pass `True` for `bytes`, is `data` a `bytes` object? If so, can we remove the parameter and just check the input type? This seems pretty error-prone.

DaoWen: Alternatively, is there any downside to *always* writing to `stdout.buffer`?

shamsimam: Done:
- directly checking the input type for string or not (bytes)
- always always writing to stdout|stderr.buffer

shamsimam:
Let's merge this after https://github.com/twosigma/Cook/pull/601 which also does binary reads

shamsimam:
ready for another round of review.

executor/tests/test_executor.py
```diff
@@ -15,11 +15,19 @@
...
+    ensure_directory, get_random_task_id, parse_message, redirect_stderr_to_file, redirect_stdout_to_file, wait_for, \
     FakeExecutorConfig, FakeMesosExecutorDriver
 
 
+def sleep_and_set_stop_signal(stop_signal, wait_seconds):
+    """Busy waits for wait_seconds seconds before setting stop_signal."""
+    for _ in range(wait_seconds):
+        if not stop_signal.isSet():
+            time.sleep(1)
+    stop_signal.set()
```
DaoWen:
```python
for _ in range(wait_seconds):
    if stop_signal.isSet():
        break # don't keep looping if it's already set
    time.sleep(1)
else: # don't bother setting if we broke out of the loop (because it's already set)
    stop_signal.set()
```

executor/tests/test_executor.py
```diff
@@ -306,12 +314,7 @@ def test_await_process_completion_killed(self):
             shutdown_grace_period_ms = 2000
 
             stop_signal = Event()
-
-            def sleep_and_set_stop_signal():
-                time.sleep(2 * cook.RUNNING_POLL_INTERVAL_SECS)
-                stop_signal.set()
-            thread = Thread(target=sleep_and_set_stop_signal, args=())
-            thread.start()
+            Thread(target=sleep_and_set_stop_signal, args=(stop_signal, 2 * cook.RUNNING_POLL_INTERVAL_SECS)).start()
```
DaoWen: Is there any reason not to use a `Timer` here? You should be able to set it as a daemon thread to avoid holding up the whole process if it wants to exit before this timer goes off.

executor/tests/test_executor.py
DaoWen: Actually, this whole looping/sleeping thing might go away if you use `Timer` instead (see comment below).

executor/tests/test_executor.py
```diff
@@ -630,12 +613,7 @@ def assertions(driver, task_id, sandbox_directory):
             assert_message(self, expected_message_1, actual_encoded_message_1)
 
         stop_signal = Event()
-
-        def sleep_and_set_stop_signal():
-            time.sleep(2 * cook.RUNNING_POLL_INTERVAL_SECS)
-            stop_signal.set()
-        thread = Thread(target=sleep_and_set_stop_signal, args=())
-        thread.start()
+        Thread(target=sleep_and_set_stop_signal, args=(stop_signal, 2 * cook.RUNNING_POLL_INTERVAL_SECS)).start()
```
DaoWen: Timer?

You could actually move the creation of the Timer into a helper method so that the implementation doesn't leak through at all of these call-sites. (You'd want to use `async` in the name to make it clear that it doesn't block the current thread.)

executor/tests/test_executor.py
```diff
@@ -661,24 +639,48 @@ def assertions(driver, task_id, sandbox_directory):
...
                 with open(stdout_name) as f:
                     file_contents = f.read()
-                    expected_string = 'Hello' * 1000000
-                    self.assertTrue(expected_string in file_contents)
+                    self.assertEqual(1000000, file_contents.count('X'))
             else:
                 self.fail('{} does not exist.'.format(stdout_name))
 
         stop_signal = Event()
+        Thread(target=sleep_and_set_stop_signal, args=(stop_signal, 60)).start()
```
DaoWen: Timer?

executor/tests/test_executor.py
```diff
@@ -661,24 +639,48 @@ def assertions(driver, task_id, sandbox_directory):
...
 
-        thread = Thread(target=sleep_and_set_stop_signal, args=())
-        thread.start()
+            stdout_name = ensure_directory('build/stdout.' + str(task_id))
+            if not os.path.isfile(stdout_name):
+                self.fail('{} does not exist.'.format(stdout_name))
 
-        command = 'for i in `seq 1000000`; do printf "Hello"; done; echo "World"'
+        stop_signal = Event()
+        Thread(target=sleep_and_set_stop_signal, args=(stop_signal, 60)).start()
```
DaoWen: Timer?

executor/tests/test_executor.py
```diff
@@ -714,34 +716,15 @@ def assertions(driver, task_id, sandbox_directory):
...
-        def sleep_and_set_stop_signal():
-            # wait upto 1 minute for the task to complete
-            for _ in range(60):
-                if not stop_signal.isSet():
-                    time.sleep(1)
-            stop_signal.set()
-
-        thread = Thread(target=sleep_and_set_stop_signal, args=())
-        thread.start()
+        Thread(target=sleep_and_set_stop_signal, args=(stop_signal, 60)).start()
```
DaoWen: Timer? (helper?)

executor/cook/io_helper.py
```diff
@@ -73,12 +77,16 @@ def print_err(string_data, flush=False, newline=True):
...
+        if isinstance(data, str):
+            stderr_buffer.write(data.encode())
+        else:
+            stderr_buffer.write(data)
         if newline:
-            sys.stderr.write(os.linesep)
+            stderr_buffer.write(os.linesep)
         if flush:
-            sys.stderr.flush()
+            stderr_buffer.flush()
```
DaoWen: I still don't like the fact that this is repeated (once for `stdout` and once for `stderr`), especially since they're not even adjacent. Even though they're pretty simple, this still makes it harder to maintain since the code is in two places. I think the fact that the `if newline` branches don't match right now is a prime example of this issue.

executor/tests/test_executor.py
shamsimam: Created healper for thread variant as it is simpler than `Timer` variant.

```def sleep_and_set_stop_signal_task(stop_signal, wait_seconds):
    """Busy waits for wait_seconds seconds before setting stop_signal."""
    def sleep_and_set_stop_signal_fn():
        for _ in range(4 * wait_seconds):
            if not stop_signal.isSet():
                time.sleep(0.25)
        stop_signal.set()

    thread = Thread(target=sleep_and_set_stop_signal_fn, args=())
    thread.daemon = True
    thread.start()


def sleep_and_set_stop_signal_timer(stop_signal, wait_seconds):
    """Busy waits for wait_seconds seconds before setting stop_signal."""
    limit = 4 * wait_seconds

    def sleep_and_set_stop_signal_fn(iteration):
        if iteration < limit:
            if not stop_signal.isSet():
                Timer(0.25, sleep_and_set_stop_signal_fn, args=(iteration + 1,)).start()
        else:
            stop_signal.set()

    Timer(0.25, sleep_and_set_stop_signal_fn, args=(0,)).start()
```

executor/tests/test_executor.py
shamsimam: See comment above.

executor/tests/test_executor.py
shamsimam: Created `sleep_and_set_stop_signal_task`. Please see above.

executor/tests/test_executor.py
shamsimam: Please see comment above.

executor/tests/test_executor.py
shamsimam: Please see comment above.

executor/tests/test_executor.py
shamsimam: Please see comment above.

executor/cook/io_helper.py
shamsimam: Created `print_to_buffer` helper function. I did not understand the `newline` branches not matching comment.

executor/tests/test_executor.py
DaoWen: This is what I was envisioning:
```python
def sleep_and_set_stop_signal_timer(stop_signal, wait_seconds):          
    """Busy waits for wait_seconds seconds before setting stop_signal."""
    t = Timer(4 * wait_seconds, stop_signal.set)                         
    t.daemon = True                                                      
    t.start()                                                            
```

shamsimam: Oops, my bad. Done.

PR #603, opened 2017-11-29
cook executor - checks for maximum length validations for progress messages

shamsimam:
checks for maximum length validations for progress messages

### DIFF
--- a/executor/cook/progress.py
+++ b/executor/cook/progress.py
@@ -44,7 +44,7 @@ class ProgressUpdater(object):
         self.lock = Lock()
 
     def has_enough_time_elapsed_since_last_update(self):
-        """Returns true if enough time (based on poll_interval_ms) has elapsed since 
+        """Returns true if enough time (based on poll_interval_ms) has elapsed since
         the last progress update (available in last_reported_time).
         """
         if self.last_reported_time is None:
@@ -65,7 +65,7 @@ class ProgressUpdater(object):
             The progress data to send.
         force_send: boolean, optional
             Defaults to false.
-            
+
         Returns
         -------
         Nothing 
@@ -87,9 +87,12 @@ class ProgressUpdater(object):
                         message_dict['progress-message'] = new_progress_str
                         progress_message = json.dumps(message_dict)
 
-                    self.send_progress_message(progress_message)
-                    self.last_reported_time = time.time()
-                    self.last_progress_data = progress_data
+                    send_success = self.send_progress_message(progress_message)
+                    if send_success:
+                        self.last_progress_data = progress_data
+                        self.last_reported_time = time.time()
+                    else:
+                        logging.info('Unable to send progress message {}'.format(progress_message))
                 else:
                     logging.debug('Not sending progress data as enough time has not elapsed since last update')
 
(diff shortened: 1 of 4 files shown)


executor/cook/progress.py
```diff
@@ -246,43 +227,59 @@ def __init__(self, task_id, config, stop_signal, task_completed_signal, send_pro
...
+            logging.info('Progress monitoring from did not complete [tag={}]'.format(self.location_tag))
 
     def track_progress(self):
         """Retrieves and sends progress updates using send_progress_update_fn.
+        It throttles the rate at which progress updates are sent.
         It sets the progress_complete_event before returning."""
         try:
+            last_reported_time = None
+
+            def has_enough_time_elapsed_since_last_update():
```
pschorf: Is this just lifting the check from the ProgressWatcher to the caller? What's the rationale for doing that?

executor/cook/progress.py
```diff
@@ -20,78 +20,58 @@ def increment_and_get(self):
...
         with self.lock:
             if progress_data is not None and self.last_progress_data != progress_data:
-                if force_send or self.has_enough_time_elapsed_since_last_update():
-                    logging.info('Sending progress message {}'.format(progress_data))
-                    message_dict = dict(progress_data)
-                    message_dict['task-id'] = self.task_id
+                logging.info('Sending progress message {}'.format(progress_data))
+                message_dict = dict(progress_data)
+                message_dict['task-id'] = self.task_id
+                progress_message = json.dumps(message_dict)
```
pschorf: Why is the maximum length checked on the JSON object, but we truncate the message in the object? It seems like it would make more sense to check the message length on the message itself.

shamsimam:
> It seems like the message length validation was already there in the existing code. Was there an issue with it? 

The fix for the max length was changing the condition to `<=` from `<` in `send_message`
https://github.com/twosigma/Cook/pull/603/files#diff-1706ae0b989558a124ae113ca8b2e2ffR88

executor/cook/progress.py
shamsimam: We chose to interpret the max length to refer to the un-encoded version.

The message itself is encrypted and we can't guarantee lengths of the encrypted version by trimming the message.

executor/cook/progress.py
shamsimam: Yes, this moves the throttle from `ProgressUpdater` to `ProgressTracker` as a refactoring. I will move it to a separate PR if you prefer.

executor/cook/progress.py
```diff
@@ -20,78 +20,58 @@ def increment_and_get(self):
...
-        if self.last_reported_time is None:
-            return True
-        else:
-            current_time = time.time()
-            time_diff_ms = (current_time - self.last_reported_time) * 1000
-            return time_diff_ms >= self.poll_interval_ms
-
-    def send_progress_update(self, progress_data, force_send=False):
+    def send_progress_update(self, progress_data):
         """Sends a progress update if enough time has elapsed since the last progress update.
```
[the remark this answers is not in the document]
shamsimam: Yes, done.

executor/cook/progress.py
shamsimam: Removed the changes from this PR

shamsimam:
**Note:** Limited the scope of this PR to only check for maximum length validations for progress messages

dposada:
RTM?

PR #609, opened 2017-11-30
adds support for floating point numbers as progress percents

shamsimam:


executor/cook/progress.py
```diff
@@ -212,22 +212,22 @@ def retrieve_progress_states(self):
...
+                    if percent_int < 0 or percent_int > 100:
+                        logging.info('Skipping "%s" as the percent is not in [0, 100]', progress_report)
+                        continue
+                    logging.debug('Updating progress to %s percent [tag=%s]', percent_str, self.location_tag)
+                    self.progress = {'progress-message': message.strip(),
+                                     'progress-percent': percent_int,
+                                     'progress-sequence': self.sequence_counter.increment_and_get()}
+                    yield self.progress
+                except:
+                    logging.exception('Skipping "%s" as a progress entry', progress_report)
```
wyegelwel: Please add `exc_info=True` or some other way to know *why* it is being skipped.

wyegelwel: Never mind, I see that logging.exception does that for you

PR #613, opened 2017-11-30
triggers termination of running task when executor is requested to di…

shamsimam:
…sconnect

### DIFF
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -273,7 +273,7 @@ def await_process_completion(stop_signal, process_info, shutdown_grace_period_ms
     while is_running(process_info):
 
         if stop_signal.isSet():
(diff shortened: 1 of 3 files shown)


executor/cook/executor.py
```diff
@@ -456,6 +456,7 @@ def reregistered(self, driver, agent_info):
     def disconnected(self, driver):
         logging.info('Mesos requested executor to disconnect')
         self.disconnect_signal.set()
+        self.stop_signal.set()
```
DaoWen: This looks right to me. Based on [the documentation for executor disconnection](http://mesos.apache.org/documentation/latest/executor-http-api/#disconnections): <i>"If framework checkpointing is disabled, the executor is not supposed to retry subscription and gracefully exit."</i>

PR #614, opened 2017-12-01
ensures single progress updater throttles send of progress messages

shamsimam:


executor/cook/progress.py
```diff
@@ -71,37 +71,42 @@ def send_progress_update(self, progress_data, force_send=False):
...
-                        num_extra_chars = len(progress_message) - self.max_message_length
-                        allowed_progress_message_length = max(len(progress_str) - num_extra_chars - 3, 0)
-                        new_progress_str = progress_str[:allowed_progress_message_length].strip() + '...'
-                        logging.info('Progress message trimmed to {}'.format(new_progress_str))
-                        message_dict['progress-message'] = new_progress_str
-                        progress_message = json.dumps(message_dict)
+            last_progress_data_sent = self.last_progress_data_sent
+            if progress_data is not None and last_progress_data_sent != progress_data:
+                last_progress_sequence = last_progress_data_sent['progress-sequence'] if last_progress_data_sent else 0
+                if last_progress_sequence < progress_data['progress-sequence']:
```
pschorf: What is this helping beyond the `self.has_enough_time_elapsed_since_last_update()` condition? Are the sequence numbers not always increasing? It seems like the equals check covers this already.

shamsimam: There is force send at the end from each of the three (or two) `ProgressTracker`s and the equals check will fail if the sequence numbers do not match. We perform the secondary sequence check to ensure we do not send an outdated progress message.

executor/cook/progress.py
```diff
@@ -71,37 +71,42 @@ def send_progress_update(self, progress_data, force_send=False):
...
+                            self.last_progress_data_sent = progress_data
+                            self.last_reported_time = time.time()
+                        else:
+                            logging.info('Unable to send progress message {}'.format(progress_message))
                     else:
-                        logging.info('Unable to send progress message {}'.format(progress_message))
+                        logging.debug('Not sending progress data as enough time has not elapsed since last update')
                 else:
-                    logging.debug('Not sending progress data as enough time has not elapsed since last update')
+                    logging.info('Skipping outdated progress data {}'.format(progress_data))
```
dposada: Might be nice for readability to invert the if checks so that you can log these messages and return early, instead of having the `else` so far below the corresponding `if`

DaoWen: I think it would be more readable if you invert the logic on these two outer branches so that the current `else` clause comes first. Since these clauses are simple (a single-line log message) and descriptive, it's helpful to see these first. As is, I end up reading over 30 lines in the true branch before hitting this one-liner in the `else` clause, at which point I have to scroll back up to see what the `if` condition was again.

executor/cook/progress.py
pschorf: I think this is worth adding a comment.

executor/cook/progress.py
```diff
@@ -71,37 +71,42 @@ def send_progress_update(self, progress_data, force_send=False):
...
+                            logging.info('Unable to decode progress message in ascii, using empty string instead')
+                            progress_str = ''
+                        message_dict['progress-message'] = progress_str
 
-                    send_success = self.send_progress_message(progress_message)
-                    if send_success:
-                        self.last_progress_data_sent = progress_data
-                        self.last_reported_time = time.time()
+                        progress_message = json.dumps(message_dict)
+                        if len(progress_message) > self.max_message_length:
```
DaoWen: Wouldn't it make more sense to perform this check on the `raw_progress_message` rather than on the stringified JSON response (which includes all the other stuff in the dictionary)? If you check the length of `raw_progress_message`, you can also avoid decoding the whole string if you're just going to truncate it later.

executor/cook/progress.py
shamsimam: Done.

executor/cook/progress.py
shamsimam: The check needs to be on the `progress_message` as that is what is sent as the framework message.

executor/cook/progress.py
shamsimam: Done.

shamsimam: Done.

executor/cook/progress.py
DaoWen: OK, that makes sense. I didn't notice that allowed_progress_message_length was computed based on the length of the whole JSON message.

shamsimam:
RTM: Build is green.

PR #618, opened 2017-12-01
logs memory usage of the executor at fixed intervals

shamsimam:
Example logs:
```
...
2017-12-01 16:42:50,168 INFO Memory usage will be logged every 3600 secs
...
2017-12-01 16:42:50,169 INFO Executor Memory usage: 14612.0 kB
2017-12-01 16:42:50,172 INFO MesosExecutorDriver is starting...
...
2017-12-01 16:42:50,242 INFO Executor Memory usage: 15092.0 kB
2017-12-01 16:42:50,248 INFO Executor exiting with code 1
```

### DIFF
--- a/executor/cook/config.py
+++ b/executor/cook/config.py
@@ -26,7 +26,7 @@ class ExecutorConfig(object):
         """
         try:
             return int(1000 * parse_duration(time_string))
-        except:
+        except Exception:
             logging.exception('Unable to extract seconds from {}'.format(time_string))
             logging.info('Defaulting time to 1 second.')
             return 1000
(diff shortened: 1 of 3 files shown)


executor/cook/config.py
```diff
@@ -35,6 +35,7 @@ def __init__(self,
                  flush_interval_secs=2,
                  max_bytes_read_per_line=1024,
                  max_message_length=512,
+                 memory_usage_interval_secs=15,
```
pschorf: I think 15 seconds is a pretty low default value, what about something like 5 minutes?

shamsimam: The default is set to `3600` in https://github.com/twosigma/Cook/pull/618/files#diff-6a568f1963475d5ea436f8b8c51555caR88

dposada:
RTM?

executor/cook/config.py
shamsimam: @pschorf happy with having the default there?

PR #623, opened 2017-12-06
Handles OSErrors in Cook Executor

shamsimam:

- adds OSError exception handler to handle 'Container Memory Limit Exceeded' scenarios
-- logs memory usage while handling `PSError`
- ensures the executor sends a single terminal status message for a task
- adds support for sending reason messages in the status updates

It is possible for individual Cook Executor threads to terminate with an OSError when memory limits are exceeded on the cgroups. For example, the thread reading files for progress messages can run out of memory and throw an `OSError`. We trigger an exit as soon as we detect this memory limit exceeded behavior (expected behavior for tasks running on Mesos when they run out memory) and continue chugging along.


dposada:
Needs rebase

shamsimam:
Rebased

pschorf:
Needs rebase again :(

shamsimam:
rebased

dposada:
> For example, when the thread reading stdout ends abruptly, this can lead to blocking on the launched subprocess as it cannot write to its stdout (blocked on the stdout thread to read data).

Isn't this no longer true since we're just letting the child process write to the files now?

shamsimam:
@dposada updated the statement. We can get `OSError`s from other locations in code too.

executor/cook/progress.py
```diff
@@ -256,14 +257,19 @@ def __update_progress(self, progress_report):
                          'progress-sequence': self.sequence_counter.increment_and_get()}
         return True
 
-    def retrieve_progress_states(self):
+    def retrieve_progress_states(self, os_error_handler):
```
pschorf: Why not rethrow the OSError and handle it in the caller instead of passing a handler?

executor/cook/progress.py
```diff
@@ -290,7 +298,7 @@ class ProgressTracker(object):
     """Helper class to track progress messages from the specified location."""
 
     def __init__(self, config, stop_signal, task_completed_signal, counter, progress_updater,
-                 progress_termination_signal, location, location_tag):
+                 progress_termination_signal, location, location_tag, os_error_handler):
```
pschorf: Same here

executor/cook/progress.py
shamsimam: Done.

executor/cook/progress.py
shamsimam: This is in the constructor and one of the calls will start a thread where we need to guard the executing function with the handler.

shamsimam:
ready for another round :)

executor/cook/executor.py
```diff
@@ -224,9 +277,14 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
     launched_process = None
     task_id = get_task_id(task)
     cio.print_and_log('Starting task {}'.format(task_id))
+    status_updater = StatusUpdater(driver, task_id)
+
+    def inner_os_error_handler(os_error):
+        os_error_handler(os_error, stop_signal, status_updater)
```
dposada: Nitpick: might be able to accomplish this with `functools.partial`

executor/tests/test_progress.py
```diff
@@ -1,61 +1,58 @@
...
-                         match_progress_update("^^^^JOB-PROGRESS:   1 done"))
-        self.assertEqual(('1', ' One percent complete'),
-                         match_progress_update("^^^^JOB-PROGRESS: 1 One percent complete"))
-        self.assertEqual(('1', '    One percent complete'),
-                         match_progress_update("^^^^JOB-PROGRESS: 1    One percent complete"))
-        self.assertEqual(('50', ' Fifty percent complete'),
-                         match_progress_update("^^^^JOB-PROGRESS: 50 Fifty percent complete"))
+            return progress_watcher.match_progress_update(input_string)
+
+        self.assertIsNone(match_progress_update(b'One percent complete'))
```
dposada: Why did these change to byte strings?

shamsimam: I changed them to match what we expect while executing the code (byte strings from the `tail` function).

executor/cook/executor.py
shamsimam: Done. Needed to change the order of the arguments.

shamsimam:
@pschorf more feedback please :)

PR #624, opened 2017-12-07
Mimic Mesos CommandExecutor kill logic of killing process tree

shamsimam:

- refactors the process management logic to cook/subprocess.py 
- adds support for killing process trees in the executor (uses [psutil](https://github.com/giampaolo/psutil)) using breadth-first ordering and `SIGSTOP` signals before retrieving child processes 

We want to align our kill logic with the Mesos [`CommandExecutor`](https://github.com/apache/mesos/blob/master/src/launcher/executor.cpp#L856)

### DIFF
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -1,17 +1,16 @@
-#!/usr/bin/env python3
-
 import json
 import logging
-import os
 import signal
-import subprocess
 import time
 from threading import Event, Thread, Timer
+
+import os
+import pymesos as pm
 
 import cook
 import cook.io_helper as cio
 import cook.progress as cp
-import pymesos as pm
+import cook.subprocess as cs
 
 
 def get_task_id(task):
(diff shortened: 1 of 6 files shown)


executor/cook/executor.py
```diff
@@ -444,7 +333,8 @@ def launch_progress_tracker(progress_location, location_tag):
     finally:
         # ensure completed_signal is set so driver can stop
         completed_signal.set()
-        kill_process_group(group_id, signal.SIGKILL)
+        if launched_process is not None and cs.is_process_running(launched_process):
```
dposada: Could `launched_process is not None` be replaced with just `launched_process`?

executor/cook/subprocess.py
```diff
@@ -0,0 +1,216 @@
...
+
+        _send_signal_to_process(process_id, signal_to_send)
+
+        signal_sent_to_all_children_successfully = True
+        if children:
+            for child in children:
+                logging.info('Found child process (id: {}) of process (id: {})'.format(child.pid, process_id))
+                if not _send_signal_to_process(child.pid, signal_to_send):
+                    logging.info('Failed to send signal {} to child process (id: {})'.format(signal_name, child.pid))
+                    signal_sent_to_all_children_successfully = False
```
dposada: Did we look into whether this does the kills in the same order as the Mesos command executor?

executor/cook/executor.py
shamsimam: Yes, will do

executor/cook/subprocess.py
shamsimam: I haven't yet.

shamsimam: I was previously using depth-first order whereas [stout](https://github.com/deric/libprocess/blob/master/3rdparty/stout/include/stout/os/killtree.hpp#L123) (the library Mesos uses) used breadth-first order. I have changed the code to match the order and also use the `SIGSTOP` and `SIGCONT` signals.

PR #625, opened 2017-12-07
Improve stdout and stderr processing speed

shamsimam:
- changes call from [`readline()`](https://docs.python.org/3/library/io.html?highlight=bufferedreader#io.IOBase.readline) to [`read1()`](https://docs.python.org/3/library/io.html?highlight=bufferedreader#io.BufferedReader.read1) to improve speed of output syncing

The stdout and stderr syncer is much slower than the Mesos `CommandExecutor` version. We wish to lower the performance gap by optimizing the stdout/stderr writers.

## Additional notes

- I did not use [`readinto`](https://docs.python.org/3/library/io.html?highlight=bufferedreader#io.BufferedIOBase.readinto) or `readinto1` suggested in [this article](http://rabexc.org/posts/io-performance-in-python) as I would then need to manually create the byte array to send to write.

- The command `python3 -c '[print("Hello:",i) for i in range(10000000)]'` runtime is down to 240 secs from 840 secs. For comparison, the Mesos command executor runs in 120 secs.
- pschorf has a C programs that writes about 2GB of data. It runs in 20 seconds on the Mesos command executor but takes around 3 minutes on the Cook Executor (used to take 9 minutes before this change).

Command:
`python3 -c '[print("Hello:",i) for i in range(100000000)]'`

=== Job: fe785ef0-f8c1-41a3-adac-deb32c7fb487 (shamsimam_job) ===
Executor     **mesos**                                    
Job Instance                          Run Time                                          Host         Instance Status
4e7f890f-3ccc-423c-9cb3-1875d184144f  **1 minute and 54 seconds** (started 26 minutes ago)  localhost-3  Success

=== Job: a11173f1-d302-444c-a26c-d604bb67cfb1 (shamsimam_job) ===
Executor     **cook (with readline version)**
Job Instance                          Run Time                                            Host         Instance Status
671aaafc-b54a-4d04-91f8-468fa48591b3  **13 minutes and 57 seconds** (started 15 minutes ago)  localhost-3  Success

=== Job: fae0f771-b6e9-4d8b-b479-500a05223479 (shamsimam_job) ===
Executor     **cook (with read1 version)**
Job Instance                          Run Time                                           Host         Instance Status
40c8e2f4-d0d7-4736-a2bc-d4f7fb2d2f27  **3 minutes and 58 seconds** (started 22 minutes ago)  localhost-3  Success

### DIFF
--- a/executor/cook/io_helper.py
+++ b/executor/cook/io_helper.py
@@ -102,15 +102,15 @@ def print_err(data, flush=False, newline=True):
     print_to_buffer(__stderr_lock__, sys.stderr.buffer, data, flush=flush, newline=newline)
 
 
-def process_output(label, out_file, out_fn, flush_fn, flush_interval_secs, max_bytes_read_per_line):
-    """Processes output piped from the out_file and prints it using the out_fn function.
+def process_output(label, out_buffer, out_fn, flush_fn, flush_interval_secs, max_bytes_per_read):
+    """Processes output piped from the out_buffer and prints it using the out_fn function.
     When done reading the file, calls flush_fn to flush the output.
 
     Parameters
     ----------
     label: string
         The string to associate in outputs
-    out_file: a file object
+    out_buffer: a file object
         Provides that output from the child process
     out_fn: function(string)
         Function to output a string
@@ -118,8 +118,8 @@ def process_output(label, out_file, out_fn, flush_fn, flush_interval_secs, max_b
         Function that flushes the output.
     flush_interval_secs: number
         The number of seconds to wait between flushing buffered data
-    max_bytes_read_per_line: int
-        The maximum number of bytes to read per call to readline().
+    max_bytes_per_read: int
+        The maximum number of bytes to read per call to read().
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


executor/cook/io_helper.py
```diff
@@ -149,7 +149,7 @@ def trigger_flush_daemon():
         logging.info('Starting to pipe {}'.format(label))
         trigger_flush_daemon()
         while True:
-            line = out_file.readline(max_bytes_read_per_line)
+            line = out_buffer.read1(max_bytes_per_read)
```
DaoWen: :+1: Based on the documentation you linked to, using `read1` definitely sounds like an improvement to me.

dposada:
RTM?

DaoWen:
I think this is good to merge as-is, but if we're still concerned about the overheads in the future, we can revisit the idea of using [`readinto1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.readinto1) with [memory views](https://docs.python.org/dev/library/stdtypes.html#memory-views) over the byte array (for writing just the portion of the byte-buffer that has new data).

PR #628, opened 2017-12-08
Removes needs for stdout and stderr tracker threads

shamsimam:

- removes helper methods and threads for stdout/stderr tracking
- uses stdout=sys.stdout and stderr=sys.stderr for subprocess arguments
-- this helps with testability as otherwise the subprocess ends up writing on the console due to stdout/stderr file descriptors
-- I have manually tested that using the file handles does not slow down the stdout syncing when compared to the mesos command executor
- launches a separate thread to perform task killing on stop_signal
- uses timed [`wait()`](https://docs.python.org/3.5/library/subprocess.html#subprocess.Popen.wait) to await process termination on SIGTERM
- uses `wait()` for process termination either normally or via SIGKILL

We want to relieve ourselves from having to deal with buffering and memory issues of piping stdout and stderr from the subprocess.

## Example

Command: `python3 -c '[print("loremipsumstringfoobar:",i) for i in range(1000000000)]'`

Output on stdout:
```
Cook Executor version 0.1.6

Starting task 19646813-9dc0-4baa-8431-303f78a59465

Forked command at 41484
loremipsumstringfoobar: 0
loremipsumstringfoobar: 1
loremipsumstringfoobar: 2
...
loremipsumstringfoobar: 44653459
loremipsumstringfoobar: 44653460
Received kill for task 19646813-9dc0-4baa-8431-303f78a59465 with grace period of 5secs

Command exited with status -15 (pid: 41484)

Executor completed execution of 19646813-9dc0-4baa-8431-303f78a59465 (state=TASK_KILLED)
```

### DIFF
--- a/executor/cook/__init__.py
+++ b/executor/cook/__init__.py
@@ -7,7 +7,6 @@ For more information on Mesos executors, see the "Working with Executors"
 section at http://mesos.apache.org/documentation/latest/app-framework-development-guide/
 """
 
-RUNNING_POLL_INTERVAL_SECS = 1
 DAEMON_GRACE_SECS = 1
 TERMINATE_GRACE_SECS = 0.1
 
--- a/executor/cook/__main__.py
+++ b/executor/cook/__main__.py
@@ -39,7 +39,7 @@ def main(args=None):
         print(__version__)
         sys.exit(0)
 
-    cio.print_out('Cook Executor version {}'.format(__version__))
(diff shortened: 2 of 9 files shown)


wyegelwel:
Did this end up working?

shamsimam:
@wyegelwel 

> Did this end up working?

It did! :)

PR #631, opened 2017-12-11
Fix auth check on group endpoint

DaoWen:

- Fix auth-check on group operations
- Simplify data-flow of context data by caching it in the `malformed?` handler
- Filter out unauthorized groups (in addition to missing groups) when using `partial` flag

The authorization check for group operations was broken, meaning anyone could kill anyone else's groups.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1325,11 +1325,11 @@
 (defn retrieve-groups
   "Returns a tuple that either has the shape:
 
-    [true {::guuids ...}]
+    [false {::guuids ...}]
 
   or:
 
-    [false {::error ...}]
+    [true {::error ...}]
 
   Given a collection of group uuids, attempts to return the corresponding
   set of existing group uuids. By default (or if the 'partial' query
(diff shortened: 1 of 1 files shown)


DaoWen:
@dposada - I tested this locally, but I didn't add any new integration tests. I'm looking into adding some tests using basic-auth to our integration test suite, but that will come in a separate PR.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1346,22 +1346,18 @@
...
-        [true {::guuids existing-groups}]
-        [false {::error (str "UUID "
-                             (str/join
-                               \space
-                               (set/difference (set requested-guuids) (set existing-groups)))
-                             " didn't correspond to a group")}]))
+          {existing-groups true missing-groups false} (group-by exists? requested-guuids)]
+      [false {::allow-partial-results? allow-partial-results
+              ::guuids existing-groups
+              ::bad-uuids missing-groups}])
```
dposada: Should we use either `guuids` or `uuids` in both of these key names?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1371,16 +1367,26 @@
...
                                           " the following groups "
                                           (str/join \space unauthorized-guuids))}])))
-     :exists? (partial retrieve-groups conn)
+     :exists? (fn [ctx]
+                (or (::allow-partial-results? ctx)
+                    (empty? (::bad-uuids ctx))
+                    [false {::error (str "The following UUIDs didn't correspond to a group: "
+                                         (str/join
+                                           \space
+                                           (::bad-uuids ctx)))}]))
```
dposada: I'm not following the logic here. Let's say `(::bad-uuids ctx)` is empty - wouldn't this cause `exists?` to return the error structure?

DaoWen: The `or` here is equivalent to this:
```clojure
(cond  ; :exists?
  (::allow-partial-results? ctx) true
  (empty? (::bad-uuids ctx)) true
  :else [false {:error ...}])

scheduler/src/cook/mesos/api.clj
DaoWen: Renamed `::bad-uuids` → `::non-existing-guuids`.

PR #634, opened 2017-12-12
Makes several CLI tests more reliable

dposada:

- `test_submit_with_command_prefix`: explicitly use a config file with no command-prefix
- `temp_config_file`: use the `COOK_SESSION_MODULE` environment variable if it's set
- `test_submit_raw_multiple`: respect the configured command-prefix
- `test_submit_raw`: respect the configured command-prefix
- `test_retries`: look for the 'Retrying' log message to count the retries
- `test_complex_commands`: respect the configured command-prefix
- `test_double_dash_for_end_of_options`: respect the configured command-prefix
- `test_tail_follow`: use `wait_for_instance` instead of `wait_for_job(..., 'running')`

So that the CLI tests can be run in multiple environments without flaking.

### DIFF
--- a/integration/tests/cook/cli.py
+++ b/integration/tests/cook/cli.py
@@ -138,12 +138,12 @@ class temp_config_file:
     """
 
     def __init__(self, config):
-        self.config = config
         session_module = os.getenv('COOK_SESSION_MODULE')
         if session_module:
-            modules_config = {'http': {'modules': {'session-module': session_module,
-                                                   'adapters-module': session_module}}}
-            self.config = {**modules_config, **config}
+            self.config = {'http': {'modules': {'session-module': session_module, 'adapters-module': session_module}}}
+            self.config.update(config)
+        else:
+            self.config = config
 
     def write_temp_json(self):
         path = tempfile.NamedTemporaryFile(delete=False).name
--- a/integration/tests/cook/cli.py
+++ b/integration/tests/cook/cli.py
@@ -245,7 +245,7 @@ def version():
         raise Exception(f'Unable to parse version from {string}')
 
 
-def config_get(key, flags):
+def config_get(key, flags=None):
     """Invokes the config subcommand to get a config value"""
     cp = cli(f'config --get {key}', flags=flags)
     return cp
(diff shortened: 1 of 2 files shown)


integration/tests/cook/cli.py
```diff
@@ -139,6 +139,11 @@ class temp_config_file:
 
     def __init__(self, config):
         self.config = config
+        session_module = os.getenv('COOK_SESSION_MODULE')
+        if session_module:
```
shamsimam: Should we use an `if-else` branch to initialize `self.config`

integration/tests/cook/cli.py
```diff
@@ -139,6 +139,11 @@ class temp_config_file:
...
+            modules_config = {'http': {'modules': {'session-module': session_module,
+                                                   'adapters-module': session_module}}}
+            self.config = {**modules_config, **config}
```
shamsimam: Nitpick: I find `modules_config.update(config)` more readable.

integration/tests/cook/cli.py
```diff
@@ -250,3 +255,9 @@ def config_set(key, value, flags):
     """Invokes the config subcommand to set a config value"""
     cp = cli(f'config {key} {value}', flags=flags)
     return cp
+
+
+def command_prefix():
```
shamsimam: +1

integration/tests/cook/test_cli.py
```diff
@@ -295,30 +294,27 @@ def test_retries(self):
         cp, uuids = cli.submit('-- ls -al', 'localhost:99999', '--verbose')
         stderr = cli.decode(cp.stderr)
         self.assertEqual(1, cp.returncode, stderr)
-        self.assertIn('Starting new HTTP connection (1)', stderr)
-        self.assertIn('Starting new HTTP connection (2)', stderr)
-        self.assertIn('Starting new HTTP connection (3)', stderr)
-        self.assertNotIn('Starting new HTTP connection (4)', stderr)
+        self.assertIn('Retrying (Retry(total=1', stderr)
+        self.assertIn('Retrying (Retry(total=0', stderr)
```
shamsimam: Nitpick: we should place the `0` statement before the `1` statement if that is the order they appear in the stderr file.

dposada: That's actually the order they show up in

integration/tests/cook/cli.py
dposada: Done

integration/tests/cook/cli.py
dposada: Done

PR #635, opened 2017-12-12
Refactor test_progress_update_submit and test_show_progress_message_custom_progress_file

shamsimam:

- Chooses the EXECUTOR_PROGRESS_OUTPUT_FILE_ENV dynamically in `test_progress_update_submit` and `test_show_progress_message_custom_progress_file`

Allows the test respect the `EXECUTOR_PROGRESS_OUTPUT_FILE_ENV` variable value when configured in the scheduler

### DIFF
--- a/integration/tests/cook/test_cli.py
+++ b/integration/tests/cook/test_cli.py
@@ -936,9 +936,7 @@ class CookCliTest(unittest.TestCase):
 
     def test_show_progress_message_custom_progress_file(self):
         executor = util.get_job_executor_type(self.cook_url)
-        settings = util.settings(self.cook_url)
-        progress_file_env = (util.get_in(settings, 'executor', 'environment', 'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV') or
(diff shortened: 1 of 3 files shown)


integration/tests/cook/test_cli.py
```diff
@@ -936,15 +936,18 @@ def test_show_progress_message(self):
 
     def test_show_progress_message_custom_progress_file(self):
         executor = util.get_job_executor_type(self.cook_url)
+        settings = util.settings(self.cook_url)
+        progress_file_env = (util.get_in(settings, 'executor', 'environment', 'EXECUTOR_PROGRESS_OUTPUT_FILE_ENV') or
+                             'EXECUTOR_PROGRESS_OUTPUT_FILE')
```
dposada: Should we extract a function since we do the same thing in the two tests?

shamsimam: Done, introduced `util.retrieve_progress_file_env`

PR #638, opened 2017-12-13
Fix NPE for nonextant group show

DaoWen:

- Fix bug with NPE on group show of invalid group
- Add an integration test for this case

Apparently we have something against NPEs.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -781,6 +781,11 @@ class CookTest(unittest.TestCase):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertEqual('success', job['state'], 'Job details: %s' % (json.dumps(job, sort_keys=True)))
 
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -781,6 +781,11 @@ def test_cancel_instance(self):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertEqual('success', job['state'], 'Job details: %s' % (json.dumps(job, sort_keys=True)))
 
+    def test_no_such_group(self):
+        group_uuid = str(uuid.uuid4())
+        resp = util.query_groups(self.cook_url, uuid=[group_uuid])
+        self.assertEqual(resp.status_code, 404)
```
dposada: Is there anything in the body that we should assert on?

DaoWen:
@dposada - I think this is ready to merge.

PR #640, opened 2017-12-14
makes message length config responsible only for progress messages

shamsimam:

- make `max_message_length` config only affect length of progress messages
- refactoring: make `send_message` expect a dictionary instead of a string

It simplifies the code to make the max message length directly apply to the progress message length. Since the length check only applies to the progress messages, we are still backward compatible in terms of consumption of the config.

### DIFF
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -102,11 +102,13 @@ class StatusUpdater(object):
 def send_message(driver, message):
     """Sends the message, if it is smaller than the max length, using the driver.
 
+    Note: This function must rethrow any OSError exceptions that it encounters.
+
     Parameters
     ----------
     driver: MesosExecutorDriver
         The driver to send the message to.
-    message: object
+    message: dictionary
         The raw message to send.
 
     Returns
@@ -115,10 +117,12 @@ def send_message(driver, message):
     """
     try:
         logging.info('Sending framework message {}'.format(message))
-        message_string = str(message).encode('utf8')
+        message_string = json.dumps(message).encode('utf8')
         encoded_message = pm.encode_data(message_string)
         driver.sendFrameworkMessage(encoded_message)
         return True
+    except OSError as os_error:
+        raise os_error
     except Exception:
         logging.exception('Error in sending message {}'.format(message))
         return False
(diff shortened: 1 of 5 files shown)


executor/cook/executor.py
```diff
@@ -305,12 +302,11 @@ def manage_task(driver, task, stop_signal, completed_signal, config):
         task_completed_signal = Event() # event to track task execution completion
         sequence_counter = cp.ProgressSequenceCounter()
 
-        def send_progress_message(message):
-            return send_message(driver, message, config.max_message_length)
-
+        send_progress_message = functools.partial(send_message, driver)
```
pschorf: I like `def` or `lambda` better

executor/cook/executor.py
```diff
@@ -99,34 +99,34 @@ def update_status(self, task_state, reason=None):
...
         encoded_message = pm.encode_data(message_string)
         driver.sendFrameworkMessage(encoded_message)
         return True
-    else:
-        log_message_template = 'Unable to send message as its length of {} exceeds allowed max length of {}'
-        logging.warning(log_message_template.format(len(message_string), max_message_length))
+    except OSError as os_error:
+        raise os_error
+    except Exception:
+        logging.exception('Error in sending message {}'.format(message))
```
dposada: Were these new `except` clauses for a reason?

shamsimam: I added the exception handling to allow the executor to chug along even if there is an error in sending a framework message. The exception handler needs to rethrow the `OSError` so that cleanup can be performed higher up.

DaoWen: FYI, you can leave the exception value anonymous if all you want to with it do is re-raise it:
```python
except OSError:
    raise
```

executor/cook/executor.py
shamsimam: Keeping this as `partial` as it aligns with our Clojure style.

executor/cook/executor.py
shamsimam: Done.

shamsimam:
@DaoWen @pschorf any more feedback?

PR #641, opened 2017-12-14
Deref Kerberos middleware fn var before adding meta-data

DaoWen:

Fix error caused by trying to attach meta-data to the Kerberos middlewar *Var* instead of the *Fn*.

Clojure doesn't magically dereference Vars for you when calling `with-meta` like it does when you try to call them like a function.


DaoWen:
@pschorf - Could you please verify that this patch works internally before merging?

shamsimam:
This patch worked for me.

PR #642, opened 2017-12-14
Filters the clusters of interest from the provided entity refs

dposada:

- `parse_entity_refs` now returns a second value, which is the list of only the clusters of interest
- `query_with_stdin_support` now returns the same second value
- sub-commands use the clusters of interest in downstream functions instead of all configured clusters

The issue that exposed this is that if a user passes an entity ref URL with the wrong base cluster URL, the "No matching data found" message was claiming that no data was found in all configured clusters, when in reality only the clusters from the provided URLs were queried.

### DIFF
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -290,6 +290,8 @@ def parse_entity_refs(clusters, ref_strings):
     Throws if an invalid entity ref string is encountered.
     """
     entity_refs = []
+    cluster_names_of_interest = set()
+    all_cluster_names = set(c['name'] for c in clusters)
     for ref_string in ref_strings:
         result = urlparse(ref_string)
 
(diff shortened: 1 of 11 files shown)


cli/cook/querying.py
```diff
@@ -330,7 +334,8 @@ def parse_entity_refs(clusters, ref_strings):
                 for uuid in query_args['uuid']:
                     entity_refs.append({'cluster': cluster_name, 'type': entity_type, 'uuid': uuid})
 
-    return entity_refs
+    clusters_of_interest = [c for c in clusters if c['name'] in cluster_names_of_interest]
+    return entity_refs, clusters_of_interest
```
DaoWen: You should probably update the docstring to reflect the new (pair) return type.

cli/cook/querying.py
```diff
@@ -356,7 +362,7 @@ def query_with_stdin_support(clusters, entity_refs, pred_jobs=None, pred_instanc
             raise Exception('You must specify at least one UUID or URL.')
 
         ref_strings = stdin.splitlines()
-        entity_refs = parse_entity_refs(clusters, ref_strings)
+        entity_refs, clusters_of_interest = parse_entity_refs(clusters, ref_strings)
 
-    query_result = query(clusters, entity_refs, pred_jobs, pred_instances, pred_groups, timeout, interval)
-    return query_result
+    query_result = query(clusters_of_interest, entity_refs, pred_jobs, pred_instances, pred_groups, timeout, interval)
+    return query_result, clusters_of_interest
```
DaoWen: Same here.

cli/cook/querying.py
dposada: Good call, done

cli/cook/querying.py

PR #643, opened 2017-12-14
Add /info endpoint (unauthenticated)

DaoWen:

Add unauthenticated `/info` endpoint that includes:
- Cook version
- Cook build commit hash
- Cook scheduler instance start time
- Current authentication scheme

Sample JSON response:
```javascript
{
    "authorization-scheme": "one-user",
    "commit": "dev",
    "start-time": "2017-12-15T01:22:12.956Z",
    "version": "1.8.4-SNAPSHOT"
}
```

Having the authentication scheme queryable on an unauthenticated endpoint will be helpful for determining the current authentication scheme when performing setup for our multi-user integration tests. The other information is helpful for checking the status of an update deployment.

### DIFF
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -47,6 +47,7 @@
            org.eclipse.jetty.server.handler.HandlerCollection
            org.eclipse.jetty.server.handler.RequestLogHandler)
   (:gen-class))
+
 (defn wrap-no-cache
   [handler]
   (fn [req]
@@ -230,6 +231,15 @@
       (log/info "Found framework id in zookeeper:" framework-id)
       framework-id)))
 
+(defn conditional-auth-bypass
+  "Skip authentication on some hard-coded endpoints."
+  [h auth-middleware]
+  (let [auth-fn (auth-middleware h)]
+    (fn filtered-auth [{:keys [uri request-method] :as req}]
+      (if (and (= "/info" uri) (= :get request-method))
+        (h req)
+        (auth-fn req)))))
+
 (def scheduler-server
   (graph/eager-compile
     {:mesos-datomic mesos-datomic
@@ -246,7 +256,7 @@
                                                        tell-jetty-about-usename
                                                        (wrap-rate-limit {:storage rate-limit-storage
                                                                          :limit user-limit})
-                                                       authorization-middleware
+                                                       (conditional-auth-bypass authorization-middleware)
                                                        wrap-stacktrace
                                                        wrap-no-cache
                                                        wrap-cookies
@@ -363,7 +373,7 @@
                                      (log/info "Using http basic authorization with validation" validation)
                                      (with-meta
                                        ((lazy-load-var 'cook.basic-auth/create-http-basic-middleware) user-password-valid?)
-                                       {:json-value "HttpBasicAuthMiddleware"}))
+                                       {:json-value "http-basic"}))
 
                                    one-user
                                    (do
@@ -372,14 +382,14 @@
                                        (fn one-user-middleware [h]
                                          (fn one-user-auth-wrapper [req]
                                            (h (assoc req :authorization/user one-user))))
-                                       {:json-value "OneUserAuthMiddleware"}))
+                                       {:json-value "one-user"}))
 
                                    kerberos
                                    (do
                                      (log/info "Using kerberos middleware")
                                      (with-meta
                                        (lazy-load-var 'cook.spnego/require-gss)
-                                       {:json-value "KerberosAuthMiddleware"}))
+                                       {:json-value "kerberos"}))
                                    :else (throw (ex-info "Missing authorization configuration" {}))))
      :rate-limit (fnk [[:config {rate-limit nil}]]
                    (let [{:keys [user-limit-per-m]
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -17,6 +17,7 @@
   (:require [camel-snake-kebab.core :refer [->snake_case ->kebab-case]]
             [cheshire.core :as cheshire]
             [clj-time.core :as t]
+            [clj-time.format :as tf]
             [clojure.set :as set]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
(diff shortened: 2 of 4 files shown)


scheduler/src/cook/components.clj
```diff
@@ -230,6 +231,16 @@
       (log/info "Found framework id in zookeeper:" framework-id)
       framework-id)))
 
+(defn auth-filter
```
shamsimam: Should this be called `bypass-auth-filter` or `skip-auth-filter`. Otherwise, the name suggests to me that the filter is going to perform the authorization.

scheduler/src/cook/components.clj
```diff
@@ -230,6 +231,16 @@
...
+  "Skip authentication on some hard-coded endpoints."
+  [h auth-middleware]
+  (let [auth-fn (auth-middleware h)]
+    (fn filtered-auth [req]
```
shamsimam: Nitpick: eagerly destructure the `uri` and `request-method` here.

scheduler/src/cook/components.clj
```diff
@@ -230,6 +231,16 @@
...
+      (if (and (= (:uri req) "/info")
```
shamsimam: Nitpick: prefer to see the constant first in the equals comparison.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1119,6 +1130,21 @@
...
+        auth-scheme (str (or (-> auth-middleware meta :json-value) auth-middleware))
+        ;; We store the start-up time (ISO-8601) for reporting on the /info endpoint
+        start-up-time (tf/unparse iso-8601-format (t/now))]
+    (base-cook-handler
+      {:allowed-methods [:get]
+       :handle-ok (fn get-info-handler [_]
+                    {:version @cook.util/version
+                     :commit @cook.util/commit
+                     :start-time start-up-time
+                     :authorization-scheme auth-scheme})})))
```
shamsimam: Nitpick: lexicographic ordering

scheduler/src/cook/mesos/api.clj
```diff
@@ -31,6 +32,7 @@
             [cook.mesos.unscheduled :as unscheduled]
             [cook.mesos.util :as util]
             [cook.mesos]
+            [cook.util]
```
shamsimam: Nitpick: can we alias this to something, e.g. `cutil`

scheduler/src/cook/components.clj
DaoWen: Renamed to `conditional-auth-bypass`.

scheduler/src/cook/components.clj
DaoWen: Done

scheduler/src/cook/components.clj
DaoWen: Done

scheduler/src/cook/mesos/api.clj
DaoWen: Done

scheduler/src/cook/mesos/api.clj
DaoWen: I'd prefer not to alias this since we have both `cook.mesos.util` and `cook.util` in use here.

DaoWen:
@dposada - Now the values here match the keys in the config file.

dposada:
Can we update the sample JSON response to reflect the new format?

DaoWen:
@dposada - I updated the sample in the PR comment. It's copy-pasted from the test Cook I'm running locally. @shamsimam will also be happy to see that the output is now lexicographically sorted! 😉

dposada:
Should we call the key `authentication-scheme` instead of `authorization-scheme`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2209,6 +2235,14 @@
                    :handler (read-jobs-handler-multiple conn framework-id is-authorized-fn
                                                         retrieve-sandbox-directory-from-agent)}}))
 
+        (c-api/context
+          "/info" []
+          (c-api/resource
+            {:get {:summary "Returns info about this Cook scheduler instance's setup. No authentication required."
```
shamsimam: I don't like the fact that the `No authentication required.` logic is sitting far away in `conditional-auth-bypass`. But not sure if there is a way to get past it.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1119,6 +1130,21 @@
...
+  "Handler for the /info endpoint"
+  [settings]
+  (let [auth-middleware (:authorization-middleware settings)
+        auth-scheme (str (or (-> auth-middleware meta :json-value) auth-middleware))
+        ;; We store the start-up time (ISO-8601) for reporting on the /info endpoint
+        start-up-time (tf/unparse iso-8601-format (t/now))]
+    (base-cook-handler
+      {:allowed-methods [:get]
+       :handle-ok (fn get-info-handler [_]
+                    {:authorization-scheme auth-scheme
```
dposada: Should this be `:authentication-scheme`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2209,6 +2235,14 @@
                    :handler (read-jobs-handler-multiple conn framework-id is-authorized-fn
                                                         retrieve-sandbox-directory-from-agent)}}))
 
+        (c-api/context
+          "/info" []
+          (c-api/resource
+            {:get {:summary "Returns info about this Cook scheduler instance's setup. No authentication required."
+                   :responses {200 {:schema CookInfo
+                                    :description "The Cook scheduler info was returned."}}
```
dposada: Nitpick: I think we should capitalize Scheduler in both places

scheduler/src/cook/mesos/api.clj
DaoWen: I agree this isn't ideal, but I don't have a good solution. I feel like pulling this out to add another level of indirection wouldn't necessarily make it easier to track down. But I did add a comment above this line that references `cook.components/conditional-auth-bypass`.

scheduler/src/cook/mesos/api.clj
DaoWen: Yes, that's a better name. Fixed.

scheduler/src/cook/mesos/api.clj
DaoWen: Done

DaoWen:
@shamsimam @dposada – RTM?

PR #644, opened 2017-12-14
Softens the UUID validity check

dposada:

- changing `is_valid_uuid` to no longer check if the string representation of the parsed UUID is equal to the input string
- adding an integration test for a UUID that exposes this: `019c34c3-13b3-b370-01a5-d1ecc9071249`

The UUID above, and others, parse as valid UUIDs, but for some reason have a different string representation:

```bash
>>> uuid_str = '019c34c3-13b3-b370-01a5-d1ecc9071249'
>>> import uuid
>>> uuid_obj = uuid.UUID(uuid_str, version=4)
>>> str(uuid_obj)
'019c34c3-13b3-4370-81a5-d1ecc9071249'
>>> str(uuid_obj) == uuid_str
False
```

At the end of the day, all we care is that Cook Scheduler accepts these UUIDs, and it does.

### DIFF
--- a/cli/cook/util.py
+++ b/cli/cook/util.py
@@ -76,11 +76,10 @@ def is_valid_uuid(uuid_to_test, version=4):
     False
     """
     try:
-        uuid_obj = uuid.UUID(uuid_to_test, version=version)
+        uuid.UUID(uuid_to_test, version=version)
+        return True
     except:
         return False
-
-    return str(uuid_obj) == uuid_to_test
 
 
 silent = False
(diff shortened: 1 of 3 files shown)


DaoWen:
Just FYI, this seems to be a result of Python overriding the version/variant bits:

> The version argument is optional; if given, the resulting UUID will have its variant and version number set according to RFC 4122, overriding bits in the given hex, bytes, bytes_le, fields, or int.[<sup>[1]</sup>](https://docs.python.org/3.6/library/uuid.html#uuid.UUID)

PR #645, opened 2017-12-18
Cook Executor:  narrows OSError handler to only trigger task exit only for out of memory

shamsimam:

- narrows OSError handler to only trigger task exit only for out of memory OSErrors
- adds exception handling logic to send_message

There are a few error classes (e.g. `socker.error`) that extend `OSError` and are currently incorrectly being treated as out of memory errors. This PR fixes this bug.

In addition, we are adding exception handling logic to sending of messages to correctly treat network error as transient errors and allow the executor to chug along.

### DIFF
--- a/executor/cook/_version.py
+++ b/executor/cook/_version.py
@@ -1,4 +1,4 @@
 # This file is read by setup.py to obtain the version.
 # Be aware that changing the format may break the parsing logic.
 
-__version__ = "0.1.6"
+__version__ = "0.1.7"
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -1,4 +1,3 @@
-import errno
 import functools
 import json
 import logging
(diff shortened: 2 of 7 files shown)


shamsimam:
@dposada 

> Would it be possible to write an integration test that confirms that non-OOM OSErrors don't trigger the shutdown?

This was discovered by socket timeout errors in sending progress messages. I will add an integration test that sends lots of progress messages, but not sure we can consistently, if at all, reproduce socket exceptions in the test.

dposada:
Can we configure the socket timeout in the executor to trigger this scenario?

pschorf:
I'll let @dposada merge this.

shamsimam:
> Can we configure the socket timeout in the executor to trigger this scenario?

I ended up writing a 'unit' test in cook executor that ensures the command executor runs to completion when all sends of status and framework messages from the driver are failing (`tests.test_executor.ExecutorTest#test_manage_task_successful_exit_despite_faulty_driver`). This test exercises the entire workflow of the Cook executor and verifies that the command was run to completion despite the presence of socket errors.

dposada:
Needs rebase

shamsimam:
Rebased.

dposada:
Are we waiting for @DaoWen to take a look?

shamsimam:
@dposada Nope, I had added all of you as reviewers, removed @DaoWen now :) Feel free to merge.

PR #646, opened 2017-12-19
Adds cat sub-command to cs

dposada:

- adds `cs cat`
- adds integration tests for it
- adds docs for it

We already have `ls` and `tail`, and `cat` is useful and simple.

### DIFF
--- a/cli/cook/mesos.py
+++ b/cli/cook/mesos.py
@@ -78,6 +78,6 @@ def read_file(instance, sandbox_dir, path, offset=None, length=None):
 
     if resp.status_code != 200:
         logging.error(f'mesos agent returned status code {resp.status_code} and body {resp.text}')
-        raise Exception('Encountered error when reading file from Mesos agent.')
+        raise Exception('Could not read the file.')
 
     return resp.json()
--- a/cli/cook/mesos.py
+++ b/cli/cook/mesos.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from urllib.parse import urlparse, parse_qs
 
 from cook import http
@@ -59,3 +60,24 @@ def retrieve_instance_sandbox_directory(instance, job):
         raise Exception(f'Found more than one Mesos executor with ID {instance_id}')
 
     return directories[0]
+
+
+def read_file(instance, sandbox_dir, path, offset=None, length=None):
+    """Calls the Mesos agent files/read API for the given path, offset, and length"""
+    logging.info(f'reading file from sandbox {sandbox_dir} with path {path} at offset {offset} and length {length}')
+    agent_url = instance_to_agent_url(instance)
+    params = {'path': os.path.join(sandbox_dir, path)}
+    if offset is not None:
+        params['offset'] = offset
+    if length is not None:
+        params['length'] = length
+
+    resp = http.__get(f'{agent_url}/files/read', params=params)
+    if resp.status_code == 404:
+        raise Exception(f"Cannot open '{path}' for reading (file was not found).")
+
+    if resp.status_code != 200:
+        logging.error(f'mesos agent returned status code {resp.status_code} and body {resp.text}')
+        raise Exception('Encountered error when reading file from Mesos agent.')
+
+    return resp.json()
--- a/cli/cook/subcommands/cat.py
+++ b/cli/cook/subcommands/cat.py
@@ -1,7 +1,8 @@
+import argparse
 from functools import partial
 
 from cook.mesos import read_file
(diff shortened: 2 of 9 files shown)


cli/cook/subcommands/cat.py
```diff
@@ -0,0 +1,43 @@
...
+
+def cat(clusters, args, _):
+    """Outputs the contents of the corresponding Mesos sandbox path by job or instance uuid."""
+    guard_no_cluster(clusters)
+    entity_refs, clusters_of_interest = parse_entity_refs(clusters, args.get('uuid'))
+    paths = args.get('path')
+
+    if len(entity_refs) > 1:
+        # argparse should prevent this, but we'll be defensive anyway
+        raise Exception(f'You can only provide a single uuid.')
```
DaoWen: I think it's good practice to avoid using "you" when reporting errors. Something less direct like *"Only a single UUID is supported" might be better. However, I think this error would be better handled by `argparse`.

Also, we should replace "UUID" with "entity refs" or something like that. More discussion on this below.

cli/cook/subcommands/cat.py
```diff
@@ -0,0 +1,43 @@
...
+
+    command_fn = partial(cat_for_instance, path=paths[0])
+    query_unique_and_run(clusters_of_interest, entity_refs[0], command_fn)
+
+
+def register(add_parser, _):
+    """Adds this sub-command's parser and returns the action function"""
+    parser = add_parser('cat', help='output files by job or instance uuid')
+    parser.add_argument('uuid', nargs=1)
+    parser.add_argument('path', nargs=1)
```
DaoWen: I think you should add a custom `type=` verification on both of these arguments so that the usage message is triggered if they pass the wrong arguments, or the wrong type of arguments.

cli/cook/subcommands/cat.py
dposada: I have the comment above the line to attempt to make it clear that `argparse` should prevent this from ever happening

cli/cook/subcommands/cat.py
```diff
@@ -0,0 +1,43 @@
...
+        raise Exception(f'You can only provide a single path.')
+
+    command_fn = partial(cat_for_instance, path=paths[0])
+    query_unique_and_run(clusters_of_interest, entity_refs[0], command_fn)
+
+
+def register(add_parser, _):
+    """Adds this sub-command's parser and returns the action function"""
+    parser = add_parser('cat', help='output files by job or instance uuid')
+    parser.add_argument('uuid', nargs=1)
```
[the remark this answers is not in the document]
dposada: Changing to:

```bash
$ cs cat --help
usage: cs cat [-h] target-entity path

positional arguments:
  target-entity  Accepts either a job or an instance UUID or URL. The latest
                 instance is selected for a job with multiple instances.
  path           The path is relative to the sandbox directory on the Mesos
                 agent where the instance runs.

optional arguments:
  -h, --help     show this help message and exit
```

cli/cook/subcommands/cat.py
DaoWen: Ah, yes, I should have noticed the comment. It sounds like this check fits better semantically as an assertion than as an explicit test+raise. E.g.:
```python
# argparse should enforce this
assert len(entity_refs) == 1, "Must provide a single job or instance reference"
```

cli/cook/mesos.py
```diff
@@ -59,3 +60,24 @@ def retrieve_instance_sandbox_directory(instance, job):
...
+    if length is not None:
+        params['length'] = length
+
+    resp = http.__get(f'{agent_url}/files/read', params=params)
+    if resp.status_code == 404:
+        raise Exception(f"Cannot open '{path}' for reading (file was not found).")
+
+    if resp.status_code != 200:
+        logging.error(f'mesos agent returned status code {resp.status_code} and body {resp.text}')
+        raise Exception('Encountered error when reading file from Mesos agent.')
```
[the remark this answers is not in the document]
dposada: Changed to "Could not read the file"

cli/cook/subcommands/cat.py
dposada: Done

cli/cook/subcommands/cat.py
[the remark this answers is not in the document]
dposada: Done

cli/cook/subcommands/cat.py
dposada: Done

dposada:
@DaoWen this is ready for another pass

PR #647, opened 2017-12-19
Avoids cs exiting on connection error with a cluster

dposada:

- if we encounter a connection error or read timeout, instead of bailing out, print an error message and continue (we might get results from other clusters)

If one of your configured clusters is down, then running, for example, `cs jobs`, should still allow you to see data from your other configured clusters.


dposada:
@DaoWen Correct.

PR #648, opened 2017-12-19
Job Client: adds support for configuring the executor field on a job and populating the executor field on an instance

shamsimam:

- adds support for configuring the executor field on a job and populating the executor field on an instance

This will enable the `JobClient` to specify the executor to use while submitting/loading a job. Additionally, it also allows displaying the executor used in an instance.

### DIFF
--- a/jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
+++ b/jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
@@ -73,7 +73,7 @@ final public class Instance {
         private Boolean _preempted;
         private String _outputURL;
         private String _hostName;
-        private String _executor;
+        private Executor _executor;
 
         /**
          * The task id must be provided prior to {@code build()}. If the instance status is not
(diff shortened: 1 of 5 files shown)


jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
```diff
@@ -242,9 +258,11 @@ public String getHostName() {
     final private Boolean _preempted;
     final private String _outputURL;
     final private String _hostName;
+    private String _executor;
```
WenboZhao: Should we use Enum for this?

shamsimam: Will do.

shamsimam: Done.

WenboZhao:
LGTM, feel free to merge. Or should I do that?

shamsimam:
Please merge, I'll rebase the other PR then.

PR #649, opened 2017-12-20
Job Client: adds support for progress percent and message to the instance

shamsimam:

- adds support for progress percent and message to the instance

Allows users of the job client to query the instance for progress message and percent.

### DIFF
--- a/jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
+++ b/jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
@@ -69,6 +69,8 @@ final public class Instance {
         private Long _startTime;
         private Long _endTime;
         private Status _status;
+        private Integer _progress;
+        private String _progressMessage;
         private Long _reasonCode;
         private Boolean _preempted;
         private String _outputURL;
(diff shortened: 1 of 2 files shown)


jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
```diff
@@ -220,6 +244,14 @@ public Status getStatus() {
             return _status;
         }
 
+        public Integer getProgress() {
```
WenboZhao: Can we add document something like "return null if progress info is not available etc. "

jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
```diff
@@ -220,6 +244,14 @@ public Status getStatus() {
...
+            return _progress;
+        }
+
+        public String getProgressMessage() {
```
WenboZhao: similar here. Otherwise, LGTM

shamsimam: Done.

jobclient/src/main/java/com/twosigma/cook/jobclient/Instance.java
shamsimam: Done.

PR #651, opened 2017-12-20
Adds batching to queries made from cs

dposada:

- groups UUIDs into batches of at most 100 when querying
- modifies 2 integration tests (`show`, `wait`) to query for 200 jobs each

We discovered a bug where querying for too many UUIDs in the same request was failing due to hitting the limit of the query string length. This change allows us to ask for as many UUIDs as needed, but in batches.

### DIFF
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -26,21 +26,16 @@ class Clusters:
 
 def __distinct(seq):
     """Remove duplicate entries from a sequence. Maintains original order."""
-    return collections.OrderedDict(zip(seq, seq)).keys()
+    seen = set()
+    seen_add = seen.add
+    return [x for x in seq if not (x in seen or seen_add(x))]
 
 
-def query_cluster(cluster, uuids, pred, timeout, interval, make_request_fn, entity_type):
+def __query_cluster(cluster, uuids, pred, timeout, interval, make_request_fn, entity_type):
     """
     Queries the given cluster for the given uuids with
     an optional predicate, pred, that must be satisfied
     """
-    if len(uuids) == 0:
-        return []
-
-    # Cook will give us back two copies if the user asks for the same UUID twice, e.g.
-    # $ cs show d38ea6bd-8a26-4ddf-8a93-5926fa2991ce d38ea6bd-8a26-4ddf-8a93-5926fa2991ce
-    # Prevent this by calling distinct:
-    uuids = __distinct(uuids)
 
     def satisfy_pred():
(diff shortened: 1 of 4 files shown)


cli/cook/querying.py
```diff
@@ -26,21 +26,16 @@ class Clusters:
 
 def __distinct(seq):
```
shamsimam: Nitpick: should this be moved to `util.py`

cli/cook/querying.py
```diff
@@ -72,6 +67,30 @@ def satisfy_pred():
     return entities
 
 
+def partition(l, n):
```
shamsimam: Nitpick: should this be moved to `util.py`

cli/cook/querying.py
```diff
@@ -72,6 +67,30 @@ def satisfy_pred():
...
+        return []
+
+    # Cook will give us back two copies if the user asks for the same UUID twice, e.g.
+    # $ cs show d38ea6bd-8a26-4ddf-8a93-5926fa2991ce d38ea6bd-8a26-4ddf-8a93-5926fa2991ce
+    # Prevent this by calling distinct:
+    uuids = __distinct(uuids)
+
+    entities = []
+    query_batch_size = 100
+    for uuid_batch in partition(uuids, query_batch_size):
```
shamsimam: Should this be inside a try-catch block and the exceptions aggregated?

cli/cook/querying.py

cli/cook/querying.py

cli/cook/querying.py
dposada: `make_data_request` (which is called by `__query_cluster`) takes care of a couple types of exceptions:

```python
def make_data_request(cluster, make_request_fn):
    """
    Makes a request (using make_request_fn), parsing the
    assumed-to-be-JSON response and handling common errors
    """
    try:
        resp = make_request_fn()
        if resp.status_code == 200:
            return resp.json()
        else:
            return []
    except requests.exceptions.ConnectionError as ce:
        logging.exception(ce)
        raise Exception(f'Encountered connection error with {cluster["name"]} ({cluster["url"]}).')
    except requests.exceptions.ReadTimeout as rt:
        logging.exception(rt)
        raise Exception(f'Encountered read timeout with {cluster["name"]} ({cluster["url"]}).')
    except IOError as ioe:
        logging.exception(ioe)
        return []
    except json.decoder.JSONDecodeError as jde:
        logging.exception(jde)
        return []
```

Any others I think we actually want to bail out of the whole operation for.

dposada:
@shamsimam RTM

PR #652, opened 2017-12-20
Job Client: adds support for configuring progress output and progress regex on a Job

shamsimam:

- adds support for configuring progress output and progress regex on a Job

Allow users of the `JobClient` to set the progress output file and regex String on a per Job basis.

### DIFF
--- a/jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -98,6 +98,8 @@ final public class Job {
         // Use LinkedHashSet to ensure the insertion order will be kept.
         private Set<Constraint> _constraints = new LinkedHashSet<>();
         private Application _application;
+        private String _progressOutputFile;
+        private String _progressRegexString;
 
         /**
          * Prior to {@code build()}, command, memory and cpus for a job must be provided.<br>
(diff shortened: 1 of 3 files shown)


jobclient/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -528,6 +530,28 @@ public Builder setApplication(Application application) {
...
             return this;
         }
+
+        /**
+         * Set the progress output file of the job expected to build.
+         *
+         * @param progressOutputFile {@link String} specifies the progress output file for the job.
+         * @return this builder.
+         */
+        public Builder setProgressOutputFile(String progressOutputFile) {
```
WenboZhao: How to use this one? Should it be relative path of MESOS_DIRECTORY ?

WenboZhao: I guess the logic is that the COOK executor will tailed the progress output file to and parse with regex to get the number?

shamsimam: It can be an absolute path or a path relative to the sandbox directory. Cook executor will use the regex to match against lines from the progress output file (and stdout and stderr) to determine progress messages. I have updated the javadocs in the two methods.

PR #653, opened 2017-12-21
Batches kill requests from cs

dposada:

- batches kill requests coming from `cs` to avoid hitting the query string length limit

Now that you can pipe from `cs jobs` to `cs kill`, it's possible that users will want to, for example, kill 1000 running jobs with something like:

```bash
$ cs jobs --all --user root --name test01 --limit 1000 -1 | cs kill
```

If we don't batch the kill requests, we end up constructing a query string that is too long.

### DIFF
--- a/cli/cook/subcommands/kill.py
+++ b/cli/cook/subcommands/kill.py
@@ -87,9 +87,6 @@ def kill_entities(query_result, clusters):
 
     def __kill(cluster, uuids, kill_fn):
         if len(uuids) > 0:
-            # success = kill_fn(cluster, uuids)
-            # if not success:
-            #     failed.extend(uuids)
             for uuid_batch in partition(uuids, kill_batch_size):
                 success = kill_fn(cluster, uuid_batch)
                 (succeeded if success else failed).extend(uuid_batch)
--- a/cli/cook/subcommands/kill.py
+++ b/cli/cook/subcommands/kill.py
@@ -48,34 +48,25 @@ def guard_against_duplicates(query_result):
         raise Exception(message)
 
 
-def __kill_entities(cluster, uuids, endpoint, param, entity_type):
+def __kill_entities(cluster, uuids, endpoint, param):
     """Attempts to kill the jobs / instances / groups with the given UUIDs on the given cluster"""
-    success_status_code = 204
-    cluster_name = cluster['name']
     resp = http.delete(cluster, endpoint, params={param: uuids})
-    if resp.status_code == success_status_code:
-        for uuid in uuids:
-            print_info(f'Killed {entity_type} {colors.bold(uuid)} on {colors.bold(cluster_name)}.')
-        return True
-    else:
-        for uuid in uuids:
-            print(colors.failed(f'Failed to kill {entity_type} {uuid} on {cluster_name}.'))
-        return False
+    return resp.status_code == 204
 
 
 def kill_jobs(cluster, uuids):
     """Attempts to kill the jobs with the given UUIDs"""
-    return __kill_entities(cluster, uuids, 'rawscheduler', 'job', 'job')
+    return __kill_entities(cluster, uuids, 'rawscheduler', 'job')
 
 
 def kill_instances(cluster, uuids):
     """Attempts to kill the job instsances with the given UUIDs"""
-    return __kill_entities(cluster, uuids, 'rawscheduler', 'instance', 'job instance')
+    return __kill_entities(cluster, uuids, 'rawscheduler', 'instance')
 
 
 def kill_groups(cluster, uuids):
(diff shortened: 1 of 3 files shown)


cli/cook/subcommands/kill.py
```diff
@@ -48,51 +48,62 @@ def add(uuid, entity_type, cluster):
...
+def __kill_entities(cluster, uuids, endpoint, param, entity_type):
+    """Attempts to kill the jobs / instances / groups with the given UUIDs on the given cluster"""
+    success_status_code = 204
+    cluster_name = cluster['name']
+    resp = http.delete(cluster, endpoint, params={param: uuids})
+    if resp.status_code == success_status_code:
+        for uuid in uuids:
+            print_info(f'Killed {entity_type} {colors.bold(uuid)} on {colors.bold(cluster_name)}.')
+        return True
+    else:
```
shamsimam: Should we include the response content as explanation of why the kill failed? Or at least log it?

cli/cook/subcommands/kill.py
```diff
@@ -48,51 +48,62 @@ def add(uuid, entity_type, cluster):
         raise Exception(message)
 
 
+def __kill_entities(cluster, uuids, endpoint, param, entity_type):
```
shamsimam: Should this function return a dictionary with `success` and `failed` entries and let the printing be done elsewhere? This allows the caller to decide to group together the success and fail calls.

integration/tests/cook/test_cli.py
```diff
@@ -1423,30 +1423,10 @@ def test_jobs_json_and_one_per_line_mutually_exclusive(self):
...
-        self.assertEqual(0, cp.returncode, cp.stderr)
-        cp, jobs = cli.jobs_json(self.cook_url, jobs_flags)
-        self.assertEqual(0, cp.returncode, cp.stderr)
-        self.assertEqual(0, len(jobs))
-        cp, jobs = cli.jobs_json(self.cook_url, f'--user {user} --name {name} --failed')
-        self.assertEqual(0, cp.returncode, cp.stderr)
-        self.assertEqual(3, len(jobs))
-
-    def test_piping_from_jobs_to_show_and_wait(self):
+    def test_piping_from_jobs_to_kill_show_wait(self):
```
shamsimam: When test run in parallel what stops jobs from other tests showing up in the results of these commands?

cli/cook/subcommands/kill.py
dposada: I moved the printing outward, but this function still just returns `True`/`False` -- let me know what you think

cli/cook/subcommands/kill.py
dposada: The response is logged in `http.delete`:

```python
def delete(cluster, endpoint, params):
    """DELETEs data corresponding to the given params on cluster at /endpoint"""
    url = __make_url(cluster, endpoint)
    resp = __delete(url, params)
    logging.info(f'DELETE response: {resp.text}')
    return resp
```

integration/tests/cook/test_cli.py
dposada: All of the listing is done by `--name`, which is unique to this test

cli/cook/subcommands/kill.py
```diff
@@ -85,20 +76,25 @@ def kill_entities(query_result, clusters):
     succeeded = []
     clusters_by_name = {c['name']: c for c in clusters}
 
-    def __kill(cluster, uuids, kill_fn):
+    def __kill(cluster, uuids, kill_fn, entity_type):
         if len(uuids) > 0:
             for uuid_batch in partition(uuids, kill_batch_size):
                 success = kill_fn(cluster, uuid_batch)
                 (succeeded if success else failed).extend(uuid_batch)
+                for uuid in uuid_batch:
```
shamsimam: With my previous comment I was expecting the prints to come all at once in the end categorized by succeeded and failed after the `__kill` calls below. The current printing happens in the same manner in which they were happening when they were inside `__kill_entities`

dposada: Ah, I misunderstood your previous comment. What's the downside of having them inside the `__kill` chain versus after it?

shamsimam: I'm okay with them being inside `__kill` but outside the partition loop as then the prints are grouped by entity type first and then success/failure. My motivation was that I wanted to see all the success prints and then all the failed prints as the failed ones are the ones I would care about.  If they are interleaved, then I have to go searching for the ones that failed.

dposada: Done

shamsimam:
Awaiting green build

dposada:
@shamsimam RTM

PR #655, opened 2017-12-22
CS: Revert "disable triggering syncing agent sandbox entries into datomic

shamsimam:

- This reverts commit f2b5847 enabling the sandbox syncing for non-cook-executor instances

The PR is split into two commits. The first commit reverts a previous commit that disabled the syncer. The second commit handles merge conflicts and adds metrics.

We would like timely sandbox directory updates for tasks that are not executed via the cook executor. This can significantly save the load time on the list pages as they do not trigger the sandbox syncing logic. Additionally, saving the sandbox directory into datomic means once the directory is available, it is consistently always rendered instead of relying on the availability of the data from the Mesos agent state.

### DIFF
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -193,7 +193,7 @@
   [{:keys [curator-framework executor-config fenzo-config framework-id get-mesos-utilization gpu-enabled? make-mesos-driver-fn
            mea-culpa-failure-limit mesos-datomic-conn mesos-datomic-mult mesos-leadership-atom mesos-pending-jobs-atom
            offer-cache offer-incubate-time-ms progress-config rebalancer-config riemann-host riemann-port
-           sandbox-publisher-state server-config task-constraints trigger-chans zk-prefix]}]
+           sandbox-syncer-state server-config task-constraints trigger-chans zk-prefix]}]
   (let [{:keys [fenzo-fitness-calculator fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-warn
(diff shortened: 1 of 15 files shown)


shamsimam:
@dpo the PR is ready for review and is split into two commits. The first commit reverts a previous commit that disabled the syncer. The second commit handles merge conflicts and adds metrics.

scheduler/src/cook/mesos.clj
```diff
@@ -189,7 +189,8 @@
    rebalancer-config        -- map, config for rebalancer. See scheduler/docs/rebalancer-config.adoc for details
    progress-config          -- map, config for progress publishing. See scheduler/docs/configuration.adoc for more details
    framework-id             -- str, the Mesos framework id from the cook settings
-   fenzo-config             -- map, config for fenzo, See scheduler/docs/configuration.adoc for more details"
+   fenzo-config             -- map, config for fenzo, See scheduler/docs/configuration.adoc for more details
+   sandbox-syncer-state     -- map, representing the sandbox syncer object"
```
dposada: Should this be `sandbox-syncer` instead of `-state`?

shamsimam:
@dposada I have looked at them internally and they look good.

scheduler/src/cook/mesos.clj
shamsimam: It is a map representing the atoms and functions that are returned as a result of starting the syncer. The map is not the sandbox syncer.

shamsimam:
Rebased and handled merge conflicts. The PR is ready for review.

PR #657, opened 2017-12-22
CS: refactors code using suggestions from kibit

shamsimam:

- Small refactoring suggested by kibit

Cleanup of code to use common Clojure idioms.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -66,9 +66,9 @@
 (sh/let-programs
   [_id "/usr/bin/id"]
   (defn uid [user-name]
-    (clojure.string/trim (_id "-u" user-name)))
+    (str/trim (_id "-u" user-name)))
   (defn gid [user-name]
-    (clojure.string/trim (_id "-g" user-name))))
+    (str/trim (_id "-g" user-name))))
 
 (defn render-error
   [ctx]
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -2028,7 +2027,7 @@
                    [false {::error (str "You are not authorized to list jobs for " user)}]
 
                    (or (< 168 since-hours-ago)
-                       (> 0 since-hours-ago))
+                       (neg? since-hours-ago))
                    [false {::error (str "since-hours-ago must be between 0 and 168 (7 days)")}]
```
DaoWen: We should add "inclusive" to this error message.

scheduler/src/cook/mesos/api.clj
```diff
@@ -2028,7 +2027,7 @@
                    [false {::error (str "You are not authorized to list jobs for " user)}]
 
                    (or (< 168 since-hours-ago)
-                       (> 0 since-hours-ago))
```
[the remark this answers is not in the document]
shamsimam: Done.

scheduler/src/cook/mesos/api.clj
shamsimam: Used `interval [0, 168]` in the error message

PR #661, opened 2017-12-28
Disable Marathon in Travis tests

DaoWen:
Requires using a custom-patched minimesos docker image until minimesos
version 0.14 is released, which fixes a NPE bug.

- Disable Marathon in the Travis minimesosFile.
- Use custom-patched minimesos-cli image to fix NPE when Marathon is disabled.

Disabling Marathon should speed up our integration tests a bit, and use fewer resources in the restricted Travis VM environment.

### DIFF
--- a/travis/minimesos
+++ b/travis/minimesos
@@ -2,9 +2,9 @@
 
 set -e
 
-MINIMESOS_TAG="0.13.0"
-PARAMS="$@"
-MINIMESOS_CLI_IMAGE="containersol/minimesos-cli"
+MINIMESOS_TAG="0.13.1"
+PARAMS="--debug $@"
+MINIMESOS_CLI_IMAGE="daowen/minimesos-cli"
 
 command_exists() {
     command -v "$@" > /dev/null 2>&1
--- a/travis/minimesosFile
+++ b/travis/minimesosFile
@@ -284,13 +284,6 @@ minimesos {
         loggingLevel = "# INHERIT FROM CLUSTER"
     }
 
-    // Minimesos currently has an NPE without marathon
-    // https://github.com/ContainerSolutions/minimesos/issues/532
-    marathon {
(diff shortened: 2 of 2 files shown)


travis/minimesos
```diff
@@ -2,9 +2,9 @@
 
 set -e
 
-MINIMESOS_TAG="0.13.0"
-PARAMS="$@"
-MINIMESOS_CLI_IMAGE="containersol/minimesos-cli"
+MINIMESOS_TAG="0.13.1"
+PARAMS="--debug $@"
+MINIMESOS_CLI_IMAGE="daowen/minimesos-cli"
```
shamsimam: Using a custom patched Minimesos makes it harder for me to play with different Mesos versions that Minimesos offers. I'm not sure how I feel about this change :(

DaoWen: I don't think that's true. This is the same version we've been using, plus the patch for [this bug](https://github.com/ContainerSolutions/minimesos/issues/532). We can ditch this as soon as the minimesos teams makes their next release (which will include the fix that I had to apply manually). This only affects the `minimesos-cli` image, which is just the controller used to start/stop the cluster of docker instances. None of the long-running containers are affected.

shamsimam: Does that mean I can play around with different versions by changing the value of `MINIMESOS_TAG`? IIRC, changing that value is how I experimented with different versions of Mesos.

DaoWen: No, you wouldn't be able to change the `MINIMESOS_TAG` value—but I don't think that's what controls the Mesos version. Are you sure you weren't editing [the `mesosVersion` number in the minimesosFile?](https://github.com/DaoWen/Cook/blob/4689f3cfb2e7dc9f5c0eaa020aacbbcacbd55818/travis/minimesosFile#L6)

shamsimam: Yes! thanks.

PR #664, opened 2018-01-02
Makes several CLI tests more reliable / faster

dposada:

- making several tests submit jobs with the name of the job being the name of the test
- making tweaks to several tests to make them more reliable and/or faster to complete

To reduce flakiness and increase our ability to troubleshoot it.


integration/tests/cook/test_cli.py
```diff
@@ -16,6 +16,11 @@
 class CookCliTest(unittest.TestCase):
     _multiprocess_can_split_ = True
 
+    def current_name(self):
+        """Returns the name of the currently running test function"""
+        test_id = self.id()
```
DaoWen: This worked in nose too? Great! 😄

PR #670, opened 2018-01-06
Add multi-user integration test support

DaoWen:

- Add support for multi-user integration tests using http basic auth.
- Refactor some test setup code into test-class setup code.
- Allow specifying a custom config file with `./scheduler/bin/run-docker.sh`.
- Use `TRAVIS_BUILD_DIR` in more places in our Travis-CI scripts to eliminate some less-clear relative paths.

We need multi-user support in some of our integration tests in order to properly test some user-specific features, e.g., permissions and usage.


integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,84 @@
...
+                        job_uuid, resp = util.submit_job(self.cook_url, command='sleep 240', **job_resources)
+                        self.assertEqual(resp.status_code, 201, resp.content)
+                        all_job_uuids.append(job_uuid)
+            # Don't query until the job starts
+            util.wait_for_jobs(self.cook_url, all_job_uuids, 'running')
+            # Check the usage for each of our users
+            for i, user in enumerate(users):
+                with user:
+                    # Get the current usage
+                    util.wait_for_job(self.cook_url, job_uuid, 'running')
```
pschorf: What is `job_uuid` here?

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,84 @@
...
+                    resp = util.user_current_usage(self.cook_url, user=user1.name)
```
pschorf: Did you mean `user` instead of `user1` here?

integration/tests/cook/util.py
```diff
@@ -13,20 +14,108 @@
...
+
+def multi_cluster_tests_enabled():
+    """
+    Returns true if the CONTINUOUS_INTEGRATION environment variable is set,
+    indicating that multiple cook scheduler instances are running.
+    """
+    return os.getenv('COOK_MULTI_CLUSTER') is not None
+
+
+def _cook_auth_scheme():
```
pschorf: Instead of caching `_cook_info` in a global, can you just memoize this function call?

scheduler/bin/run-docker.sh
```diff
@@ -78,7 +78,7 @@ docker create \
     -e "COOK_DATOMIC_URI=${COOK_DATOMIC_URI}" \
     -e "COOK_LOG_FILE=log/cook-${COOK_PORT}.log" \
     -v ${DIR}/../log:/opt/cook/log \
-    cook-scheduler:latest
+    cook-scheduler:latest ${COOK_CONFIG}
```
pschorf: Can we set a default value for this (could just be empty string) instead of possibly having it undefined?

DaoWen:
@pschorf - Updated the PR comment to better motivate the `TRAVIS_BUILD_DIR` change.

scheduler/http-basic-container-config.edn
```diff
@@ -0,0 +1,38 @@
...
+         :framework-id #config/env "COOK_FRAMEWORK_ID"}
+ :unhandled-exceptions {:log-level :error}
+ :metrics {:jmx true}
+ :nrepl {:enabled? true
+         :port #config/env-int "COOK_NREPL_PORT"}
+ :log {:file #config/env "COOK_LOG_FILE"
+       :levels {"datomic.db" :warn
+                "datomic.peer" :warn
+                "datomic.kv-cluster" :warn
+                :default :info}}}
```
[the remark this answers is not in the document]
DaoWen: I combined the config files used by Travis, but I'd prefer not merging these because they're used inside a docker container. I tried that out locally, and I had to modify `run-docker.sh` to pass additional environment variables from the host through to the container. I'd rather not have to worry about correctly setting those environment variables when I'm testing. We have other sample config files in the scheduler folder, so I don't think one more will hurt.

pschorf:
Needs rebase

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,83 @@
+import logging
+import pytest
+import unittest
+
+from tests.cook import util
+
+@unittest.skipUnless(util.multi_user_tests_enabled(), "Requires using multi-user coniguration (e.g., BasicAuth) for Cook Scheduler")
```
dposada: Nitpick: Single quotes instead of double

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,83 @@
...
+        self.cook_url = type(self).cook_url
+        self.logger = logging.getLogger(__name__)
+        self.user_factory = util.UserFactory(self)
+
+    def test_job_delete_permission(self):
+        user1, user2 = self.user_factory.new_users(2)
+        with user1:
+            job_uuid, resp = util.submit_job(self.cook_url, command='sleep 10')
+        try:
+            self.logger.debug(f"User 1: {user1.name}")
```
dposada: Nitpick: single quotes

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,83 @@
...
+            util.kill_jobs(self.cook_url, [job_uuid])
+
+    def test_group_delete_permission(self):
+        user1, user2 = self.user_factory.new_users(2)
+        with user1:
+            group_spec = util.minimal_group()
+            group_uuid = group_spec['uuid']
+            job_uuid, resp = util.submit_job(self.cook_url, command='sleep 10', group=group_uuid)
+        try:
+            self.logger.debug(f"User 1: {user1.name}")
```
dposada: Nitpick: single quotes

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,83 @@
...
+        user1, user2 = self.user_factory.new_users(2)
+        with user1:
+            job_uuid, resp = util.submit_job(self.cook_url, command='sleep 10')
+        try:
+            self.logger.debug(f"User 1: {user1.name}")
+            self.assertEqual(resp.status_code, 201, resp.text)
+            with user2:
+                util.kill_jobs(self.cook_url, [job_uuid], expected_status_code=403)
+            with user1:
+                util.kill_jobs(self.cook_url, [job_uuid])
```
dposada: Should we assert after this that the job was killed?

integration/tests/cook/test_multi_user.py
```diff
@@ -0,0 +1,83 @@
...
+            group_spec = util.minimal_group()
+            group_uuid = group_spec['uuid']
+            job_uuid, resp = util.submit_job(self.cook_url, command='sleep 10', group=group_uuid)
+        try:
+            self.logger.debug(f"User 1: {user1.name}")
+            self.assertEqual(resp.status_code, 201, resp.text)
+            with user2:
+                util.kill_groups(self.cook_url, [group_uuid], expected_status_code=403)
+            with user1:
+                util.kill_groups(self.cook_url, [group_uuid])
```
dposada: Should we assert after this that the group was killed?

scheduler/http-basic-container-config.edn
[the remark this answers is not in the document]
dposada: Instead of adding another config, would it make sense to have multiple "run docker" scripts, e.g. `run-docker-basic-auth.sh`?

integration/tests/cook/test_multi_user.py
DaoWen: Done

integration/tests/cook/test_multi_user.py
DaoWen: Done

DaoWen:
@dposada - I consolidated the scheduler config files used by `run-docker.sh` using environment variables, and added the same flags to the `run-docker.sh` script that we have in the Travis script for setting up the environment within the docker container. I think that's definitely an improvement over having more config files. Thanks for the feedback!

PR #672, opened 2018-01-09
Add required "reason" param to swagger for user limit delete ops

DaoWen:

Add the required `reason` query-string parameter to the Swagger annotations on the quota and share `DELETE` endpoints.

The required parameter should be documented, and we'd prefer a 400 (malformed) error over a 500 (internal server exception) error.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1737,14 +1737,15 @@
 
 ;; /share and /quota
 (def UserParam {:user s/Str})
+(def ReasonParam {:reason s/Str})
+(def UserLimitChangeParams (merge UserParam ReasonParam))
 
 (def UserLimitsResponse
   {String s/Num})
 
 (defn set-limit-params
   [limit-type]
-  {:body-params (merge UserParam {limit-type {s/Keyword s/Num}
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -1737,14 +1737,15 @@
...
+(def UserLimitDestroyParams (merge UserParam ReasonParam))
 
 (def UserLimitsResponse
   {String s/Num})
 
 (defn set-limit-params
   [limit-type]
-  {:body-params (merge UserParam {limit-type {s/Keyword s/Num}
-                                  :reason s/Str})})
+  {:body-params (merge UserParam ReasonParam {limit-type {s/Keyword s/Num}})})
```
pschorf: Should we use `UserLimitDestroyParams` here as well?

DaoWen: I was thinking that too, but the name didn't fit, and I didn't like "UserAndReason" as an identifier... I just came up with a better name though: `UserLimitChangeParams`

DaoWen:
@dposada - There will be some tests for this in the impersonation-testing patch, but I added some test code here as well.

PR #684, opened 2018-01-16
Adds integration test for submitting job with no name

dposada:

- adding `test_submit_with_no_name`, which asserts that you can submit a job with no `name` field, and the scheduler will accept it and auto-populate the name to `cookjob`

We want to ensure that clients that may be submitting name-less jobs will continue to be able to do so.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1716,6 +1716,8 @@ class CookTest(unittest.TestCase):
         self.assertEqual(resp.status_code, 400, resp.text)
 
     def test_submit_with_no_name(self):
+        # We need to manually set the 'uuid' to avoid having the
+        # job name automatically set by the submit_job function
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -1707,3 +1714,15 @@ def test_user_limits_change(self):
         # reset user share fails (malformed) if no reason is given
         resp = util.reset_limit(self.cook_url, 'share', user, reason=None)
         self.assertEqual(resp.status_code, 400, resp.text)
+
+    def test_submit_with_no_name(self):
+        job_with_no_name = {'uuid': str(uuid.uuid4()),
```
DaoWen: You should probably add a comment here to clarify that you need to manually set the `'uuid'` to avoid having the job name automatically set by the `submit_job` function.

dposada: Good call, done

dposada:
This is RTM

PR #687, opened 2018-01-17
Add Kerberos support for multi-user integration tests

DaoWen:

Refactor to support Kerberos authentication in multi-user integration tests.

This allows us to run our automated integration tests on Kerberos-enabled clusters.

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -65,6 +65,36 @@ $ bin/run-integration.sh --port 12321 --port-2 22321
 By default, this script mounts the local directory in the docker container so there is no need to rebuild when changing the tests.
 You can disable this behavior using the `--no-mount` option.
 
+## Running multi-user tests with Kerberos
+
+Running the multi-user integration tests with Kerberos requires that the test framework
+has a way to get valid Kerberos authentication tickets for the many users from the tests.
+We use several environment variables to make this possible.
+
+* `COOK_TEST_USER_PREFIX`:
+The numbers 0-9 will be appended to this username prefix to generate usernames for the tests.
+We currently assume that no Kerberos-enabled test will need more than 10 unique test users.
+
+* `COOK_ADMIN_USER_NAME`:
+The username to be used in tests specifically requiring an admin user.
+This user should *not* have impersonation priviledges.
+
+* `COOK_IMPERSONATOR_USER_NAME`:
+The username to be used in tests specifically requiring an impersonator user.
+This user should *not* have administrator priviledges.
+
+* `COOK_KERBEROS_TEST_AUTH_CMD`:
+Command string (run in a shell subprocess) to obtain a test user's Kerberos ticket.
+The command can contain the special strings `{{COOK_USER}}`,
+which will be replaced with the target test user's name,
+and `{{COOK_SCHEDULER_URL}}`,
+which will be replaced with the Cook Scheduler URL being used by the current tests.
+The command must print (to `stdout`) the `Authorization` header value for the target user.
+
+When using Kerberos, multi-user integration tests must *not* be run in parallel.
+This restriction is due to only having 10 unique test user names,
+which are reused across each of the multi-user tests.
+
 ## Credits
 
 This package was created with (Cookiecutter)[https://github.com/audreyr/cookiecutter] and the (audreyr/cookiecutter-pypackage)[https://github.com/audreyr/cookiecutter-pypackage] project template.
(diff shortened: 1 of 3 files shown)


integration/tests/cook/util.py
```diff
@@ -34,7 +35,18 @@ def has_docker_service():
...
+_default_admin_name = 'root'
+_default_impersonator_name = 'poser'
+
+
+# Shell command used to obtain Kerberos credentials for a given test user
+_kerberos_missing_cmd =  'echo "MISSING COOK_KERBEROS_TEST_AUTH_CMD" && exit 1'
+_kerberos_auth_cmd = os.getenv('COOK_KERBEROS_TEST_AUTH_CMD', _kerberos_missing_cmd)
+
+
+class _AuthenticatedUser(object):
```
pschorf: Is it worth having this base class just for the logger lines? It seems like overkill to me.

integration/tests/cook/util.py
```diff
@@ -43,46 +55,114 @@ class User(object):
...
 
 
+class _KerberosUser(_AuthenticatedUser):
+    """
+    Object representing a Cook user with Kerberos credentials.
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
+        subcommand = f'{_kerberos_auth_cmd} {name}'
```
pschorf: Do you mean to append the username here? I don't think that's in the documentation

DaoWen:
@dposada - This is the Kerberos patch I mentioned in #667. I've tested it internally.

integration/tests/cook/util.py
DaoWen: You're right. I'll delete that.

integration/tests/cook/util.py
DaoWen: I was originally thinking there would be more common code in here... I ended up moving a lot of it into the UserFactory instead though. I kind of like having this—it defines the interface—but I guess it's not really needed in duck-typed Python... Do you feel really strongly that I should remove this?

pschorf: I don't know if I'd say really strongly, but I'd prefer to delete it. The duplicated logger code is more clear to me than calling out to the super methods in both children. There's no problem adding it back later if needed.

integration/README.md
```diff
@@ -65,6 +65,36 @@ $ bin/run-integration.sh --port 12321 --port-2 22321
...
+The command can contain the special strings `{{COOK_USER}}`,
+which will be replaced with the target test user's name,
+and `{{COOK_SCHEDULER_URL}}`,
+which will be replaced with the Cook Scheduler URL being used by the current tests.
+The command must print (to `stdout`) the `Authorization` header value for the target user.
+
+When using Kerberos, multi-user integration tests must *not* be run in parallel.
+This restriction is due to only having 10 unique test user names,
+which are reused across each of the multi-user tests.
+
```
dposada: 👍

integration/tests/cook/util.py
```diff
@@ -43,46 +57,114 @@ class User(object):
...
+        elif kerberos_enabled():
+            self.user_class = _KerberosUser
+        else:
+            raise NotImplementedError(f'Unsupported user authentication scheme: {_cook_auth_scheme()}')
+        # Set up generator for new user objects
+        if test_handle:
+            test_id = test_handle.id()
+            test_base_name = test_id[test_id.rindex('.test_')+6:].lower()
+            base_name = os.getenv('COOK_TEST_USER_PREFIX', f'{test_base_name}_')
+            self.__user_generator = ( f'{base_name}{i}' for i in range(1000000) )
```
dposada: I would favor not having the extra spaces at the start and end of the generator

integration/tests/cook/util.py
```diff
@@ -43,46 +57,114 @@ class User(object):
...
+    def admin(self):
+        """Return the administrator user"""
+        name = os.getenv('COOK_ADMIN_USER_NAME', _default_admin_name)
+        return self.user_class(name)
+
+    @functools.lru_cache()
+    def impersonator(self):
+        """Return the impersonator user"""
+        name = os.getenv('COOK_IMPERSONATOR_USER_NAME', _default_impersonator_name)
+        return self.user_class(name)
```
dposada: Can we check against `/settings` that the users chosen by these functions actually have the capabilities they're supposed to?

DaoWen: That information isn't exposed in the `/settings` data, but the assumptions are documented in the README. There's always going to be some point where we just have to assume the person running the tests configured things correctly.

dposada: Gotcha

PR #689, opened 2018-01-18
Reject submitted jobs that exceed user's quota

DaoWen:

If a single job requires more resources than the user's entire quota, then reject the job.

It seems better to reject an unschedulable job at submission time, rather than letting the user submit a job that can never be scheduled (assuming the user's quota doesn't change).

This is particularly useful in scenarios where an organization has multiple Cook clusters, but the user only has quota on some of those clusters. Currently, a user could submit a bunch of jobs to a cluster where they have no quota, and the jobs just sit there forever. After this change, all jobs would be rejected on any clusters where the user has no quota.

### DIFF
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -88,7 +88,7 @@ class MultiUserCookTest(unittest.TestCase):
                 with self.user_factory.admin():
                     util.kill_jobs(self.cook_url, all_job_uuids)
 
-    def test_job_too_much_cpu(self):
+    def test_job_cpu_quota(self):
         admin = self.user_factory.admin()
         user = self.user_factory.new_user()
         try:
@@ -123,7 +123,7 @@ class MultiUserCookTest(unittest.TestCase):
             with admin:
                 util.reset_limit(self.cook_url, 'quota', user.name)
 
-    def test_job_too_much_mem(self):
+    def test_job_mem_quota(self):
         admin = self.user_factory.admin()
         user = self.user_factory.new_user()
         try:
@@ -158,7 +158,7 @@ class MultiUserCookTest(unittest.TestCase):
             with admin:
                 util.reset_limit(self.cook_url, 'quota', user.name)
 
-    def test_job_no_quota(self):
+    def test_job_count_quota(self):
         admin = self.user_factory.admin()
         user = self.user_factory.new_user()
         try:
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -84,3 +84,41 @@ class MultiUserCookTest(unittest.TestCase):
             # Terminate all of the jobs
             if all_job_uuids:
                 util.kill_jobs(self.cook_url, all_job_uuids)
+
+    def test_job_too_big(self):
+        admin = self.user_factory.admin()
+        user = self.user_factory.new_user()
+        try:
+            # User with no quota can't submit jobs
+            with admin:
+                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=0)
+                self.assertEqual(resp.status_code, 201, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url)
+                self.assertEqual(resp.status_code, 422, msg=resp.text)
+            # User with tiny quota can't submit bigger jobs, but can submit tiny jobs
+            with admin:
+                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=4)
+                self.assertEqual(resp.status_code, 201, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url, cpus=5)
+                self.assertEqual(resp.status_code, 422, msg=resp.text)
+                _, resp = util.submit_job(self.cook_url, cpus=4)
+                self.assertEqual(resp.status_code, 201, msg=resp.text)
+            # User with zero-jobs quota can't submit any jobs
+            with admin:
+                resp = util.set_limit(self.cook_url, 'quota', user.name, count=0)
+                self.assertEqual(resp.status_code, 201, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url, cpus=0.1)
+                self.assertEqual(resp.status_code, 422, msg=resp.text)
+            # Reset user's quota back to default, then user can submit jobs again
+            with admin:
+                resp = util.reset_limit(self.cook_url, 'quota', user.name)
+                self.assertEqual(resp.status_code, 204, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url)
+                self.assertEqual(resp.status_code, 201, msg=resp.text)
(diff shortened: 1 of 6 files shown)


DaoWen:
I ran into several bugs in existing code when testing this change. Added some band-aids to get the tests to pass, but there are still several outstanding issues in #691 that should be fixed.

Also, the quota unit tests were not being run because the namespace was misnamed... 😖

integration/tests/cook/test_multi_user.py
```diff
@@ -84,3 +84,41 @@ def test_multi_user_usage(self):
...
+                self.assertEqual(resp.status_code, 422, msg=resp.text)
+            # User with tiny quota can't submit bigger jobs, but can submit tiny jobs
+            with admin:
+                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=4)
+                self.assertEqual(resp.status_code, 201, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url, cpus=5)
+                self.assertEqual(resp.status_code, 422, msg=resp.text)
+                _, resp = util.submit_job(self.cook_url, cpus=4)
+                self.assertEqual(resp.status_code, 201, msg=resp.text)
```
dposada: Maybe we should make this case even tinier so that we don't actually use 4 cpus on the cluster when we run the test

integration/tests/cook/test_multi_user.py
```diff
@@ -84,3 +84,41 @@ def test_multi_user_usage(self):
...
+            # Reset user's quota back to default, then user can submit jobs again
+            with admin:
+                resp = util.reset_limit(self.cook_url, 'quota', user.name)
+                self.assertEqual(resp.status_code, 204, resp.text)
+            with user:
+                _, resp = util.submit_job(self.cook_url)
+                self.assertEqual(resp.status_code, 201, msg=resp.text)
+        finally:
+            with admin:
+                util.reset_limit(self.cook_url, 'quota', user.name)
```
dposada: I wonder if we should replicate some or all of the scenarios but with mem instead of cpus

integration/tests/cook/util.py
```diff
@@ -869,14 +869,15 @@ def query_queue(cook_url):
     return session.get(f'{cook_url}/queue')
 
 
-def set_limit(cook_url, limit_type, user, mem=None, cpus=None, gpus=None, jobs=None, reason='testing'):
+def set_limit(cook_url, limit_type, user, mem=None, cpus=None, gpus=None, count=None, reason='testing'):
```
dposada: Seems like a docstring would be helpful here

scheduler/src/cook/mesos/api.clj
```diff
@@ -108,9 +108,14 @@
 (def ZeroInt
   (s/both s/Int (s/pred zero? 'zero?)))
 
-(def PosNum
+(s/defschema PosNum
```
dposada: What effect does `s/defschema` have?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1291,6 +1296,33 @@
       true
       [false {::error "You are not authorized to create jobs"}])))
 
+(defn no-job-exceeds-quota?
+  "Check if any of the submitted jobs exceed the user's total quota,
+   in which case the job would never be schedulable (unless the quota is increased).
+
+   If none of the jobs individually seem to exceed the user's quota, returns true;
+   otherwise, the following error structure is returned:
```
dposada: If the jobs are not grouped, should we only block the individual jobs that exceed quota?

scheduler/src/cook/mesos/api.clj
DaoWen: I think it attaches some metadata that's checked when generating the Swagger HTML page. We should probably change all of our schema definitions to use this macro.

scheduler/src/cook/mesos/api.clj
DaoWen: I thought about that, but I concluded that the return status would be too messy in that case. Either we have to do something like a multi-status return code (HTTP 207 Multi-Status), or return a status that's misleading about part of the request (either we say it failed when some jobs *were* successfully submitted, or we say it was successful when in fact some jobs were rejected). Simply rejecting the whole request outright seemed significantly cleaner than either of those options.

integration/tests/cook/util.py
DaoWen: Added

integration/tests/cook/test_multi_user.py
DaoWen: Done

scheduler/src/cook/mesos/api.clj
dposada: Makes sense

scheduler/src/cook/mesos/api.clj
```diff
@@ -1758,7 +1792,7 @@
 
 (defn set-limit-params
   [limit-type]
-  {:body-params (merge UserLimitChangeParams {limit-type {s/Keyword s/Num}})})
+  {:body-params (merge UserLimitChangeParams {limit-type {s/Keyword NonNegNum}})})
```
dposada: Should we add an integration test for this?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -554,6 +554,15 @@
...
 
+    (testing "quota is checked on job submission"
+      (let [job (assoc (basic-job) "cpus" 32.0)
+            job-resp (h (merge quota-req-attrs
+                               {:uri "/rawscheduler"
+                                :authorization/user "foo"
+                                :request-method :post
+                                :body-params {"jobs" [job]}}))]
+        (is (= 422 (:status job-resp)))))
+
```
dposada: Doesn't the integration test cover this? Are we getting any additional value out of the unit test?

scheduler/test/cook/test/mesos/quota.clj
```diff
@@ -29,6 +29,8 @@
     (quota/set-quota! conn "u1" "too many CPUs" :cpus 5.0)
     (quota/set-quota! conn "u1" "higher count" :count 6)
     (quota/set-quota! conn "u2" "custom limits" :cpus 5.0  :mem 10.0)
+    (quota/set-quota! conn "u3" "needs no GPUs" :gpus 0.0)
+    (quota/set-quota! conn "u4" "no jobs allowed" :count 0)
```
dposada: Why were these added? Do we need additional assertions for this?

scheduler/test/cook/test/mesos/api.clj
DaoWen: I added this because I was debugging locally and this was much faster than doing integration tests, but now that it's stable it's just redundant. I'll remove it.

DaoWen: I deleted the redundant CPUs test, but it occurred to me that we can test GPUs here (which are not being tested in the integration tests), so now this is a GPU-quota test.

PR #698, opened 2018-01-24
Simplify JobClient setup and dependences on its JAR

DaoWen:

- Refactor Scheduler project so only the tests depend on the Java JobClient JAR.
- Use Maven pom.xml file rather than project.clj for Java JobClient project.
- Merge the Scheduler and JobClient unit test jobs in Travis since both are in fact running JobClient tests.

This makes it simpler to run the JobClient unit tests (they didn't work correctly with lein due to JMockit complications). The dependency refactoring means that you won't get an error about missing the JobClient JAR except when running the Scheduler unit tests. The JobClient unit tests only take a few seconds to run, so merging those with the Scheduler unit tests should also shave a bunch of redundant bootstrapping time from our Travis builds without really loosing any parallelism gains.


scheduler/project.clj
```diff
@@ -176,8 +172,14 @@
     :resource-paths ["test-resources"]
     :source-paths []}
 
+   :test
+   {:dependencies [[criterium "0.4.4"]
+                   [org.mockito/mockito-core "1.10.19"]
+                   [twosigma/cook-jobclient "0.1.2-SNAPSHOT"]
+                   [org.clojure/test.check "0.6.1"]]}
```
shamsimam: Lexicographic ordering :(

jobclient/README.md
```diff
@@ -46,4 +46,13 @@ public class HostPlacementExample {
 }
 ``` 
 
+# Running the Tests
+
+The test coverage isn't very good at the moment, but here's how you run the existing unit tests:
```
shamsimam: Imho, we don't need to mention the code coverage as that will change over time.

jobclient/pom.xml
```diff
@@ -3,7 +3,7 @@
   <groupId>twosigma</groupId>
   <artifactId>cook-jobclient</artifactId>
   <packaging>jar</packaging>
-  <version>0.2.0-SNAPSHOT</version>
+  <version>0.1.2-SNAPSHOT</version>
```
shamsimam: Was the previous version from this `pom.xml` never used?

jobclient/pom.xml
```diff
@@ -90,17 +76,18 @@
       <groupId>log4j</groupId>
       <artifactId>log4j</artifactId>
       <version>1.2.17</version>
-  </dependency>
-  <dependency>
+    </dependency>
+    <dependency>
       <groupId>org.jmockit</groupId>
       <artifactId>jmockit</artifactId>
       <version>1.18</version>
```
shamsimam: Can we extract this version into a Maven property and also use it in the surefire plugin argument?

jobclient/project.clj
```diff
@@ -1,15 +0,0 @@
-;; This project file exists to make it easier to develop the the job client
```
shamsimam: 💯

jobclient/pom.xml
DaoWen: Apparently. `0.1.2-SNAPSHOT` is the version number that was in project.clj.

jobclient/README.md
DaoWen: Removed.

jobclient/pom.xml
DaoWen: Done.

scheduler/project.clj
DaoWen: Fixed.

jobclient/project.clj
DaoWen: Yeah, that comment didn't make much sense to me either...

PR #702, opened 2018-01-25
Handles NoSuchProcess thrown by psutil.Process

shamsimam:

-  handles NoSuchProcess thrown by `psutil`.Process (instead of `ProcessLookupError`)

- It correctly catches the exception thrown by the `psutil` library and avoids an unexpected error while trying to terminate a task.

### DIFF
--- a/executor/cook/_version.py
+++ b/executor/cook/_version.py
@@ -1,4 +1,4 @@
 # This file is read by setup.py to obtain the version.
 # Be aware that changing the format may break the parsing logic.
 
-__version__ = "0.1.7"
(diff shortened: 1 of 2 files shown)


executor/cook/_version.py
```diff
@@ -1,4 +1,4 @@
 # This file is read by setup.py to obtain the version.
 # Be aware that changing the format may break the parsing logic.
 
-__version__ = "0.1.7"
+__version__ = "0.1.8"
```
DaoWen: :+1:

PR #705, opened 2018-01-26
Add docker mark to docker-related integration tests

DaoWen:

Add the `docker` pytest mark to docker-related integration tests.

Allows us to filter using `pytest -m docker` or `pytest -m 'not docker'` to run or exclude tests using docker.


DaoWen:
@dposada - I thought I added these marks back when I added the `skipIf` annotation to the port-mapping test. These missing marks explain why filtering with `-m` wasn't working when we tried that the other day.

PR #707, opened 2018-01-28
Enable parallel multi-user integration tests with Kerberos

DaoWen:

Reworks how tests users are generated across xdist workers so that there aren't user collisions when restricted to a limited number of test users with Kerberos.

This lets us run all of the tests together even when using Kerberos, rather than having to run the multi-user tests separately so that they could be run sequentially.

### DIFF
--- a/integration/README.md
+++ b/integration/README.md
@@ -92,7 +92,7 @@ which will be replaced with the Cook Scheduler URL being used by the current tes
 The command must print (to `stdout`) the `Authorization` header value for the target user.
 
 * `COOK_MAX_TEST_USERS`:
-Setting this environment variable limits the number of test user allocated to each test runner process.
+Setting this environment variable limits the number of test users allocated to each test runner process.
 This is most likely necessary in any scenario where users are actually authenticated by the underlying OS.
 
 ## Credits
--- a/integration/README.md
+++ b/integration/README.md
@@ -91,9 +91,9 @@ and `{{COOK_SCHEDULER_URL}}`,
 which will be replaced with the Cook Scheduler URL being used by the current tests.
 The command must print (to `stdout`) the `Authorization` header value for the target user.
 
-When using Kerberos, multi-user integration tests must *not* be run in parallel.
-This restriction is due to only having 10 unique test user names,
-which are reused across each of the multi-user tests.
+* `COOK_MAX_TEST_USERS`:
+Setting this environment variable limits the number of test user allocated to each test runner process.
+This is most likely necessary in any scenario where users are actually authenticated by the underlying OS.
 
 ## Credits
 
--- a/integration/tests/cook/conftest.py
+++ b/integration/tests/cook/conftest.py
@@ -1,0 +1,20 @@
+import os
+import subprocess
+import threading
+import time
+
+from tests.cook import util
+
+def sudo_check(username):
+    sudo_ok = (0 == subprocess.call(f'sudo -nu {username} echo CACHED SUDO', shell=True))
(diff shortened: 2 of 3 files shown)


integration/README.md
```diff
@@ -91,9 +91,9 @@ and `{{COOK_SCHEDULER_URL}}`,
 which will be replaced with the Cook Scheduler URL being used by the current tests.
 The command must print (to `stdout`) the `Authorization` header value for the target user.
 
-When using Kerberos, multi-user integration tests must *not* be run in parallel.
-This restriction is due to only having 10 unique test user names,
-which are reused across each of the multi-user tests.
+* `COOK_MAX_TEST_USERS`:
+Setting this environment variable limits the number of test user allocated to each test runner process.
```
dposada: test user**s**

integration/tests/cook/conftest.py
```diff
@@ -0,0 +1,20 @@
+import os
+import subprocess
+import threading
+import time
+
+from tests.cook import util
+
+def sudo_check(username):
```
dposada: A docstring might be nice here

integration/tests/cook/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+
+def _sudo_checker_task(username):
+    while True:
+        sudo_check(username)
+        time.sleep(60)
+
+if util.kerberos_enabled() and os.getenv('COOK_MAX_TEST_USERS'):
+    username = next(util._test_user_names())
+    sudo_check(username)
+    threading.Thread(target=_sudo_checker_task, args=[username], daemon=True).start()
```
dposada: Why `conftest.py`? What does the name mean?

integration/tests/cook/util.py
```diff
@@ -42,9 +43,41 @@ def has_docker_service():
...
+@functools.lru_cache()
+def _test_user_ids():
+    """
+    Get the numeric user suffixes for this test worker.
+    Returns the range 0 to 1 million if COOK_MAX_TEST_USERS is not set.
+    If this is a distributed run with a limited number of users,
+    e.g., 10 per worker, then this function returns range(0, 10) for worker 0,
+    or range(20, 30) for worker 2.
+    """
+    pytest_worker = os.getenv('PYTEST_XDIST_WORKER')
```
dposada: What sets this environment variable? Might be nice to add a comment

integration/tests/cook/util.py
```diff
@@ -42,9 +43,41 @@ def has_docker_service():
...
+    else:
+        return range(1000000)
+
+
+def _test_user_names(test_name_prefix=None):
+    """
+    Returns a generator of unique test user names, with form {PREFIX}{ID}.
+    The COOK_TEST_USER_PREFIX environment variable is used by default;
+    otherwise, the test_name_prefix value is used as the PREFIX.
+    """
```
dposada: 👍

integration/tests/cook/util.py
DaoWen: The pytest-xdist (parallel testing) plugin sets this for each of its worker sub-processes.

integration/tests/cook/conftest.py
DaoWen: I'll add a comment to explain this. It's a pytest thing:
https://docs.pytest.org/en/2.7.3/plugins.html#conftest-py-plugins

integration/README.md
DaoWen: Fixed

integration/tests/cook/conftest.py
DaoWen: Added

PR #710, opened 2018-01-29
Converts monitor riemann events to codahale metrics

dposada:

- using codahale counters to track running and waiting jobs, memory, and cpus by user
- removing the explicit sends to riemann (of course you can still use the riemann reporter)

We want these metrics to be available for the non-riemann reporters as well (e.g. graphite).

## Example

Here's an example of the counters, as reported by the console reporter:

```
-- Counters --------------------------------------------------------------------
...
hungry.users
             count = 0
...
running.all.cpus
             count = 1
running.all.jobs
             count = 1
running.all.mem
             count = 128
running.root.cpus
             count = 1
running.root.jobs
             count = 1
running.root.mem
             count = 128
satisfied.users
             count = 0
starved.all.cpus
             count = 6
starved.all.jobs
             count = 1
starved.all.mem
             count = 128
starved.root.cpus
             count = 6
starved.root.jobs
             count = 1
starved.root.mem
             count = 128
starved.users
             count = 1
total.users
             count = 1
waiting.all.cpus
             count = 6
waiting.all.jobs
             count = 1
waiting.all.mem
             count = 128
waiting.root.cpus
             count = 6
waiting.root.jobs
             count = 1
waiting.root.mem
             count = 128
```


scheduler/src/cook/mesos/api.clj
```diff
@@ -1968,6 +1969,7 @@
     (instance? ServerSocket v) (str v)
     (instance? Var v) (str v)
     (instance? RiemannReporter v) (str v)
+    (instance? ConsoleReporter v) (str v)
```
shamsimam: Why is this needed? If `ConsoleReporter` is needed, then should we make a single clause covering `com.codahale.metrics.ScheduledReporter` and remove the clauses for `ConsoleReporter ` and `RiemannReporter `?

scheduler/src/cook/mesos/monitor.clj
```diff
@@ -53,32 +47,32 @@
...
   "Given a map from users to their stats, associcate a special user
    \"all\" for the sum of all users stats."
   [db stats]
   (if (seq stats)
     (->> (vals stats)
          (apply merge-with +)
          (assoc stats "all"))
-    {"all" (zipmap (conj (util/get-all-resource-types db) :count)
-                   (repeat 0.0))}))
+    {"all" {:cpus 0, :mem 0, :jobs 0}}))
```
shamsimam: Why not use `util/get-all-resource-types` to get all the resources? Isn't this list missing `:gpus`?

scheduler/src/cook/mesos/monitor.clj
```diff
@@ -89,27 +83,61 @@
...
+  [state stats state->previous-stats-atom]
+  (let [previous-stats (get @state->previous-stats-atom state)
+        previous-users (set (keys previous-stats))
+        current-users (set (keys stats))
+        users-to-clear (difference previous-users current-users)]
+    (run! (fn [user]
+            (run! (fn [[type _]]
+                    (-> [state user (name type)]
+                        counters/counter
+                        counters/clear!))
```
shamsimam: Can we delete the counter instead of just clearing it?

scheduler/test/cook/test/mesos/monitor.clj
```diff
@@ -0,0 +1,194 @@
...
+    (is (= 0 (counter ["starved" "all" "mem"])))
+    (is (= 0 (counter ["starved" "alice" "jobs"])))
+    (is (= 0 (counter ["starved" "alice" "cpus"])))
+    (is (= 0 (counter ["starved" "alice" "mem"])))
+    (is (= 0 (counter ["starved" "bob" "jobs"])))
+    (is (= 0 (counter ["starved" "bob" "cpus"])))
+    (is (= 0 (counter ["starved" "bob" "mem"])))
+    (is (= 0 (counter ["total" "users"])))
+    (is (= 0 (counter ["starved" "users"])))
+    (is (= 0 (counter ["hungry" "users"])))
```
shamsimam: Can we add a case where `hungry-users` is non-zero and then transitions to zero after the user's task starts running?

scheduler/src/cook/mesos.clj
```diff
@@ -241,8 +239,7 @@
                                     (mesomatic.scheduler/start! driver)
                                     (reset! current-driver driver)
 
-                                    (when riemann-host
-                                      (cook.mesos.monitor/riemann-reporter mesos-datomic-conn :riemann-host riemann-host :riemann-port riemann-port))
+                                    (cook.mesos.monitor/start-collecting-stats mesos-datomic-conn)
```
shamsimam: Should we make the interval configurable?

scheduler/src/cook/mesos/monitor.clj
```diff
@@ -53,32 +47,32 @@
...
   (if (seq stats)
     (->> (vals stats)
          (apply merge-with +)
          (assoc stats "all"))
-    {"all" (zipmap (conj (util/get-all-resource-types db) :count)
-                   (repeat 0.0))}))
+    {"all" {:cpus 0, :mem 0, :jobs 0}}))
 
-(defn get-starved-job-stats
+(defn- get-starved-job-stats
```
shamsimam: Should we run this and see if there are any effects on performance due to the additional datomic queries via `get-share`?

integration/tests/cook/test_basic.py
```diff
@@ -1374,6 +1374,7 @@ def test_unscheduled_jobs_partial(self):
         finally:
             util.kill_jobs(self.cook_url, [job_uuid_1])
 
+    @pytest.mark.xfail(reason='Sometimes fails on Travis')
```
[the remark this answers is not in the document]
dposada: I just created issue #713

scheduler/src/cook/mesos/monitor.clj
dposada: The `:gpus` were already not getting tallied up due to this `(select-keys [:cpus :mem])`:

https://github.com/twosigma/Cook/pull/710/files#diff-09dd7646946b40f7cce49e291e1897cfL55

We could add `:gpus`, but I would vote to make that a separate enhancement PR rather than include the change here.

shamsimam: sgtm

scheduler/src/cook/mesos.clj
dposada: Done

scheduler/src/cook/mesos/api.clj
dposada: This is needed because otherwise calls to `/settings` fail when the console reporter is turned on.

Combined to use `ScheduledReporter`.

scheduler/src/cook/mesos/monitor.clj
dposada: Done

scheduler/test/cook/test/mesos/monitor.clj
dposada: Done

scheduler/src/cook/mesos/monitor.clj
dposada: Thanks for pointing this out. I changed this to do a single datomic query using `get-shares` for all of the waiting users, instead of doing a query for each individual waiting user

shamsimam:
@dposada let me know when this ready for another round of review.

dposada:
@shamsimam it's ready now, thanks

dposada:

dposada:
@pschorf Have you had a chance to review this?

pschorf:
Is it possible to remove the Riemann dependencies from project.clj?

dposada:
Can we merge this?

PR #712, opened 2018-01-30
Lift start-up-time out of the /info handler function

### LINKED ISSUE #711: /info endpoint "start-time" value is wrong
The `start-time` value retrieved from the `/info` endpoint is the time of the request rather than the time the server started up. This appears to be a weird "feature" of Liberator/Compojure, where the endpoint specification is re-evaluated on each request: https://github.com/weavejester/compojure/issues/148

DaoWen:

Lift the "start-time" timestamp out of the handler-construction function.

The timestamp was being re-evaluated to `now` on each request.

Fixes #711.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1141,6 +1141,7 @@
       (-> instance
           (assoc :job (select-keys job [:uuid :name :status :state :user]))))))
 
+;; We store the start-up time (ISO-8601) for reporting on the /info endpoint
 (def start-up-time (tf/unparse iso-8601-format (t/now)))
 
 (defn cook-info-handler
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -1141,13 +1141,14 @@
...
 
+(def start-up-time (tf/unparse iso-8601-format (t/now)))
+
 (defn cook-info-handler
   "Handler for the /info endpoint"
   [settings]
   (let [auth-middleware (:authorization-middleware settings)
-        auth-scheme (str (or (-> auth-middleware meta :json-value) auth-middleware))
+        auth-scheme (str (or (-> auth-middleware meta :json-value) auth-middleware))]
         ;; We store the start-up time (ISO-8601) for reporting on the /info endpoint
```
dposada: The comment should move up too, no?

DaoWen: Yes, that would probably make sense...

PR #716, opened 2018-02-05
Improves error handling on startup

dposada:

- printing the full cause trace when there is an error on startup

To make startup errors (e.g. a bogus configuration) easier to identify and correct.


dposada:
Can we merge this?

PR #719, opened 2018-02-09
Converts pending jobs query from using metatransaction to commit latch

dposada:

- using `util/get-pending-job-ents` instead of using the metatransaction-filtered `db`
- adding `util/get-running-job-ents` and using it for running jobs

Running queries using the metatransaction-filtered `db` is very slow.

### DIFF
--- a/scheduler/src/cook/mesos/monitor.clj
+++ b/scheduler/src/cook/mesos/monitor.clj
@@ -21,8 +21,7 @@
             [clojure.tools.logging :as log]
             [cook.mesos.share :as share]
             [cook.mesos.util :as util]
-            [datomic.api :as d :refer (q)]
-            [metatransaction.core :refer (db)]
+            [datomic.api :as d :refer (db, q)]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/monitor.clj
```diff
@@ -21,8 +21,7 @@
             [clojure.tools.logging :as log]
             [cook.mesos.share :as share]
             [cook.mesos.util :as util]
-            [datomic.api :as d :refer (q)]
-            [metatransaction.core :refer (db)]
+            [datomic.api :as d :refer (db, q)]
```
wyegelwel: Nitpick, please don't refer db since it is overloaded

dposada: Done, thanks

dposada:
@wyegelwel can we merge?

wyegelwel:
Am I ok to merge?

PR #720, opened 2018-02-09
CS: Ensures the task filters for the sandbox syncer occur on a single thread

shamsimam:

- Move the logic to query datomic for presence of sandbox directories into a single thread

On large clusters, the sandbox syncer can spin up many threads (in the order of hundreds) that trigger datomic queries. Datomic has a read concurrency setting which causes these queries to start becoming a performance concern if many threads make read queries that end up hitting backing storage. As a result, the sandbox syncer starts performing poorly as well as affecting the performance of other read queries in the cook scheduler. With the single reader thread version for the syncer, we can minimize the effect we have on datomic read performance.

## Relevant Links

[Datomic Peer properties](https://docs.datomic.com/on-prem/system-properties.html#peer-properties)

### DIFF
--- a/scheduler/src/cook/mesos/sandbox.clj
+++ b/scheduler/src/cook/mesos/sandbox.clj
@@ -38,8 +38,9 @@
 (def sandbox-updater-tx-rate (meters/meter ["cook-mesos" "scheduler" "sandbox-updater-tx-rate"]))
 (def sandbox-updater-unfiltered-count (counters/counter ["cook-mesos" "scheduler" "sandbox-updater-unfiltered-count"]))
 
-(defn- threadsafe-counter-reset
-  "Resets the value of the counter to the provided value."
+(defn- counter-reset
+  "Resets the value of the counter to the provided value.
+   Note: This function is not thread safe, it can introduce data races when called in parallel."
   [the-counter new-value]
   (counters/clear! the-counter)
   (when-not (zero? new-value)
@@ -55,7 +56,7 @@
    Since we expect instances to have the same sandbox, we do not check for value equality."
   [task-id->sandbox published-task-id->sandbox]
   (let [task-id->sandbox' (apply dissoc task-id->sandbox (keys published-task-id->sandbox))]
-    (threadsafe-counter-reset sandbox-aggregator-pending-count (count task-id->sandbox'))
+    (counter-reset sandbox-aggregator-pending-count (count task-id->sandbox'))
     task-id->sandbox'))
 
 (defn aggregate-sandbox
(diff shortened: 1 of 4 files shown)


dposada:
Could you link to the Datomic documentation on this topic?

shamsimam:
@dposada added a link to the datomic peer properties. Unfortunately, I couldn't find any other document discussing this issue. 

I have found with experimentation that triggering parallel queries that are greater than 2X the read concurrency limit cause the queries to slow down dramatically.

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -36,6 +36,14 @@
 (def sandbox-updater-publish-rate (meters/meter ["cook-mesos" "scheduler" "sandbox-updater-publish-rate"]))
 (def sandbox-updater-tx-duration (timers/timer ["cook-mesos" "scheduler" "sandbox-updater-tx-duration"]))
 (def sandbox-updater-tx-rate (meters/meter ["cook-mesos" "scheduler" "sandbox-updater-tx-rate"]))
+(def sandbox-updater-unfiltered-count (counters/counter ["cook-mesos" "scheduler" "sandbox-updater-unfiltered-count"]))
+
+(defn- threadsafe-counter-reset
```
dposada: Why is the word `threadsafe` in the function name?

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -265,29 +265,87 @@
...
               :else
               (recur remaining-hostnames pending-sync-agent-send-performed))))))
-    {:error-handler (fn sandbox-publisher-error-handler [ex]
+    {:error-handler (fn start-host-sandbox-syncer-error-handler [ex]
                       (log/error ex "Sync of sandbox directories on pending hosts failed"))}))
 
+(defn- remove-task-ids-with-sandbox
+  "Filters the task-ids which have a sandbox directory from task-id->sandbox."
+  [datomic-conn task-id->sandbox]
+  (letfn [(retrieve-task-ids-with-sandbox [datomic-db task-ids]
```
dposada: I thought we decided to favor `let` over `letfn`

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -265,29 +265,87 @@
...
+            (dissoc! task-id->sandbox task-id))
+          (update-filter-entries-histogram! [task-id->sandbox]
+            (histograms/update! sandbox-updater-filter-entries (count task-id->sandbox))
+            task-id->sandbox)]
+    (timers/time!
+      sandbox-updater-filter-duration
+      (->> (keys task-id->sandbox)
+           (retrieve-task-ids-with-sandbox (d/db datomic-conn))
+           (reduce remove-task-id-with-sandbox (transient task-id->sandbox))
+           persistent!
```
dposada: Why do we need `transient`/`persistent!` here?

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -149,44 +156,37 @@
...
+           (let [result-map (apply dissoc unprocessed-task-id->sandbox-in-atom processed-task-ids)]
+             (threadsafe-counter-reset sandbox-updater-unfiltered-count (count result-map))
+             result-map))))
 
 (defn refresh-agent-cache-entry
   "If the entry for the specified agent is not cached:
    - Triggers building an indexed version of all task-id to sandbox directory for the specified agent;
    - After the indexed version is built, the tasks which have sandbox directories are removed;
    - The filtered version of tasks without sandbox directories is then synced into the task-id->sandbox-agent.
    The function call is a no-op if the specified agent already exists in the cache."
```
dposada: I think this docstring is now misleading, correct?

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Removed, I intended to name it thread unsafe :( 
I updated the docstring to reflect this unsafe behavior.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: I undid the move of this function so that it doesn't appear in the diff. We can address this when we actually change the function.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: We used transient here as we can end up whittling down the map from a size of hundreds or thousands to tens.

scheduler/src/cook/mesos/sandbox.clj
shamsimam: Updated the docstring.

shamsimam:
@dposada ready for next round of feedback

shamsimam:
Reopening in the hopes of triggering Travis build.

dposada:
The code LGTM, have we confirmed that this helps overall performance?

shamsimam:
@dposada I am measuring performance internally.

shamsimam:
@dposada I measured the performance under load and the sandbox syncer, as well as other components (e.g. ranker, handle status and framework messages), continued to perform fine.

PR #723, opened 2018-02-13
CS: introduces configurable limit before cook executor is no longer used

shamsimam:
…to launch instances for a given job

- A configurable limit before cook executor is no longer used to run instances

As Cook Executor gains in stability and popularity, users no longer want the Mesos Executor to run their jobs. With this increased limit we will try to run jobs using Cook Executor and only fallback to the Mesos executor if we fail multiple times while trying to run the job using Cook executor.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -71,18 +71,29 @@ class CookTest(unittest.TestCase):
                             lambda job: len(job['instances']) > 4)
         except BaseException as e:
             self.logger.debug("Didn't reach desired instance count: {}".format(e))
+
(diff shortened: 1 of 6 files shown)


dposada:
Maybe we should never fall back to the mesos executor. As users start to depend on the features provided by the cook executor, maybe not having those features is as bad as the job failing.

shamsimam:
We spoke offline. The plan is to release with this knob set at 5. If we see there are very few tasks using the Mesos executor, eventually we will remove support for this knob (in a follow-up PR) and run all jobs with the Cook Executor.

shamsimam:
@dposada removed `retry-limit` from the `configuration.adoc` documentation.

PR #724, opened 2018-02-14
CS: ensures scheduling continues even if we are unable to convert a job entity into a map

shamsimam:

- Error handling in the `job-ent->map` function
- Usages of `job-ent->map` now handle nil return values

If `job-ent->map` fails it can end up preventing jobs from being scheduled as the ranker has not ranked all queued jobs.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -1025,24 +1025,28 @@
     (when (seq running-tasks)
       (log/info "Preparing to reconcile" (count running-tasks) "tasks")
       ;; TODO: When turning on periodic reconcilation, probably want to move this to startup
-      (doseq [[task-id] running-tasks
-              :let [task-ent (d/entity db [:instance/task-id task-id])
-                    hostname (:instance/hostname task-ent)
-                    job (util/job-ent->map (:job/_instance task-ent))
-                    task-request (make-task-request db job :task-id task-id)]]
-        ;; Need to lock on fenzo when accessing taskAssigner because taskAssigner and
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1025,24 +1025,28 @@
...
+                                       ;; Need to lock on fenzo when accessing taskAssigner because taskAssigner and
+                                       ;; scheduleOnce can not be called at the same time.
+                                       (locking fenzo
+                                         (.. fenzo
+                                             (getTaskAssigner)
+                                             (call task-request hostname)))
+                                       [task-id])))
+                                 (remove nil?))]
+        (when (not= (count running-tasks) (count processed-tasks))
+          (log/info "Skipping reconciling" (- (count running-tasks) (count processed-tasks)) "tasks"))
```
dposada: Maybe this should be `log/error`?

shamsimam: We already logged the error in `job-ent->map`

dposada: Wouldn't we want this line to show up in the ERROR-only log?

shamsimam: @pschorf tie-breaker?

pschorf: I'd say it's an error

shamsimam: Done.

shamsimam:
@pschorf ready for another round of review.

PR #727, opened 2018-02-17
CS: runs the sandbox syncer without the need for datomic reads for task filtering

shamsimam:

- Avoid using datomic queries to filter entries received from the host
- Instead 'remember' the tasks which requested sandbox directory lookup and use them in the filter

We want to avoid making datomic queries where we can to avoid introducing performance issues.

### DIFF
--- a/scheduler/src/cook/mesos/sandbox.clj
+++ b/scheduler/src/cook/mesos/sandbox.clj
@@ -19,6 +19,8 @@
             [clj-time.core :as time]
             [clj-time.periodic :as periodic]
             [clojure.core.cache :as cache]
+            [clojure.set :as set]
+            [clojure.string :as str]
             [clojure.tools.logging :as log]
             [datomic.api :as d]
             [metrics.counters :as counters]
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -158,78 +151,93 @@
...
+                   (assoc unprocessed-host->task-ids-in-atom host unprocessed-task-ids')
+                   (dissoc unprocessed-host->task-ids-in-atom host)))
+               unprocessed-host->task-ids-in-atom))))
+  (histograms/update! sandbox-updater-unprocessed-entries (counters/value sandbox-updater-unprocessed-count)))
+
+(defn- process-task-id->sandbox-directory-on-host
+  "Processes the result of the indexed task-id->sandbox-directory on a host.
+   First, the tasks which have pending sandbox directory requests are filtered based on existing requests.
+   The filtered version of tasks without sandbox directories is then synced into the task-id->sandbox-agent."
+  [{:keys [task-id->sandbox-agent unprocessed-host->task-ids-atom]} host unprocessed-task-ids-prev task-id->sandbox-directory]
```
pschorf: What is the `unprocessed-task-ids-prev` for? I don't think I understand the logic here.

shamsimam: I updated `refresh-agent-cache-entry` to further handle this possibility by ensuring the sync of the host is triggered when this scnerio occurs.

scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -158,78 +151,96 @@
...
-             result-map))))
+  "Removes the processed-task-ids from the unprocessed-host->task-ids-atom for a given host."
+  [unprocessed-host->task-ids-atom host task-ids]
+  (swap! unprocessed-host->task-ids-atom
+         (fn [unprocessed-host->task-ids-in-atom]
+           (let [unprocessed-task-ids (get unprocessed-host->task-ids-in-atom host #{})
+                 unprocessed-task-ids' (set/difference unprocessed-task-ids task-ids)]
+             (if (not= unprocessed-task-ids unprocessed-task-ids')
+               (do
+                 (->> (- (count unprocessed-task-ids') (count unprocessed-task-ids))
```
pschorf: Nitpick: I don't think the `->>` is really helping much here.

PR #729, opened 2018-02-20
Support reserving hosts from the rebalancer

pschorf:
- Introduces `rebalancer-reservation-atom` to reserve hosts for jobs which preempted several other jobs
- Reserves hosts after each rebalance cycle
- Releases reservations when all jobs reserving the host have been scheduled

It's possible when preempting multiple tasks to get a new offer which only represents part of the total resources prevented. Then, smaller, lower-priority jobs could be scheduled on this offer which delays scheduling for the actual job the rebalancer was trying to launch.

### DIFF
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -202,6 +202,7 @@
         datomic-report-chan (async/chan (async/sliding-buffer 4096))
         mesos-heartbeat-chan (async/chan (async/buffer 4096))
         current-driver (atom nil)
+        rebalancer-reservation-atom (atom {})
         leader-selector (LeaderSelector.
                           curator-framework
                           zk-prefix
@@ -216,26 +217,27 @@
                                 (try
                                   (let [{:keys [scheduler view-incubating-offers]}
                                         (sched/create-datomic-scheduler
-                                          mesos-datomic-conn
-                                          current-driver
-                                          mesos-pending-jobs-atom
-                                          offer-cache
-                                          mesos-heartbeat-chan
-                                          offer-incubate-time-ms
-                                          mea-culpa-failure-limit
-                                          fenzo-max-jobs-considered
-                                          fenzo-scaleback
-                                          fenzo-floor-iterations-before-warn
-                                          fenzo-floor-iterations-before-reset
-                                          fenzo-fitness-calculator
-                                          task-constraints
-                                          gpu-enabled?
-                                          good-enough-fitness
-                                          framework-id
-                                          sandbox-syncer-state
-                                          executor-config
-                                          progress-config
-                                          trigger-chans)
+                                         {:conn mesos-datomic-conn
+                                          :driver-atom current-driver
+                                          :pending-jobs-atom mesos-pending-jobs-atom
+                                          :offer-cache offer-cache
+                                          :heartbeat-ch mesos-heartbeat-chan
+                                          :offer-incubate-time-ms offer-incubate-time-ms
+                                          :mea-culpa-failure-limit mea-culpa-failure-limit
+                                          :fenzo-max-jobs-considered fenzo-max-jobs-considered
+                                          :fenzo-scaleback fenzo-scaleback
+                                          :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
+                                          :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
+                                          :fenzo-fitness-calculator fenzo-fitness-calculator
+                                          :task-constraints task-constraints
+                                          :gpu-enabled? gpu-enabled?
+                                          :good-enough-fitness good-enough-fitness
+                                          :framework-id framework-id
+                                          :sandbox-syncer-state sandbox-syncer-state
+                                          :executor-config executor-config
+                                          :progress-config progress-config
+                                          :trigger-chans trigger-chans
+                                          :rebalancer-reservation-atom rebalancer-reservation-atom})
                                         driver (make-mesos-driver-fn scheduler framework-id)]
                                     (mesomatic.scheduler/start! driver)
                                     (reset! current-driver driver)
@@ -253,6 +255,7 @@
                                                                               :get-mesos-utilization get-mesos-utilization
                                                                               :offer-cache offer-cache
                                                                               :pending-jobs-atom mesos-pending-jobs-atom
+                                                                              :rebalancer-reservation-atom rebalancer-reservation-atom
                                                                               :trigger-chan rebalancer-trigger-chan
                                                                               :view-incubating-offers view-incubating-offers})
                                     (counters/inc! mesos-leader)
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -219,25 +219,25 @@
                                         (sched/create-datomic-scheduler
                                          {:conn mesos-datomic-conn
                                           :driver-atom current-driver
-                                          :pending-jobs-atom mesos-pending-jobs-atom
-                                          :offer-cache offer-cache
-                                          :heartbeat-ch mesos-heartbeat-chan
-                                          :offer-incubate-time-ms offer-incubate-time-ms
-                                          :mea-culpa-failure-limit mea-culpa-failure-limit
+                                          :executor-config executor-config
+                                          :fenzo-fitness-calculator fenzo-fitness-calculator
+                                          :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
+                                          :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
                                           :fenzo-max-jobs-considered fenzo-max-jobs-considered
                                           :fenzo-scaleback fenzo-scaleback
-                                          :fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-warn
-                                          :fenzo-floor-iterations-before-reset fenzo-floor-iterations-before-reset
-                                          :fenzo-fitness-calculator fenzo-fitness-calculator
-                                          :task-constraints task-constraints
-                                          :gpu-enabled? gpu-enabled?
-                                          :good-enough-fitness good-enough-fitness
                                           :framework-id framework-id
-                                          :sandbox-syncer-state sandbox-syncer-state
-                                          :executor-config executor-config
+                                          :good-enough-fitness good-enough-fitness
+                                          :gpu-enabled? gpu-enabled?
+                                          :heartbeat-ch mesos-heartbeat-chan
+                                          :mea-culpa-failure-limit mea-culpa-failure-limit
+                                          :offer-cache offer-cache
+                                          :offer-incubate-time-ms offer-incubate-time-ms
+                                          :pending-jobs-atom mesos-pending-jobs-atom
                                           :progress-config progress-config
-                                          :trigger-chans trigger-chans
-                                          :rebalancer-reservation-atom rebalancer-reservation-atom})
+                                          :rebalancer-reservation-atom rebalancer-reservation-atom
+                                          :sandbox-syncer-state sandbox-syncer-state
+                                          :task-constraints task-constraints
+                                          :trigger-chans trigger-chans})
                                         driver (make-mesos-driver-fn scheduler framework-id)]
                                     (mesomatic.scheduler/start! driver)
                                     (reset! current-driver driver)
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -187,13 +187,10 @@
 
 (defn build-rebalancer-reservation-constraint
   "Constructs a rebalancer-reservation-constraint"
-  [job job-uuid->reserved-hostname]
-  (let [reserved-hosts (-> (hash-set)
-                           (into (vals job-uuid->reserved-hostname))
-                           (disj (job-uuid->reserved-hostname (:job/uuid job))))]
-    (-> reserved-hosts
-        ->rebalancer-reservation-constraint
-        fenzoize-job-constraint)))
+  [reserved-hosts]
+  (-> reserved-hosts
+      ->rebalancer-reservation-constraint
+      fenzoize-job-constraint))
 
 (defn make-rebalancer-job-constraints
   "Returns a sequence of all job constraints for 'job', in rebalancer-compatible (rebalancer.clj)
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -121,7 +121,7 @@
     [this _ vm-attributes]
     (let [target-hostname (get vm-attributes "HOSTNAME")]
       [(not (contains? reserved-hosts target-hostname))
-       "Host was reserved for rebalancing"]))
+       "Host is reserved for rebalancing"]))
   (job-constraint-evaluate
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -121,7 +121,7 @@
     [this _ vm-attributes]
     (let [target-hostname (get vm-attributes "HOSTNAME")]
       [(not (contains? reserved-hosts target-hostname))
-       "Host is reserved for rebalancing"]))
+       "Host is temporarily reserved"]))
   (job-constraint-evaluate
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -114,6 +114,18 @@
   (let [needs-gpus? (job-needs-gpus? job)]
     (->gpu-host-constraint job needs-gpus?)))
 
+(defrecord rebalancer-reservation-constraint [reserved-hosts]
+  JobConstraint
+  (job-constraint-name [this] (get-class-name this))
+  (job-constraint-evaluate
+    [this _ vm-attributes]
+    (let [target-hostname (get vm-attributes "HOSTNAME")]
+      [(not (contains? reserved-hosts target-hostname))
+       "Host was reserved for rebalancing"]))
+  (job-constraint-evaluate
+    [this _ vm-attributes _]
+    (job-constraint-evaluate this _ vm-attributes)))
+
 (defrecord user-defined-constraint [constraints]
   JobConstraint
   (job-constraint-name [this] (get-class-name this))
@@ -172,6 +184,16 @@
   (for [constraint-constructor job-constraint-constructors
         :let [constraint (constraint-constructor job)]]
     (fenzoize-job-constraint constraint)))
+
+(defn build-rebalancer-reservation-constraint
+  "Constructs a rebalancer-reservation-constraint"
+  [job job-uuid->reserved-hostname]
+  (let [reserved-hosts (-> (hash-set)
+                           (into (vals job-uuid->reserved-hostname))
+                           (disj (job-uuid->reserved-hostname (:job/uuid job))))]
+    (-> reserved-hosts
+        ->rebalancer-reservation-constraint
+        fenzoize-job-constraint)))
 
 (defn make-rebalancer-job-constraints
   "Returns a sequence of all job constraints for 'job', in rebalancer-compatible (rebalancer.clj)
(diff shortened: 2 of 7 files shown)


scheduler/src/cook/mesos.clj
```diff
@@ -216,26 +217,27 @@
...
+                                          :fenzo-fitness-calculator fenzo-fitness-calculator
+                                          :task-constraints task-constraints
+                                          :gpu-enabled? gpu-enabled?
+                                          :good-enough-fitness good-enough-fitness
+                                          :framework-id framework-id
+                                          :sandbox-syncer-state sandbox-syncer-state
+                                          :executor-config executor-config
+                                          :progress-config progress-config
+                                          :trigger-chans trigger-chans
+                                          :rebalancer-reservation-atom rebalancer-reservation-atom})
```
shamsimam: Nitpick: lexicographic ordering of keys.

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -114,6 +114,18 @@
...
     (->gpu-host-constraint job needs-gpus?)))
 
+(defrecord rebalancer-reservation-constraint [reserved-hosts]
+  JobConstraint
+  (job-constraint-name [this] (get-class-name this))
+  (job-constraint-evaluate
+    [this _ vm-attributes]
+    (let [target-hostname (get vm-attributes "HOSTNAME")]
+      [(not (contains? reserved-hosts target-hostname))
+       "Host was reserved for rebalancing"]))
```
shamsimam: Nitpick: "Host is reserved for rebalancing`.

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -114,6 +114,18 @@
   (let [needs-gpus? (job-needs-gpus? job)]
     (->gpu-host-constraint job needs-gpus?)))
 
+(defrecord rebalancer-reservation-constraint [reserved-hosts]
```
shamsimam: Should this be `host->job-ids` so we can add which jobs the host is reserved for in the message? If so, we should change `job-uuid->reserved-host` to `reserved-host->job-uuids` in the reservation atom.

shamsimam: Missing unit tests for `rebalancer-reservation-constraint`.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
             :to-make-room-for pending-job)]
     [state nil]))
 
+(defn reserve-hosts! [state preemption-decisions]
+  (let [multiple-tasks (filter #(< 1 (count (:task %))) preemption-decisions)
```
shamsimam: Nitpick: rename to `multiple-task-decisions`

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
...
+        reservations (into {} (map (fn [d] [(:job/uuid (:to-make-room-for d))
+                                            (:hostname d)])
+                                   multiple-tasks))]
```
shamsimam: Nitpick: This  block would be easier to read for me if we used `->>` of if we had a `(map-from-seq fk fv es)` in our utils.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
             :to-make-room-for pending-job)]
     [state nil]))
 
+(defn reserve-hosts! [state preemption-decisions]
```
shamsimam: Missing unit tests, it would have helped me understand the structure of `state` and `preemption-decisions`

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
...
+  (let [multiple-tasks (filter #(< 1 (count (:task %))) preemption-decisions)
+        reservations (into {} (map (fn [d] [(:job/uuid (:to-make-room-for d))
+                                            (:hostname d)])
+                                   multiple-tasks))]
+    (swap! state (fn [{:keys [jobs-launched]}]
+                   {:reservations (apply dissoc reservations jobs-launched)
+                    :jobs-launched #{}}))))
```
shamsimam: Why is the function clearing out `jobs-launched`? Maybe we can clarify with a comment?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -503,7 +517,8 @@
 
 (defn start-rebalancer!
   [{:keys [config conn driver get-mesos-utilization pending-jobs-atom offer-cache
-           trigger-chan view-incubating-offers view-mature-offers]}]
+           trigger-chan view-incubating-offers view-mature-offers
+           rebalancer-reservation-atom]}]
```
shamsimam: Nitpick: lexicographic order

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Nitpick: missing docstring

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
             :to-make-room-for pending-job)]
     [state nil]))
 
+(defn reserve-hosts! [state preemption-decisions]
+  (let [multiple-tasks (filter #(< 1 (count (:task %))) preemption-decisions)
+        reservations (into {} (map (fn [d] [(:job/uuid (:to-make-room-for d))
+                                            (:hostname d)])
+                                   multiple-tasks))]
+    (swap! state (fn [{:keys [jobs-launched]}]
+                   {:reservations (apply dissoc reservations jobs-launched)
```
shamsimam: Nitpick, should we rename `reservations` to `job-uuid->reserved-host`

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -579,18 +579,21 @@
...
+               running-cotask-cache (atom (cache/fifo-cache-factory {} :threshold 1))
+               job-uuid->reserved-host {}}}]
+  (let [constraints (-> (constraints/make-fenzo-job-constraints job)
+                        (into
                           (remove nil?
                                   (mapv (fn make-group-constraints [group]
                                           (constraints/make-fenzo-group-constraint
                                            db group #(guuid->considerable-cotask-ids (:group/uuid group)) running-cotask-cache))
                                         (:group/_job job))))
+                        (conj (constraints/build-rebalancer-reservation-constraint job job-uuid->reserved-host)))
```
shamsimam: Should we make sure that cheaper to compute constraints are placed earlier in the list of constraints? Would it be helpful if the `rebalancer-reservation-constraint` was conj-ed earlier before the `into`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -619,10 +622,12 @@
         considerable->task-id (plumbing.core/map-from-keys (fn [_] (str (java.util.UUID/randomUUID))) considerable)
         guuid->considerable-cotask-ids (util/make-guuid->considerable-cotask-ids considerable->task-id)
         running-cotask-cache (atom (cache/fifo-cache-factory {} :threshold (max 1 (count considerable))))
+        job-uuid->reserved-host (or (:reservations @rebalancer-reservation-atom) {})
         ; Important that requests maintains the same order as considerable
         requests (mapv (fn [job]
                          (make-task-request db job :task-id (considerable->task-id job) :guuid->considerable-cotask-ids guuid->considerable-cotask-ids
-                                            :running-cotask-cache running-cotask-cache))
+                                            :running-cotask-cache running-cotask-cache
+                                            :job-uuid->reserved-host job-uuid->reserved-host))
```
shamsimam: Nitpick: lexicographic ordering of optional arguments. Also, it would be nice if we can align them.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -831,10 +836,21 @@
                 (getTaskAssigner)
                 (call task-request hostname))))))))
 
+(defn- update-host-reservations [rebalancer-reservation-atom matches]
+  (let [matched-uuids (->> matches
+                          (mapcat #(-> % :tasks))
+                          (map #(-> % .getRequest :job))
+                          (map :job/uuid)
+                          (into (hash-set)))]
```
shamsimam: Can we have multiple matches for the same job?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,22 @@
             :to-make-room-for pending-job)]
     [state nil]))
 
+(defn reserve-hosts! [state preemption-decisions]
+  (let [multiple-tasks (filter #(< 1 (count (:task %))) preemption-decisions)
+        reservations (into {} (map (fn [d] [(:job/uuid (:to-make-room-for d))
+                                            (:hostname d)])
+                                   multiple-tasks))]
+    (swap! state (fn [{:keys [jobs-launched]}]
```
shamsimam: Should this be `launched-jobs`? Also can we please rename to `launched-job-uuids`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -831,10 +836,21 @@
...
+    (swap! rebalancer-reservation-atom (fn [{:keys [launched-jobs reservations]}]
+                                         {:reservations (apply dissoc reservations matched-uuids)
+                                          :launched-jobs (into matched-uuids launched-jobs)}))))
```
shamsimam: Nitpick: lexicographic ordering.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Nitpick: lexicographic ordering.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -831,10 +836,21 @@
                 (getTaskAssigner)
                 (call task-request hostname))))))))
 
+(defn- update-host-reservations [rebalancer-reservation-atom matches]
+  (let [matched-uuids (->> matches
```
shamsimam: Nitpick: rename to `matched-job-uuids`

scheduler/src/cook/mesos/constraints.clj
shamsimam: Should we consider passing in the raw `job-uuid->reserved-hostname` here (instead of building `reserved-hosts` in `build-rebalancer-reservation-constraint`) and then being smarter in `job-constraint-evaluate` while computing the return value? It will save us having to create the `reserved-hosts` for every job.

pschorf: I don't want to add the hostname to the string because we've had issues in this code before allocating tons of string objects for failure messages which are ultimately unused.

How were you thinking of changing the implementation? I was trying to front load the cost on constructing the constraint as opposed to the matching, which could happen much more often.

Added unit test.

scheduler/src/cook/mesos/constraints.clj
pschorf: Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: If only we had a type system to tell us these kind of things...

Done

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Added a comment. These are in my opinion ordered correctly.

scheduler/src/cook/mesos/rebalancer.clj
pschorf: done

scheduler/src/cook/mesos/scheduler.clj
pschorf: Done

scheduler/src/cook/mesos/scheduler.clj
pschorf: Reordered. Not going to align them.

scheduler/src/cook/mesos/scheduler.clj
pschorf: Done

scheduler/src/cook/mesos/scheduler.clj
pschorf: No

scheduler/src/cook/mesos/scheduler.clj
pschorf: done

scheduler/src/cook/mesos/scheduler.clj
shamsimam: thanks, I meant justify, not align :)

scheduler/src/cook/mesos/constraints.clj
shamsimam: >  I was trying to front load the cost on constructing the constraint as opposed to the matching, which could happen much more often.

If we had `hostname->job-uuids`, we could use the same one for all jobs. Inside the job we would just check if the hostname entry exists and ensure the job is in the set of job-uuids for a given host.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -831,10 +838,21 @@
                 (getTaskAssigner)
                 (call task-request hostname))))))))
 
+(defn update-host-reservations! [rebalancer-reservation-atom matches]
+  (let [matched-job-uuids (->> matches
```
shamsimam: Nitpick: formatting is off.

pschorf: Done

pschorf:
I implemented an optimization where we create the set of reserved hosts once per match and then remove the single host (if present) for the job.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -619,10 +622,16 @@
...
         running-cotask-cache (atom (cache/fifo-cache-factory {} :threshold (max 1 (count considerable))))
+        job-uuid->reserved-host (or (:job-uuid->reserved-host @rebalancer-reservation-atom) {})
+        reserved-hosts (into (hash-set) (vals job-uuid->reserved-host))
         ; Important that requests maintains the same order as considerable
         requests (mapv (fn [job]
-                         (make-task-request db job :task-id (considerable->task-id job) :guuid->considerable-cotask-ids guuid->considerable-cotask-ids
-                                            :running-cotask-cache running-cotask-cache))
+                         (make-task-request db job
+                                            :guuid->considerable-cotask-ids guuid->considerable-cotask-ids
+                                            :reserved-hosts (disj reserved-hosts (job-uuid->reserved-host (:job/uuid job)))
```
shamsimam: Nitpick: lexicographic order :(

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -382,12 +382,28 @@
...
+  "Reserves all hosts in preemption-decisions which will preempt more than one task"
+  [rebalancer-reservation-atom preemption-decisions]
+  (let [multiple-task-decisions (filter #(< 1 (count (:task %))) preemption-decisions)
+        reservations (->> multiple-task-decisions
+                          (map (fn [d] [(:job/uuid (:to-make-room-for d))
+                                        (:hostname d)]))
+                          (into {}))]
+    (swap! rebalancer-reservation-atom (fn [{:keys [launched-job-uuids]}]
+                                        ; In case one of the jobs the rebalancer has launched already while computing
+                                        ; the pre-emption decisions, remove it's reservation from the map. Then
```
dposada: I don't understand the grammar here:

> In case one of the jobs the rebalancer has launched already while computing the pre-emption decisions, remove it's reservation from the map

scheduler/src/cook/mesos/constraints.clj
```diff
@@ -114,6 +114,18 @@
...
+  JobConstraint
+  (job-constraint-name [this] (get-class-name this))
+  (job-constraint-evaluate
+    [this _ vm-attributes]
+    (let [target-hostname (get vm-attributes "HOSTNAME")]
+      [(not (contains? reserved-hosts target-hostname))
+       "Host is reserved for rebalancing"]))
```
dposada: Where does this string get consumed/exposed?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -831,10 +840,21 @@
                 (getTaskAssigner)
                 (call task-request hostname))))))))
 
+(defn update-host-reservations! [rebalancer-reservation-atom matches]
```
dposada: Would be nice to have a docstring here

scheduler/test/cook/test/mesos/rebalancer.clj
```diff
@@ -1153,5 +1158,161 @@
...
+                                  (util/get-pending-job-ents db)
+                                  {"hostA" {:cpus 0.0 :mem 0.0 :gpus 0.0}}
+                                  reservations
+                                  params)]
+        (is (= "hostA" hostname))
+        (is (= job2 (:db/id to-make-room-for)))
+        (is (= [task1] (map :db/id task)))
+        (is (= {} (:job-uuid->reserved-host @reservations)))
+        (is (= #{} (:launched-job-uuids @reservations))))))
+  (testing "does not reserve host for job already launched"
```
dposada: Nitpick: I would favor a blank line before each `testing` form

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1473,20 +1472,30 @@
...
+                                            job-3 (d/entity test-db (create-dummy-job conn :uuid (get-uuid "job-3")
+                                                                                      :group group-ent-id :name "job-3" :ncpus 7 :memory 4096))
+                                            job-4 (d/entity test-db (create-dummy-job conn :uuid (get-uuid "job-4")
+                                                                                      :group group-ent-id :name "job-4" :ncpus 11 :memory 1024))
+                                            job-5 (d/entity test-db (create-dummy-job conn :uuid (get-uuid "job-5")
+                                                                                      :group group-ent-id :name "job-5" :ncpus 5 :memory 2048 :gpus 2))
+                                            job-6 (d/entity test-db (create-dummy-job conn :uuid (get-uuid "job-6")
+                                                                                      :group group-ent-id :name "job-6" :ncpus 19 :memory 1024 :gpus 4))
```
dposada: Nitpick: I would favor one key/value pair per line

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Me neither. Fixed.

scheduler/src/cook/mesos/scheduler.clj
pschorf: Fixed

scheduler/src/cook/mesos/constraints.clj
pschorf: This will be recorded and shown on the `unscheduled_jobs` endpoint if triggered by a user.

scheduler/src/cook/mesos/scheduler.clj
pschorf: Done

scheduler/test/cook/test/mesos/rebalancer.clj
pschorf: Done

scheduler/test/cook/test/mesos/scheduler.clj
pschorf: Done

scheduler/src/cook/mesos/constraints.clj
dposada: Are there cases where I would call `unscheduled_jobs` and see a bunch of "Host is reserved for rebalancing" messages? If so, that seems like it would be confusing

pschorf: These would be aggregated by host, so you'd see something like:
`[{:reason \"Not enough CPU available\" :host_count 2}
   {:reason \"Not enough Memory available\" :host_count 3}
   {:reason \"Job already ran on this host.\" :host_count 4}
   {:reason \"Host is reserved for rebalancing\" :host_count 4}]`

I'm happy to change the message if you can think of something better

dposada: Ah, got it, that makes sense.

PR #730, opened 2018-02-20
Outgoing/quicky opt

scrosby:

- Add a timer
- Remove a list membership that only has one element.
-


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -707,13 +709,15 @@
 (defn generate-user-usage-map
   "Returns a mapping from user to usage stats"
   [unfiltered-db]
-  (->> (util/get-running-task-ents unfiltered-db)
+      (timers/time!
```
pschorf: Indentation looks off here

pschorf:
Can you fill out the PR template?

dposada:
@scrosby Can you fill out the "Why are we making these changes?" section?

PR #731, opened 2018-02-20
Added resource and job category caching.

scrosby:

- Cache job resources, task->user mapping, and job categories.
- 
- 

Improves performance of various scheduler tasks.

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -42,7 +42,7 @@
 
 (defn lookup-cache!
   "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
-  any misses. Caches only postive hits where both functions return non-nil"
+  any misses. Caches only positive hits where both functions return non-nil"
   [^Cache cache extract-key-fn miss-fn item]
   (if-let [key (extract-key-fn item)]
       (if-let [result (.getIfPresent cache key)]
@@ -56,7 +56,7 @@
 
 (defn lookup-cache-datomic-entity!
   "Specialized function for caching where datomic entities are the key.
-  Extracts :db/id so that we don't keep the entity alive in the cache"
+  Extracts :db/id so that we don't keep the entity alive in the cache."
   [cache miss-fn entity]
   (lookup-cache! cache :db/id miss-fn entity))
 
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -24,7 +24,46 @@
             [metatransaction.core :refer (db)]
             [metrics.timers :as timers]
             [plumbing.core :as pc :refer (map-vals map-keys)])
-  (:import [java.util Date]))
+  (:import
+    [com.google.common.cache Cache CacheBuilder]
+    [java.util.concurrent TimeUnit]
+    [java.util Date]))
+
+
+(defn new-cache []
+  "Build a new cache"
+  (-> (CacheBuilder/newBuilder)
+      (.maximumSize 1000000)
+      ;; if its not been accessed in 2 hours, whatever is going on, its not being visted by the
+      ;; scheduler loop anymore. E.g., its probably failed/done and won't be needed. So,
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache!
+  "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+  any misses. Caches only postive hits where both functions return non-nil"
+  [^Cache cache extract-key-fn miss-fn item]
+  (if-let [key (extract-key-fn item)]
+      (if-let [result (.getIfPresent cache key)]
+        result ; we got a hit.
+        (let [new-result (miss-fn item)]
+          ; Only cache non-nil
+          (when new-result
+            (.put cache key new-result))
+          new-result))
+      (miss-fn item)))
+
+(defn lookup-cache-datomic-entity!
+  "Specialized function for caching where datomic entities are the key.
+  Extracts :db/id so that we don't keep the entity alive in the cache"
+  [cache miss-fn entity]
+  (lookup-cache! cache :db/id miss-fn entity))
+
+(defonce ^Cache job-ent->resources-cache (new-cache))
+(defonce ^Cache categorize-job-cache (new-cache))
+(defonce ^Cache task-ent->user-cache (new-cache))
+(defonce ^Cache task->feature-vector-cache (new-cache))
 
 (defn get-all-resource-types
   "Return a list of all supported resources types. Example, :cpus :mem :gpus ..."
@@ -36,13 +75,16 @@
        (map first)))
 
 (defn categorize-job
-  "Return the category of the job. Currently jobs can be :normal or :gpu. This
-   is used to give separate queues for scarce & non-scarce resources"
-  [job]
-  (let [resources (:job/resource job)]
-    (if (some #(= :resource.type/gpus (:resource/type %)) resources)
-      :gpu
-      :normal)))
+      "Return the category of the job. Currently jobs can be :normal or :gpu. This
+       is used to give separate queues for scarce & non-scarce resources"
+      [job]
+      (let [categorize-job-miss
+            (fn [job]
+                (let [resources (:job/resource job)]
+                     (if (some #(= :resource.type/gpus (:resource/type %)) resources)
+                       :gpu
+                       :normal)))]
+           (lookup-cache-datomic-entity! categorize-job-cache categorize-job-miss job)))
 
 (defn without-ns
   [k]
@@ -173,19 +215,22 @@
           (:job/label job-ent)))
 
 (defn job-ent->resources
-  "Take a job entity and return a resource map. NOTE: the keys must be same as mesos resource keys"
-  [job-ent]
-  (reduce (fn [m r]
-            (let [resource (keyword (name (:resource/type r)))]
-              (condp contains? resource
-                #{:cpus :mem :gpus} (assoc m resource (:resource/amount r))
-                #{:uri} (update-in m [:uris] (fnil conj [])
-                                   {:cache (:resource.uri/cache? r false)
-                                    :executable (:resource.uri/executable? r false)
-                                    :value (:resource.uri/value r)
-                                    :extract (:resource.uri/extract? r false)}))))
-          {:ports (:job/ports job-ent 0)}
-          (:job/resource job-ent)))
+      "Take a job entity and return a resource map. NOTE: the keys must be same as mesos resource keys"
+      [job]
+      (let [job-ent->resources-miss
+              (fn [job-ent]
+                  (reduce (fn [m r]
+                              (let [resource (keyword (name (:resource/type r)))]
+                                   (condp contains? resource
+                                          #{:cpus :mem :gpus} (assoc m resource (:resource/amount r))
+                                          #{:uri} (update-in m [:uris] (fnil conj [])
+                                                             {:cache (:resource.uri/cache? r false)
+                                                              :executable (:resource.uri/executable? r false)
+                                                              :value (:resource.uri/value r)
+                                                              :extract (:resource.uri/extract? r false)}))))
+                          {:ports (:job/ports job-ent 0)}
+                          (:job/resource job-ent)))]
+           (lookup-cache-datomic-entity! job-ent->resources-cache job-ent->resources-miss job)))
 
 (defn job-ent->attempts-consumed
   "Determines the amount of attempts consumed by a job-ent."
@@ -437,23 +482,29 @@
          (when slave-id {:instance/slave-id slave-id})))
 
 (defn task-ent->user
-  [task-ent]
-  (get-in task-ent [:job/_instance :job/user]))
+      [task-ent]
+      (let [task-ent->user-miss
+              (fn [task-ent]
+                  (get-in task-ent [:job/_instance :job/user]))]
+           (lookup-cache-datomic-entity! task-ent->user-cache task-ent->user-miss task-ent)))
 
 (def ^:const default-job-priority 50)
 
 
 (defn task->feature-vector
-  [task]
-  "Vector of comparable features of a task.
-   Last two elements are aribitary tie breakers.
-   Use :db/id because they guarantee uniqueness for different entities
-   (:db/id task) is not sufficient because synthetic task entities don't have :db/id
-    This assumes there are at most one synthetic task for a job, otherwise uniqueness invariant will break"
-  [(- (:job/priority (:job/_instance task) default-job-priority))
-   (:instance/start-time task (java.util.Date. Long/MAX_VALUE))
-   (:db/id task)
-   (:db/id (:job/_instance task))])
+      "Vector of comparable features of a task.
+       Last two elements are aribitary tie breakers.
+       Use :db/id because they guarantee uniqueness for different entities
+       (:db/id task) is not sufficient because synthetic task entities don't have :db/id
+       This assumes there are at most one synthetic task for a job, otherwise uniqueness invariant will break"
+      [task]
+      (let [task->feature-vector-miss
+            (fn [task]
+                [(- (:job/priority (:job/_instance task) default-job-priority))
+                 (:instance/start-time task (java.util.Date. Long/MAX_VALUE))
+                 (:db/id task)
+                 (:db/id (:job/_instance task))])]
+           (lookup-cache-datomic-entity! task->feature-vector-cache task->feature-vector-miss task)))
 
 (defn same-user-task-comparator
   "Comparator to order same user's tasks"
--- a/scheduler/test/cook/test/mesos/api.clj
+++ b/scheduler/test/cook/test/mesos/api.clj
@@ -28,7 +28,7 @@
             [cook.mesos.reason :as reason]
             [cook.mesos.scheduler :as sched]
             [cook.mesos.util :as util]
-            [cook.test.testutil :refer [restore-fresh-database! create-dummy-job create-dummy-instance]]
+            [cook.test.testutil :refer [restore-fresh-database! flush-caches! create-dummy-job create-dummy-instance]]
             [datomic.api :as d :refer [q db]]
             [mesomatic.scheduler :as msched]
             [schema.core :as s])
@@ -1140,6 +1140,7 @@
         (is (thrown? Exception (s/validate api/Job (assoc min-job :expected-runtime 3 :max-runtime 2))))))))
 
 (deftest test-create-jobs!
+  (cook.test.testutil/flush-caches!)
   (let [expected-job-map
         (fn
           ; Converts the provided job and framework-id (framework-id) to the job-map we expect to get back from
--- a/scheduler/test/cook/test/mesos/util.clj
+++ b/scheduler/test/cook/test/mesos/util.clj
@@ -63,6 +63,24 @@
     ;; u3 has no jobs (running or otherwise)
     (is (= 0 (count (util/get-user-running-job-ents (db conn) "u3"))))))
 
+(deftest test-cache
+  (let [cache (util/new-cache)
+        extract-fn #(if (odd? %) nil %)
+        miss-fn #(if (> % 100) nil %)]
+    ;; u1 has 2 jobs running
+    (is (= 1 (util/lookup-cache! cache extract-fn miss-fn 1))) ; Should not be cached. Nil from extractor.
+    (is (= 2 (util/lookup-cache! cache extract-fn miss-fn 2))) ; Should be cached.
+    (is (= nil (.getIfPresent cache 1)))
+    (is (= 2 (.getIfPresent cache 2)))
+    (is (= nil (util/lookup-cache! cache extract-fn miss-fn 101))) ; Should not be cached. Nil from miss function
+    (is (= nil (util/lookup-cache! cache extract-fn miss-fn 102))) ; Should not be cached. Nil from miss function
+    (is (= nil (.getIfPresent cache 101)))
+    (is (= nil (.getIfPresent cache 102)))
+    (is (= 4 (util/lookup-cache! cache extract-fn miss-fn 4))) ; Should be cached.
+    (is (= 2 (.getIfPresent cache 2)))
+    (is (= 4 (.getIfPresent cache 4)))))
+
+
 (deftest test-get-pending-job-ents
   (let [uri "datomic:mem://test-get-pending-job-ents"
         conn (restore-fresh-database! uri)]
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -30,6 +30,7 @@
                  [com.rpl/specter "1.0.1"]
 
                  ;;Utility
+                 [com.google.guava/guava "16.0"]
                  [amalloy/ring-buffer "1.1"]
                  [listora/ring-congestion "0.1.2"]
                  [lonocloud/synthread "1.0.4"]
--- a/scheduler/test/cook/test/mesos/scheduler.clj
+++ b/scheduler/test/cook/test/mesos/scheduler.clj
@@ -319,6 +319,7 @@
   (def t2-1 (d/resolve-tempid db-after tempids t2-1)))
 
 (deftest test-extract-job-resources
+  (cook.test.testutil/flush-caches!)
   (let [resources (util/job-ent->resources (d/entity (db c) j1))]
     (is (= 1.0 (:cpus resources)))
     (is (= 1000.0 (:mem resources)))))
--- a/scheduler/test/cook/test/testutil.clj
+++ b/scheduler/test/cook/test/testutil.clj
@@ -23,6 +23,7 @@
             [cook.impersonation :refer (create-impersonation-middleware)]
             [cook.mesos.api :as api]
             [cook.mesos.schema :as schema]
+            [cook.mesos.util :as util]
             [datomic.api :as d :refer (q db)]
             [plumbing.core :refer [mapply]]
             [qbits.jet.server :refer (run-jetty)]
@@ -63,12 +64,21 @@
        (finally
          (async/close! exit-chan#)))))
 
+(defn flush-caches!
+  "Flush the caches. Needed in unit tests. We centralize initialization by using it to initialize the caches too."
+  []
+  (.invalidateAll util/job-ent->resources-cache)
+  (.invalidateAll util/categorize-job-cache)
+  (.invalidateAll util/task-ent->user-cache)
+  (.invalidateAll util/task->feature-vector-cache))
+
 (defn restore-fresh-database!
   "Completely delete all data, start a fresh database and apply transactions if
    provided.
 
    Return a connection to the fresh database."
   [uri & txn]
+  (flush-caches!)
   (d/delete-database uri)
   (d/create-database uri)
   (let [conn (d/connect uri)]


scheduler/src/cook/mesos/util.clj
```diff
@@ -26,6 +26,44 @@
...
+      we don't keep the entity alive in the cache"
+      [cache miss-fn entity]
+      (lookup-cache-generic! cache :db/id miss-fn entity))
+
+; Fake. Are actually initialized in flush-caches
+(defonce job-resources-cache (atom nil))
+(defonce job-category-cache (atom nil))
+(defonce task-user-cache (atom nil))
+
+(defn flush-caches
```
pschorf: Can this move to the unit test file?

scheduler/src/cook/mesos/util.clj
```diff
@@ -26,6 +26,44 @@
...
+             (miss-fn item))))
+
+(defn lookup-cache-datomic-entity!
+      "Specialized function for caching where datomic entities are the key. Extracts :db/id so that
+      we don't keep the entity alive in the cache"
+      [cache miss-fn entity]
+      (lookup-cache-generic! cache :db/id miss-fn entity))
+
+; Fake. Are actually initialized in flush-caches
+(defonce job-resources-cache (atom nil))
```
pschorf: Can we initialize the caches here? It doesn't seem like it would be any faster to `def` them with nil and call `flush-caches` with 1,000,000 while loading the file.

scheduler/src/cook/mesos/util.clj
```diff
@@ -35,7 +73,7 @@
           db)
        (map first)))
 
-(defn categorize-job
+(defn- categorize-job-miss
```
pschorf: Move this into a `let-fn` inside categorize-job.

scheduler/src/cook/mesos/util.clj
```diff
@@ -172,7 +216,7 @@
           {}
           (:job/label job-ent)))
 
-(defn job-ent->resources
+(defn job-ent->resources-miss
```
pschorf: Move this into a `let-fn` inside `job-ent->resources`

scheduler/src/cook/mesos/util.clj
```diff
@@ -436,10 +484,14 @@
          (when hostname {:instance/hostname hostname})
          (when slave-id {:instance/slave-id slave-id})))
 
-(defn task-ent->user
+(defn task-ent->user-miss
```
pschorf: Move this to a `let-fn` inside `task-ent->user`

scheduler/src/cook/mesos/util.clj
```diff
@@ -26,6 +26,44 @@
             [plumbing.core :as pc :refer (map-vals map-keys)])
   (:import [java.util Date]))
 
+(defn lookup-cache-generic!
```
pschorf: Are there any uses for this besides `lookup-cache-datomic-entity!`? If so, I'd rather just have the one function (which can be private)

scrosby: Right now, no. But its not unlikely I may add a cache thats not based on lookup-cache-datomic-entity!

pschorf: Given our conversation yesterday, that sounds good to me. This could also be an interface for a guava cache as well, so it's fine with me to leave it.

scheduler/src/cook/mesos/util.clj
dposada: I thought we favor `let` over `let-fn`

scheduler/src/cook/mesos/util.clj
```diff
@@ -26,6 +26,44 @@
...
+      "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+      any misses. Caches only postive hits where both functions return non-nil"
+      [cache extract-key-fn miss-fn item]
+      (let [key (extract-key-fn item)]
+           (if key
```
dposada: Can we use `if-let` here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -26,6 +26,44 @@
...
+             (cache/lookup (swap! cache
+                                  #(if (cache/has? % key)
+                                     (cache/hit % key)
+                                     (let [result (miss-fn item)]
+                                          ; Only cache non-nil
+                                          (if result
```
dposada: Can we use `if-let` here?

scheduler/src/cook/mesos/util.clj
dposada: Do we need `generic` in the name? Maybe `cache-lookup!` is a good name

scheduler/src/cook/mesos/util.clj
```diff
@@ -35,7 +73,7 @@
           db)
        (map first)))
 
-(defn categorize-job
```
dposada: Was this being used anywhere?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,44 +24,58 @@
             [metatransaction.core :refer (db)]
             [metrics.timers :as timers]
             [plumbing.core :as pc :refer (map-vals map-keys)])
-  (:import [java.util Date]))
+  (:import [java.util Date]
+           [com.google.common.cache CacheBuilder]
+           [java.util.concurrent TimeUnit]))
```
dposada: please lexicographically sort the `:import`s

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,44 +24,58 @@
...
+
+
+(defn new-cache []
+  "Build a new cache"
+  (-> (CacheBuilder/newBuilder)
+      (.maximumSize 1000000)
+      ;; if its not been accessed in 2 hours, whatever is going on, its not being visted by the
+      ;; scheduler loop anymore. E.g., its probably failed/done and won't be needed. So,
+      ;; lets kick it out to keep cache small.
```
dposada: Maybe we should put this info in the docstring?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,44 +24,58 @@
...
-                           key)
-             (miss-fn item))))
+  "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+  any misses. Caches only postive hits where both functions return non-nil"
+  [cache-atom extract-key-fn miss-fn item]
+  (let [key (extract-key-fn item)
+        cache (deref cache-atom)]
+    (if key
+      (let [result (.getIfPresent cache key)]
+        (if result
```
dposada: Can we use `if-let` here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,44 +24,58 @@
...
+          result ; we got a hit.
```
dposada: In my opinion, `; we got a hit.` is not too useful

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,44 +24,58 @@
...
+          (let [new-result (miss-fn item)]
+            ; Only cache non-nil
+            (if new-result
```
dposada: Can we use `if-let` here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -73,20 +87,17 @@
...
-      "Return the category of the job. Currently jobs can be :normal or :gpu. This
-       is used to give separate queues for scarce & non-scarce resources"
-      [job]
-      (lookup-cache-datomic-entity! job-category-cache categorize-job-miss job))
+  (letfn [(categorize-job-miss
+            [job]
+            (let [resources (:job/resource job)]
+              (if (some #(= :resource.type/gpus (:resource/type %)) resources)
+                :gpu
+                :normal)))]
```
dposada: We generally favor `let` over `letfn`

scheduler/src/cook/mesos/util.clj
```diff
@@ -484,28 +494,30 @@
...
          (when slave-id {:instance/slave-id slave-id})))
 
-(defn task-ent->user-miss
-  [task-ent]
-  (get-in task-ent [:job/_instance :job/user]))
-
 (defn task-ent->user
   [task-ent]
-  (lookup-cache-datomic-entity! task-user-cache task-ent->user-miss task-ent))
+  (letfn [( task-ent->user-miss
```
dposada: superfluous space in `[( task-ent->user-miss`

scheduler/src/cook/mesos/util.clj
scrosby: ACK

scheduler/src/cook/mesos/util.clj
scrosby: As discussed, we'll revisit how to thread these (and the resulting caches) in a followup JIRA. Dependency injection, globals, or something similar.

scheduler/src/cook/mesos/util.clj
scrosby: ACK

scheduler/src/cook/mesos/util.clj
scrosby: Not obviously. We want to unconditionally return new-result, so it needs to leave the scope of the 'if'.

scheduler/src/cook/mesos/util.clj
dposada: What I meant was that maybe the info in the comment should live in the docstring instead

scheduler/src/cook/mesos/util.clj
```diff
@@ -438,22 +496,28 @@
...
-   (:db/id (:job/_instance task))])
+   This assumes there are at most one synthetic task for a job, otherwise uniqueness invariant will break"
+  [task]
+  (letfn [(task->feature-vector-miss
+             [task]
+             [(- (:job/priority (:job/_instance task) default-job-priority))
+              (:instance/start-time task (java.util.Date. Long/MAX_VALUE))
+              (:db/id task)
+              (:db/id (:job/_instance task))])]
+          (lookup-cache-datomic-entity! feature-vector-cache task->feature-vector-miss task)))
```
pschorf: These might be synthetic task entities from `create-task-ent` which won't have a `:db/id` field. You could use some logic like `(or (:db/id task) (:db/id (:job/_instance task)))` to pull either the task ID if present or the job ID. Since it's a map, I don't think this will be doing any datomic reads. 

I think it's also valid as a key because that mirrors the logic in the feature vector which is being cached.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,60 @@
...
+             (miss-fn item))))
+
+(defn lookup-cache-datomic-entity!
+      "Specialized function for caching where datomic entities are the key. Extracts :db/id so that
+      we don't keep the entity alive in the cache"
+      [cache miss-fn entity]
+      (lookup-cache-generic! cache :db/id miss-fn entity))
+
+
+; Fake. Are actually initialized in flush-caches
```
dposada: Why "Fake"?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,60 @@
...
+(defonce job-resources-cache (atom nil))
+(defonce job-category-cache (atom nil))
+(defonce task-user-cache (atom nil))
+(defonce feature-vector-cache (atom nil))
+
+(defn flush-caches
+  "Flush the caches. Needed in unit tests. We centralize initialization by using it to initialize the caches too."
```
dposada: If this is only for unit tests, should we move it to the unit tests?

scheduler/src/cook/mesos/util.clj
scrosby: In flush caches, I need to reconstruct the caches. rather than have two initializers, this lets me only get one, and I am indicating this with a comment that the real initialization happens at XXXXX.

scheduler/src/cook/mesos/util.clj
scrosby: the only reason it exists is to fix unit tests. But invoked here, and belongs in this namespace.

dposada: Why not put it in the unit tests file?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,60 @@
...
+  "Build a new cache"
+  (-> (CacheBuilder/newBuilder)
+      (.maximumSize 1000000)
+      ;; if its not been accessed in 2 hours, whatever is going on, its not being visted by the
+      ;; scheduler loop anymore. E.g., its probably failed/done and won't be needed. So,
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache-generic!
```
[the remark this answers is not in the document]
pschorf: 👍

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -63,6 +63,26 @@
...
+    (is (= nil (.getIfPresent (deref cache) 1)))
+    (is (= 2 (.getIfPresent (deref cache) 2)))
+    (is (= nil (util/lookup-cache-generic! cache extract-fn miss-fn 101))) ; Should not be cached. Nil from miss function
+    (is (= nil (util/lookup-cache-generic! cache extract-fn miss-fn 102))) ; Should not be cached. Nil from miss function
+    (is (= nil (.getIfPresent (deref cache) 101)))
+    (is (= nil (.getIfPresent (deref cache) 102)))
+    (is (= 4 (util/lookup-cache-generic! cache extract-fn miss-fn 4))) ; Should be cached.
+    (is (= 2 (.getIfPresent (deref cache) 2)))
+    (is (= 4 (.getIfPresent (deref cache) 4)))))
+
```
pschorf: Nitpick: lots of newlines between these

wyegelwel:
@scrosby, how did you decide to cache these elements? 

I'm a little concerned these caches are going to add memory overhead to Cook. Do you have any measure for memory impact?

scheduler/src/cook/mesos/util.clj
wyegelwel: My prior is that it will need a lot less time than 2 hours. Can we measure when last access was?

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
scrosby: Made the rename. Not sure what you're asking for thoughts about....

scheduler/src/cook/mesos/util.clj
scrosby: Fixed another way.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+  "Build a new cache"
+  (-> (CacheBuilder/newBuilder)
+      (.maximumSize 1000000)
+      ;; if its not been accessed in 2 hours, whatever is going on, its not being visted by the
+      ;; scheduler loop anymore. E.g., its probably failed/done and won't be needed. So,
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache!
```
shamsimam: We should report metrics on the hit rate per cache, it may need an extra argument to be passed in to name the metric.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+      "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+      any misses. Caches only postive hits where both functions return non-nil"
+      [cache-atom extract-key-fn miss-fn item]
+      (let [key (extract-key-fn item)
+            ^Cache cache (deref cache-atom)]
+           (if key
+             (if-let [result (.getIfPresent cache key)]
```
shamsimam: Nitpick: formatting looks off.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache!
+      "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+      any misses. Caches only postive hits where both functions return non-nil"
+      [cache-atom extract-key-fn miss-fn item]
+      (let [key (extract-key-fn item)
+            ^Cache cache (deref cache-atom)]
```
shamsimam: Nitpick: Can we use `@cache-atom`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+           (if key
+             (if-let [result (.getIfPresent cache key)]
+                     result ; we got a hit.
+                     (let [new-result (miss-fn item)]
+                          ; Only cache non-nil
+                          (if new-result
```
shamsimam: Can we use `when` here as the else part is missing.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+                            (.put cache key new-result))
+                          new-result))
+             (miss-fn item))))
+
+(defn lookup-cache-datomic-entity!
```
shamsimam: Nitpick: formatting looks off.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+      ;; if its not been accessed in 2 hours, whatever is going on, its not being visted by the
+      ;; scheduler loop anymore. E.g., its probably failed/done and won't be needed. So,
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache!
+      "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+      any misses. Caches only postive hits where both functions return non-nil"
+      [cache-atom extract-key-fn miss-fn item]
```
shamsimam: Nitpick: `extract-key-fn` could not be an argument in this function, the caller can send the extracted key.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+
+(defn lookup-cache-datomic-entity!
+      "Specialized function for caching where datomic entities are the key. Extracts :db/id so that
+      we don't keep the entity alive in the cache"
+      [cache miss-fn entity]
+      (lookup-cache! cache :db/id miss-fn entity))
+
+
+; Fake. Are actually initialized in flush-caches
+(defonce job-resources-cache (atom (new-cache)))
```
shamsimam: Can we name this `job-entity-id->resources-cache`? Also, similar `->` mapping for the names below.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+(defonce job-category-cache (atom (new-cache)))
+(defonce task-user-cache (atom (new-cache)))
+(defonce feature-vector-cache (atom (new-cache)))
+
+(defn flush-caches
+  "Flush the caches. Needed in unit tests. We centralize initialization by using it to initialize the caches too."
+  []
+  (swap! job-resources-cache (fn [_] (new-cache)))
```
shamsimam: We should use `reset!` instead of `swap!` here.

scheduler/src/cook/mesos/util.clj
```diff
@@ -39,10 +91,13 @@
   "Return the category of the job. Currently jobs can be :normal or :gpu. This
    is used to give separate queues for scarce & non-scarce resources"
   [job]
-  (let [resources (:job/resource job)]
-    (if (some #(= :resource.type/gpus (:resource/type %)) resources)
-      :gpu
-      :normal)))
+      (let [categorize-job-miss
```
shamsimam: Nitpick: formatting looks off.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+      (lookup-cache! cache :db/id miss-fn entity))
+
+
+; Fake. Are actually initialized in flush-caches
+(defonce job-resources-cache (atom (new-cache)))
+(defonce job-category-cache (atom (new-cache)))
+(defonce task-user-cache (atom (new-cache)))
+(defonce feature-vector-cache (atom (new-cache)))
+
+(defn flush-caches
```
shamsimam: Nitpick: can we name this `flush-caches!`?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1140,6 +1140,7 @@
         (is (thrown? Exception (s/validate api/Job (assoc min-job :expected-runtime 3 :max-runtime 2))))))))
 
 (deftest test-create-jobs!
+   (util/flush-caches)
```
shamsimam: Should a similar call not exist for all other unit tests that trigger these cache operations? Alternatively, if each test creates its own job, should there be no cache hits so there should be no need to flush the caches?

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -63,6 +63,24 @@
     ;; u3 has no jobs (running or otherwise)
     (is (= 0 (count (util/get-user-running-job-ents (db conn) "u3"))))))
 
+(deftest test-cache
```
shamsimam: Should we also test for eviction of entries in the cache?

scrosby:
@wyegelwel 

> how did you decide to cache these elements?

These functions showed up in profiles. Overall benefit of this is about 30% faster rank jobs, a bit faster in offer processing, etc, in dftstaging.

> I'm a little concerned these caches are going to add memory overhead to Cook. Do you have any measure for memory impact?
No specific measure, thats why they're limited to 1MB max size, but with the last-access 2 hour expiration, I expect they shouldn't be much more than the size of the queue + running jobs + last 2 hours of completed jobs in total size. (you concur?) I can collect measurement estimates with a heap dump, but dftstaging is too different from dft for me to believe in anything I measure. We can bound the size though as 1M * bytes-in-element, which from the code would seem to be at most 400 bytes or so. (I'm caching dictionaries with 2-5 keys)

scheduler/src/cook/mesos/util.clj
scrosby: Track under https://github.com/twosigma/Cook/issues/739

scheduler/src/cook/mesos/util.clj
scrosby: Thats how cursive formats if's.

scheduler/src/cook/mesos/util.clj
scrosby: Cursive formatted this.

scheduler/src/cook/mesos/util.clj
scrosby: It works better this way, as re-using extract-key-fn can be reused across multiple caches. What you're asking for is essentially to inline the invocation of this into all of the callers, versus passing it.

scheduler/src/cook/mesos/util.clj
shamsimam: The docstring should be indented at 2 spaces.

scheduler/test/cook/test/mesos/api.clj
scrosby: Most unit tests get it done via invoking restore-fresh-database! in testutils.clj, which now flushes the caches.

scheduler/test/cook/test/mesos/util.clj
scrosby: I'm willing to say 'trust guava'? :)

shamsimam: It basically ends up testing whether we have invoked guava correctly to configure the eviction policy :)

scheduler/src/cook/mesos/util.clj
scrosby: @dposada Its an implementation detail, I don't see why we should expose it in a docstring.
@wyegelwel Probalby fine. I chose 2 hours to be uncontroversial. Do you have a specific timeout we should use for expiratoin? We could try to measure it, but its going to be workload and configuration dependent, so we'd want to measure in production.  IMO, not worth deep diving that far.

scheduler/test/cook/test/mesos/util.clj
scrosby: Yup.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+             (miss-fn item))))
+
+(defn lookup-cache-datomic-entity!
+      "Specialized function for caching where datomic entities are the key. Extracts :db/id so that
+      we don't keep the entity alive in the cache"
+      [cache miss-fn entity]
+      (lookup-cache! cache :db/id miss-fn entity))
+
+
+; Fake. Are actually initialized in flush-caches
```
sradack: If we're running in production, the intializers inside the `defonce` do actually initialize the cache though, right?  What is fake about this?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+(defonce job-ent->resources-cache (atom (new-cache)))
```
sradack: Should these be dynamic vars instead of atoms?  We could use `binding` in the unit tests, which is scoped around the code that need independent caches, as opposed to remembering to flush before a test because "we know" someone polluted the cache already.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,59 @@
...
+(defonce categorize-job-cache (atom (new-cache)))
+(defonce task-ent->user-cache (atom (new-cache)))
+(defonce task->feature-vector-cache (atom (new-cache)))
+
+(defn flush-caches!
+      "Flush the caches. Needed in unit tests. We centralize initialization by using it to initialize the caches too."
+      []
+      (reset! job-ent->resources-cache (new-cache))
```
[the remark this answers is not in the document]
scrosby: Doh. Fixed. Yes we want invalidateAll.

scheduler/src/cook/mesos/util.clj
scrosby: Stale comment. Fixed.

scheduler/src/cook/mesos/util.clj
scrosby: No. We want these to be shared across threads. E.g., as a side effect, these caches make offer processing and the metrics dashboard faster/cheaper.

scheduler/src/cook/mesos/util.clj
```diff
@@ -36,13 +85,16 @@
...
+      "Return the category of the job. Currently jobs can be :normal or :gpu. This
+       is used to give separate queues for scarce & non-scarce resources"
+      [job]
+      (let [categorize-job-miss
+            (fn [job]
+                (let [resources (:job/resource job)]
+                     (if (some #(= :resource.type/gpus (:resource/type %)) resources)
+                       :gpu
+                       :normal)))]
+           (lookup-cache-datomic-entity! categorize-job-cache categorize-job-miss job)))
```
sradack: Is `job` a Datomic entity?  If so, it has a builtin reference to `db` and we don't want to cache that.  @pschorf removed a huge memory leak before related to this.

sradack: If it's not a Datomic entity, we should rename the function.

scrosby: Yes, it is a datomic entity. But that's OK. lookup-cache-datomic-entity! is a specialized version of lookup-cache-generic! that can use datomic entities as cache keys without the memory leak.

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,56 @@
             [metatransaction.core :refer (db)]
             [metrics.timers :as timers]
             [plumbing.core :as pc :refer (map-vals map-keys)])
-  (:import [java.util Date]))
+  (:import
+    [com.google.common.cache Cache]
+    [com.google.common.cache CacheBuilder]
```
dposada: Can we combine these two onto a single line?

scheduler/src/cook/mesos/util.clj
dposada: Did we get @wyegelwel's concern resolved?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,56 @@
...
+      ;; lets kick it out to keep cache small.
+      (.expireAfterAccess 2 TimeUnit/HOURS)
+      (.build)))
+
+(defn lookup-cache!
+  "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+  any misses. Caches only postive hits where both functions return non-nil"
+  [^Cache cache extract-key-fn miss-fn item]
+  (let [key (extract-key-fn item)]
+    (if key
```
dposada: Can we use `if-let` here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -24,7 +24,56 @@
...
+(defonce ^Cache task-ent->user-cache (new-cache))
+(defonce ^Cache task->feature-vector-cache (new-cache))
+
+(defn flush-caches!
+  "Flush the caches. Needed in unit tests. We centralize initialization by using it to initialize the caches too."
+  []
+  (.invalidateAll job-ent->resources-cache)
+  (.invalidateAll categorize-job-cache)
+  (.invalidateAll task-ent->user-cache)
+  (.invalidateAll task->feature-vector-cache))
```
dposada: Is this called anywhere other than from unit tests? If not, can we move it to the unit tests?

scheduler/src/cook/mesos/util.clj
scrosby: Done

scheduler/src/cook/mesos/util.clj
scrosby: Thanks. Done.

scheduler/src/cook/mesos/util.clj

scheduler/src/cook/mesos/util.clj
wyegelwel: Sorry, missed Scott's comment. If it is easy to measure, I think it is worth measuring. If not, then as long as we are measuring the overall size than we can probably punt on the 2 hours unless the cache is very large.

scheduler/src/cook/mesos/util.clj
dposada: They still seem to be on separate lines

scheduler/src/cook/mesos/util.clj
dposada: I don't see the `if-let`

scheduler/src/cook/mesos/util.clj
dposada: It still seems to be here

scheduler/src/cook/mesos/util.clj
scrosby: Hadn't pushed when you saw that. Was waiting to roll up feedback and push all at once.

scheduler/src/cook/mesos/util.clj
scrosby: We have an open JIRA to add metrics collection from the caches -- #739.

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1140,6 +1140,7 @@
         (is (thrown? Exception (s/validate api/Job (assoc min-job :expected-runtime 3 :max-runtime 2))))))))
 
 (deftest test-create-jobs!
+  (cook.test.testutil/flush-caches!)
```
dposada: Why are we `refer`ing `flush-caches!` but also specifying the namespace when we call it?

scheduler/src/cook/mesos/util.clj
wyegelwel: Ok, good enough for me. I think #739 should get merged before we cut a release with this feature (or have it defaulted off)

PR #734, opened 2018-02-26
Adds histograms by failure reason for run/cpu/mem time

dposada:

- introducing new histograms by failure reason for run time, cpu time, and mem time
- updating the histograms when a task fails

We want to enable better visibility into the amount of time and resources spent on various failures, e.g. "preempted by rebalancer".

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -171,6 +171,24 @@
                       (catch Exception e
                         (log/debug e "Error reading a string from mesos status data. Is it in the format we expect?")))))})
 
+(defn update-reason-histos!
+  "Updates histograms for run time, cpu time, and memory time,
+  where the histograms have the failure reason in the title"
+  [db mesos-reason instance-runtime {:keys [cpus mem]}]
+  (let [reason (->> mesos-reason
+                    (reason/mesos-reason->cook-reason-entity-id db)
+                    (d/entity db)
+                    :reason/name
+                    name)
+        update-histo! (fn update-histo! [s v]
+                        (histograms/update!
+                          (histograms/histogram
+                            ["cook-mesos" "scheduler" (str "hist-task-fail-" s "-reason-" reason)])
+                          v))]
+    (update-histo! "times" instance-runtime)
+    (update-histo! "cpu-times" (* instance-runtime cpus))
+    (update-histo! "mem-times" (* instance-runtime mem))))
+
 (defn handle-status-update
   "Takes a status update from mesos."
   [conn driver ^TaskScheduler fenzo sync-agent-sandboxes-fn status]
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -171,6 +171,24 @@
                       (catch Exception e
                         (log/debug e "Error reading a string from mesos status data. Is it in the format we expect?")))))})
 
+(defn update-reason-histos!
+  "Updates histograms for run time, cpu time, and memory time,
+  where the histograms have the failure reason in the title"
+  [db mesos-reason instance-runtime {:keys [cpus mem]}]
```
shamsimam: Should we check whether `reason` is `nil`?

dposada: `mesos-reason->cook-reason-entity-id` returns the `:mesos-unknown` reason if the reason from mesos is `nil`

dposada:
I would like to have integration tests, but we don't currently have the integration test infrastructure for asserting on metrics (I think it's worth considering, but probably outside the scope of this PR).

As for unit tests, we don't typically write unit tests just for checking that metrics were set. I don't know that it's worth it for this case.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -171,6 +171,24 @@
...
+  (let [reason (->> mesos-reason
+                    (reason/mesos-reason->cook-reason-entity-id db)
+                    (d/entity db)
+                    :reason/name
+                    name)
+        update-histo! (fn update-histo! [s v]
+                        (histograms/update!
+                          (histograms/histogram
+                            ["cook-mesos" "scheduler" (str "hist-task-fail-" s "-reason-" reason)])
```
shamsimam: Can we create nesting on reason and then the times?

dposada: Sure thing. Here's what they look like now:

```bash
cook-mesos.scheduler.hist-task-fail.mesos-command-executor-failed.cpu-times
             count = 1
               min = 402
               max = 402
              mean = 402.00
            stddev = 0.00
            median = 402.00
              75% <= 402.00
              95% <= 402.00
              98% <= 402.00
              99% <= 402.00
            99.9% <= 402.00
cook-mesos.scheduler.hist-task-fail.mesos-command-executor-failed.mem-times
             count = 1
               min = 51456
               max = 51456
              mean = 51456.00
            stddev = 0.00
            median = 51456.00
              75% <= 51456.00
              95% <= 51456.00
              98% <= 51456.00
              99% <= 51456.00
            99.9% <= 51456.00
cook-mesos.scheduler.hist-task-fail.mesos-command-executor-failed.times
             count = 1
               min = 402
               max = 402
              mean = 402.00
            stddev = 0.00
            median = 402.00
              75% <= 402.00
              95% <= 402.00
              98% <= 402.00
              99% <= 402.00
            99.9% <= 402.00
```

PR #736, opened 2018-02-27
Refactors config code into config namespace

dposada:

- `lazy-load-var` was defined twice (once in `cook.components`, again in `cook.util`), consolidating
- moving all config-related code from `cook.components` to the new `cook.config`
- deleting some dead code

For readability / organization. Having all of the config code in one place makes it easier to read the code.


dposada:
@shamsimam I added back the variadic args to `-main` -- can we merge?

PR #738, opened 2018-02-28
cs: adds logging for the iteration where the rebalance was a no-op

shamsimam:

- log when the rebalance was a no-op
- minor refactoring to sort keys and remove unused values

Helps us quickly identify why a rebalance iteration was a no-op.

### DIFF
--- a/scheduler/src/cook/mesos/rebalancer.clj
+++ b/scheduler/src/cook/mesos/rebalancer.clj
@@ -522,13 +522,13 @@
               recognized-params))]))))
 
 (defn start-rebalancer!
-  [{:keys [config conn driver get-mesos-utilization pending-jobs-atom offer-cache
-           rebalancer-reservation-atom trigger-chan view-incubating-offers view-mature-offers]}]
+  [{:keys [config conn driver get-mesos-utilization offer-cache pending-jobs-atom
+           rebalancer-reservation-atom trigger-chan view-incubating-offers]}]
   (binding [metrics-dru-scale (:dru-scale config)]
     (update-datomic-params-from-config! conn config)
     (util/chime-at-ch
       trigger-chan
-      (fn []
+      (fn trigger-rebalance-iteration []
         (log/info "Rebalance cycle starting")
         (let [params (read-datomic-params conn)
               utilization (get-mesos-utilization)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -538,19 +538,22 @@
                                                  (select-keys (keywordize-keys (:resources v))
                                                               [:cpus :mem :gpus])]))
                                          (into {}))]
-          (when (and (seq params)
-                     (> utilization (:min-utilization-threshold params)))
+          (if (and (seq params)
+                   (> utilization (:min-utilization-threshold params)))
```
dposada: Can we destructure this?

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -538,19 +538,22 @@
...
+                                        :compute-pending-job-dru compute-pending-normal-job-dru))
               (rebalance! conn driver offer-cache gpu-pending-jobs host->spare-resources
                           rebalancer-reservation-atom
-                          (assoc params :compute-pending-job-dru compute-pending-gpu-job-dru
-                                        :category :gpu))))))
-        {:error-handler (fn [ex] (log/error ex "Rebalance failed"))})
-      #(async/close! trigger-chan)))
+                          (assoc params :category :gpu
+                                        :compute-pending-job-dru compute-pending-gpu-job-dru)))
+            (log/info "Rebalance iteration is a no-op"
```
dposada: Maybe "Skipping rebalancing due to low cluster utilization"?

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: Done.

scheduler/src/cook/mesos/rebalancer.clj
shamsimam: That may not be true as `min-utilization-threshold` may be nil.

shamsimam:
@dposada build is green, can we merge?

scheduler/src/cook/mesos/rebalancer.clj
pschorf: Should we just disable the rebalancer in that case? I like @dposada 's message better.

shamsimam: We can't disable the rebalancer as the data might be written to the DB after Cook has been running for a while. I'll change the messaging.

shamsimam: Changed the log message.

dposada:
I am good with merging. Should we wait for @pschorf to approve?

PR #741, opened 2018-02-28
Adds mount minimally

dposada:
This is a proof-of-concept PR to demonstrate how we might use [`mount`](https://github.com/tolitius/mount) to manage our dependencies. This PR does not attempt to change everything over to using `mount`. Instead, it attempts to show "the smallest thing that could possibly work". 

- adding [`mount`](https://github.com/tolitius/mount) library
- adding `mount/defstate`s for:
  - the config settings
  - the datomic connection
- fully taking advantage of the above `defstate`s from `cook.mesos.monitor/start-collecting-stats`
- investigating how unit tests should work after the above changes

The current state of affairs is that all dependencies need to be threaded through from `cook.components` to their final point of consumption. This makes it hard to add dependencies and hard to read certain parts of the code. We want to find something better.

## What is left to do?

- [ ] currently the `datomic/conn` `defstate` is mostly consumed at the top level; to take full advantage, it should be removed as an argument from the various call chains and consumed at the leaf levels
- [ ] same as above for the `config/config` `defstate` (see https://github.com/dposada/Cook/pull/3 for one deep-dive example of this)
- [ ] add other `defstate`s in useful places

## Random

This is something kind of cool that you can do with `mount`:

```bash
(require '[mount.tools.graph :as mount-graph])
=> nil
(mount-graph/states-with-deps)
=>
({:name "#'cook.config/config", :order 1, :status #{:started}, :deps #{}}
 {:name "#'cook.datomic/conn", :order 1, :status #{:stopped}, :deps #{"#'cook.config/config"}})
```

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -108,7 +108,10 @@
                   :exclusions [org.slf4j/slf4j-log4j12
                                org.slf4j/log4j
                                log4j]]
-                 [org.apache.curator/curator-test "2.7.1"]]
+                 [org.apache.curator/curator-test "2.7.1"]
+
+                 ;; Dependency management
+                 [mount "0.1.12"]]
 
   :repositories {"maven2" {:url "https://files.couchbase.com/maven2/"}
                  "sonatype-oss-public" "https://oss.sonatype.org/content/groups/public/"}
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -145,13 +145,13 @@
                    ; using a profiles.clj file that defines a profile
                    ; which pulls in datomic-pro
                    [com.datomic/datomic-free "0.9.5206"
-                    :exclusions [org.slf4j/slf4j-api
-                                 com.fasterxml.jackson.core/jackson-core
+                    :exclusions [com.fasterxml.jackson.core/jackson-core
+                                 joda-time
                                  org.slf4j/jcl-over-slf4j
                                  org.slf4j/jul-to-slf4j
                                  org.slf4j/log4j-over-slf4j
-                                 org.slf4j/slf4j-nop
-                                 joda-time]]
+                                 org.slf4j/slf4j-api
+                                 org.slf4j/slf4j-nop]]
                    ; Similarly, one could use an older version of the
                    ; mesomatic library in environments that require it
                    [wyegelwe/mesomatic "1.0.1-r0-SNAPSHOT"]]}
@@ -159,13 +159,13 @@
    :uberjar
    {:aot [cook.components]
     :dependencies [[com.datomic/datomic-free "0.9.5206"
-                    :exclusions [org.slf4j/slf4j-api
-                                 com.fasterxml.jackson.core/jackson-core
+                    :exclusions [com.fasterxml.jackson.core/jackson-core
+                                 joda-time
                                  org.slf4j/jcl-over-slf4j
                                  org.slf4j/jul-to-slf4j
                                  org.slf4j/log4j-over-slf4j
-                                 org.slf4j/slf4j-nop
-                                 joda-time]]]}
+                                 org.slf4j/slf4j-api
+                                 org.slf4j/slf4j-nop]]]}
 
    :dev
    {:dependencies [[criterium "0.4.4"]
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -157,7 +157,15 @@
                    [wyegelwe/mesomatic "1.0.1-r0-SNAPSHOT"]]}
 
    :uberjar
-   {:aot [cook.components]}
+   {:aot [cook.components]
+    :dependencies [[com.datomic/datomic-free "0.9.5206"
+                    :exclusions [org.slf4j/slf4j-api
+                                 com.fasterxml.jackson.core/jackson-core
+                                 org.slf4j/jcl-over-slf4j
+                                 org.slf4j/jul-to-slf4j
+                                 org.slf4j/log4j-over-slf4j
+                                 org.slf4j/slf4j-nop
+                                 joda-time]]]}
 
    :dev
    {:dependencies [[criterium "0.4.4"]
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -13,19 +13,14 @@
 ;; limitations under the License.
 ;;
 (ns cook.components
-  (:require [clj-logging-config.log4j :as log4j-conf]
-            [clj-time.core :as t]
-            [clojure.core.cache :as cache]
-            [clojure.edn :as edn]
+  (:require [clojure.core.cache :as cache]
             [clojure.pprint :refer (pprint)]
-            [clojure.stacktrace :as stacktrace]
-            [clojure.string :as str]
             [clojure.tools.logging :as log]
             [compojure.core :refer (GET POST routes context)]
             [compojure.route :as route]
-            [congestion.limits :refer (RateLimit)]
             [congestion.middleware :refer (wrap-rate-limit ip-rate-limit)]
             [congestion.storage :as storage]
+            [cook.config :as config]
             [cook.curator :as curator]
             [cook.impersonation :refer (impersonation-authorized-wrapper)]
             [cook.util :as util]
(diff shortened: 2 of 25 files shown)


scheduler/test/cook/test/mesos.clj
```diff
@@ -18,11 +18,8 @@
   (:require [clojure.core.async :as async]
             [cook.curator :as curator]
             [cook.datomic]
-            [cook.mesos :as mesos]
```
shamsimam: We should this cleanup in a separate PR.

scheduler/src/cook/datomic.clj
```diff
@@ -17,8 +17,34 @@
...
+  [[:settings mesos-datomic-uri]]
+  ((util/lazy-load-var 'datomic.api/create-database) mesos-datomic-uri)
+  (let [conn ((util/lazy-load-var 'datomic.api/connect) mesos-datomic-uri)]
+    (doseq [txn (deref (util/lazy-load-var 'cook.mesos.schema/work-item-schema))]
+      (deref ((util/lazy-load-var 'datomic.api/transact) conn txn))
+      ((util/lazy-load-var 'metatransaction.core/install-metatransaction-support) conn)
+      ((util/lazy-load-var 'metatransaction.utils/install-utils-support) conn))
+    conn))
+
+(defn disconnect
```
shamsimam: New and good functionality :)

scheduler/src/cook/datomic.clj
```diff
@@ -17,8 +17,34 @@
...
+            [cook.config :refer (config)]
+            [cook.util :as util]
+            [datomic.api :as d :refer (q)]
+            [mount.core :as mount]
+            [plumbing.core :refer (defnk)])
+  (:import (clojure.lang Agent)
+           (datomic Connection)
+           (java.util.concurrent BlockingQueue TimeUnit)))
+
+(defnk create-connection
```
shamsimam: Nitpick: Can we stick to using `defn`?

scheduler/src/cook/mesos.clj
```diff
@@ -191,12 +190,11 @@
...
    framework-id                  -- str, the Mesos framework id from the cook settings
    fenzo-config                  -- map, config for fenzo, See scheduler/docs/configuration.adoc for more details
-   sandbox-syncer-state          -- map, representing the sandbox syncer object
-   user-metrics-interval-seconds -- long, time in seconds between collecting and counting per-user job metrics"
+   sandbox-syncer-state          -- map, representing the sandbox syncer object"
   [{:keys [curator-framework executor-config fenzo-config framework-id get-mesos-utilization gpu-enabled? make-mesos-driver-fn
            mea-culpa-failure-limit mesos-datomic-conn mesos-datomic-mult mesos-leadership-atom mesos-pending-jobs-atom
            offer-cache offer-incubate-time-ms optimizer-config progress-config rebalancer-config
-           sandbox-syncer-state server-config task-constraints trigger-chans user-metrics-interval-seconds zk-prefix]}]
+           sandbox-syncer-state server-config task-constraints trigger-chans zk-prefix]}]
```
shamsimam: Yay!

scheduler/test/cook/test/mesos.clj
dposada: :+1: done

scheduler/src/cook/datomic.clj
dposada: Done

scheduler/src/cook/components.clj
```diff
@@ -288,11 +277,12 @@
...
   "Entry point for Cook. Initializes configuration settings,
   instruments the JVM, and starts up the scheduler and API."
-  [config & _]
+  [config-file-path & _]
+  (println "Cook" @util/version "( commit" @util/commit ")")
   (try
-    (let [settings (config/init-settings config)
-          _ (metrics-jvm/instrument-jvm)
-          server (scheduler-server settings)]
+    (mount/start-with-args (cook.config/read-config config-file-path))
```
shamsimam: Could this be just the `config-file-path`? Then the necessary reading and initialization stays local to the `cook.config` namespace.

scheduler/src/cook/mesos/monitor.clj
```diff
@@ -155,14 +157,15 @@
   "Starts a periodic timer to collect stats about running, waiting, and starved jobs per user.
 
    Return a function which can be used to stop collecting stats if invoked."
-  [mesos-conn interval-seconds]
```
shamsimam: Yay! 

Can we add a unit test for this to see how the code looks like when mocking `mount` initialized `def`s?

scheduler/src/cook/components.clj
dposada: I had it that way originally, but I thought it would be nice to be able to call `mount/start-with-args` with a map instead of a file (for example, when testing in the REPL).

scheduler/src/cook/mesos/monitor.clj
dposada: Done, I added `test-start-collecting-stats`, let me know what you think

scheduler/test/cook/test/testutil.clj
```diff
@@ -237,3 +238,20 @@
                              {:level :info
                               :out (ConsoleAppender. (PatternLayout. "%d{ISO8601} %-5p %c [%t] - %m%n"))}))
   (log/info "Running" (:var m)))
+
+(let [minimal-config {:unhandled-exceptions {}
```
shamsimam: Lexicographic ordering :(

scheduler/test/cook/test/testutil.clj
```diff
@@ -237,3 +238,20 @@
...
+                      :scheduler {}
+                      :zookeeper {:local? true}
+                      :port nil
+                      :metrics {}
+                      :nrepl {}}]
+  (defn setup
+    "Given an optional config map, initializes the config state"
+    [& {:keys [config], :or nil}]
+    (mount/stop)
+    (mount/start-with-args (merge minimal-config config) #'cook.config/config)))
```
shamsimam: Do we want mount to initialize the entire dependency graph for each test we run? 

Is this less severe in the repl which triggers reloads of the `defstate`s from the loaded namespaces?

scheduler/test/cook/test/mesos/monitor.clj
```diff
@@ -265,4 +266,18 @@
...
+
+(deftest test-start-collecting-stats
+  (setup :config {:metrics {:user-metrics-interval-seconds 1}})
+  (let [period (atom nil)
+        millis-between (fn [a b]
+                         (-> a
+                             ^Interval (time/interval b)
+                             .toDuration
+                             .getMillis))]
+    (with-redefs [chime/chime-at (fn [p _ _] (reset! period p))]
```
shamsimam: Does `with-redefs` of `datomic/conn` work?

scheduler/test/cook/test/testutil.clj
dposada: > Do we want mount to initialize the entire dependency graph for each test we run?

I don't think so. I think we want each test to be specific about what it initializes. Currently, the `setup` function only initializes `cook.config/config`, but I expect it to change to allow for initializing other states as needed.

> Is this less severe in the repl which triggers reloads of the defstates from the loaded namespaces?

I don't understand this question, can you re-state?

scheduler/test/cook/test/mesos/monitor.clj
dposada: I'm not sure. Why do you want to, though? Using the in-memory datomic connection seems to work really well.

scheduler/test/cook/test/testutil.clj
dposada: My deepest apologies. Done

scheduler/project.clj
```diff
@@ -154,7 +157,15 @@
                    [wyegelwe/mesomatic "1.0.1-r0-SNAPSHOT"]]}
 
    :uberjar
-   {:aot [cook.components]}
+   {:aot [cook.components]
+    :dependencies [[com.datomic/datomic-free "0.9.5206"
+                    :exclusions [org.slf4j/slf4j-api
```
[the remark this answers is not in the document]
dposada: Done

scheduler/test/cook/test/testutil.clj
shamsimam: I want to see how it feels to use `with-redefs` with `mount` initialized state as that can be a common usage pattern.

scheduler/test/cook/test/mesos/monitor.clj
shamsimam: We spoke offline as we want to see how it feels to use this pattern.

scheduler/src/cook/components.clj
shamsimam: Okay

scheduler/test/cook/test/testutil.clj
shamsimam: > I think we want each test to be specific about what it initializes.

But initializing an individual state will trigger upstream dependencies to have to be initialized first? I think being able to use `with-redefs` can avoid that concern.

dposada: I added a unit test, `test-redef-config`, showing that a `defstate` can be `with-redef`'d, let me know what you think

scheduler/test/cook/test/mesos/monitor.clj
dposada: See `test-redef-config`

dposada:
@pschorf This is ready for more opinions :)

dposada:

dposada:
@wyegelwel If you have time, I would welcome your feedback on this

dposada:
@pschorf Have you had a chance to review?

pschorf:
This looks good to me.

PR #742, opened 2018-02-28
cs: improves the `dru/init-user->dru-divisors` performance …

shamsimam:
…by making fewer datomic queries from `share/get-share`

- introduce `share/get-share-by-types` which makes a single query to load shares for a user

For 250 users, reporting last time of 10 iterations and using in-memory db and 10K running tasks and 25K pending jobs

Before: `Elapsed time: 366.445606 msecs` (With a datomic connection backed by persistent storage this number is between 3 to 5 seconds)
After:  `Elapsed time: 163.840391 msecs`

The 2-3X improvement is expected as it is making 2-3X fewer DB queries.

There are existing unit tests verifying this change in `cook.test.mesos.share/test'`.

### DIFF
--- a/scheduler/src/cook/mesos/share.clj
+++ b/scheduler/src/cook/mesos/share.clj
@@ -24,8 +24,14 @@
 (def default-user "default")
 
 (defn- resource-type->datomic-resource-type
+  "Converts the resource type to a datomic resource type keyword, e.g. :cpus to :resource.type/cpus"
   [type]
   (keyword "resource.type" (name type)))
+
+(defn- datomic-resource-type->resource-type
+  "Converts the datomic resource type to a resource type keyword, e.g. :resource.type/cpus to :cpus"
+  [type]
+  (keyword (name type)))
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/share.clj
```diff
@@ -44,27 +50,30 @@
...
+  (let [query '[:find ?t ?a
+                :in $ ?u [?t ...]
+                :where
+                [?e :share/user ?u]
+                [?e :share/resource ?r]
+                [?r :resource/type ?t]
+                [?r :resource/amount ?a]]
+        datomic-resource-types (map resource-type->datomic-resource-type types)
+        default-type->share (pc/map-from-keys
+                              (fn [type] (get type->share type Double/MAX_VALUE) )
```
pschorf: nitpick: extra space

shamsimam: Fixed.

PR #744, opened 2018-03-01
cs: adds support for Mesos run-as-user

shamsimam:

- this sets the user that Cook will override and set as the user to run tasks on Mesos with

Allows us to mimic submitting jobs from many users.

### DIFF
--- a/scheduler/docs/configuration.adoc
+++ b/scheduler/docs/configuration.adoc
@@ -136,6 +136,11 @@ We'll look at the configurable options in turn:
   This sets the role that Cook will connect to Mesos with. Default: `*`
   You can omit this property unless you've enabled security features in Mesos. The value should be in authorized list for the current `:principal` in `register_frameworks` Action in Mesos Authorization file. See http://mesos.apache.org/documentation/latest/authorization/ for details.
 
+`:run-as-user`::
+  When configured, this sets the user that Cook will override and set as the user to run tasks on Mesos with.
+  You can omit this property, in which case the user configured in the job will be used as user to run specific jobs on Mesos.
+  Cook's scheduling algorithm continues to use the user specified in the job to compute job schedules.
+
 `:framework-name`::
   This sets part of the name of the framework that Cook will register to Mesos. Default: Cook
   When connecting to Mesos, Cook will use a framework name like "YourFrameworkName-e254483".  It will append the current git hash to the value you specify here.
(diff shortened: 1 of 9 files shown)


scheduler/docs/configuration.adoc
```diff
@@ -136,6 +136,11 @@ We'll look at the configurable options in turn:
   This sets the role that Cook will connect to Mesos with. Default: `*`
   You can omit this property unless you've enabled security features in Mesos. The value should be in authorized list for the current `:principal` in `register_frameworks` Action in Mesos Authorization file. See http://mesos.apache.org/documentation/latest/authorization/ for details.
 
+`:run-as-user`::
+  When configured, this sets the user that Cook will override and set as the user to run tasks on Mesos with.
+  You can omit this property, in which case the user configured in the job will be used as user to run specific jobs on Mesos.
```
dposada: "in which case the user configured in the job will be used as user to run specific jobs on Mesos" ->
"in which case the user configured in the job will be used as the user to run the job (the default behavior)"

scheduler/src/cook/config.clj
```diff
@@ -259,6 +259,10 @@
                         principal)
      :mesos-role (fnk [[:config [:mesos {role "*"}]]]
                    role)
+     :mesos-run-as-user (fnk [[:config [:mesos {run-as-user nil}]]]
+                          (when run-as-user
+                            (log/info "Tasks launched in Mesos will ignore user specified in the job and run as" run-as-user))
```
dposada: Maybe `log/warn`?

scheduler/test/cook/test/mesos/task.clj
```diff
@@ -332,21 +333,40 @@
...
+                    :executor :executor/custom
+                    :executor-key :custom-executor
+                    :framework-id framework-id
+                    :labels {}
+                    :name (format "dummy_job_%s_%s" (:job/user job-ent) task-id)
+                    :num-ports 0
+                    :resources {:cpus 1.0, :mem 10.0}
+                    :task-id task-id}
+                   (dissoc task-metadata :data)))
+            (is (= (pr-str {:instance "0"}) (-> task-metadata :data (String. "UTF-8"))))))))
```
dposada: For both of these tests, how about if we only assert on the `:command` `:user`, since that's what we are interested in here?

shamsimam: Since it is a different invocation, I assert on the entire output in both versions of the tests.

scheduler/src/cook/config.clj
shamsimam: Done.

scheduler/docs/configuration.adoc
shamsimam: Done.

scheduler/test/cook/test/mesos/task.clj
dposada: We talked offline, agree to disagree :)

PR #757, opened 2018-03-09
Adds deeper integration with mount

dposada:

- removing the executor config from all the call chains it is passed to, and using the `defstate` instead
- fixing the corresponding unit tests

As another step towards not passing config around.

### DIFF
--- a/scheduler/src/cook/mesos/task.clj
+++ b/scheduler/src/cook/mesos/task.clj
@@ -17,11 +17,12 @@
   (:require [clojure.data.json :as json]
             [clojure.set :as set]
             [clojure.tools.logging :as log]
-            [cook.config :refer (config, executor-config)]
+            [cook.config :as config]
             [cook.mesos.util :as util]
             [mesomatic.types :as mtypes]
             [plumbing.core :refer (map-vals)])
-  (:import com.netflix.fenzo.TaskAssignmentResult))
+  (:import com.google.protobuf.ByteString
+           com.netflix.fenzo.TaskAssignmentResult))
 
 (def cook-executor-name "cook_executor")
 (def cook-executor-source "cook_scheduler_executor")
@@ -40,7 +41,7 @@
    b. This is going to be the first instance of the job, and
    c. The job UUID hash mod 100 yields less than portion percent."
   [job-ent]
-  (let [{:keys [portion retry-limit]} (executor-config)]
+  (let [{:keys [portion retry-limit]} (config/executor-config)]
     (and (nil? (:job/executor job-ent))
          (number? retry-limit)
          (> retry-limit (count (:job/instance job-ent)))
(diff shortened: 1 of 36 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -52,17 +54,18 @@
    2. The Cook executor command has been configured,
    3. Either :job/executor is explicitly enabled
       Or: the job is a cook-executor candidate (see cook-executor-candidate?)."
-  [job-ent {:keys [command portion retry-limit]}]
+  [job-ent]
   (and (not (use-custom-executor? job-ent))
-       command
+       (-> config :settings :executor :command)
```
shamsimam: How do you feel about `def-state`-ing `executor-config`? Then  we do not need to navigate the config to retrieve the command.

scheduler/test/cook/test/mesos/task.clj
```diff
@@ -324,6 +324,7 @@
                         :extract false
                         :value "file:///path/to/cook-executor"}}
         mesos-run-as-user nil]
+    (setup :config {:executor executor})
```
shamsimam: If we `def-state` `executor-config`, then we can change this to a `with-redef` and can avoid calling `setup` of 'knowing' where in config to place the executor config.

scheduler/src/cook/mesos/task.clj
dposada: Should every sub-element in the config be its own `defstate`?

shamsimam: Every sub-component could be, that let's us track dependencies better rather and also avoids passing the responsibility to the consumer to navigate the settings and retrieving the element it needs.

dposada: Discussed offline, instead of `defstate`ing the executor config, I added a new function, `cook.config/executor-config`, which returns the executor config from the full config, and can be `with-redef`'d

scheduler/test/cook/test/mesos/task.clj
dposada: See comment above

dposada:
@shamsimam This is ready for another round

scheduler/src/cook/mesos/task.clj
```diff
@@ -17,7 +17,7 @@
   (:require [clojure.data.json :as json]
             [clojure.set :as set]
             [clojure.tools.logging :as log]
-            [cook.config :refer (config)]
+            [cook.config :refer (config, executor-config)]
```
shamsimam: Nitpick: I have a preference for qualifying calls to the method with the alias, e.g. `[cook.config :as config]` and `(config/executor-config)`

dposada: Done

dposada:
@shamsimam Regarding `cook-executor-candidate?`, I can see it going either way. If you feel strongly about it not accessing the config, I can change it. My thought going into this refactor was to push all config-reading to the actual consumer of the config.

shamsimam:
@dposada I'm a little wary about changing all consumers to use the config helper function. But I'm good either way. I'll merge this and we can refactor it later if needed.

PR #761, opened 2018-03-13
Outgoing/caching2

scrosby:

- Store synthetic ID's in pending job tasks
- 
- 

This increases the cache hit rate for the job->feature-vector cache. Without this, these items are uncachable.

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -505,19 +505,22 @@
 
 
 (defn task->feature-vector
-      "Vector of comparable features of a task.
-       Last two elements are aribitary tie breakers.
-       Use :db/id because they guarantee uniqueness for different entities
-       (:db/id task) is not sufficient because synthetic task entities don't have :db/id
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -489,7 +489,9 @@
 (defn create-task-ent
   "Takes a pending job entity and returns a synthetic running task entity for that job"
   [pending-job-ent & {:keys [hostname slave-id] :or {hostname nil slave-id nil}}]
-  (merge {:job/_instance pending-job-ent
+  (merge {; Synthetic ID's so that we don't incur uncachable task-ent's in task->feature-vector.
+          :instance/synthetic-id (:db/id pending-job-ent)
```
shamsimam: Instead of introducing this field can the `extract-key` not look into `:job/_instance` and retrieve the id?

scrosby: Good suggestion. Done. Simplifes the code enough I rebased it.

PR #762, opened 2018-03-13
Use d/squuid to derive task ID's do that they're more tightly clustered.

scrosby:
- Change UUID generation for tasks to be more temporal.
- 
- 
- 

Increases the hit rate and the locality for storage of these datomic ID's.


shamsimam:
Awaiting green build

scrosby:
Build is green.

PR #763, opened 2018-03-13
Change JVM options. remove MaxPermSize and add +UseStringDeduplication

scrosby:
Switch to G1GC and turn on string deduplication. 
Remove deprecated JDK option

- In a test with a 500k jobs in queue, there were 92M strings in memory, with the char arrays accounting for a third of the heap.
- We have been using G1GC internally for a while, so expect that change to be safe.

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -207,7 +207,9 @@
 
   :main cook.components
   :jvm-opts ["-Dpython.cachedir.skip=true"
-             "-XX:MaxPermSize=500M"
+             "-XX:+UseG1GC"
+             "-XX:+UseStringDeduplication"
+             "-XX:+PrintStringDeduplicationStatistics"
              ;"-Dsun.security.jgss.native=true"
              ;"-Dsun.security.jgss.lib=/opt/mitkrb5/lib/libgssapi_krb5.so"
              ;"-Djavax.security.auth.useSubjectCredsOnly=false"


shamsimam:
Do you have numbers for the 500K jobs after these changes?

scrosby:
Not yet. dftstaging isn't freed up yet. This is technically dead code for configuring production as the command line options for production don't come from this file, but come from our internal ansible launcher. I have a pull request to change that config, then we can test on dftstaging and then I'll have numbers....

scrosby:
Something ate my old response.. Anyways, this code is only used for lein test and internal use. In production, the ansible configuration is used to construct the command line. So (AFAIK), I can't test and give real numbers until we deploy that change and (unless you are aware of a way to do a one-off launch with a custom command line)

scrosby:
Turned it on in a test enviornment. With 100k jobs in the queue, the number of char arrays is a third of the number of strings, before they were nearly identical. Memory usage for strings has dropped by a lot. UUID storage now dominates the heap.

scheduler/project.clj
```diff
@@ -207,7 +207,9 @@
 
   :main cook.components
   :jvm-opts ["-Dpython.cachedir.skip=true"
-             "-XX:MaxPermSize=500M"
+             "-XX:+UseG1GC"
+             "-XX:+UseStringDeduplication"
+             "-XX:+PrintStringDeduplicationStatistics"
```
dposada: I had to comment out `-XX:+UseStringDeduplication` and `-XX:+PrintStringDeduplicationStatistics` in order to get my local cook scheduler to start up. Is there something I'm missing @scrosby?

PR #768, opened 2018-03-19
Add names to some jobs in unit tests.

scrosby:
- Name some jobs.
- 

- Was having trouble figuring out why my test was failing. This makes it easier.


shamsimam:
`create-dummy-job` already takes a `:name` parameter, why not pass it directly, e.g.
```
job1 (create-named-dummy-job "job1" conn
  :custom-executor? false
  :job-state state
  :name "dummy_job-job1"
  :submit-time (Date.)
  :user "u1")
```

scrosby:
Yes! Thanks. Respinning this to add in the missing names.

PR #769, opened 2018-03-19
Expand test coverage with name-filter-fn.

scrosby:
- Add some unit tests for named-filter-fn == nil as well as non-null.
- 
- 

- We test it == (constantly true), but name-filter-str->name-filter-fn only sets it to a regexp matcher closure or nil. 
- This exercises the special case nil in the query logic which wouldn't be exercised with the 'constantly true' existing implementation.

### DIFF
--- a/scheduler/test/cook/test/mesos/util.clj
+++ b/scheduler/test/cook/test/mesos/util.clj
@@ -398,21 +398,50 @@
             _ (Thread/sleep 5)
             end-time (Date.)
             states [(name state)]
-            name (constantly true)]
+            match-any-name-fn (constantly true)
+            match-no-name-fn (constantly false)
+            ]
         (testing (str "get " state " jobs")
-          (is (= 2 (count (util/get-jobs-by-user-and-states (d/db conn) "u1" states start-time half-way-time 10 name nil))))
-          (is (= 1 (count (util/get-jobs-by-user-and-states (d/db conn) "u1" states start-time half-way-time 1 name nil))))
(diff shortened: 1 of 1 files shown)


scheduler/test/cook/test/mesos/util.clj
```diff
@@ -393,21 +393,50 @@
             _ (Thread/sleep 5)
             end-time (Date.)
             states [(name state)]
-            name (constantly true)]
+            namet (constantly true)
+            namef (constantly false)
```
dposada: Maybe `match-any-name` and `match-no-name`?

scheduler/test/cook/test/mesos/util.clj
```diff
@@ -393,21 +393,50 @@
...
+            ]
```
dposada: The close bracket should be on the previous line

scheduler/test/cook/test/mesos/util.clj
scrosby: Renamed to match-any-name-fn and match-no-name-fn.

dposada:
Needs rebase

scheduler/test/cook/test/mesos/util.clj
dposada: @scrosby Thoughts on this?

dposada: @scrosby Thoughts on this?

scrosby: Fixed in recent push.

PR #771, opened 2018-03-20
New query functions (take 2)

scrosby:
Vastly improved query functions. I tested the new git-active-jobs-by-use-and-state on a few thousand production queries and it was identical each time. 

- get-active-jobs-by-user-and-state is rewritten on top of seek-datums and is vastly faster, I see up p50's 5x-100x faster. p95's are 2x-5x faster.
- get-completed-jobs-by-user has had small tweaks made to improve it; removing reflection and minimizing datomic queries.
- Collect a histogram of metrics of time-to-search, as a function of the time range being searched.

- Much greater performance.

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -65,6 +65,7 @@
 (defonce ^Cache job-ent->resources-cache (new-cache))
 (defonce ^Cache categorize-job-cache (new-cache))
 (defonce ^Cache task-ent->user-cache (new-cache))
+(defonce ^Cache job-ent->user-cache (new-cache))
 (defonce ^Cache task->feature-vector-cache (new-cache))
 
 (defn get-all-resource-types
@@ -322,6 +323,13 @@
 
 (timers/deftimer [cook-mesos scheduler get-completed-jobs-by-user-duration])
 
+(defn job-ent->user
+  [job-ent]
+  (let [job-ent->user-miss
+        (fn [job-ent]
+          (:job/user job-ent))]
+    (lookup-cache-datomic-entity! job-ent->user-cache job-ent->user-miss job-ent)))
+
 (defn job-ent->state
   "Given a job entity, returns the corresponding 'state', which means
   calling with a completed job will return either success or failed,
@@ -338,20 +346,67 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+;; get-jobs-by-user-and-state-and-submit-by-state is a bit opaque
+;; because it is reaching into datomic internals. Here is a quick explanation.
+;; seek-datoms provides a pointer into the raw datomic indices
+;; that we can then seek through. We set the pointer to look
+;; through the avet index, with attribute :job/state, seek to
+;; state and then seek to the entity id that *would* have been
+;; created at expanded start. This works because the submission
+;; time (which we sort on) is correlated with the entitiy id.
+;; We then seek through the list of all jobs in that state,
+;; then filtering to the target user.
+;;
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
+;;
+;; This function is O(# jobs in that state in the time range)
+(defn get-jobs-by-user-and-state-and-submit-by-state
+  "Returns all jobs for a particular user in the specified state
+   and timeframe, without a custom executor. Returns lazy output."
+  [db user start end state-keyword]
+  (let [;; Expand the time range so that clock skew between cook
+        ;; and datomic doesn't cause us to miss jobs
+        ;; 1 hour was picked because a skew larger than that would be
+        ;; suspicious
+        expanded-start (Date. (- (.getTime start)
+                                 (-> 1 t/hours t/in-millis)))
+        expanded-end (Date. (+ (.getTime end)
+                               (-> 1 t/hours t/in-millis)))
+        entid-start (d/entid-at db :db.part/user expanded-start)
+        entid-end (d/entid-at db :db.part/user expanded-end)
+        job-state-entid (d/entid db :job/state)
+        state-entid (d/entid db state-keyword)
+        ^long start-ms (.getTime start)
+        ^long end-ms (.getTime end)]
+    (->> (d/seek-datoms db :avet :job/state state-entid entid-start)
+         (take-while #(and (< (:e %) entid-end)
+                           (= (:a %) job-state-entid)
+                           (= (:v %) state-entid)))
+         (map :e)
+         (map (partial d/entity db))
+         (filter #(= (job-ent->user %) user))
+         (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                    (and (<= start-ms submit-ms)
+                                                  (< submit-ms end-ms)))))))
+
 ;; get-completed-jobs-by-user is a bit opaque because it is
 ;; reaching into datomic internals. Here is a quick explanation.
-;; seek-datoms provides a pointer into the raw datomic indices 
+;; seek-datoms provides a pointer into the raw datomic indices
 ;; that we can then seek through. We set the pointer to look
-;; through the avet index, with attribute :job/user, seek to 
+;; through the avet index, with attribute :job/user, seek to
 ;; user and then seek to the entity id that *would* have been
-;; created at expanded start. 
+;; created at expanded start.
 ;; This works because the submission time and job/user field
 ;; are set at the same time, in "real" time. This means that
 ;; jobs submitted after `start` will have been created after
 ;; expanded start
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
 (defn get-completed-jobs-by-user
   "Returns all completed job entities for a particular user
-   in the specified timeframe, without a custom executor."
+   in the specified timeframe, without a custom executor. Supports looking up based
+   on task state 'success' and 'failed' if passed into 'state'"
   [db user start end limit state name-filter-fn include-custom-executor?]
   (timers/time!
     get-completed-jobs-by-user-duration
@@ -363,22 +418,29 @@
                                    (-> 1 t/hours t/in-millis)))
           expanded-end (Date. (+ (.getTime end)
                                  (-> 1 t/hours t/in-millis)))
+          entid-start (d/entid-at db :db.part/user expanded-start)
+          entid-end (d/entid-at db :db.part/user expanded-end)
+          job-user-entid (d/entid db :job/user)
+          ^long start-ms (.getTime start)
+          ^long end-ms (.getTime end)
           jobs
-          (->> (d/seek-datoms db :avet :job/user user (d/entid-at db :db.part/user expanded-start))
-               (take-while #(and (< (.e %) (d/entid-at db :db.part/user expanded-end))
-                                 (= (.a %) (d/entid db :job/user))
-                                 (= (.v %) user)))
-               (map #(.e %))
+          (->> (d/seek-datoms db :avet :job/user user entid-start)
+               (take-while #(and (< (:e %) entid-end)
+                                 (= (:a %) job-user-entid)
+                                 (= (:v %) user)))
+               (map #(:e %))
                (map (partial d/entity db))
-               (filter #(<= (.getTime start) (.getTime (:job/submit-time %))))
-               (filter #(< (.getTime (:job/submit-time %)) (.getTime end)))
+               (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                          (and (<= start-ms submit-ms)
+                               (< submit-ms end-ms))))
                (filter #(= :job.state/completed (:job/state %))))]
       (->>
         (cond->> jobs
                  (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
                  (instance-states state) (filter #(= state (job-ent->state %)))
                  name-filter-fn (filter #(name-filter-fn (:job/name %))))
-        (take limit)))))
+        (take limit)
+        doall))))
 
 (defn uncommitted?
   "Returns true if the given job's commit latch is not committed"
@@ -391,30 +453,37 @@
 ;; running and waiting jobs, so it's implemented that way to keep things simple.
 (defn get-active-jobs-by-user-and-state
   "Returns all jobs for a particular user in the specified state
-   and timeframe, without a custom executor.
-   Note that this query is not performant for completed jobs, use
-   get-completed-jobs-by-user instead."
-  [db user start end state name-filter-fn include-custom-executor?]
+   and timeframe, without a custom executor. This query works for state
+   waiting and running only."
+  [db user start end limit state name-filter-fn include-custom-executor?]
   (let [state-keyword (case state
                         "running" :job.state/running
                         "waiting" :job.state/waiting)
-        jobs
-        (timers/time!
-          (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-duration")])
-          (->> (q '[:find [?j ...]
-                    :in $ ?user ?state ?start ?end
-                    :where
-                    [?j :job/state ?state]
-                    [?j :job/user ?user]
-                    [?j :job/submit-time ?t]
-                    [(<= ?start ?t)]
-                    [(< ?t ?end)]]
-                  db user state-keyword start end)
-               (map (partial d/entity db))))]
-    (cond->> jobs
-             (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
-             name-filter-fn (filter #(name-filter-fn (:job/name %)))
-             (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))))
+        ;; This is used to compute the histogram bucket to put in. take the time range in hours, with a one minute leeway. E.g., this counts
+        ;; a requested time of 24 hours + 3 seconds into the 24-hour bucket.
+        millis-diff (max 0 (- (.getTime end) (.getTime start) (-> 1 t/minutes t/in-millis)))
+        hours-diff (/ millis-diff (-> 1 t/hours t/in-millis))
+        ; A histogram bucket, so that we can see the query time distribution as a function of the range of time being searched over.
+        bucket (cond
+                 (<= hours-diff 6) 6
+                 (<= hours-diff 12) 12
+                 (<= hours-diff 24) 24
+                 (<= hours-diff 36) 36
+                 (<= hours-diff 48) 48
+                 (<= hours-diff 72) 72
+                 :else 999)]
+    (timers/time!
+      (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-duration")])
+      (timers/time!
+        (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-by-state-" bucket "-duration")])
+        (->>
+          (cond->> (get-jobs-by-user-and-state-and-submit-by-state db user start end state-keyword)
+                   (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
+                   name-filter-fn (filter #(name-filter-fn (:job/name %)))
+                   (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))
+          ; No need to sort. We're traversing in entity_id order, so in time order.
+          (take limit)
+          doall)))))
 
 (defn get-jobs-by-user-and-states
   "Returns all jobs for a particular user in the specified states
@@ -424,12 +493,14 @@
                             (if (#{"completed" "success" "failed"} state)
                               (get-completed-jobs-by-user db user start end limit state
                                                           name-filter-fn include-custom-executor?)
-                              (get-active-jobs-by-user-and-state db user start end state
+                              (get-active-jobs-by-user-and-state db user start end limit state
                                                                  name-filter-fn include-custom-executor?)))
         jobs-by-state (mapcat get-jobs-by-state states)]
-    (->> jobs-by-state
-         (sort-by :job/submit-time)
-         (take limit))))
+    (timers/time!
+      (timers/timer ["cook-mesos" "scheduler" "get-jobs-by-user-and-states-duration"])
+      (->> jobs-by-state
+           (sort-by :job/submit-time)
+           (take limit)))))
 
 
 (defn jobs-by-user-and-state
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -459,33 +459,17 @@
   [db user start end limit state name-filter-fn include-custom-executor?]
   (let [state-keyword (case state
                         "running" :job.state/running
-                        "waiting" :job.state/waiting)
-        ;; This is used to compute the histogram bucket to put in. take the
-        ;; time range in hours, with a one minute leeway. E.g., this counts
-        ;; a requested time of 24 hours + 3 seconds into the 24-hour bucket.
-        millis-diff (max 0 (- (.getTime end) (.getTime start) (-> 1 t/minutes t/in-millis)))
-        hours-diff (/ millis-diff (-> 1 t/hours t/in-millis))
-        ; A histogram bucket, so that we can see the query time distribution as a function of the range of time being searched over.
-        bucket (cond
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -337,20 +345,45 @@
...
+                                                    (= (:v %) state-entid)))
+                                  (map :e)
+                                  (map (partial d/entity db))
+                                  (filter #(= (job-ent->user %) user))
+                                  (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                                             (and (<= start-ms submit-ms)
+                                                  (< submit-ms end-ms))))
+                                  (filter #(false? (:job/custom-executor %))))]
+                  ;; If we have a limit, take it here (before doall, so that our metrics are robust)
+                  (if limit (doall (take limit result)) result))))
```
wyegelwel: Why do we only use `doall` in the case of limit. If there is no limit, then our metrics will be skewed.

Also, I think you can remove the comment.

scheduler/src/cook/mesos/util.clj
```diff
@@ -362,51 +395,55 @@
...
+                 (<= hours-diff 36) 36
+                 (<= hours-diff 48) 48
+                 (<= hours-diff 72) 72
+                 :else 999)
+        ; If we have a name filter, we cannot push the limit into the query, but have to do it after the filter.
+        adaptedlimit (if name-filter-fn nil limit)
+        jobs (get-jobs-by-user-and-state-and-submit-by-state db user start end state-keyword adaptedlimit bucket)]
+    (cond->> jobs
+             name-filter-fn (filter #(name-filter-fn (:job/name %)))
+             name-filter-fn (take limit))))
```
wyegelwel: Should this be checking if there is limit as opposed to name-filter-fn?

wyegelwel: Oh, I see. I didn't read the comment on adaptedlimit.

I would recommend not doing the limit in the `get-jobs-by-user-and-state-and-submit-by-state` in any case and handling it out here in the caller. As it stands, you are required to take some odd steps to make it work (the adaptedlimit) which wouldn't be necessary if you pushed off the functionality until later.

scheduler/src/cook/mesos/util.clj
```diff
@@ -337,20 +345,45 @@
...
+                                  (map (partial d/entity db))
+                                  (filter #(= (job-ent->user %) user))
+                                  (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                                             (and (<= start-ms submit-ms)
+                                                  (< submit-ms end-ms))))
+                                  (filter #(false? (:job/custom-executor %))))]
+                  ;; If we have a limit, take it here (before doall, so that our metrics are robust)
+                  (if limit (doall (take limit result)) result))))
+
+;; This differs from get-active-jobs-by-user-and-state as it is also looking up based on task state.
```
wyegelwel: This doesn't capture the only reason it differs. You should also include the comment block you removed below about why this one is querying on user instead of state.

scheduler/src/cook/mesos/util.clj
```diff
@@ -337,20 +345,45 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
-;; get-completed-jobs-by-user is a bit opaque because it is
```
wyegelwel: I think this comment is probably worth keeping. I believe it was added because reviewers when it was added requested an explanation of why the code worked. @dposada what do you think?

scheduler/src/cook/mesos/util.clj
wyegelwel: See my comment below. I recommend you remove the limit on this utility function and have the limit be done by the caller.

scheduler/src/cook/mesos/util.clj
```diff
@@ -362,51 +395,55 @@
...
-;; The standard datomic query performs reasonably well on the smaller set of
-;; running and waiting jobs, so it's implemented that way to keep things simple.
 (defn get-active-jobs-by-user-and-state
   "Returns all jobs for a particular user in the specified state
-   and timeframe, without a custom executor.
-   Note that this query is not performant for completed jobs, use
-   get-completed-jobs-by-user instead."
-  [db user start end state name-filter-fn]
+   and timeframe, without a custom executor. This query looks for all
+   jobs by job state only (i.e., no 'success' or 'failed')"
```
wyegelwel: I recommend removing the line "This query looks for all jobs ..." and instead say "running" and "waiting" are valid state inputs. Also, it is probably worth referencing the get-completed-jobs-by-user.

scheduler/src/cook/mesos/util.clj
```diff
@@ -362,51 +395,55 @@
...
-                    [?j :job/user ?user]
-                    [?j :job/submit-time ?t]
-                    [(<= ?start ?t)]
-                    [(< ?t ?end)]
-                    [?j :job/custom-executor false]]
-                  db user state-keyword start end)
-               (map (partial d/entity db))))]
-    (cond->> jobs name-filter-fn (filter #(name-filter-fn (:job/name %))))))
+        millis-diff (max 0 (- (.getTime end) (.getTime start) 60000)) ; 60000ms is so that we get queries that are just a tiny bit over a bucket.
+        hours-diff (/ millis-diff 3600000)
```
wyegelwel: `Please use the `(-> 1 t/hours t/in-millis)` so it is easier to understand what time magnitude we are using

scheduler/src/cook/mesos/util.clj
```diff
@@ -362,51 +395,55 @@
...
-                    [?j :job/state ?state]
-                    [?j :job/user ?user]
-                    [?j :job/submit-time ?t]
-                    [(<= ?start ?t)]
-                    [(< ?t ?end)]
-                    [?j :job/custom-executor false]]
-                  db user state-keyword start end)
-               (map (partial d/entity db))))]
-    (cond->> jobs name-filter-fn (filter #(name-filter-fn (:job/name %))))))
+        millis-diff (max 0 (- (.getTime end) (.getTime start) 60000)) ; 60000ms is so that we get queries that are just a tiny bit over a bucket.
```
wyegelwel: Can you explain the 60000ms thing? I'm pretty confused.

scheduler/src/cook/mesos/util.clj
dposada: I agree, this comment seems very helpful to me

scheduler/src/cook/mesos/util.clj
scrosby: The reasoning was to unlazy as soon as possible ot minimize metric skew. Had I not done this I'd be measuring the same '1ms' metrics that we see for 'completd jobs' in production. :)

In any case, good eye. I have a better fix for this and the next.

scheduler/src/cook/mesos/util.clj
scrosby: See above response.

scheduler/src/cook/mesos/util.clj
scrosby: Resurrected

scheduler/src/cook/mesos/util.clj
scrosby: Fixed

scheduler/src/cook/mesos/util.clj
scrosby: Done

scheduler/src/cook/mesos/util.clj
scrosby: ACK

scrosby:
FYI: Had to rebase this with significant changes to deal with conflicting patch.

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,20 +346,67 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+;; get-jobs-by-user-and-state-and-submit-by-state is a bit opaque
```
dposada: This function name is a bit of a mouthful. Maybe just `get-jobs-by-user-and-state`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,20 +346,67 @@
...
+;; We then seek through the list of all jobs in that state,
+;; then filtering to the target user.
+;;
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
+;;
+;; This function is O(# jobs in that state in the time range)
+(defn get-jobs-by-user-and-state-and-submit-by-state
+  "Returns all jobs for a particular user in the specified state
+   and timeframe, without a custom executor. Returns lazy output."
```
dposada: This isn't doing any filtering based on custom executor

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,20 +346,67 @@
...
+    (->> (d/seek-datoms db :avet :job/state state-entid entid-start)
+         (take-while #(and (< (:e %) entid-end)
+                           (= (:a %) job-state-entid)
+                           (= (:v %) state-entid)))
+         (map :e)
+         (map (partial d/entity db))
+         (filter #(= (job-ent->user %) user))
+         (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                    (and (<= start-ms submit-ms)
+                                                  (< submit-ms end-ms)))))))
```
dposada: Seems like indentation is off here

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,20 +346,67 @@
...
+;; that we can then seek through. We set the pointer to look
+;; through the avet index, with attribute :job/state, seek to
+;; state and then seek to the entity id that *would* have been
+;; created at expanded start. This works because the submission
+;; time (which we sort on) is correlated with the entitiy id.
+;; We then seek through the list of all jobs in that state,
+;; then filtering to the target user.
+;;
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
```
dposada: It seems to say this "differs" from `get-active-jobs-by-user-and-state` given that `get-active-jobs-by-user-and-state` calls this function

scheduler/src/cook/mesos/util.clj
```diff
@@ -363,22 +418,29 @@
...
-          (->> (d/seek-datoms db :avet :job/user user (d/entid-at db :db.part/user expanded-start))
-               (take-while #(and (< (.e %) (d/entid-at db :db.part/user expanded-end))
-                                 (= (.a %) (d/entid db :job/user))
-                                 (= (.v %) user)))
-               (map #(.e %))
+          (->> (d/seek-datoms db :avet :job/user user entid-start)
+               (take-while #(and (< (:e %) entid-end)
+                                 (= (:a %) job-user-entid)
+                                 (= (:v %) user)))
+               (map #(:e %))
```
dposada: Can't this just be `(map :e)`

scheduler/src/cook/mesos/util.clj
```diff
@@ -363,22 +418,29 @@
...
+                               (< submit-ms end-ms))))
                (filter #(= :job.state/completed (:job/state %))))]
       (->>
         (cond->> jobs
                  (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
                  (instance-states state) (filter #(= state (job-ent->state %)))
                  name-filter-fn (filter #(name-filter-fn (:job/name %))))
-        (take limit)))))
+        (take limit)
+        doall))))
```
dposada: Why do we need the `doall`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -391,30 +453,37 @@
 ;; running and waiting jobs, so it's implemented that way to keep things simple.
 (defn get-active-jobs-by-user-and-state
   "Returns all jobs for a particular user in the specified state
-   and timeframe, without a custom executor.
-   Note that this query is not performant for completed jobs, use
-   get-completed-jobs-by-user instead."
-  [db user start end state name-filter-fn include-custom-executor?]
+   and timeframe, without a custom executor. This query works for state
```
dposada: "without a custom executor" isn't really accurate since the caller gets to choose

scheduler/src/cook/mesos/util.clj
```diff
@@ -391,30 +453,37 @@
...
-                    [?j :job/submit-time ?t]
-                    [(<= ?start ?t)]
-                    [(< ?t ?end)]]
-                  db user state-keyword start end)
-               (map (partial d/entity db))))]
-    (cond->> jobs
-             (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
-             name-filter-fn (filter #(name-filter-fn (:job/name %)))
-             (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))))
+        ;; This is used to compute the histogram bucket to put in. take the time range in hours, with a one minute leeway. E.g., this counts
```
dposada: Can we make this line shorter?

scheduler/src/cook/mesos/util.clj
```diff
@@ -391,30 +453,37 @@
...
+                 (<= hours-diff 12) 12
+                 (<= hours-diff 24) 24
+                 (<= hours-diff 36) 36
+                 (<= hours-diff 48) 48
+                 (<= hours-diff 72) 72
+                 :else 999)]
+    (timers/time!
+      (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-duration")])
+      (timers/time!
+        (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-by-state-" bucket "-duration")])
```
dposada: Why are we doing the bucket timer for active but not for completed?

scheduler/src/cook/mesos/util.clj
```diff
@@ -391,30 +453,37 @@
...
+        (->>
+          (cond->> (get-jobs-by-user-and-state-and-submit-by-state db user start end state-keyword)
+                   (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
+                   name-filter-fn (filter #(name-filter-fn (:job/name %)))
+                   (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))
+          ; No need to sort. We're traversing in entity_id order, so in time order.
```
dposada: Why the comment about sorting here but not in the completed version?

scheduler/src/cook/mesos/util.clj
```diff
@@ -322,6 +323,13 @@
 
 (timers/deftimer [cook-mesos scheduler get-completed-jobs-by-user-duration])
 
+(defn job-ent->user
```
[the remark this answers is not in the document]
scrosby: Done

scheduler/src/cook/mesos/util.clj
```diff
@@ -322,6 +323,13 @@
...
+  [job-ent]
+  (let [job-ent->user-miss
+        (fn [job-ent]
+          (:job/user job-ent))]
```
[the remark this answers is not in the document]

scheduler/src/cook/mesos/util.clj
```diff
@@ -65,6 +65,7 @@
 (defonce ^Cache job-ent->resources-cache (new-cache))
 (defonce ^Cache categorize-job-cache (new-cache))
 (defonce ^Cache task-ent->user-cache (new-cache))
+(defonce ^Cache job-ent->user-cache (new-cache))
```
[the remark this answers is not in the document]
scrosby: Yes, but I'd like to split it off from this work, e.g., in particular, we need to wire it into cache flushing. Tracking it under https://github.com/twosigma/Cook/issues/779

scheduler/src/cook/mesos/util.clj
scrosby: Renamed shorter

scheduler/src/cook/mesos/util.clj
scrosby: Thats done by the caller, get-active-jobs-by-user-and-state as a postprocess.

scheduler/src/cook/mesos/util.clj
scrosby: Fixed.

scheduler/src/cook/mesos/util.clj
scrosby: Good catch. Merge/rebase bug.

scheduler/src/cook/mesos/util.clj
scrosby: Yes.

scheduler/src/cook/mesos/util.clj
scrosby: Force the metric to not be junk. If we didn't, then this entire body is lazy and our metrics would 'improve' to about 1ms.

scrosby: https://github.com/twosigma/Cook/issues/780

scheduler/src/cook/mesos/util.clj
scrosby: Done

scheduler/src/cook/mesos/util.clj
scrosby: We can as part of https://github.com/twosigma/Cook/issues/780. Right now, no point, completed jobs metric is junk.

scheduler/src/cook/mesos/util.clj
scrosby: Copied.

scrosby:
I'll ask about exporting my replay code. For obvious reasons, it can only be run internally, so was developed internally. 

The comparsion test&evaluation code can be seen can be seen in my wip/newquery branch. https://github.com/scrosby/Cook/tree/wip/newquery There's nothing deep there. I'm happy to put it someplace in the repository, but it'll bitrot rapidly.

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
dposada: Sounds good

scheduler/src/cook/mesos/util.clj
dposada: Right, I was just commenting on the docstring saying "without a custom executor"

scheduler/src/cook/mesos/util.clj
dposada: 👍

scheduler/src/cook/mesos/util.clj
dposada: Doesn't it make sense to introduce the bucket timers at the same time for active and completed though?

dposada:
Where in your `newquery` branch should I be looking?

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
dposada: I still don't see the docstring

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
dposada: I don't see this change

dposada:
@scrosby Are you missing a push? A few of the comments you marked as "Done" don't seem to have changed

scrosby:
Except for one unfixed comment, everything is up to date with respect to my tree. The before/after evaluation code is up at https://github.com/scrosby/Cook/tree/wip/newquery

scheduler/src/cook/mesos/util.clj
```diff
@@ -391,45 +449,54 @@
...
         jobs-by-state (mapcat get-jobs-by-state states)]
-    (->> jobs-by-state
-         (sort-by :job/submit-time)
-         (take limit))))
+    (timers/time!
+      (timers/timer ["cook-mesos" "scheduler" "get-jobs-by-user-and-states-duration"])
+      (->> jobs-by-state
+           (sort-by :job/submit-time)
+           (take limit)
+           (doall)))))
```
dposada: You don't need the parentheses around `doall` here

scheduler/src/cook/mesos/util.clj
```diff
@@ -363,22 +413,30 @@
...
+               (take-while #(and (< (:e %) entid-end)
+                                 (= (:a %) job-user-entid)
+                                 (= (:v %) user)))
+               (map :e)
                (map (partial d/entity db))
-               (filter #(<= (.getTime start) (.getTime (:job/submit-time %))))
-               (filter #(< (.getTime (:job/submit-time %)) (.getTime end)))
+               (filter #(let [^long submit-ms (.getTime (:job/submit-time %))]
+                          (and (<= start-ms submit-ms)
+                               (< submit-ms end-ms))))
```
dposada: Seems like it would be useful to `defn` this function since we use it twice, e.g. `(submitted-in-range? start end job)`

scheduler/src/cook/mesos/util.clj
scrosby: We need to inject doall's in completed in the right spots so that we get good answers. I agree that we want to do the work, but don't want to grow the scope of this JIRA.

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
scrosby: I pushed after making all of the changes. See update message below.

scheduler/src/cook/mesos/util.clj

scheduler/src/cook/mesos/util.clj
scrosby: Done

wyegelwel:
Do we have tests for these functions already?

scrosby:
Thats odd. The failing tests are the result of merging this with the (new head, with the metrics), but when the revert was committed, the tests weren't rerun.

@wyegelwel Yes, these are tested as part of the normal query tests. E.g., https://github.com/twosigma/Cook/blob/master/scheduler/test/cook/test/mesos/util.clj#L355

I also did a before&after test running 4k production queries (via the REPL) and checked that both versions returned the same data. (That regression discovered a long-standing bug about commit executors.)

scheduler/src/cook/mesos/util.clj
dposada: I would vote to make this consistent as part of this PR, in the sense that we either have the bucket timers for both or for neither.

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,21 +344,73 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+
+(defn job-submit-time-between
+  "Returns true if the job-ent's :job/submit-time is between start-ms (inclusive) and
+  end-ms (exclusive)"
+  [job-ent ^long start-ms ^long end-ms]
+  (let [^Date submit-time (:job/submit-time job-ent)
```
dposada: Can we destructure this from the function arg?

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,21 +344,72 @@
...
+;; through the avet index, with attribute :job/user, seek to
 ;; user and then seek to the entity id that *would* have been
-;; created at expanded start. 
+;; created at expanded start.
 ;; This works because the submission time and job/user field
 ;; are set at the same time, in "real" time. This means that
 ;; jobs submitted after `start` will have been created after
 ;; expanded start
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
```
pschorf: Can you add a comment that this is O(number of jobs the user has submitted) to mirror the one above?

scheduler/src/cook/mesos/util.clj
pschorf: I also think this could be broken out into a helper function if we were doing both at the same time. I'd probably vote to just leave it out for now.

scrosby: Code ripped out.

scheduler/src/cook/mesos/util.clj
```diff
@@ -338,21 +344,73 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+
+(defn job-submit-time-between
+  "Returns true if the job-ent's :job/submit-time is between start-ms (inclusive) and
+  end-ms (exclusive)"
+  [{^Date submit-time :job/submit-time} ^long start-ms ^long end-ms]
```
dposada: You should be able to do `{:keys [job/submit-time]}`

scheduler/src/cook/mesos/util.clj
scrosby: Done

PR #773, opened 2018-03-21
Kill all running tasks for completed jobs

pschorf:
- When starting `monitor-tx-report-queue`, first query for any running instances for completed jobs and attempt to kill them

Some instances may have been "missed" during the leadership change and will still be running when the new leader takes over and never killed.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -490,7 +490,7 @@
                            [?i :instance/task-id ?task-id]]
                          query-db [:instance.status/unknown :instance.status/running])]
     (doseq [[task-id] tasks-to-kill]
-      (if-let [driver @driver-ref]
+      (when-let [driver @driver-ref]
         (try
           (log/info "Attempting to kill task" task-id "from already completed job")
           (meters/mark! tx-report-queue-tasks-killed)
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -504,31 +504,31 @@
                            (async/go
                              (timers/start-stop-time! ; Use this in go blocks, time! doesn't play nice
                                tx-report-queue-processing-duration
-                               (let [{:keys [tx-data db-after db-before]} tx-report
-                                     db (db conn)]
+                               (let [{:keys [tx-data db-after]} tx-report]
                                  (when (< query-basis (d/basis-t db-after))
-                                   (meters/mark! tx-report-queue-datoms (count tx-data))
-                                   ;; Monitoring whether a job is completed.
-                                   (doseq [{:keys [e a v]} tx-data]
-                                     (try
-                                       (when (and (= a (d/entid db :job/state))
-                                                  (= v (d/entid db :job.state/completed)))
-                                         (meters/mark! tx-report-queue-job-complete)
-                                         (doseq [[task-id] (q '[:find ?task-id
-                                                                :in $ ?job [?status ...]
-                                                                :where
-                                                                [?job :job/instance ?i]
-                                                                [?i :instance/status ?status]
-                                                                [?i :instance/task-id ?task-id]]
-                                                              db e [:instance.status/unknown
-                                                                    :instance.status/running])]
-                                           (if-let [driver @driver-ref]
-                                             (do (log/info "Attempting to kill task" task-id "due to job completion")
-                                                 (meters/mark! tx-report-queue-tasks-killed)
-                                                 (mesos/kill-task! driver {:value task-id}))
-                                             (log/error "Couldn't kill task" task-id "due to no Mesos driver!"))))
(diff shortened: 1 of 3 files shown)


scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -2175,5 +2175,36 @@
...
+          instance-ent (d/entity (d/db conn) instance-id-1)]
+      (is (= :job.state/completed (:job/state job-ent)))
+      (is (= :instance.status/unknown (:instance/status instance-ent))))
+    (let [[report-mult close-fn] (cook.datomic/create-tx-report-mult conn)
+          transaction-chan (async/chan (async/sliding-buffer 4096))
+          _ (async/tap report-mult transaction-chan)
+          kill-fn (cook.mesos.scheduler/monitor-tx-report-queue transaction-chan conn driver-ref)]
+      (cook.mesos/kill-job conn [(:job/uuid (d/entity (d/db conn) job-id-2))])
+      ;; TODO - find a better way to wait for this
+      (Thread/sleep 100)
```
dposada: Can we use the "wait until" pattern to wait until `killed-tasks` has what we expect?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -478,38 +478,57 @@
...
+        tasks-to-kill (q '[:find ?task-id
+                           :in $ [?status ...]
+                           :where
+                           [?i :instance/status ?status]
+                           [?job :job/instance ?i]
+                           [?job :job/state :job.state/completed]
+                           [?i :instance/task-id ?task-id]]
+                         query-db [:instance.status/unknown :instance.status/running])]
+    (doseq [[task-id] tasks-to-kill]
+      (if-let [driver @driver-ref]
```
dposada: Can this be `when-let`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -478,38 +478,57 @@
...
+        (try
+          (log/info "Attempting to kill task" task-id "from already completed job")
```
dposada: Would it be useful to log the job uuid as well?

scheduler/src/cook/mesos/scheduler.clj
pschorf: Done

scheduler/src/cook/mesos/scheduler.clj
pschorf: I think we could log it, but not sure if the cost of querying for it is worth it since it's easy when debugging to go from one to the other, and it's not going to change.

scheduler/test/cook/test/mesos/scheduler.clj
pschorf: IMHO, it's overkill here. The 100 ms should be plenty, I just was trying to think of something to programmatically make sure we have gone through the entire transaction queue

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -478,38 +478,57 @@
...
     (async/go
       (loop []
         (async/alt!
           tx-report-chan ([tx-report]
                            (async/go
                              (timers/start-stop-time! ; Use this in go blocks, time! doesn't play nice
                                tx-report-queue-processing-duration
-                               (let [{:keys [tx-data db-before]} tx-report
+                               (let [{:keys [tx-data db-after db-before]} tx-report
                                      db (db conn)]
```
shamsimam: nitpick: Should the creation of `db` be moved inside the following `when`?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -478,38 +478,57 @@
...
+            (log/error e (str "Failed to kill task" task-id))))))
     (async/go
       (loop []
         (async/alt!
           tx-report-chan ([tx-report]
                            (async/go
                              (timers/start-stop-time! ; Use this in go blocks, time! doesn't play nice
                                tx-report-queue-processing-duration
-                               (let [{:keys [tx-data db-before]} tx-report
+                               (let [{:keys [tx-data db-after db-before]} tx-report
```
shamsimam: nitpick: We can move declaration of `query-basis` into this `let` block.

scheduler/src/cook/mesos/scheduler.clj
dposada: Makes sense

scheduler/test/cook/test/mesos/scheduler.clj
dposada: What's the downside to using "wait until"?

dposada: It just seems like every sleep we introduce is asking for flakiness at some point down the road, maybe on Travis

pschorf: Agreed, added

PR #781, opened 2018-03-23
Lets capture the time it takes to get a list endpoint out.

scrosby:

-  Put a metric on the actual response generation time for the `/list` API call

- We don't actually metric this user-visible time.


scrosby:
Any suggestions on the best way to test this?

shamsimam:
@scrosby I don't think we really test output of metrics.

dposada:
This was merged with a broken build and master is breaking now, e.g. https://travis-ci.org/twosigma/Cook/jobs/357585822. Can we revert?

PR #784, opened 2018-03-26
Lets capture the time it takes to get a list endpoint out.

scrosby:

-  Put a metric on the actual response generation time for the `/list` API call

- We don't actually metric this user-visible time.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -2159,8 +2159,10 @@
     :handle-malformed ::error
     :handle-forbidden ::error
     :handle-ok (fn [ctx]
-                 (let [job-uuids (list-jobs db false ctx)]
-                   (mapv (partial fetch-job-map db framework-id) job-uuids)))))
+                 (timers/time!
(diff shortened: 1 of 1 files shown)


dposada:
Is this still wip?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2159,8 +2159,10 @@
     :handle-malformed ::error
     :handle-forbidden ::error
     :handle-ok (fn [ctx]
-                 (let [job-uuids (list-jobs db false ctx)]
-                   (mapv (partial fetch-job-map db framework-id) job-uuids)))))
+                 (timers/time!
+                   (timers/timer ["cook-scheduler" "handler" "list-endpoint-duration"])
+                   (let [job-uuids (list-jobs db false ctx)]
+                     (doall (mapv (partial fetch-job-map db framework-id) job-uuids)))))))
```
dposada: I don't think `mapv` is lazy

pschorf: 👍

scrosby: Its not. :) Fixed.

Issue #793, opened 2018-03-29
Gpu support in CLI needed


dPeS:
Hi,

When I added `--gpus` option here [1]:

`cs submit --executor mesos --gpus 1 sleep 10`
works ok (task is finished on gpu host), but:

```
[centos@ip-192-168-1-20]$ sudo cs submit --executor mesos --gpus 0 sleep 1
Attempting to submit on KKK cluster...
Job submission failed on KKK:
{"jobs": [{"gpus": "(not (pos? 0))"}]}
Job submission failed on all of your configured clusters.
```
any clues?

[1] https://github.com/twosigma/Cook/blob/master/cli/cook/subcommands/submit.py#L226

dposada:
Hey @dPeS, thanks for the issue. The error you're getting is due to the fact that we have a server-side schema validation that doesn't allow gpus to be non-positive:

https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/api.clj#L250

Can you leave off the `--gpus` flag altogether for the case where you don't need gpus?

dPeS:
Hi Daniel,

Sure I'm good here - that doesn't block me - just reporting.
Please remember to add `--gpus` option to `submit.py`

dposada:
Sounds good. I will take a look at this. Glad to hear you're not blocked.

dPeS:
@dposada thanks, so I'm good here!

dposada:
Resolved by PR #794

PR #794, opened 2018-03-29
Adds support for submitting jobs with gpus

dposada:

- adding `--gpus` flag to `cs submit`
- adding an integration test for `--gpus`

This allows users to submit jobs with gpus using `cs`. See issue #793 for more context.

### DIFF
--- a/cli/cook/subcommands/submit.py
+++ b/cli/cook/subcommands/submit.py
@@ -8,7 +8,7 @@ import uuid
 import requests
 
 from cook import colors, http, metrics, version
-from cook.util import deep_merge, is_valid_uuid, read_lines, print_info, current_user, guard_no_cluster
+from cook.util import deep_merge, is_valid_uuid, read_lines, print_info, current_user, guard_no_cluster, check_positive
 
 
 def parse_raw_job_spec(job, r):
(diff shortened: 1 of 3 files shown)


dposada:
@dPeS - Any chance you can take this for a spin and see if it works in your environment?

dPeS:
@dposada works as expected... but I see 1 issue here:

`cs submit --gpus 1 sleep 1` works ok now, but:

`cs submit sleep 1` waits infinitely since there is no host without GPU

is it expected?

dposada:
Currently, non-gpu jobs won't get scheduled on agents with gpus (@pschorf correct me if I'm wrong)

pschorf:
@dPeS That's correct. Currently, we have logic to prevent scheduling jobs without GPUs on GPU-enabled hosts to prevent tying up the non-GPU resources on the host. If that's an issue in your environment we can look into making that configurable.

pschorf:
@DaoWen any feedback before I merge this?

dPeS:
@pschorf hi, that would be useful for my case... it's not a bloker since I could spawn p2.xlarge + some non gpu instance, but in future `configuration` option will be cool

dposada:
I created #803 to track this.

Thanks for testing this PR, @dPeS!

PR #800, opened 2018-04-03
Wip/remove metatransaction

scrosby:
- In my performance optimization work, I've encountered problems where code that runs fast in test & dev fails to show the expected performance in production. Further analysis appears to link the slow API /list queries to metatransactions and requesting very recent data.
- On further investigation, we no longer use metatransactions, and the former uses of them have been replaced with commit-latch for a long time, so lets rip out the code.
- 

- Code cleanup
- Removing unnecessary complexity and code
- Performance


scheduler/src/cook/mesos.clj
```diff
@@ -32,54 +32,13 @@
...
-            [metatransaction.utils :as dutils]
             [metrics.counters :as counters]
             [swiss.arrows :refer :all])
   (:import [org.apache.curator.framework.recipes.leader LeaderSelector LeaderSelectorListener]
            org.apache.curator.framework.state.ConnectionState))
 
 ;; ============================================================================
 ;; mesos scheduler etc.
-
-(defn submit-to-mesos
```
scrosby: AFAICT, this is dead code; the only use being to create the fn stored under :submitter below. However, I find no references of :submitter anywhere. This appears to be a helper function for batch submission by clients that write to the DB directly, and wired to metatransaction via dutil/transact-with-retries!!.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1528,7 +1527,7 @@
                                 vectorize
                                 (mapv #(UUID/fromString %)))
           allow-partial-results (get-in ctx [:request :query-params :partial])
-          exists? #(group-exists? (db conn) %)
+          exists? #(group-exists? (d/db conn) %)
```
wyegelwel: This could let you read uncommitted data

scheduler/src/cook/mesos/api.clj
```diff
@@ -1572,12 +1571,12 @@
                                          (str/join \space (::non-existing-guuids ctx)))}]))
      :handle-ok (fn [ctx]
                   (if (Boolean/valueOf (get-in ctx [:request :query-params "detailed"]))
-                    (mapv #(merge (fetch-group-map (db conn) %)
-                                  (fetch-group-job-details (db conn) %))
+                    (mapv #(merge (fetch-group-map (d/db conn) %)
```
wyegelwel: This could let you query for uncommitted groups

scheduler/src/cook/mesos/api.clj
```diff
@@ -2584,7 +2583,7 @@
       (ANY "/running" []
         (running-jobs conn is-authorized-fn))
       (ANY "/list" []
-        (list-resource (db conn) framework-id is-authorized-fn)))
+        (list-resource (d/db conn) framework-id is-authorized-fn)))
```
wyegelwel: This would let you read list uncommitted jobs

scrosby: In this case, we should be OK. For job.state/waiting, we explicitly look at commit-latch. For running/completed, it doesn't matter.

scheduler/src/cook/mesos/util.clj
```diff
@@ -376,6 +376,23 @@
                  name-filter-fn (filter #(name-filter-fn (:job/name %))))
         (take limit)))))
 
+(defn committed?
```
dposada: We already have [`uncommitted?`](https://github.com/twosigma/Cook/blob/5bb3c2cab43b1d63c8635cf28429173e06890cd8/scheduler/src/cook/mesos/util.clj#L384); is there an opportunity for code re-use between them?

Issue #755, opened 2018-03-08
Cook pools design


dposada:
This is a discussion for the new Cook Scheduler feature called a "pool". 

# Motivation

We want to support scheduling jobs on heterogeneous clusters (e.g. some preemptible machines, some non-preemptible machines) with a single Cook instance while preserving the independence of scheduling provided by separate ranking and matching cycles for each of these types of machines.

# Overview
Users will be able to submit jobs to a to a pre-defined set of pools by adding a field on the job specification:

```
"pool": "non-preemptible"
```

Additionally, Mesos agents could be tagged with the pool that they support scheduling on:

```
"pool": "non-preemptible"
```

Internally, Cook will schedule jobs independently between pools. Each pool will support different quotas, shares, etc. Jobs in different pools will be ranked and matched separately.

# Implementation Details

Internally, Cook will run separate ranking and matching loops for each pool. The jobs and offers will be broken down by pool to speed matching and improve resiliency. For instance, if one pool is not receiving offers for an extended period and has a large number of jobs queued, it will not adversely impact scheduling on other pools. A separate Fenzo matcher will be started for each pool to avoid the need to synchronize on access to a shared Fenzo resource.

PR #801, opened 2018-04-05
Adds support for pool on job submission

dposada:

- adding `pool` as an optional, top-level field on `/job` `POST`s
- adding validation for jobs submitted with a pool

For full context, see issue #755. This PR is focused only on adding the ability to submit jobs to a specific pool.

## What's not included?

- this PR does not yet make the quota check on job submission pool-aware (issue #802)
- this PR does not introduce any pool-specific scheduling

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -308,7 +308,8 @@
   "Schema for a description of a job (as returned by the API)."
   (let [ParentGroup {:uuid s/Uuid, :name s/Str}]
     (-> JobResponseBase
-        (assoc (s/optional-key :groups) [ParentGroup]))))
+        (assoc (s/optional-key :groups) [ParentGroup])
+        (assoc (s/optional-key :pool) s/Str))))
 
 (def InstanceResponse
   "Schema for a description of a job instance (as returned by the API)."
@@ -601,7 +602,7 @@
                     executor (assoc :job/executor executor)
                     progress-output-file (assoc :job/progress-output-file progress-output-file)
                     progress-regex-string (assoc :job/progress-regex-string progress-regex-string)
-                    pool (assoc :job/pool pool))]
+                    pool (assoc :job/pool (:db/id pool)))]
 
     ;; TODO batch these transactions to improve performance
     (-> ports
@@ -843,6 +844,7 @@
         executor (:job/executor job)
         progress-output-file (:job/progress-output-file job)
         progress-regex-string (:job/progress-regex-string job)
+        pool (:job/pool job)
         state (util/job-ent->state job)
         constraints (->> job
                          :job/constraint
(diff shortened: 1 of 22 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -417,12 +418,21 @@
...
    (s/optional-key :groups) [Group]})
 
+(def RawSchedulerRequestDeprecated
+  "Schema for a request to the raw scheduler endpoint."
+  JobSubmission)
+
+(def JobSubmissionRequest
+  "Schema for a POST request to the /jobs endpoint."
+  (assoc JobSubmission
+    (s/optional-key :pool) s/Str))
```
pschorf: I know we talked about this earlier, but now I'm getting cold feet on having this at the top level. It feels weird when you look at `make-job-txn` for instance, which now has an additional argument where it didn't before. How do you feel about moving this down into each job spec?

dposada: Some questions come up if move the pool into the job spec:

- should we allow jobs for multiple pools to be submitted together?
- two jobs in the same group probably can't be in different pools?
- if job A cannot be submitted on pool X due to the quota check, should we still allow job B from the same `POST` to be submitted pool Y?

Maybe it would be simpler to just not allow different pools in the same submission. But then, what's the point of requiring users to specify the pool on each job? 

Would it make more sense to put the pool in the URL, like `curl -XPOST localhost:12321/jobs/pool-name -d ...`, instead of having it in the query string?

dposada: @pschorf Even if we do move the pool into the job spec, I think we will want to keep the extra `pool` argument in `make-job-txn`. The reason is that by the time we get to `make-job-txn`, we're already supposed to have validated the job, which means we've checked that its pool name is valid and the corresponding pool is accepting submissions. We will have resolved the provided name to a pool entity, which we want to pass directly to `make-job-txn` instead of forcing it to re-load the entity by name. It feels odd to add the pool entity to the `Job` schema (especially because we would need to `dissoc` it from the schema for the response side), so I think it actually makes sense to keep the `pool` argument separate.

Overall, I'm not convinced it's a bad idea to have the pool at the top level. Conceptually, you can defend it by saying that you are submitting a set of jobs and groups to a pool.

What do you think?

dposada: @pschorf Thoughts on this?

scheduler/src/cook/mesos/pool.clj
```diff
@@ -11,3 +11,18 @@
               :where
               [?e :pool/name]]
             db)))
+
+(defn pool-by-name
```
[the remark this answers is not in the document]
dposada: Done

dposada:
Can we merge this?

PR #806, opened 2018-04-09
Provides /usage sub-map with all pools in the default case

dposada:

- adding `pools` sub-map to `/usage`, which contains usage info for all pools, when the pool is not explicitly provided

We decided, as explained in issue #755, that it makes sense to default to providing the information for all pools, instead of just for the default pool.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -2036,7 +2036,16 @@
                              :running-jobs job-uuids}
                      :usage usage})
          :ungrouped (let [[job-uuids usage] (get breakdowns nil)]
-                      {:running-jobs job-uuids :usage (or usage zero-usage)})}))))
+                      {:running-jobs job-uuids
+                       :usage (or usage zero-usage)})}))))
+
+(defn no-usage-map
+  "Returns a resource usage map showing no usage"
+  [with-group-breakdown?]
+  (cond-> {:total-usage zero-usage}
+          with-group-breakdown? (assoc :grouped []
+                                       :ungrouped {:running-jobs []
+                                                   :usage zero-usage})))
 
 (defn get-user-usage
   "Query a user's current resource usage based on running jobs."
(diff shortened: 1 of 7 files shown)


dposada:
@pschorf Thoughts on this?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2000,41 +2001,75 @@
...
-           :ungrouped (let [[job-uuids usage] (get breakdowns nil)]
-                        {:running-jobs job-uuids :usage (or usage zero-usage)})})))))
+        pool (get-in ctx [:request :query-params :pool])]
+    (if pool
+      (let [jobs (util/get-user-running-job-ents db user pool)]
+        (user-usage with-group-breakdown? jobs))
+      (let [jobs (util/get-user-running-job-ents db user)
+            default-pool-name (config/default-pool)
+            pool-name->jobs (group-by
+                              #(let [pool (:job/pool %)]
```
pschorf: Can you use `if-let` here?

scheduler/src/cook/mesos/util.clj
```diff
@@ -471,23 +471,37 @@
...
-               :where
-               ;; Note: We're assuming that many users will have significantly more
-               ;; completed jobs than there are jobs currently running in the system.
-               ;; If not, we might want to swap these two constraints.
-               [?j :job/state :job.state/running]
-               [?j :job/user ?user]
-               [(cook.mesos.pool/check-pool $ ?j :job/pool ?pool-name ?requesting-default-pool)]]
-             db user pool-name' requesting-default-pool?)
-           (map (partial d/entity db))))))
+  ([db user]
```
pschorf: Does this make sense to split into a separate function? I think intuitively a 2 argument version of this function would return the jobs in the default pool, not all jobs. In any case, we should update the docstring.

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1632,3 +1636,34 @@
       (is (nil? (:job/name job-entity)))
       (is (= job-uuid (:uuid job-map-for-api)))
       (is (= "cookjob" (:name job-map-for-api))))))
+
+(deftest test-get-user-usage-no-usage
```
pschorf: Can you add some more tests, for instance where they have usage in some pools but not others?

scheduler/src/cook/mesos/api.clj
dposada: Done, thanks

scheduler/src/cook/mesos/util.clj
dposada: Done

scheduler/test/cook/test/mesos/api.clj
dposada: Done

dposada:
Can we merge this?

PR #810, opened 2018-04-12
Sets monitor metrics to 0 instead of removing them

dposada:

- setting the counters in `monitor.clj` to 0 instead of removing them

When using singlestat panels in Grafana, the `current` function uses the last non-null value from the series. As a result, the panels can show, for example, cpus and memory in use when the user in question actually has nothing running. Explicitly setting the counters to 0 resolves this.


dposada:
@pschorf Thoughts on this?

PR #818, opened 2018-04-19
Guards against invalid default pool configuration

### LINKED ISSUE #813: Require default pool config if there are pools in the database
If there are pools in the database, we should require on startup that there's a default pool configured.

dposada:

- moving the seeding of pools to before starting cook scheduler, so that they're already there at startup
- making the pool-seeding setup the database schema
- adding guards on cook scheduler startup against:
  - pools in the database, but no default pool configured
  - the configured default pool does not match any pool in the database

We want to prevent the situation where there are pools in the database, but we don't have a valid configuration for which of those pools is the default. Resolves #813.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1671,7 +1671,7 @@ class CookTest(unittest.TestCase):
         job_resources = {'cpus': 0.1, 'mem': 123}
         job_uuid, resp = util.submit_job(self.cook_url, command='sleep 120', **job_resources)
         self.assertEqual(resp.status_code, 201, resp.content)
-        pools, _ = util.pools(self.cook_url)
+        pools, _ = util.all_pools(self.cook_url)
         try:
             user = util.get_user(self.cook_url, job_uuid)
             # Don't query until the job starts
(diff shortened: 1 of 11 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -1632,6 +1632,7 @@ def query():
 
     def test_retrieve_jobs_with_deprecated_api(self):
         pools, _ = util.pools(self.cook_url)
```
shamsimam: How about naming them to `all_pools` and `active_pools`?

scheduler/test/cook/test/mesos/pool.clj
```diff
@@ -0,0 +1,22 @@
+(ns cook.test.mesos.pool
+  (:require [clojure.test :refer :all]
+            [cook.config :as config]
+            [cook.mesos.pool :as pool])
+  (:import (clojure.lang ExceptionInfo)))
+
+(deftest test-guard-invalid-default-pool
```
shamsimam: Can you regex match on the message strings of the exceptions to verify the expected path was triggered?

integration/tests/cook/test_basic.py

scheduler/test/cook/test/mesos/pool.clj
dposada: Done

shamsimam:
Awaiting green build

dposada:
@shamsimam Build is now green

PR #829, opened 2018-04-30
Adds debugging info to test_instance_stats_success

dposada:

- asserting that each job actually succeeded
- dumping the sandbox logs in the `finally` block

We've seen this test fail with:

```bash
    def test_instance_stats_success(self):
        name = str(uuid.uuid4())
        job_uuid_1, resp = util.submit_job(self.cook_url, command='exit 0', name=name, cpus=0.1, mem=32)
        job_uuid_2, resp = util.submit_job(self.cook_url, command='sleep 1', name=name, cpus=0.2, mem=64)
        job_uuid_3, resp = util.submit_job(self.cook_url, command='sleep 2', name=name, cpus=0.4, mem=128)
        job_uuids = [job_uuid_1, job_uuid_2, job_uuid_3]
        try:
            util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
            instances = [util.wait_for_instance(self.cook_url, j) for j in job_uuids]
            start_time = min(i['start_time'] for i in instances)
            end_time = max(i['start_time'] for i in instances)
            stats, _ = util.get_instance_stats(self.cook_url,
                                               status='success',
                                               start=util.to_iso(start_time),
                                               end=util.to_iso(end_time + 1),
                                               name=name)
            user = util.get_user(self.cook_url, job_uuid_1)
            stats_overall = stats['overall']
>           self.assertEqual(3, stats_overall['count'])
E           AssertionError: 3 != 2
```

This seems to indicate that one of the jobs did not succeed, and this should help us track down the reason.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2025,9 +2025,11 @@ class CookTest(unittest.TestCase):
         job_uuid_3, resp = util.submit_job(self.cook_url, command='sleep 2', name=name, cpus=0.4, mem=128)
         self.assertEqual(resp.status_code, 201, msg=resp.content)
         job_uuids = [job_uuid_1, job_uuid_2, job_uuid_3]
+        util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
+        instances = [util.wait_for_instance(self.cook_url, j) for j in job_uuids]
         try:
-            util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
-            instances = [util.wait_for_instance(self.cook_url, j) for j in job_uuids]
+            for instance in instances:
+                self.assertEqual('success', instance['parent']['state'])
             start_time = min(i['start_time'] for i in instances)
             end_time = max(i['start_time'] for i in instances)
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2025,9 +2025,11 @@ def test_instance_stats_success(self):
         job_uuid_3, resp = util.submit_job(self.cook_url, command='sleep 2', name=name, cpus=0.4, mem=128)
         self.assertEqual(resp.status_code, 201, msg=resp.content)
         job_uuids = [job_uuid_1, job_uuid_2, job_uuid_3]
+        util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
+        instances = [util.wait_for_instance(self.cook_url, j) for j in job_uuids]
```
DaoWen: Since these two lines were moved out of the `try` block, we no longer kill the jobs if they fail. Right?

Maybe we should add a second try/catch-raise layer for the sandbox dumping, since dumping the sandbox only when the test failed might be an improvement as well.

dposada: Good call, done

PR #835, opened 2018-05-02
Adds support for pool param to /jobs

dposada:

- adding `pool` as an optional param to `/jobs`

This is part of the overall pools project.

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -18,7 +18,7 @@
             [clj-time.core :as t]
             [clj-time.format :as tf]
             [clj-time.periodic :refer [periodic-seq]]
-            [cook.config :as config]
+            [cook.mesos.pool :as pool]
             [cook.mesos.schema :as schema]
             [clojure.core.async :as async]
             [clojure.core.cache :as cache]
@@ -26,12 +26,11 @@
             [datomic.api :as d :refer (q)]
             [metatransaction.core :refer (db)]
             [metrics.timers :as timers]
-            [plumbing.core :as pc :refer (map-vals map-keys)]
-            [cook.mesos.pool :as pool])
+            [plumbing.core :as pc :refer (map-vals map-keys)])
   (:import
     [com.google.common.cache Cache CacheBuilder]
     [java.util.concurrent TimeUnit]
-    [java.util Date UUID]))
+    [java.util Date]))
 
 
 (defn new-cache []


scheduler/src/cook/mesos/util.clj
```diff
@@ -395,22 +399,25 @@
...
+                    :in $ ?user ?state ?start ?end ?pool-name ?default-pool
                     :where
                     [?j :job/state ?state]
                     [?j :job/user ?user]
                     [?j :job/submit-time ?t]
                     [(<= ?start ?t)]
-                    [(< ?t ?end)]]
-                  db user state-keyword start end)
+                    [(< ?t ?end)]
+                    [(cook.mesos.pool/check-pool-for-listing $ ?j :job/pool ?pool-name ?default-pool)]]
```
pschorf: Probably doesn't matter now, but should we think about putting this first instead of last and indexing on it to help smaller pools?

dposada: Putting the pools first would probably help in clusters with many pools of differing sizes, but it would probably hurt clusters with no pools and many jobs in one status.

dposada:
Performance tests showed that this change does not have any major effect on the performance of the list queries (more detail was discussed offline).

dposada:

dposada:

scheduler/src/cook/mesos/util.clj
```diff
@@ -26,7 +26,8 @@
             [datomic.api :as d :refer (q)]
             [metatransaction.core :refer (db)]
             [metrics.timers :as timers]
-            [plumbing.core :as pc :refer (map-vals map-keys)])
+            [plumbing.core :as pc :refer (map-vals map-keys)]
+            [cook.mesos.pool :as pool])
```
pschorf: I hate to say this, but please sort this :)

scheduler/src/cook/mesos/pool.clj
```diff
@@ -7,12 +7,26 @@
...
+
+(def nil-pool (str (UUID/randomUUID)))
+
+(defn check-pool-for-listing
+  "Returns true if either the provided pool-name is the 'nil' pool, or if
+  pool/check-pool returns true. This allows us to return jobs in all pools when the user
+  does not specify a pool."
+  ([entity entity->pool pool-name default-pool?]
+   (or
+     (= nil-pool pool-name)
```
pschorf: Why can't this just be `(nil? pool-name)` and we can avoid the `nil-pool`

dposada: Because Datomic won't let you have `nil` as a "source" in the datalog query used by `get-active-jobs-by-user-and-state`

pschorf: Ah, gotcha. Annoying.

scheduler/src/cook/mesos/util.clj
dposada: Done

PR #838, opened 2018-05-03
Return 409 (Conflict) when call to /retry is a no-op

pschorf:
- Return a 409 (Conflict) when a write to /retry does not retry the job or change the max-retries

Reduce user confusion around /retry requests which don't affect the job.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2228,10 +2228,7 @@ class CookTest(unittest.TestCase):
         try:
             util.wait_for_job(self.cook_url, uuid, 'running')
             resp = util.retry_jobs(self.cook_url, job=uuid, assert_response=False, retries=1)
-            self.assertEqual(409, resp.status_code, msg=resp.content)
-            job = util.load_job(self.cook_url, uuid)
-            self.assertEqual('completed', job['status'], json.dumps(job, indent=2))
-            self.assertEqual(1, len(job['instances']), json.dumps(job, indent=2))
+            self.assertEqual(201, resp.status_code, msg=resp.content)
         finally:
             util.kill_jobs(self.cook_url, [uuid])
 
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2255,10 +2255,11 @@ class CookTest(unittest.TestCase):
 
     def test_set_retries_to_attempts_conflict(self):
         uuid, resp = util.submit_job(self.cook_url, command='sleep 30', max_retries=5, disable_mea_culpa_retries=True)
-        util.wait_for_instance(self.cook_url, uuid)
+        util.wait_until(lambda: util.load_job(self.cook_url, uuid),
+                        lambda j: (len(j['instances']) == 1) and (j['instances'][0]['status'] == 'running'))
         util.kill_jobs(self.cook_url, [uuid])
         def instances_complete(job):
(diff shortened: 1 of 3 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2214,10 +2214,58 @@ def test_decrease_retries_below_attempts(self):
...
+                {'group': group_uuid,
+                 'command': 'exit 1',
+                 'max_retries': 1,
+                 'disable_mea_culpa_retries': True}]
+        jobs, resp = util.submit_jobs(self.cook_url, jobs)
+        util.wait_for_jobs(self.cook_url, jobs, 'completed')
+        resp = util.retry_jobs(self.cook_url, groups=[group_uuid], assert_response=False, retries=2)
+        self.assertEqual(409, resp.status_code, resp.content)
+        self.assertTrue(jobs[0] in str(resp.content), resp.content)
+        self.assertFalse(jobs[1] in str(resp.content), resp.content)
```
dposada: Should we assert on the status and number of instances of each job?

dposada: Does the response have just a string message, or is it JSON? Should we make it JSON?

integration/tests/cook/test_basic.py
```diff
@@ -2214,10 +2214,58 @@ def test_decrease_retries_below_attempts(self):
...
+
+    def test_set_retries_to_attempts_conflict(self):
+        uuid, resp = util.submit_job(self.cook_url, command='sleep 30', max_retries=5, disable_mea_culpa_retries=True)
+        util.wait_for_instance(self.cook_url, uuid)
+        util.kill_jobs(self.cook_url, [uuid])
+        def instances_complete(job):
+            return all([i['status'] != 'running' for i in job['instances']])
```
dposada: Would it be better to wait for the instance status to be `'failed'`, since instances can have status `'unknown'`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1884,6 +1885,25 @@
...
+  "Checks whether a 409 conflict should be returned during a retry operation. This should occur when one of the jobs:
+   - Is already completed
+   - Has already retried 'retries' times"
+  [conn ctx]
+  (let [jobs (::jobs ctx)
+        retries (get-in ctx [:request :body-params :retries])
+        db (d/db conn)]
+    (if (not (nil? retries))
+      (let [jobs-not-updated (filter (fn [uuid]
+                                       (let [job (d/entity (d/db conn) [:job/uuid uuid])]
```
dposada: Can we use the `db` from the `let` here?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1884,6 +1885,25 @@
...
+                                         (and (= :job.state/completed (:job/state job))
```
dposada: Destructure `:job/state`?

pschorf: Discussed offline, I did the restructuring

integration/tests/cook/test_basic.py
pschorf: I think this got fixed in a subsequent PR

integration/tests/cook/test_basic.py
dposada: Aren't we doing it in the other tests?

PR #840, opened 2018-05-07
Add some new/fixed metrics

scrosby:

- Add a metric on fetch-job-map time
- Add a metric on /list API response size
- Add a metric on /list API time to fetch all job maps

- Better track what is going on. Try to explain the slow p95 lookup time for /list API queries as being from large responses.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -837,55 +837,57 @@
 
 (defn fetch-job-map
   [db framework-id job-uuid]
-  (let [job (d/entity db [:job/uuid job-uuid])
-        resources (util/job-ent->resources job)
-        groups (:group/_job job)
-        application (:job/application job)
-        expected-runtime (:job/expected-runtime job)
-        executor (:job/executor job)
-        progress-output-file (:job/progress-output-file job)
-        progress-regex-string (:job/progress-regex-string job)
-        pool (:job/pool job)
-        state (util/job-ent->state job)
-        constraints (->> job
-                         :job/constraint
-                         (map util/remove-datomic-namespacing)
-                         (map (fn [{:keys [attribute operator pattern]}]
-                                (->> [attribute (str/upper-case (name operator)) pattern]
-                                     (map str)))))
-        instances (map #(fetch-instance-map db %1) (:job/instance job))
-        submit-time (when (:job/submit-time job) ; due to a bug, submit time may not exist for some jobs
-                     (.getTime (:job/submit-time job)))
-        job-map {:command (:job/command job)
-                 :constraints constraints
-                 :cpus (:cpus resources)
-                 :disable_mea_culpa_retries (:job/disable-mea-culpa-retries job false)
-                 :env (util/job-ent->env job)
-                 :framework_id framework-id
-                 :gpus (int (:gpus resources 0))
-                 :instances instances
-                 :labels (util/job-ent->label job)
-                 :max_retries (:job/max-retries job) ; consistent with input
-                 :max_runtime (:job/max-runtime job Long/MAX_VALUE) ; consistent with input
-                 :mem (:mem resources)
-                 :name (:job/name job "cookjob")
-                 :ports (:job/ports job 0)
-                 :priority (:job/priority job util/default-job-priority)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -837,55 +837,57 @@
...
-    (cond-> job-map
-            groups (assoc :groups (map #(str (:group/uuid %)) groups))
-            application (assoc :application (util/remove-datomic-namespacing application))
-            expected-runtime (assoc :expected-runtime expected-runtime)
-            executor (assoc :executor (name executor))
-            progress-output-file (assoc :progress-output-file progress-output-file)
-            progress-regex-string (assoc :progress-regex-string progress-regex-string)
-            pool (assoc :pool (:pool/name pool)))))
+  (timers/time!
+    (timers/timer ["cook-scheduler" "internal" "fetch-job-map"])
```
dposada: Why "internal"? Do we use that elsewhere?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2242,7 +2244,12 @@
                  (timers/time!
                    (timers/timer ["cook-scheduler" "handler" "list-endpoint-duration"])
                    (let [job-uuids (list-jobs db false ctx)]
-                     (mapv (partial fetch-job-map db framework-id) job-uuids))))))
+                     (histograms/update!
+                       (histograms/histogram ["cook-scheduler" "handler" "list-endpoint-response-count"])
+                       (count job-uuids))
```
dposada: Why are we doing the `(count job-uuids)`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -2242,7 +2244,12 @@
...
+                     (timers/time!
+                       (timers/timer ["cook-scheduler" "handler" "list-endpoint-fetch-job-map-duration"])
+                       (mapv (partial fetch-job-map db framework-id) job-uuids)))))))
```
dposada: Should we add this for `/jobs` as well?

scheduler/src/cook/mesos/api.clj
scrosby: No. I'm happy with any other name. I thought that 'internal' was appropriate because this internal is accessed via many paths. E.g., rawscheduler, and another half-dozen.

scheduler/src/cook/mesos/api.clj
scrosby: I'm histogramming the response sizes (as counted by # jobs in the response) as a metric. E.g., to discover that 10% of our queries are returning 10000 jobs in the response. :)

pschorf:
Should we performance test this before we merge it?

scheduler/src/cook/mesos/api.clj
scrosby: Not sure exactly what you're referring to by 'this'?

scheduler/src/cook/mesos/api.clj
dposada: We already have `cook-scheduler.cook-mesos.api.list-response-job-count.*` - doesn't that have what you want?

scheduler/src/cook/mesos/api.clj
dposada: By "this" I mean the new metric(s) that you're introducing for `/list`

scheduler/src/cook/mesos/api.clj
scrosby: I just noticed that, sitting in list-jobs, about 15 minutes ago and am updating the review with this nuked.

pschorf:
My mistake, I had misread part of the change.

scheduler/src/cook/mesos/api.clj
dposada: What are the other names we use for this part of the metric name? Do any of them work well for this one?

dposada: SGTM

dposada: @scrosby Are we planning on renaming to `cook-mesos internal ...`?

scrosby: Yes. Been busy with the other testing, the writeup, and such. Now ready to resume this. :)

PR #845, opened 2018-05-10
Defaults tail path to stdout

dposada:

- allowing the user to not specify a path, in which case `stdout` is used

It's nice when running, for example, `cs tail -f <uuid>`, to not have to specify the path.


dposada:

PR #846, opened 2018-05-10
Fixes race in test_list_jobs_by_completion_state_with_jobs_endpoint

dposada:

- moving the assertion that status == success to after the `wait_for_job`

Otherwise, there's a race between the job finishing and the assertion running.


dposada:

PR #847, opened 2018-05-14
New query functions with metatransactions turned off and VAET index.

scrosby:

- The same as PR#771, except two additional patches
- Turn off metatransactions: https://github.com/scrosby/Cook/commit/f673b8d060673b02c9ee769857f3cee5fe2e1f51
- Change index to VAET https://github.com/scrosby/Cook/commit/0e7979ef0d0357a56bea2354f4954b27a92ac4f6

- Improved performance


dposada:
Needs rebase

scheduler/src/cook/mesos/util.clj
```diff
@@ -323,6 +324,11 @@
 
 (timers/deftimer [cook-mesos scheduler get-completed-jobs-by-user-duration])
 
+(defn job-ent->user
+  "Given a job entity, return the user the job runs as."
+  [job-ent]
+    (lookup-cache-datomic-entity! job-ent->user-cache :job/user job-ent))
```
dposada: Indentation seems off here

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,73 @@
...
 (def ^:const instance-states #{"success" "failed"})
 
+
+(defn job-submit-time-between
+  "Returns true if the job-ent's :job/submit-time is between start-ms (inclusive) and
+  end-ms (exclusive)"
+  [{:keys [^Date job/submit-time]} ^long start-ms ^long end-ms]
+  (let [submit-ms (.getTime submit-time)]
+                    (and (<= start-ms submit-ms)
+                         (< submit-ms end-ms))))
```
dposada: Indentation seems off here

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,73 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+
+(defn job-submit-time-between
```
dposada: Maybe `job-submitted-in-range?`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,73 @@
...
+  [{:keys [^Date job/submit-time]} ^long start-ms ^long end-ms]
+  (let [submit-ms (.getTime submit-time)]
+                    (and (<= start-ms submit-ms)
+                         (< submit-ms end-ms))))
+
+;; get-jobs-by-user-and-state-and-submit is a bit opaque
+;; because it is reaching into datomic internals. Here is a quick explanation.
+;; seek-datoms provides a pointer into the raw datomic indices
+;; that we can then seek through. We set the pointer to look
+;; through the avet index, with attribute :job/state, seek to
```
dposada: This says `avet`, but the code below seems to be using `vaet`. Do we mean to be using `vaet`? If so, it would be good to explain why `vaet` is a better choice here.

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,73 @@
...
+;; through the avet index, with attribute :job/user, seek to
 ;; user and then seek to the entity id that *would* have been
-;; created at expanded start. 
+;; created at expanded start.
 ;; This works because the submission time and job/user field
 ;; are set at the same time, in "real" time. This means that
 ;; jobs submitted after `start` will have been created after
 ;; expanded start
+;; This differs from get-active-jobs-by-user-and-state as it
+;; is also looking up based on task state.
```
dposada: I think this could be clearer. What does "it" refer to? And why is this distinction relevant?

scheduler/src/cook/mesos/util.clj
```diff
@@ -364,25 +422,31 @@
...
+          job-user-entid (d/entid db :job/user)
+          start-ms (.getTime start)
+          end-ms (.getTime end)
           jobs
-          (->> (d/seek-datoms db :avet :job/user user (d/entid-at db :db.part/user expanded-start))
-               (take-while #(and (< (.e %) (d/entid-at db :db.part/user expanded-end))
-                                 (= (.a %) (d/entid db :job/user))
-                                 (= (.v %) user)))
-               (map #(.e %))
+          (->> (d/seek-datoms db :avet :job/user user entid-start)
```
dposada: Why are we using `avet` here but `vaet` above? If this is intentional, I think we should explain why.

scheduler/src/cook/mesos/util.clj
```diff
@@ -364,25 +422,31 @@
...
                (filter #(pool/check-pool-for-listing % :job/pool pool-name' default-pool?)))]
       (->>
         (cond->> jobs
                  (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
                  (instance-states state) (filter #(= state (job-ent->state %)))
                  name-filter-fn (filter #(name-filter-fn (:job/name %))))
-        (take limit)))))
+        ; No need to sort. We're traversing in entity_id order, so in time order.
+        (take limit)
+        doall))))
```
dposada: The function above mentions in the docstring that it's returning lazy output. Should we mention in this function's docstring that it's returning non-lazy output?

scheduler/src/cook/mesos/util.clj
```diff
@@ -395,48 +459,41 @@
...
+      (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-duration")])
+      (->>
+        (cond->> (get-jobs-by-user-and-state-and-submit db user start end state-keyword)
+                 (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
+                 name-filter-fn (filter #(name-filter-fn (:job/name %)))
+                 (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))
+        (filter #(pool/check-pool-for-listing % :job/pool pool-name' default-pool?))
+        ; No need to sort. We're traversing in entity_id order, so in time order.
+        (take limit)
+        doall))))
```
dposada: Same question here about whether we should mention in the docstring that this is returning a realized collection

scheduler/src/cook/mesos/util.clj
scrosby: Comment fixed. See description a top of PR as to why.

scheduler/src/cook/mesos/util.clj
scrosby: Because the queries are different. Added a comment to the affect below.

scheduler/src/cook/mesos/util.clj
```diff
@@ -395,48 +459,41 @@
...
-               (map (partial d/entity db))))]
-    (cond->> jobs
-             (not include-custom-executor?) (filter #(false? (:job/custom-executor %)))
-             name-filter-fn (filter #(name-filter-fn (:job/name %)))
-             (and include-custom-executor? (= :job.state/waiting state-keyword)) (remove uncommitted?))))
+        pool-name' (or pool-name pool/nil-pool)]
+    (timers/time!
+      (timers/timer ["cook-mesos" "scheduler" (str "get-" (name state) "-jobs-by-user-duration")])
+      (->>
+        (cond->> (get-jobs-by-user-and-state-and-submit db user start end state-keyword)
```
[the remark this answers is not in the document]
scrosby: Taste. Small self-contained functions with a sane semantic boundary are good.

scheduler/src/cook/mesos/util.clj
scrosby: Honestly, no. I think non-lazy should be the usual assumption. Are we thinking of adding 'not lazy' to every function thats not lazy? 

I can strip it off the other function, but I don't think it harms anything.

scheduler/src/cook/mesos/util.clj
scrosby: See other answer

scrosby:
PR Description updated.

scheduler/src/cook/mesos/util.clj
```diff
@@ -66,6 +66,7 @@
 (defonce ^Cache job-ent->resources-cache (new-cache))
 (defonce ^Cache categorize-job-cache (new-cache))
 (defonce ^Cache task-ent->user-cache (new-cache))
+(defonce ^Cache job-ent->user-cache (new-cache))
```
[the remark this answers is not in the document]
dposada: I'm fine with making it a separate PR, but given that these caches are the reason we began talking about implementing something like mount, it would be good to follow through on that. Can we do the mount-ification PR right after this one?

scheduler/src/cook/mesos/util.clj
dposada: Do we know why `vaet` is faster and has less variance?

scheduler/src/cook/mesos/util.clj
dposada: > Are we thinking of adding 'not lazy' to every function thats not lazy?

I'm not advocating for that.

The comment was mostly about consistency, since the two functions are so closely related. If you think it's clear now, I'm good with it.

scheduler/src/cook/mesos/util.clj
```diff
@@ -346,7 +346,7 @@
 (def ^:const instance-states #{"success" "failed"})
 
 
-(defn job-submit-time-between
+(defn job-submitted-in-range
```
dposada: How about `job-submitted-in-range?`?

scheduler/src/cook/mesos/util.clj
```diff
@@ -357,8 +357,8 @@
 ;; get-jobs-by-user-and-state-and-submit is a bit opaque
 ;; because it is reaching into datomic internals. Here is a quick explanation.
 ;; seek-datoms provides a pointer into the raw datomic indices
-;; that we can then seek through. We set the pointer to look
-;; through the avet index, with attribute :job/state, seek to
+;; that we can then seek through. We set the pointer to look through the
+;; vaet index, with value :job.state/{running,waiting,completed}, seek to
```
dposada: Isn't this function only called for running and waiting, not completed?

scheduler/src/cook/mesos/util.clj
```diff
@@ -478,6 +479,11 @@
         (take limit)
         doall))))
 
+
+;; Users have many fewer running/waiting jobs than completed jobs, so they need different queries.
+;; For searches for jobs in the active/running state, enumerating jobs based on state is more robustly
+;; a 'small set' than enumearting by user. Enumerating all jobs by user is more likely to be a
```
dposada: "enumerating"

scheduler/src/cook/mesos/util.clj
```diff
@@ -478,6 +479,11 @@
         (take limit)
         doall))))
 
+
+;; Users have many fewer running/waiting jobs than completed jobs, so they need different queries.
+;; For searches for jobs in the active/running state, enumerating jobs based on state is more robustly
```
dposada: Did you mean "running/waiting" here?

scheduler/src/cook/mesos/util.clj
dposada: The comment below explains why we're using user here and state above, but it doesn't explain why we're using `avet` here but `vaet` above

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
scrosby: Sure, Lets get this out and then we can have that discussion. :)

scheduler/src/cook/mesos/util.clj
scrosby: Was fixed.

scheduler/src/cook/mesos/util.clj
scrosby: No. Conjectures only.

scheduler/src/cook/mesos/util.clj
scrosby: Was also updated. See next 2 lines.

scheduler/src/cook/mesos/util.clj
scrosby: I don't have an answer other than 'because those are the indexes to optimially return results given the query type that we are performing.'

scrosby: Datomic doesn't do query optimization. We have to. Fortunately, the data shapes are uniform enough that we can do it statically and do a fair enough job, and those are the most sane static queries to use.

scheduler/src/cook/mesos/util.clj
scrosby: It can work for any of them, whether its the right query to use is another question. (Datomic doesn't do query optimization. We have to. Fortunately, the data shapes are uniform enough that we can do it statically and do a fair enough job.)

scheduler/src/cook/mesos/util.clj
scrosby: *NOD*

scheduler/src/cook/mesos/util.clj
dposada: Got it. I wonder if we should add the fact that we don't know why it's better (and maybe the conjectures) to the comment in the code. I think our future selves (and other future team members) would thank us for that

scheduler/src/cook/mesos/util.clj
dposada: What I meant to point out is that the comment states:

> We set the pointer to look through the vaet index, with value :job.state/{running,waiting,completed}

But we don't actually use the value `completed` with this function.

scheduler/src/cook/mesos/util.clj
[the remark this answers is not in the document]
dposada: SGTM, do we have a Github issue to track this?

scheduler/src/cook/mesos/util.clj
dposada: It doesn't look fixed to me

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,74 @@
...
+                         (< submit-ms end-ms))))
+
+;; get-jobs-by-user-and-state-and-submit is a bit opaque
+;; because it is reaching into datomic internals. Here is a quick explanation.
+;; seek-datoms provides a pointer into the raw datomic indices
+;; that we can then seek through. We set the pointer to look through the
+;; vaet index, with value :job.state/{running,waiting,completed}, seek to
+;; state and then seek to the entity id that *would* have been
+;; created at expanded start. This works because the submission
+;; time (which we sort on) is correlated with the entitiy id.
```
dposada: "entity"

scheduler/src/cook/mesos/util.clj
scrosby: Ok. Changed.

scheduler/src/cook/mesos/util.clj
scrosby: Changed.

scheduler/src/cook/mesos/util.clj
scrosby: Changed.

scheduler/src/cook/mesos/util.clj
```diff
@@ -339,21 +345,78 @@
 (def ^:const job-states #{"running" "waiting" "completed"})
 (def ^:const instance-states #{"success" "failed"})
 
+
+(defn job-submitted-in-range
```
dposada: Maybe `jobs-submitted-in-range?`?

scrosby: Changed

PR #756, opened 2018-03-09
Adds support for including custom executor jobs in /jobs

dposada:
## Changes proposed in this PR

- adding the ability to list jobs on `/jobs`
- `/jobs` will not filter out custom executor jobs

## Why are we making these changes?

We want to give users and tools the option of including custom executor jobs when listing (while maintaining backwards compatibility).

Issue #617, opened 2017-12-01
Store entities (not UUIDs) in Liberator context


DaoWen:
We're currently caching the UUIDs of our jobs/instances/groups in the Liberator context when processing the decision handlers for our REST endpoints. That choice results in a lot of repeated UUID→Entity conversions, which probably aren't too expensive, but likely aren't negligible either.

We could eliminate a lot of duplicate computation in the handlers by storing datomic entities (rather than UUIDs) in the context. I'm not sure if this will result in a measurable speedup, but it seems worth exploring. Even if there's no speed improvement, I think the code complexity reduction would be a clear win.

PR #848, opened 2018-05-14
Optimize the /list fetching code to not round-trip to UUID.

scrosby:
- Don't round-trip jobs to UUID's then back to job entities in the /list path. Get the entities and populate them.

- We think it'll be a modest performance increase; should avoid an extra random IO per job in the result set.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1269,7 +1269,7 @@
 (histograms/defhistogram [cook-mesos api list-request-param-limit])
 (histograms/defhistogram [cook-mesos api list-response-job-count])
 
-(defn list-jobents
+(defn list-jobs
   "Queries using the params from ctx and returns the jobs that were found as datomic entities."
   [db include-custom-executor? ctx]
   (timers/time!
@@ -1301,7 +1301,7 @@
 
 (defn jobs-list-exist?
   [conn ctx]
-  [true {::jobs-entities (list-jobents (d/db conn) true ctx)}])
+  [true {::jobs-entities (list-jobs (d/db conn) true ctx)}])
 
 (defn read-jobs-handler
   [conn is-authorized-fn resource-attrs]
@@ -2296,7 +2296,7 @@
     :handle-ok (fn [ctx]
                  (timers/time!
                    (timers/timer ["cook-scheduler" "handler" "list-endpoint-duration"])
-                   (let [job-ents (list-jobents db false ctx)]
+                   (let [job-ents (list-jobs db false ctx)]
                      (mapv (partial fetch-job-map-from-entity db framework-id) job-ents))))))
 
 ;;
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -1119,22 +1119,31 @@
   (mapv (partial fetch-job-map (db conn) framework-id) (::jobs ctx)))
 
 (defn render-jobs-for-response
+  "This rendes for response. Fills in in :group UUID's and names as well as map-ifies jobs
+  It will examine the ctx for ::jobs (containing UUID's) or ::job-entities (containing
+  datomic job entities) and return the merged set."
   [conn framework-id ctx]
   (let [db (db conn)
-
         fetch-group
         (fn fetch-group [group-uuid]
           (let [group (d/entity db [:group/uuid (UUID/fromString group-uuid)])]
             {:uuid group-uuid
              :name (:group/name group)}))
-
-        fetch-job
+        fetch-job-from-uuid
         (fn fetch-job [job-uuid]
           (let [job (fetch-job-map db framework-id job-uuid)
                 groups (mapv fetch-group (:groups job))]
+            (assoc job :groups groups)))
+        fetch-job-from-entity
+        (fn fetch-job [job-ent]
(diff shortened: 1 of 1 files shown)


DaoWen:
This PR partially addresses issue #617.

scheduler/src/cook/mesos/api.clj
```diff
@@ -1273,20 +1276,24 @@
...
+                      job-ents
+                      (take limit job-ents))]
       (histograms/update! list-request-param-time-range-ms (- end-ms start-ms'))
       (histograms/update! list-request-param-limit limit)
-      (histograms/update! list-response-job-count (count job-uuids))
-      job-uuids)))
+      (histograms/update! list-response-job-count (count job-ents))
+      job-ents)))
+
+(defn list-jobs
```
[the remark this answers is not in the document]
scrosby: Yes; Line 1300, jobs-list-exists?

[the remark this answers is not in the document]
dposada: Shouldn't that path also switch to using `list-jobents`?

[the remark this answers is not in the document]
dposada: Sure, but `/list` is the deprecated endpoint, and `/jobs` is the "new" endpoint. It seems like if we're choosing to one but not the other, we should choose `/jobs`

scrosby:
This PR addresses #617. the additional refactors to not have /jobs endpoint use the list-jobs function addresses part of #617.

scrosby:
@dposada  Is there a unit test /jobs/#UUID that exeercises this codepath for https://github.com/twosigma/Cook/pull/848/files#diff-6611b39b65ff49438a4d940b51c85cc0R2549 ? 
I didn't find one in a search, nor on your commit d236ec3b36c7538b79b11a6bf2fdbeefb38e17ee  #756 
I'd like to validate that this works correctly.

scheduler/src/cook/mesos/api.clj
```diff
@@ -891,6 +890,10 @@
               progress-regex-string (assoc :progress-regex-string progress-regex-string)
               pool (assoc :pool (:pool/name pool))))))
 
+(defn fetch-job-map
+  [db framework-id job-uuid]
+  (fetch-job-map-from-entity db framework-id (d/entity db [:job/uuid job-uuid])))
```
dposada: Does anything still use `fetch-job-map` after this change?

scheduler/src/cook/mesos/api.clj
```diff
@@ -1116,22 +1119,31 @@
   (mapv (partial fetch-job-map (db conn) framework-id) (::jobs ctx)))
 
 (defn render-jobs-for-response
+  "This rendes for response. Fills in in :group UUID's and names as well as map-ifies jobs
+  It will examine the ctx for ::jobs (containing UUID's) or ::job-entities (containing
+  datomic job entities) and return the merged set."
```
dposada: Why do we need both `::jobs` and `::job-entities`?

scrosby: ::jobs has UUID's (coming from the /job/<UUID> and from /job&uuid=XXXX&uuid=XXXX) and ::job-entities has entities. Both have different logic as they filter through read-jobs-handler, so are separated until the very end.

scheduler/src/cook/mesos/api.clj
scrosby: Over a dozen uses.

dposada: What are they? Should they switch over to using the new function?

scrosby: render-jobs-for-response-deprecated -- /rawscheduler handler
render-instances-for-response 
11 times as a convenience function in unit tests.

scheduler/src/cook/mesos/api.clj
dposada: Is it possible to have them work the same way so that we don't need both entries in the context map?

scheduler/src/cook/mesos/api.clj
scrosby: Thats a scope question..

If we do that, then effectively we're effectively rescoping this PR to do all of #617. Do we want to do that?

scheduler/src/cook/mesos/api.clj
scrosby: Yes, if we refactor the /jobs API endpoint implementation. It'll take time for me to get up to speed on liberator's semantics, because I'm going to have to face the ::params ::query-params, etc. All of the filtering and other functions in the &uuid=XXXX are all wired to UUID's. And if we're going to do that, we could greatly simplify it the implementation & semantics with a small API change.

Do we want really want to rescope this PR that much?

scheduler/src/cook/mesos/api.clj
dposada: I would vote to either do that in this PR or do it immediately after this in a follow-up PR. Basically, I wouldn't want the two functions to stay around for very long.

scheduler/src/cook/mesos/api.clj
dposada: > Thus the proposal for a change.

It's not clear to me what the change is that you're proposing.

dposada:
@scrosby Do we want to keep this PR open?

dposada:
@scrosby Can we close this PR?

PR #853, opened 2018-05-15
Fix Travis-CI apt-get cached installation

DaoWen:

Fix the `apt-get` commands in our Mesos install script.

Cached installs of Mesos are no longer working because the cached debian packages now have conflicts with the default packages in the default travis-ci debian image.


dposada:
@DaoWen Can we go ahead and merge, or do you want to wait for @pschorf to review?

PR #854, opened 2018-05-15
Adds optional waiting to cs tail

dposada:

- adding `--wait` flag to `cs tail`

To allow waiting until the job is running and the file being tailed is available when running, for example, `cs tail -f ...` on a pending job.

### DIFF
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -260,7 +260,8 @@ def query_unique_and_run(clusters, entity_ref, command_fn, wait_seconds=None):
 
     if wait_seconds:
         r = tenacity.Retrying(wait=tenacity.wait_fixed(5),
-                              stop=tenacity.stop_after_delay(wait_seconds))
+                              stop=tenacity.stop_after_delay(wait_seconds),
+                              reraise=True)
         r.call(query_unique_and_run)
     else:
         query_unique_and_run()
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -8,9 +8,8 @@ from functools import partial
 from operator import itemgetter
 from urllib.parse import urlparse, parse_qs
 
-import tenacity
-
 from cook import http, colors, mesos, progress
+from cook.exceptions import CookRetriableException
 from cook.util import is_valid_uuid, wait_until, print_info, distinct, partition
 
 
@@ -236,7 +235,7 @@ def __get_latest_instance(job):
             instance = max(instances, key=itemgetter('start_time'))
             return instance
 
-    raise Exception(f'Job {job["uuid"]} currently has no instances.')
+    raise CookRetriableException(f'Job {job["uuid"]} currently has no instances.')
 
 
 def query_unique_and_run(clusters, entity_ref, command_fn, wait=False):
@@ -259,7 +258,11 @@ def query_unique_and_run(clusters, entity_ref, command_fn, wait=False):
             raise Exception(f'Encountered error when querying for {entity_ref}.')
 
     if wait:
-        r = tenacity.Retrying(wait=tenacity.wait_fixed(5))
+        # Importing tenacity locally to prevent startup time
+        # from increasing in the default (i.e. don't wait) case
+        import tenacity
+        r = tenacity.Retrying(wait=tenacity.wait_fixed(5),
+                              retry=tenacity.retry_if_exception_type(CookRetriableException))
         r.call(query_unique_and_run)
     else:
         query_unique_and_run()
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -8,6 +8,8 @@ from functools import partial
 from operator import itemgetter
 from urllib.parse import urlparse, parse_qs
 
+import tenacity
+
 from cook import http, colors, mesos, progress
 from cook.util import is_valid_uuid, wait_until, print_info, distinct, partition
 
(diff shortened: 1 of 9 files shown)


cli/cook/subcommands/tail.py
```diff
@@ -124,14 +126,20 @@ def tail(clusters, args, _):
     lines = args.get('lines')
     follow = args.get('follow')
     sleep_interval = args.get('sleep-interval')
+    wait_seconds = args.get('wait')
+
+    if wait_seconds == -1:
+        wait_seconds = None
+    elif not wait_seconds:
```
DaoWen: In Python, `not 0.0` is `True`. I don't think that leads to the behavior we want here.

cli/cook/subcommands/tail.py
```diff
@@ -124,14 +126,20 @@ def tail(clusters, args, _):
     lines = args.get('lines')
     follow = args.get('follow')
     sleep_interval = args.get('sleep-interval')
+    wait_seconds = args.get('wait')
+
+    if wait_seconds == -1:
```
DaoWen: I think we should change this to `if wait_seconds <= 0`.

cli/cook/subcommands/tail.py
```diff
@@ -143,6 +151,9 @@ def register(add_parser, add_defaults):
     parser.add_argument('--sleep-interval', '-s',
                         help=f'with -f, sleep for N seconds (default {DEFAULT_FOLLOW_SLEEP_SECS}) between iterations',
                         metavar='N', type=float)
+    parser.add_argument('--wait', '-w',
+                        help=f'wait for up to N seconds (default is to wait indefinitely)',
+                        metavar='N', type=float, nargs='?', default=-1)
```
DaoWen: It looks like either the help string or the default is wrong here.

I think the default should be `float('inf')` (i.e., wait indefinitely—which matches the current help string description), and the user can do `-w0` if they don't want to wait.

cli/cook/subcommands/tail.py
```diff
@@ -143,6 +151,9 @@ def register(add_parser, add_defaults):
     parser.add_argument('--sleep-interval', '-s',
                         help=f'with -f, sleep for N seconds (default {DEFAULT_FOLLOW_SLEEP_SECS}) between iterations',
                         metavar='N', type=float)
+    parser.add_argument('--wait', '-w',
+                        help=f'wait for up to N seconds (default is to wait indefinitely)',
```
DaoWen: This doesn't need to be an f-string.

cli/cook/subcommands/tail.py
```diff
@@ -124,14 +126,20 @@ def tail(clusters, args, _):
     lines = args.get('lines')
     follow = args.get('follow')
     sleep_interval = args.get('sleep-interval')
+    wait_seconds = args.get('wait')
+
+    if wait_seconds == -1:
+        wait_seconds = None
+    elif not wait_seconds:
+        wait_seconds = sys.maxsize
```
DaoWen: I think we should change the default, then we can get rid of this branch altogether. See below.

cli/cook/subcommands/tail.py
DaoWen: FYI, it doesn't look like there'd be any issue with passing `float('inf')` into the tenacity retry annotation: https://github.com/jd/tenacity/blob/a86237f4d0dee13d328e819d7d9bf038c8bb0a77/tenacity/stop.py#L90-L97

cli/cook/subcommands/tail.py
dposada: Done, thanks

cli/cook/subcommands/tail.py
dposada: This is removed now

cli/cook/querying.py
```diff
@@ -8,6 +8,8 @@
 from operator import itemgetter
 from urllib.parse import urlparse, parse_qs
 
+import tenacity
```
sradack: Are we worried that this might increase startup time?

cli/cook/querying.py
```diff
@@ -237,22 +239,30 @@ def __get_latest_instance(job):
...
+            directory = mesos.retrieve_instance_sandbox_directory(instance, job)
+            command_fn(instance, directory)
+        else:
+            # This should not happen, because query_unique should
+            # only return a map with type "job" or type "instance"
+            raise Exception(f'Encountered error when querying for {entity_ref}.')
+
+    if wait:
+        r = tenacity.Retrying(wait=tenacity.wait_fixed(5))
+        r.call(query_unique_and_run)
```
sradack: Does this eventually give up?  Should we consider retrying only for specific exceptions?

cli/cook/querying.py
dposada: Changed it to a local import that only happens in the `--wait` case

cli/cook/querying.py
dposada: This never gives up. I introduced `CookRetriableException` and switched to using it in the specific spots where we want retrying.

DaoWen: I like this change. Now we won't get into an infinite retry loop when there's some unexpected error.

dposada: How about if we add a separate flag (e.g. `--wait-for '5 minutes'`) for the eventually giving up case, and leave the `--wait` as an indefinite wait? It's definitely true that those exceptions may not get resolved. As a user, though, I may want this to block until they do get resolved. For example, I submit a job, and it may take a long time for it to start, and then once it starts it may take a long time for my file to appear, but I would prefer to just wait until those things are true without having to decide if I should wait 30 minutes or 90 minutes or longer.

dposada: @sradack What are your thoughts on this?

dposada: @sradack ping

dposada: @sradack I added a 1-day timeout on the `--wait`, and it re-raises the last exception

PR #861, opened 2018-05-22
Release v1.17.0

pschorf:


scheduler/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file
  
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.17.0] - 2018-05-22
+### Added
+- Pool support to /jons, from @dposada
```
DaoWen: `/jons` → `/jobs`

pschorf: Fixed

dposada:
From #847:

> This is a functional change for non-internal users; if you want atomic submission for batches of jobs, you'll need to use :commit-latch to do it. If you are not, then partially committed groups of jobs (committed by metatransactions) will be visible in /list queries.

pschorf:
I don't think this is worth calling out as a breaking change, since it requires you to "integrate" with the scheduler by writing jobs directly to the database. If you use the scheduler APIs to submit jobs, there's no impact.

PR #865, opened 2018-05-22
Adds support for running test jobs with docker

dposada:

- if `COOK_TEST_DOCKER_IMAGE` is set, use it in `util.minimal_job`
- if `COOK_TEST_SKIP_SANDBOX_DUMP` is set, don't attempt to dump sandbox contents
- in several tests, switch from checking if the cook executor is configured to checking if the cook executor was actually used before assuming that the exit code is present
- introducing `util.has_ephemeral_hosts` which returns `True` if either `COOK_TEST_EPHEMERAL_HOSTS` is set, or if the estimated completion constraint is configured
- in several tests, if the cluster under test has ephemeral hosts, then bail out instead of using `HOSTNAME` `EQUALS` constraints
- if `COOK_TEST_DEFAULT_TIMEOUT_MS` is set, use it instead of the default-default timeout of 2 minutes

We would like to be able to run our integration tests using docker containers for the test jobs.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1945,7 +1945,7 @@ class CookTest(util.CookTest):
         resp = util.set_limit(self.cook_url, 'quota', user, cpus=10, reason=None)
         self.assertEqual(resp.status_code, 400, resp.text)
         # reset user quota back to default
-        resp = util.reset_limit(self.cook_url, 'quota', user)
+        resp = util.reset_limit(self.cook_url, 'quota', user, reason=self.current_name())
         self.assertEqual(resp.status_code, 204, resp.text)
         # reset user quota fails (malformed) if no reason is given
         resp = util.reset_limit(self.cook_url, 'quota', user, reason=None)
@@ -1957,7 +1957,7 @@ class CookTest(util.CookTest):
         resp = util.set_limit(self.cook_url, 'share', user, cpus=10, reason=None)
         self.assertEqual(resp.status_code, 400, resp.text)
         # reset user share back to default
-        resp = util.reset_limit(self.cook_url, 'share', user)
+        resp = util.reset_limit(self.cook_url, 'share', user, reason=self.current_name())
         self.assertEqual(resp.status_code, 204, resp.text)
         # reset user share fails (malformed) if no reason is given
         resp = util.reset_limit(self.cook_url, 'share', user, reason=None)
@@ -1987,7 +1987,7 @@ class CookTest(util.CookTest):
                 self.assertEqual(100, resp.json()['cpus'], resp.text)
 
                 # Delete the default pool limit (no pool argument)
-                resp = util.reset_limit(self.cook_url, limit, user)
+                resp = util.reset_limit(self.cook_url, limit, user, reason=self.current_name())
                 self.assertEqual(resp.status_code, 204, resp.text)
 
                 # Check that the default is returned for the default pool
@@ -2006,7 +2006,7 @@ class CookTest(util.CookTest):
                     default_cpus = resp.json()['cpus']
 
                     # delete the pool's limit
-                    resp = util.reset_limit(self.cook_url, limit, user, pool=pool)
+                    resp = util.reset_limit(self.cook_url, limit, user, pool=pool, reason=self.current_name())
                     self.assertEqual(resp.status_code, 204, resp.text)
 
                     # check that the default value is returned
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2080,8 +2080,12 @@ class CookTest(util.CookTest):
         self.assertEqual(resp.status_code, 201, msg=resp.content)
         job_uuids = [job_uuid_1, job_uuid_2, job_uuid_3]
         try:
-            util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
-            instances = [util.wait_for_instance(self.cook_url, j) for j in job_uuids]
+            jobs = util.wait_for_jobs(self.cook_url, job_uuids, 'completed')
+            instances = []
+            for job in jobs:
+                for instance in job['instances']:
+                    instance['parent'] = job
+                    instances.append(instance)
             start_time = min(i['start_time'] for i in instances)
             end_time = max(i['start_time'] for i in instances)
             stats, _ = util.get_instance_stats(self.cook_url,
@@ -2091,9 +2095,9 @@ class CookTest(util.CookTest):
                                                name=name)
             user = util.get_user(self.cook_url, job_uuid_1)
             stats_overall = stats['overall']
-            self.assertEqual(3, stats_overall['count'])
-            self.assertEqual(3, stats['by-reason']['Command exited non-zero']['count'])
-            self.assertEqual(3, stats['by-user-and-reason'][user]['Command exited non-zero']['count'])
+            self.assertEqual(len(instances), stats_overall['count'])
+            self.assertEqual(len(instances), stats['by-reason']['Command exited non-zero']['count'])
+            self.assertEqual(len(instances), stats['by-user-and-reason'][user]['Command exited non-zero']['count'])
             run_times = [(i['end_time'] - i['start_time']) / 1000 for i in instances]
             run_time_seconds = stats_overall['run-time-seconds']
             percentiles = run_time_seconds['percentiles']
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1116,18 +1116,15 @@ class CookTest(unittest.TestCase):
         self.assertEqual(resp.status_code, 500)
 
     @pytest.mark.xfail
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_constraints(self):
         """
         Marked as explicit due to:
         RuntimeError: Job ... had status running - expected completed
         """
-        if util.has_ephemeral_hosts(self.cook_url):
-            # If the cluster under test has ephemeral hosts, then it's generally a bad
-            # idea to use HOSTNAME EQUALS constraints, because it can cause the process
-            # responsible for launching hosts to launch hosts that never get used
-            self.logger.info('Bailing out because the cluster has ephemeral hosts')
-            return
-
         state = util.get_mesos_state(self.mesos_url)
         hosts = [agent['hostname'] for agent in state['slaves']][:10]
 
@@ -1439,14 +1436,11 @@ class CookTest(unittest.TestCase):
         resp = util.query_groups(self.cook_url)
         self.assertEqual(400, resp.status_code)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_queue_endpoint(self):
-        if util.has_ephemeral_hosts(self.cook_url):
-            # If the cluster under test has ephemeral hosts, then it's generally a bad
-            # idea to use HOSTNAME EQUALS constraints, because it can cause the process
-            # responsible for launching hosts to launch hosts that never get used
-            self.logger.info('Bailing out because the cluster has ephemeral hosts')
-            return
-
         constraints = [["HOSTNAME", "EQUALS", "lol won't get scheduled"]]
         group = {'uuid': str(uuid.uuid4())}
         job_spec = {'group': group['uuid'],
@@ -1553,14 +1547,11 @@ class CookTest(unittest.TestCase):
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
             mesos.dump_sandbox_files(util.session, instance, job)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_unscheduled_jobs(self):
-        if util.has_ephemeral_hosts(self.cook_url):
-            # If the cluster under test has ephemeral hosts, then it's generally a bad
-            # idea to use HOSTNAME EQUALS constraints, because it can cause the process
-            # responsible for launching hosts to launch hosts that never get used
-            self.logger.info('Bailing out because the cluster has ephemeral hosts')
-            return
-
         unsatisfiable_constraint = ['HOSTNAME', 'EQUALS', 'fakehost']
         job_uuid_1, resp = util.submit_job(self.cook_url, command='ls', constraints=[unsatisfiable_constraint])
         self.assertEqual(resp.status_code, 201, resp.content)
@@ -1591,14 +1582,11 @@ class CookTest(unittest.TestCase):
         finally:
             util.kill_jobs(self.cook_url, [job_uuid_1, job_uuid_2])
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_unscheduled_jobs_partial(self):
-        if util.has_ephemeral_hosts(self.cook_url):
-            # If the cluster under test has ephemeral hosts, then it's generally a bad
-            # idea to use HOSTNAME EQUALS constraints, because it can cause the process
-            # responsible for launching hosts to launch hosts that never get used
-            self.logger.info('Bailing out because the cluster has ephemeral hosts')
-            return
-
         unsatisfiable_constraint = ['HOSTNAME', 'EQUALS', 'fakehost']
         job_uuid_1, resp = util.submit_job(self.cook_url, command='ls', constraints=[unsatisfiable_constraint])
         self.assertEqual(resp.status_code, 201, resp.content)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -66,7 +66,7 @@ class CookTest(unittest.TestCase):
             message = json.dumps(instance, sort_keys=True)
             self.assertEqual(0, instance['exit_code'], message)
         else:
-            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
+            self.logger.info(f'Exit code not checked because cook executor was not used for {instance}')
 
     def test_no_cook_executor_on_subsequent_instances(self):
         retry_limit = util.get_in(util.settings(self.cook_url), 'executor', 'retry-limit')
@@ -106,18 +106,15 @@ class CookTest(unittest.TestCase):
                 message = f'Trailing instance {i}: {json.dumps(job_instance, sort_keys=True)}'
                 self.assertEqual('mesos', job_instance['executor'], message)
 
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_disable_mea_culpa(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
-        if job_executor_type != 'cook':
-            return
+        self.assertEqual('cook', job_executor_type)
         uuid, resp = util.submit_job(self.cook_url, command='sleep 30', env={'EXECUTOR_TEST_EXIT': '1'},
                                      disable_mea_culpa_retries=True)
         try:
             instance = util.wait_for_instance(self.cook_url, uuid)
-            if instance['executor'] != 'cook':
-                self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
-                return
-
+            self.assertEqual('cook', instance['executor'])
             job = util.wait_for_job(self.cook_url, uuid, 'completed')
             self.assertEqual(job['state'], 'failed', json.dumps(job, indent=2))
             self.assertEqual(job['retries_remaining'], 0, json.dumps(job, indent=2))
@@ -127,18 +124,14 @@ class CookTest(unittest.TestCase):
         finally:
             util.kill_jobs(self.cook_url, [uuid])
 
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_mea_culpa_retries(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
-        if job_executor_type != 'cook':
-            return
-
+        self.assertEqual('cook', job_executor_type)
         uuid, resp = util.submit_job(self.cook_url, command='sleep 30', env={'EXECUTOR_TEST_EXIT': '1'})
         try:
             instance = util.wait_for_instance(self.cook_url, uuid)
-            if instance['executor'] != 'cook':
-                self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
-                return
-
+            self.assertEqual('cook', instance['executor'])
             job = util.wait_until(lambda: util.load_job(self.cook_url, uuid),
                                   lambda job: len(job['instances']) > 1 and any(
                                       [i['status'] == 'failed' for i in job['instances']]))
@@ -210,8 +203,9 @@ class CookTest(unittest.TestCase):
             message = json.dumps(instance, sort_keys=True)
             self.assertEqual(1, instance['exit_code'], message)
         else:
-            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
+            self.logger.info(f'Exit code not checked because cook executor was not used for {instance}')
 
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_progress_update_submit(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
         progress_file_env = util.retrieve_progress_file_env(self.cook_url)
@@ -230,17 +224,15 @@ class CookTest(unittest.TestCase):
         message = json.dumps(instance, sort_keys=True)
         self.assertIsNotNone(instance['output_url'], message)
         self.assertIsNotNone(instance['sandbox_directory'], message)
+        self.assertEqual('cook', instance['executor'])
+        util.sleep_for_publish_interval(self.cook_url)
+        instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+        message = json.dumps(instance, sort_keys=True)
+        self.assertEqual(0, instance['exit_code'], message)
+        self.assertEqual(25, instance['progress'], message)
+        self.assertEqual('Twenty-five percent in progress.txt', instance['progress_message'], message)
 
-        if instance['executor'] == 'cook':
-            util.sleep_for_publish_interval(self.cook_url)
-            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
-            message = json.dumps(instance, sort_keys=True)
-            self.assertEqual(0, instance['exit_code'], message)
-            self.assertEqual(25, instance['progress'], message)
-            self.assertEqual('Twenty-five percent in progress.txt', instance['progress_message'], message)
-        else:
-            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
-
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_configurable_progress_update_submit(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
         command = 'echo "message: 25 Twenty-five percent" > progress_file.txt; sleep 1; exit 0'
@@ -261,17 +253,15 @@ class CookTest(unittest.TestCase):
         message = json.dumps(instance, sort_keys=True)
         self.assertIsNotNone(instance['output_url'], message)
         self.assertIsNotNone(instance['sandbox_directory'], message)
+        self.assertEqual('cook', instance['executor'])
+        util.sleep_for_publish_interval(self.cook_url)
+        instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+        message = json.dumps(instance, sort_keys=True)
+        self.assertEqual(0, instance['exit_code'], message)
+        self.assertEqual(25, instance['progress'], message)
+        self.assertEqual('Twenty-five percent', instance['progress_message'], message)
 
-        if instance['executor'] == 'cook':
-            util.sleep_for_publish_interval(self.cook_url)
-            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
-            message = json.dumps(instance, sort_keys=True)
-            self.assertEqual(0, instance['exit_code'], message)
-            self.assertEqual(25, instance['progress'], message)
-            self.assertEqual('Twenty-five percent', instance['progress_message'], message)
-        else:
-            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
-
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_multiple_progress_updates_submit(self):
         job_executor_type = util.get_job_executor_type(self.cook_url)
         line_1 = util.progress_line(self.cook_url, 25, 'Twenty-five percent')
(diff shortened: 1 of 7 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -56,15 +56,17 @@ def test_basic_submit(self):
...
+        message = json.dumps(instance, sort_keys=True)
+        self.assertIsNotNone(instance['output_url'], message)
+        self.assertIsNotNone(instance['sandbox_directory'], message)
+
+        if instance['executor'] == 'cook':
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -110,12 +112,20 @@ def test_disable_mea_culpa(self):
...
                                      disable_mea_culpa_retries=True)
-        job = util.wait_for_job(self.cook_url, uuid, 'completed')
-        self.assertEqual(job['state'], 'failed', json.dumps(job, indent=2))
-        self.assertEqual(job['retries_remaining'], 0, json.dumps(job, indent=2))
-        instances = job['instances']
-        self.assertEqual(1, len(instances), json.dumps(job, indent=2))
-        self.assertEqual('Mesos executor terminated', instances[0]['reason_string'], json.dumps(job, indent=2))
+        try:
+            instance = util.wait_for_instance(self.cook_url, uuid)
+            if instance['executor'] != 'cook':
```
shamsimam: I think this test should have a `skipIf` annotation as there is no point in running the test if Cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -124,6 +134,11 @@ def test_mea_culpa_retries(self):
 
         uuid, resp = util.submit_job(self.cook_url, command='sleep 30', env={'EXECUTOR_TEST_EXIT': '1'})
         try:
+            instance = util.wait_for_instance(self.cook_url, uuid)
+            if instance['executor'] != 'cook':
```
shamsimam: I think this test should have a `skipIf` annotation as there is no point in running the test if Cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -185,15 +200,17 @@ def test_failing_submit(self):
...
-        if job_executor_type == 'cook':
-            job = util.wait_for_exit_code(self.cook_url, job_uuid)
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(1, job['instances'][0]['exit_code'], message)
+        if instance['executor'] == 'cook':
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(1, instance['exit_code'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -206,26 +223,23 @@ def test_progress_update_submit(self):
...
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(25, job['instances'][0]['progress'], message)
-            self.assertEqual('Twenty-five percent in progress.txt', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(25, instance['progress'], message)
+            self.assertEqual('Twenty-five percent in progress.txt', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code and progress info not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -243,22 +257,20 @@ def test_configurable_progress_update_submit(self):
...
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(25, job['instances'][0]['progress'], message)
-            self.assertEqual('Twenty-five percent', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(25, instance['progress'], message)
+            self.assertEqual('Twenty-five percent', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code and progress info not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -206,26 +223,23 @@ def test_progress_update_submit(self):
...
-            self.assertEqual('Twenty-five percent in progress.txt', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(25, instance['progress'], message)
+            self.assertEqual('Twenty-five percent in progress.txt', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
 
     def test_configurable_progress_update_submit(self):
```
shamsimam: I think we should skip this test if cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -185,15 +200,17 @@ def test_failing_submit(self):
...
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(1, job['instances'][0]['exit_code'], message)
+        if instance['executor'] == 'cook':
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(1, instance['exit_code'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
 
     def test_progress_update_submit(self):
```
shamsimam: I think we should skip this test if cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -243,22 +257,20 @@ def test_configurable_progress_update_submit(self):
...
-            self.assertEqual('Twenty-five percent', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(25, instance['progress'], message)
+            self.assertEqual('Twenty-five percent', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
 
     def test_multiple_progress_updates_submit(self):
```
shamsimam: I think we should skip this test if cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -277,22 +289,20 @@ def test_multiple_progress_updates_submit(self):
...
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(75, job['instances'][0]['progress'], message)
-            self.assertEqual('Seventy-five percent', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(75, instance['progress'], message)
+            self.assertEqual('Seventy-five percent', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code and progress info not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -309,22 +319,20 @@ def progress_string(a):
...
-            message = json.dumps(job['instances'][0], sort_keys=True)
-            self.assertEqual(80, job['instances'][0]['progress'], message)
-            self.assertEqual('80%', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(80, instance['progress'], message)
+            self.assertEqual('80%', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code and progress info not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -277,22 +289,20 @@ def test_multiple_progress_updates_submit(self):
...
-            self.assertEqual('Seventy-five percent', job['instances'][0]['progress_message'], message)
+            instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+            message = json.dumps(instance, sort_keys=True)
+            self.assertEqual(0, instance['exit_code'], message)
+            self.assertEqual(75, instance['progress'], message)
+            self.assertEqual('Seventy-five percent', instance['progress_message'], message)
+        else:
+            self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
 
     def test_multiple_rapid_progress_updates_submit(self):
```
shamsimam: I think we should skip this test if cook executor is not in use.

integration/tests/cook/test_basic.py
```diff
@@ -370,16 +378,18 @@ def test_max_runtime_exceeded(self):
...
-            if job_executor_type == 'cook':
-                job = util.wait_for_exit_code(self.cook_url, job_uuid)
-                message = json.dumps(job['instances'][0], sort_keys=True)
-                self.assertNotEqual(0, job['instances'][0]['exit_code'], message)
+            if instance['executor'] == 'cook':
+                instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+                message = json.dumps(instance, sort_keys=True)
+                self.assertNotEqual(0, instance['exit_code'], message)
+            else:
+                self.logger.info(f'Bailing out because the cook executor is not being used: {instance}')
```
shamsimam: Should we say `exit code not checked as cook executor was not used for {instance}`?

integration/tests/cook/test_basic.py
```diff
@@ -1106,6 +1116,10 @@ def test_error_while_creating_job(self):
         self.assertEqual(resp.status_code, 500)
 
     @pytest.mark.xfail
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_basic.py
```diff
@@ -1422,6 +1436,10 @@ def test_400_on_group_query_without_uuid(self):
         resp = util.query_groups(self.cook_url)
         self.assertEqual(400, resp.status_code)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_basic.py
```diff
@@ -1530,6 +1547,10 @@ def get_docker_info():
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
             mesos.dump_sandbox_files(util.session, instance, job)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_basic.py
```diff
@@ -1561,6 +1582,10 @@ def check_unscheduled_reason():
         finally:
             util.kill_jobs(self.cook_url, [job_uuid_1, job_uuid_2])
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_cli.py
```diff
@@ -429,6 +429,10 @@ def list_jobs(self, name, user, *states):
         cp, jobs = cli.jobs_json(self.cook_url, '--name %s --user %s %s' % (name, user, state_flags))
         return cp, jobs
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_cli.py
```diff
@@ -623,6 +627,10 @@ def test_ssh_job_uuid(self):
         self.assertIn(f'-t {hostname} cd', stdout)
         self.assertIn('; bash', stdout)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_cli.py
```diff
@@ -1606,6 +1614,10 @@ def test_cat_bogus_uuid(self):
         self.assertEqual(1, cp.returncode, cp.stdout)
         self.assertIn('No matching data found', cli.decode(cp.stderr))
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
```
shamsimam: Can we have this string in a constant somewhere and refer to the constant in the annotation?

integration/tests/cook/test_basic.py
```diff
@@ -1530,6 +1547,10 @@ def get_docker_info():
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
             mesos.dump_sandbox_files(util.session, instance, job)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_unscheduled_jobs(self):
         unsatisfiable_constraint = ['HOSTNAME', 'EQUALS', 'fakehost']
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/test_basic.py
```diff
@@ -1561,6 +1582,10 @@ def check_unscheduled_reason():
         finally:
             util.kill_jobs(self.cook_url, [job_uuid_1, job_uuid_2])
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_unscheduled_jobs_partial(self):
         unsatisfiable_constraint = ['HOSTNAME', 'EQUALS', 'fakehost']
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/test_cli.py
```diff
@@ -429,6 +429,10 @@ def list_jobs(self, name, user, *states):
         cp, jobs = cli.jobs_json(self.cook_url, '--name %s --user %s %s' % (name, user, state_flags))
         return cp, jobs
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_list_by_state(self):
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/test_cli.py
```diff
@@ -623,6 +627,10 @@ def test_ssh_job_uuid(self):
         self.assertIn(f'-t {hostname} cd', stdout)
         self.assertIn('; bash', stdout)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_ssh_no_instances(self):
         raw_job = {'command': 'ls', 'constraints': [['HOSTNAME', 'EQUALS', 'will not get scheduled']]}
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/test_cli.py
```diff
@@ -1606,6 +1614,10 @@ def test_cat_bogus_uuid(self):
         self.assertEqual(1, cp.returncode, cp.stdout)
         self.assertIn('No matching data found', cli.decode(cp.stderr))
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_cat_job_with_no_instances(self):
         raw_job = {'command': 'ls', 'constraints': [['HOSTNAME', 'EQUALS', 'will not get scheduled']]}
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/test_master_slave.py
```diff
@@ -1,12 +1,13 @@
 import logging
 import os
-import pytest
-import time
 import unittest
 
+import pytest
```
shamsimam: Why move this import lower down?

integration/tests/cook/test_master_slave.py
```diff
@@ -24,6 +25,10 @@ def setUp(self):
         self.slave_url = type(self).slave_url
         self.logger = logging.getLogger(__name__)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_get_queue(self):
         job_uuid, resp = util.submit_job(self.master_url, constraints=[["HOSTNAME",
```
shamsimam: Is it possible to use a different constraint that causes jobs to not be scheduled and avoid the issue of launching hosts that never get used?

integration/tests/cook/util.py
```diff
@@ -681,15 +696,16 @@ def predicate(job):
...
-                logger.info(
-                    f"Job {job_id} instance {inst['task_id']} has sandbox directory {inst['sandbox_directory']}.")
-                return True
+            for inst in job['instances']:
+                if 'sandbox_directory' not in inst:
+                    logger.info(f"Job {job_id} instance {inst['task_id']} has no sandbox directory.")
+                else:
+                    logger.info(
+                        f"Job {job_id} instance {inst['task_id']} has sandbox directory {inst['sandbox_directory']}.")
+                    job['instance-with-sandbox'] = inst
```
shamsimam: Should we handle `truthy` values returned from the predicate in `wait_until`? Then you could have just returned `inst` from this function.

integration/tests/cook/util.py
```diff
@@ -1045,3 +1059,24 @@ def active_pools(cook_url):
...
+def _cook_estimated_completion_constraint():
+    """Get the estimated completion constraint config from the /settings endpoint"""
+    cook_url = retrieve_cook_url()
+    _wait_for_cook(cook_url)
+    estimated_completion_constraint = settings(cook_url).get('estimated-completion-constraint', None)
+    logger.info(f"Cook's estimated completion constraint is {estimated_completion_constraint}")
+    return estimated_completion_constraint
+
+
+def has_ephemeral_hosts():
```
shamsimam: Should we cache this result to avoid repeated invocations?

integration/tests/cook/test_basic.py
pschorf: I think the issue here is that it might be configured but not used for a particular instance (like in pub cloud) which you wouldn't know until running it. Maybe we can expose this via settings and then make the change?

integration/tests/cook/test_basic.py
```diff
@@ -1106,6 +1116,10 @@ def test_error_while_creating_job(self):
         self.assertEqual(resp.status_code, 500)
 
     @pytest.mark.xfail
+    @unittest.skipIf(util.has_ephemeral_hosts(), 'If the cluster under test has ephemeral hosts, then it is generally '
+                                                 'a bad idea to use HOSTNAME EQUALS constraints, because it can cause '
+                                                 'the process responsible for launching hosts to launch hosts that '
+                                                 'never get used')
     def test_constraints(self):
```
pschorf: I know it's not part of your PR, but can you rename this so it's specifically referring to the hostname equals constraint? I was concerned we were turning off something more generic.

integration/tests/cook/test_basic.py
pschorf: 👍

integration/tests/cook/test_basic.py
pschorf: :+1:

integration/tests/cook/util.py
pschorf: I'd prefer not to cache this, it's just looking up an env var in test setup. Seems like overkill to me.

integration/tests/cook/test_basic.py
dposada: @pschorf What do you mean by

> we can expose this via settings

dposada: @shamsimam and I discussed and landed on adding an environment variable, e.g. `COOK_TEST_COOK_EXECUTOR`, which can be used to skip tests like this that assume the cook executor is in use

pschorf: What about using a combo of checking if it's configured and you're not submitting via docker? Would that cover it? We already have so many environment variables.

integration/tests/cook/test_master_slave.py
dposada: From https://www.python.org/dev/peps/pep-0008/#imports:

> Imports should be grouped in the following order:
>
> Standard library imports.
> Related third party imports.
> Local application/library specific imports.
>
> You should put a blank line between each group of imports.

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py
dposada: Good call, done (see `util.is_cook_executor_in_use`)

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

integration/tests/cook/test_cli.py
dposada: Done

integration/tests/cook/test_cli.py
dposada: Done

integration/tests/cook/test_cli.py
dposada: Done

integration/tests/cook/test_basic.py
dposada: I changed this test to build up a queue and then assert that the "There are X jobs ahead in the queue" reason is listed

integration/tests/cook/test_basic.py
dposada: This didn't actually need the constraint, so I removed it

integration/tests/cook/test_cli.py
dposada: The CLI tests don't currently support docker jobs anyway, so I removed the `skipIf` until we address that in a separate PR

integration/tests/cook/test_cli.py
dposada: The CLI tests don't currently support docker jobs anyway, so I removed the `skipIf` until we address that in a separate PR

integration/tests/cook/test_cli.py
dposada: The CLI tests don't currently support docker jobs anyway, so I removed the `skipIf` until we address that in a separate PR

integration/tests/cook/test_master_slave.py
dposada: I changed this test to build up a queue

integration/tests/cook/util.py
dposada: That seems like a pretty major change. A lot of our predicates are probably just returning `True`/`False`, yet the caller is expecting `wait_until` to return the result of the `query` function.

integration/tests/cook/test_basic.py

dposada:
@pschorf @shamsimam This is ready for another pass

integration/tests/cook/test_basic.py
```diff
@@ -206,27 +217,22 @@ def test_progress_update_submit(self):
...
+        self.assertEqual('cook', instance['executor'])
+        util.sleep_for_publish_interval(self.cook_url)
+        instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+        message = json.dumps(instance, sort_keys=True)
+        self.assertEqual(0, instance['exit_code'], message)
+        self.assertEqual(25, instance['progress'], message)
+        self.assertEqual('Twenty-five percent in progress.txt', instance['progress_message'], message)
+
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_configurable_progress_update_submit(self):
```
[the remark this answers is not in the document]
dposada: It is (`@unittest.skipUnless(util.is_cook_executor_in_use(), ...)`)

integration/tests/cook/test_basic.py
```diff
@@ -243,23 +249,19 @@ def test_configurable_progress_update_submit(self):
...
+        self.assertEqual('cook', instance['executor'])
+        util.sleep_for_publish_interval(self.cook_url)
+        instance = util.wait_for_exit_code(self.cook_url, job_uuid)
+        message = json.dumps(instance, sort_keys=True)
+        self.assertEqual(0, instance['exit_code'], message)
+        self.assertEqual(25, instance['progress'], message)
+        self.assertEqual('Twenty-five percent', instance['progress_message'], message)
+
+    @unittest.skipUnless(util.is_cook_executor_in_use(), 'Test assumes the Cook Executor is in use')
     def test_multiple_progress_updates_submit(self):
```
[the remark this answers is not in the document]
dposada: It is (`@unittest.skipUnless(util.is_cook_executor_in_use(), ...)`)

integration/tests/cook/util.py
dposada: Why would the helper function be better?

integration/tests/cook/test_basic.py
```diff
@@ -1423,41 +1419,37 @@ def test_400_on_group_query_without_uuid(self):
...
 
     def test_queue_endpoint(self):
-        constraints = [["HOSTNAME", "EQUALS", "lol won't get scheduled"]]
         group = {'uuid': str(uuid.uuid4())}
-        job_spec = {'group': group['uuid'],
-                    'constraints': constraints}
-        uuids, resp = util.submit_jobs(self.cook_url, job_spec, 1, groups=[group])
-        job_uuid = uuids[0]
+        job_spec = {'group': group['uuid'], 'command': 'sleep 30'}
+        uuids, resp = util.submit_jobs(self.cook_url, job_spec, clones=100, groups=[group])
```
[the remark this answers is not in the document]
pschorf: What about using an attribute equals constraint on a bogus attribute? Like `[uuid4(), "EQUALS", uuid4()]`?

integration/tests/cook/test_basic.py
```diff
@@ -1531,12 +1523,13 @@ def get_docker_info():
...
 
     def test_unscheduled_jobs(self):
-        unsatisfiable_constraint = ['HOSTNAME', 'EQUALS', 'fakehost']
-        job_uuid_1, resp = util.submit_job(self.cook_url, command='ls', constraints=[unsatisfiable_constraint])
-        self.assertEqual(resp.status_code, 201, resp.content)
-        job_uuid_2, resp = util.submit_job(self.cook_url, command='ls', constraints=[unsatisfiable_constraint])
+        uuids, resp = util.submit_jobs(self.cook_url, {'command': 'sleep 30', 'priority': 100}, clones=100)
         self.assertEqual(resp.status_code, 201, resp.content)
         try:
+            job_uuid_1, resp = util.submit_job(self.cook_url, command='ls', priority=1)
```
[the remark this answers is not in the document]
pschorf: I'd prefer the attribute approach if it works.

integration/tests/cook/util.py
shamsimam: We don't rely on the predicate function changing its input to extract the return value. We could just re-call the helper to extract the return value.

dposada: > We don't rely on the predicate function changing its input to extract the return value

Is your concern a functional concern (it won't always work), or a stylistic concern?

shamsimam: It is a stylistic concern. I also worry it sets a precedent for future usage patterns.

integration/tests/cook/test_basic.py
```diff
@@ -1119,18 +1119,22 @@ def test_hostname_equals_job_constraint(self):
             host_to_job_uuid = {}
             for hostname in hosts:
                 constraints = [["HOSTNAME", "EQUALS", hostname]]
-                job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints)
+                job_uuid, resp = util.submit_job(self.cook_url, constraints=constraints,
+                                                 name='test_hostname_equals_job_constraint')
```
shamsimam: 1. Can we use the [`id()` method](https://docs.python.org/3/library/unittest.html#unittest.TestCase.id) to generate this name?

2. Should we make this strategy the default name (when one is not explicitly provided) for all our jobs in the tests?

integration/tests/cook/util.py
dposada: I'm not really worried about this, and I think the current way is simpler, but if you feel strongly, I can change it.

integration/tests/cook/test_basic.py
dposada: I switched this to use the `current_name()` function that is already in use by some of the CLI tests. As for number 2, that's a little trickier, because we generate the default fields in `util`, which doesn't have access to the `id()` method. I would vote to move that effort to a separate PR.

integration/tests/cook/test_basic.py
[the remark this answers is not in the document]
dposada: I couldn't get this to work (the jobs would get scheduled despite the bogus group constraint)

integration/tests/cook/test_basic.py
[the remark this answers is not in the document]
dposada: I couldn't get this to work (the jobs would get scheduled despite the bogus group constraint)

integration/tests/cook/test_basic.py
shamsimam: Sounds good.

integration/tests/cook/util.py
dposada: Done

dposada:
Build is now green, can we merge?

PR #869, opened 2018-05-23
Minor cleanup in impersonation tests

dposada:

- removing unused `itertools` import
- wrapping overly long line
- removing unused local variable

To keep things clean.


dposada:
@DaoWen Can we merge this?

PR #871, opened 2018-06-01
Fix bug causing reconciliation to fail

pschorf:
- Fix bug with `processed-tasks` in `reconcile-tasks`

Reconciliation is currently not working


shamsimam:
Awaiting green build.

PR #874, opened 2018-06-05
removes superfluous logs from the sandbox publisher

shamsimam:

- removes superfluous logs from the sandbox publisher

Avoids confusion due to the presence of the logs on multiple Cook nodes in a cluster. Now the logs will only appear if the syncer is publishing sandbox directories.


shamsimam:
@dposada build is green

PR #879, opened 2018-06-11
Injects http module dependencies with function call

dposada:

- instead of expecting the user to configure the http modules, allowing them to be passed programmatically

To simplify the workflow for specifying custom http modules.


cli/cook/__main__.py
```diff
@@ -36,4 +36,6 @@ def sigint_handler(_, __):
 signal.signal(signal.SIGINT, sigint_handler)
 
 if __name__ == '__main__':
+    import requests
+    http.inject_http_modules(requests, requests.adapters)
```
leifwalsh: Since these are defaulted in `configure()` anyway, I don't think you need to call this here.

dposada: Done

dposada:
@leifwalsh I don't see `importlib` in `install_requires`

leifwalsh:
Oh weird, I guess you were getting those transitively then.

leifwalsh:
Thank you! Let me know when you’ve pushed a new sdist to PyPI.

dposada:
@leifwalsh https://pypi.org/project/cook-client/2.7.0/

PR #885, opened 2018-06-14
Adds travis build with pools turned off

dposada:

- adding `--pools={on/off}` to `run_integration.sh`
- adding a build to the travis job that runs with `--pools=off`

We want to ensure that changes we make for pools don't break environments that don't have pools.

### DIFF
--- a/scheduler/src/cook/mesos/pool.clj
+++ b/scheduler/src/cook/mesos/pool.clj
@@ -71,7 +71,7 @@
   (let [pools (all-pools db)
         default-pool-name (config/default-pool)]
     (log/info "Pools in the database:" pools ", default pool:" default-pool-name)
-    (if (and default-pool-name (not (str/blank? default-pool-name)))
+    (if default-pool-name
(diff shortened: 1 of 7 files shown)


dposada:
@pschorf Thoughts on this?

scheduler/src/cook/mesos/pool.clj
```diff
@@ -70,7 +71,7 @@
   (let [pools (all-pools db)
         default-pool-name (config/default-pool)]
     (log/info "Pools in the database:" pools ", default pool:" default-pool-name)
-    (if default-pool-name
```
pschorf: Would this be better in the config section, so everyone else doesn't have to worry about it?

dposada: By "this", do you meany the `blank?` check?

pschorf: Right, replacing blank with nil. Sorry for the lack of clarity.

dposada: Done

dposada:

PR #887, opened 2018-06-14
Removes utilization check from rebalancer

dposada:

- removing the `min-utilization-threshold` config option
- removing the `mesos-master-hosts` config option
- removing `cook.mesos/get-mesos-utilization`
- in the rebalancer, removing the utilization check (i.e. run the rebalancer regardless of cluster utilization)

The utilization check was useful before the rebalancer took "spare" resources into account when determining if tasks need to be preempted. Now that these "spare" resources are accounted for (via incubating offers), it doesn't make much sense to only rebalance when cluster utilization is below a certain threshold.


integration/tests/cook/util.py
```diff
@@ -318,13 +318,10 @@ def retrieve_mesos_url(varname='MESOS_PORT', value='5050'):
     mesos_url = os.getenv('COOK_MESOS_LEADER_URL')
     if mesos_url is None:
         mesos_port = os.getenv(varname, value)
-        cook_url = retrieve_cook_url()
-        _wait_for_cook(cook_url)
-        mesos_master_hosts = settings(cook_url).get('mesos-master-hosts', ['localhost'])
-        resp = session.get('http://%s:%s/redirect' % (mesos_master_hosts[0], mesos_port), allow_redirects=False)
+        resp = session.get(f'http://localhost:{mesos_port}/redirect', allow_redirects=False)
```
wyegelwel: Why is it ok to assume localhost here? I'm imagining a case where you run the integration tests on a production cluster and then the tests that check the number of hosts. What am I missing?

scheduler/docs/configuration.adoc
```diff
@@ -111,10 +111,6 @@ We'll look at the configurable options in turn:
   This option sets the Mesos master connection string.
   For example, if you are running Mesos with a Zookeeper node on the local machine (a common development setup), you'd use the connection string `zk://localhost:2181/mesos`.
 
-`:master-hosts`::
```
wyegelwel: Do you think it is worthwhile to have a deprecated or removed section in our config doc so users who upgrade aren't confused why the setting they used to use is no longer documented. Perhaps as an alternative, having the init in the scheduler print that the config is no longer used? 

I don't think it is a big deal so I'm kind of on the fence about it.

integration/tests/cook/util.py
dposada: You are correct, this is only a decent assumption when running locally. For other environments, we will have to set the `COOK_MESOS_LEADER_URL` environment variable (see 3 lines above).

scheduler/docs/configuration.adoc
dposada: I like the idea of printing that it's no longer used. I will add that.

integration/tests/cook/util.py
wyegelwel: ah ha! That is what I was missing. Thanks, I knew it was something stupid on my part.

scheduler/docs/configuration.adoc
dposada: Done

scheduler/src/cook/config.clj
```diff
@@ -267,17 +267,10 @@
                             (when scheduler
                               (or (:good-enough-fitness scheduler) 0.8)))
      :mesos-master (fnk [[:config {mesos nil}]]
+                     (when (:master-hosts mesos)
+                       (log/warn "The :master-hosts configuration field is no longer used"))
```
shamsimam: 👍

PR #889, opened 2018-06-18
Add support for x-forwarded-proto in CORS requests

pschorf:
- Adds support for `x-forwarded-proto` header for validating CORS requests

Support configurations where Cook could be running behind a reverse proxy.

### DIFF
--- a/scheduler/src/cook/cors.clj
+++ b/scheduler/src/cook/cors.clj
@@ -8,10 +8,10 @@
   "Returns true if the request is from the same origin as the provided origin header"
   [{:keys [headers scheme]}]
   (let [{:strs [host origin x-forwarded-proto]} headers
-        scheme (or x-forwarded-proto
-                   (when scheme (name scheme)))]
-    (when (and host origin scheme)
-      (= origin (str scheme "://" host)))))
+        forwarded-or-scheme (or x-forwarded-proto
+                                (when scheme (name scheme)))]
+    (when (and host origin forwarded-or-scheme)
+      (= origin (str forwarded-or-scheme "://" host)))))
 
 (defn request-allowed?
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/cors.clj
```diff
@@ -7,9 +7,11 @@
 (defn same-origin?
   "Returns true if the request is from the same origin as the provided origin header"
   [{:keys [headers scheme]}]
-  (let [{:strs [host origin]} headers]
+  (let [{:strs [host origin x-forwarded-proto]} headers
+        scheme (or x-forwarded-proto
+                   (when scheme (name scheme)))]
```
dposada: Do we need the `when scheme` check? Is there a scenario where both `x-forwarded-proto` and `scheme` are `nil`?

scheduler/src/cook/cors.clj
```diff
@@ -7,9 +7,11 @@
 (defn same-origin?
   "Returns true if the request is from the same origin as the provided origin header"
   [{:keys [headers scheme]}]
-  (let [{:strs [host origin]} headers]
+  (let [{:strs [host origin x-forwarded-proto]} headers
+        scheme (or x-forwarded-proto
```
dposada: Should we name this something other than `scheme` to avoid confusion with the passed-in `scheme`?

scheduler/src/cook/cors.clj
pschorf: I don't _think_ scheme would ever be nil, but `name` throws an NPE if the argument is null, so I wanted to be defensive.

scheduler/src/cook/cors.clj
pschorf: Will address.

pschorf: Done

PR #906, opened 2018-07-02
Deletes ranker.clj code and tests

dposada:

- deleting `ranker.clj` code and tests

`ranker.clj` was added as the first step in doing incremental ranking but we ended up not doing it.


wyegelwel:
It may make sense to just remove it. ranker.clj was added as the first step in doing incremental ranking but we ended up not doing it.

PR #910, opened 2018-07-09
Makes test_group_kill_simple more reliable

dposada:

- adding a retry to the "slow" job

We've seen this test fail due to the "slow" job getting a transient failure, e.g. `Invalid task`.


dposada:
@DaoWen thanks! can we merge?

PR #914, opened 2018-07-19
Add data local fitness calculator

pschorf:
- Adds a new fitness calculator incorporating data locality

We want to support integrating Cook Scheduler with a service which will compute the cost of running jobs on different machines based on the effects of data locality. Once we have the service integrated, we could use this fitness calculator to enforce the scheduling constraints.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2468,3 +2468,10 @@ class CookTest(util.CookTest):
             else:
                 resp = util.get_limit(self.cook_url, limit, user)
                 self.assertFalse('pools' in resp.json())
+
+
+    def test_data_local_support(self):
+        uuid, resp = util.submit_job(self.cook_url, supports_data_locality=True)
+        self.assertEqual(201, resp.status_code, resp.text)
+        job = util.load_job(self.cook_url, uuid)
+        self.assertEqual(True, job['supports_data_locality'], job)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2471,7 +2471,7 @@ class CookTest(util.CookTest):
 
 
     def test_data_local_support(self):
-        uuid, resp = util.submit_job(self.cook_url, supports_data_locality=True)
+        uuid, resp = util.submit_job(self.cook_url, data_local=True)
         self.assertEqual(201, resp.status_code, resp.text)
         job = util.load_job(self.cook_url, uuid)
-        self.assertEqual(True, job['supports_data_locality'], job)
+        self.assertEqual(True, job['data_local'], job)
--- a/scheduler/src/cook/mesos/data_locality.clj
+++ b/scheduler/src/cook/mesos/data_locality.clj
@@ -34,8 +34,8 @@
   (getName [this] (-> this .getClass .getSimpleName))
   (calculateFitness [this task-request target-vm tracker-state]
     (let [base-fitness (.calculateFitness base-calculator task-request target-vm tracker-state)
-          {:keys [job/uuid job/supports-data-locality]} (:job task-request)]
-      (if supports-data-locality
+          {:keys [job/uuid job/data-local]} (:job task-request)]
+      (if data-local
         (let [normalized-fitness (get-normalized-fitness uuid (.getHostname target-vm) maximum-cost)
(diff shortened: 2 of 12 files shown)


scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1208,6 +1208,7 @@
                    :retries_remaining 1
                    :state "waiting"
                    :status "waiting"
+                   :supports_data_locality false
```
shamsimam: Should `supports_data_locality` only be displayed if the user explicitly specified it on the job?

scheduler/src/cook/mesos/data_locality.clj
```diff
@@ -0,0 +1,30 @@
...
+
+(def data-local-costs (atom {}))
+
+(defn get-normalized-fitness
+  "Returns the fitness for the job to run on the given host, normalized by max-cost"
+  [uuid hostname max-cost]
+  (let [cost (get-in @data-local-costs [uuid hostname] max-cost)]
+    (- 1 (double (/ cost max-cost)))))
+
+(deftype DataLocalFitnessCalculator [^VMTaskFitnessCalculator base-calculator data-locality-weight maximum-cost]
```
shamsimam: I would prefer `DataLocalFitnessCalculator` returning only the data locality cost directly and not having a reference to `base-calculator`.  We can have a separate `WeightedAverageFitnessCalculator` that contains the reference to both `base-calculator` and `data-local-calculator` to compute the weighted average fitness.

pschorf:
Any other feedback @dposada @shamsimam ?

scheduler/src/cook/mesos/data_locality.clj
```diff
@@ -0,0 +1,47 @@
...
+    "Updates the current data local costs. Costs larger than the configured max-cost will be capped at max-cost
+     and negative costs will be reset to 0."
+    [job->host->cost]
+    (let [{:keys [maximum-cost]} (config/data-local-fitness-config)]
+      (->> job->host->cost
+           (pc/map-vals (fn [host->cost]
+                          (pc/map-vals (fn [cost] (-> cost (min maximum-cost) (max 0)))
+                                       host->cost)))
+           (reset! data-local-costs))))
+  (defn get-data-local-costs
```
shamsimam: Nitpick: new line

pschorf: Done

scheduler/src/cook/mesos/schema.clj
```diff
@@ -186,6 +186,12 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
     :db/valueType :db.type/ref
     :db/cardinality :db.cardinality/one
     :db.install/_attribute :db.part/db}
+   {:db/id (d/tempid :db.part/db)
+    :db/ident :job/supports-data-locality
```
dposada: "Supports" seems a little bit odd. Maybe "participates in" (as the `:db/doc` states)? Or maybe we just shorten to `:job/data-local`?

integration/tests/cook/test_basic.py
```diff
@@ -2468,3 +2468,10 @@ def test_pools_in_default_limit_response(self):
             else:
                 resp = util.get_limit(self.cook_url, limit, user)
                 self.assertFalse('pools' in resp.json())
+
+
+    def test_data_local_support(self):
+        uuid, resp = util.submit_job(self.cook_url, supports_data_locality=True)
+        self.assertEqual(201, resp.status_code, resp.text)
+        job = util.load_job(self.cook_url, uuid)
+        self.assertEqual(True, job['supports_data_locality'], job)
```
dposada: Should we add an integration test that asserts that data-locality costs are being taken into account?

pschorf: I'm working on one in the next PR using a fake service. The closest thing I could think of in this PR was running through a match cycle in the scheduler unit tests.

dposada: :+1: for next PR with fake service

scheduler/src/cook/mesos/schema.clj
pschorf: Done

pschorf:
Build is green

PR #918, opened 2018-07-23
Outgoing/temporal UUID

scrosby:

- Use the datomic temporal UUID in two spots in the cook scheduler, the task_id and commit latch. 
- Make temporal UUID code in our JobClient API that can be used by our customers. 
- Change two UUID generation stacks to use datomic/squuid (the datomic temporal UUID generator)

- I'm changing the task ID UUID on general principals. We know from earlier information that random UUID's are expensive, so remove an extra one.


jobclient/src/main/java/com/twosigma/cook/jobclient/JobClient.java
```diff
@@ -44,6 +44,7 @@
 import org.apache.http.StatusLine;
 import org.apache.http.auth.AuthSchemeProvider;
 import org.apache.http.auth.AuthScope;
+
```
pschorf: Nitpick: remove empty line

PR #919, opened 2018-07-23
Use `nil-pool` instead of a new random UUID

pschorf:
- Use `nil-pool` instead of a new random UUID as the default

Save calling `(UUID/randomUUID)` unnecessarily. Also add a comment explaining why this is the default.


dposada:

pschorf:
Should be good to go!

PR #920, opened 2018-07-23
Update cook API to use temporally clustered UUID's.

scrosby:
- Use the datomic temporal UUID in two spots in the cook scheduler, the task_id and commit latch.

- This includes the cook scheduler parts of the change.
- I'm changing the task ID UUID on general principals. We know from earlier information that random UUID's are expensive, so remove an extra one.


shamsimam:
Awaiting green build

shamsimam:
Awaiting green build.

PR #921, opened 2018-07-23
Add UUID generator to JobClient.

scrosby:
- Make temporal UUID code in our JobClient API that can be used by our customers.

- This includes code that customers can call that makes appropriate UUID's.


scrosby:
Yes. This doesn't change any performance --- the code here is technically dead as Cook doesn't use it. But its part of a set of changes that will improve performance to the described level.

PR #922, opened 2018-07-24
Run preemption test serially

pschorf:
- Run preemption test serially

Reduce flakiness by lowering contention on mesos for tests which have specific resource demands.


integration/tests/cook/test_multi_user.py
```diff
@@ -214,6 +214,7 @@ def test_job_count_quota(self):
                 util.reset_limit(self.cook_url, 'quota', user.name, reason=self.current_name())
 
     @unittest.skipUnless(util.is_preemption_enabled(), 'Preemption is not enabled on the cluster')
+    @pytest.mark.serial
```
dposada: Might be good to add a comment about why this should be serial

PR #925, opened 2018-07-25
generalizes the sandbox aggregator functions

shamsimam:

- generalizes the sandbox aggregator functions (in preparation for exit-code)

We would like to batch `exit-code` datomic writes and reuse the sandbox aggregator logic to achieve this.

### DIFF
--- a/scheduler/src/cook/mesos/sandbox.clj
+++ b/scheduler/src/cook/mesos/sandbox.clj
@@ -31,41 +31,41 @@
 (defn clear-agent-state
   "Clears the published entries from the aggregated state of the agent.
    Since we expect instances to have the same field value, we do not check for value equality."
-  [task-id->field-value field-name published-task-id->field-value]
+  [task-id->value field-name published-task-id->value]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -28,77 +28,70 @@
...
 (defn clear-agent-state
   "Clears the published entries from the aggregated state of the agent.
-   Since we expect instances to have the same sandbox, we do not check for value equality."
-  [task-id->sandbox published-task-id->sandbox]
-  (let [task-id->sandbox' (apply dissoc task-id->sandbox (keys published-task-id->sandbox))]
-    (counters/clear! sandbox-aggregator-pending-count)
-    (counters/inc! sandbox-aggregator-pending-count (count task-id->sandbox'))
-    task-id->sandbox'))
+   Since we expect instances to have the same field value, we do not check for value equality."
+  [task-id->field-value field-name published-task-id->field-value]
```
pschorf: What about just `value` instead of `field-value` for both arguments?

shamsimam: Renamed `field-value` to `value`.

PR #931, opened 2018-08-04
Adds exit-code-syncer to handle a high rate of incoming exit-code messages in a graceful manner

shamsimam:

- adds exit-code-syncer to handle a high rate of incoming exit-code messages in a graceful manner

We would like to aggregate sending exit code datomic updates for tasks to avoid potential load issues based on our previous experience with the sandbox syncer.

### DIFF
--- a/scheduler/src/cook/mesos/sandbox.clj
+++ b/scheduler/src/cook/mesos/sandbox.clj
@@ -333,3 +333,35 @@
     (assoc publisher-state
       :publisher-cancel-fn publisher-cancel-fn
       :syncer-cancel-fn syncer-cancel-fn)))
+
+(defn aggregate-exit-code
+  "Sends a message to the agent to update the exit-code information."
+  [{:keys [task-id->exit-code-agent]} task-id exit-code]
+  (when task-id->exit-code-agent
+    (send task-id->exit-code-agent aggregate-instance-field "exit-code" task-id exit-code)))
+
+(defn start-exit-code-publisher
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -333,3 +333,35 @@
...
+                      (log/error ex "Instance exit-code directory publish failed"))}))
+
+(defn prepare-exit-code-publisher
+  "This function initializes the exit-code publisher as well as helper function to send individual
+   exit-code entries.
+   It returns a map with the following entries:
+   :publisher-cancel-fn - fn that take no arguments and that terminates the publisher.
+   :task-id->exit-code-agent - The agent that manages the task-id->exit-code aggregation and publishing."
+  [datomic-conn publish-batch-size publish-interval-ms]
+  (let [task-id->exit-code-agent (agent {}) ;; stores all the pending task-id->exit-code state
```
pschorf: Should we use an agent or some kind of backpressure method here?

shamsimam: None of our agent processes currently support backpressure as the task completion rates are not that high. I would vote for supporting backpressure in all our relevant agents as a separate PR.

shamsimam: @pschorf are you good with this?

pschorf: Yeah, that's fine with me.

shamsimam:
Rebased.

shamsimam:
Build is green now :)

PR #947, opened 2018-08-30
Adds more logging to test_unique_host_constraint

dposada:

- adding a log with the number of total, running, and waiting jobs

To make failures easier to track down.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1649,7 +1649,7 @@ class CookTest(util.CookTest):
                     # All of the jobs should be running
                     return num_running == num_jobs_total
 
-            jobs = util.wait_until(query, num_running_predicate, max_wait_ms=60000)
+            jobs = util.wait_until(query, num_running_predicate)
             hosts = [job['instances'][0]['hostname'] for job in jobs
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -1640,14 +1640,16 @@ def num_running_predicate(response):
...
+                                 f'and {num_waiting} waiting job(s)')
                 if num_jobs_total == num_hosts + 1:
                     # One job should not be scheduled
                     return (num_running == num_jobs_total - 1) and (num_waiting == 1)
                 else:
                     # All of the jobs should be running
                     return num_running == num_jobs_total
 
-            jobs = util.wait_until(query, num_running_predicate, max_wait_ms=60000)
+            jobs = util.wait_until(query, num_running_predicate)
```
pschorf: Why did the timeout change?

dposada: 60 seconds isn't enough in some of our environments, and if you use the default, you can change it with an environment variable, which is nice

pschorf: 👍

pschorf:
Build is red

PR #949, opened 2018-08-31
Fixes pool bug in task reconciliation

dposada:

- seeding running jobs in datomic before starting cook scheduler
- adding `test_reconciliation`, which tests that the seeded jobs get killed by the reconciler
- fixing the task reconciler to grab fenzo from the pool-to-fenzo map

This bug was introduced when we added pool scheduling. The test should help prevent future task reconciliation bugs from going unnoticed.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -9,10 +9,12 @@ import time
 import unittest
 import uuid
 from collections import Counter
+from datetime import datetime
 from urllib.parse import urlparse
 
 import dateutil.parser
 import pytest
+from pytz import utc
 from retrying import retry
 
 from tests.cook import reasons, mesos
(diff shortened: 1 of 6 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2473,3 +2474,30 @@ def test_data_local_support(self):
...
+        up, so that when it does start up and the reconciler runs, it can reconcile those tasks
+        (i.e. discover that they are not actually running in Mesos and mark them as failed). See
+        scheduler/datomic/data/seed_running_jobs.clj for the details of this job-seeding.
+        """
+        info = util.scheduler_info(self.cook_url)
+        cook_start_dt = dateutil.parser.parse(info['start-time'])
+        cook_start_ms = (cook_start_dt - datetime.fromtimestamp(0, tz=utc)).total_seconds() * 1000
+        start = int(cook_start_ms - 240000)
+        end = util.current_milli_time()
+        resp = util.jobs(self.cook_url, user='seed_running_jobs_user',
```
pschorf: Instead of querying for the jobs, what if we put the UUIDs in an env. variable (which we could use to trigger running the test.) I doubt this would be flaky, but it lets you avoid the weird timestamp math.

dposada: Yeah, I'm not super happy with the timestamp math, but I think the environment variable approach would make the workflow when running locally be not ideal. You'd have to have the environment variable defined both when you seed the jobs and when you're running the test (which might be from an IDE).

PR #953, opened 2018-09-05
Remove -XX:MaxPermSize, obsolete in JDK8 and above.

scrosby:

- Remove this option
- 
- 

- This option is obsolete in JDK8 and above. It suppresses an error message that occurs on many lein commands.

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -208,7 +208,6 @@
 
   :main cook.components
   :jvm-opts ["-Dpython.cachedir.skip=true"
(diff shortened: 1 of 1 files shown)


scrosby:
@dposada Last time I wanted to do this, it impacted your workflow. Is your workflow now JDK8-based?

dposada:
If you're able to run with `scheduler/bin/build-docker-image.sh && scheduler/bin/run-docker.sh`, then I should be good

scrosby:
That works fine; I've been doing it for integration tests with no problem.

PR #954, opened 2018-09-05
Docker file build changes for integration tests.

scrosby:

- Slight refactor in how we build docker images. 
- Avoids the pip install + requisite CPU time. 

Avoids 200mb of downloads + 200 seconds to generate docker images for
the docker-based integration test workflow.

A docker image build and integration test run in it goes from
350 to 150 seconds.

### DIFF
--- a/integration/Dockerfile
+++ b/integration/Dockerfile
@@ -1,8 +1,11 @@
 FROM python:3.6
 
+WORKDIR /opt/cook/integration
+COPY requirements.txt /opt/cook/integration
 ADD cli.tar.gz /opt/cook/cli/
+RUN pip install -r requirements.txt
 COPY . /opt/cook/integration
 WORKDIR /opt/cook/integration
-RUN pip install -r requirements.txt
+
 
 ENTRYPOINT ["pytest"]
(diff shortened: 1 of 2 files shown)


dposada:
@scrosby Out of curiosity, why are you using this? I was under the impression this was only needed on macos

integration/Dockerfile
```diff
@@ -1,8 +1,11 @@
 FROM python:3.6
 
+WORKDIR /opt/cook/integration
+COPY requirements.txt /opt/cook/integration
```
pschorf: Why did this move? Can we remove the other WORKDIR line?

integration/Dockerfile
```diff
@@ -1,8 +1,11 @@
...
 ADD cli.tar.gz /opt/cook/cli/
+RUN pip install -r requirements.txt
 COPY . /opt/cook/integration
 WORKDIR /opt/cook/integration
```
DaoWen: Since `WORKDIR` is now run as the first command, this duplicate line should be deleted.

integration/Dockerfile
DaoWen: This moved so that we have the latest copy of `requirements.txt` when running `pip` below.

The idea is that `requirements.txt` probably changes much less frequently than the rest of the test code in the directory, so moving this and the `pip` command up gives us more layers that can remain unchanged on a rebuild.

pschorf: Makes sense, but to be fair whenever you actually run the tests, they're mounted in a volume so there's no need to build the container unless the requirements.txt file changed.

pschorf:
Closing this PR due to inactivity, please reopen if needed.

PR #956, opened 2018-09-06
Fixes quota bug when not using pools

dposada:

- turning on `--auth=http-basic` for the travis run without pools
- fixing a bug in quota-checking when running without pools

The travis change allows multi-user tests (e.g. `test_preemption`) to run in the no-pools build. The bug went unnoticed because these tests were previously not running on that build.

### DIFF
--- a/scheduler/src/cook/mesos/scheduler.clj
+++ b/scheduler/src/cook/mesos/scheduler.clj
@@ -890,7 +890,7 @@
                                                             (cache/hit c slave-id)
                                                             (cache/miss c slave-id attrs)))))
                       _ (log/debug "In" pool "pool, passing following offers to handle-resource-offers!" offers)
-                      using-pools? (config/default-pool)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -890,7 +890,8 @@
                                                             (cache/hit c slave-id)
                                                             (cache/miss c slave-id attrs)))))
                       _ (log/debug "In" pool "pool, passing following offers to handle-resource-offers!" offers)
-                      user->quota (quota/create-user->quota-fn (d/db conn) (name pool))
+                      using-pools? (config/default-pool)
```
shamsimam: Nitpick: This name is confusing as `using-pools?` is not a boolean. Can we change it to `(not (nil? (config/default-pool)))`?

dposada:
@shamsimam Can we merge this?

PR #960, opened 2018-09-10
Removes assumption that cook-pool attribute is present

dposada:

- removing the assumption that the `cook-pool` attribute is present on Mesos agents

We don't want consuming tests to fail simply because agents don't have this attribute set.

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1212,8 +1212,7 @@ def hosts_to_consider(cook_url, mesos_url):
     state = get_mesos_state(mesos_url)
     slaves = state['slaves']
     pool = default_pool(cook_url)
-    slaves = [s for s in slaves if 'cook-pool' in s['attributes'] and s['attributes']['cook-pool'] == pool] \
-        if pool else slaves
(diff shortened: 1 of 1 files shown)


dposada:
@pschorf The travis build is actually green, it's just not updating here:

https://travis-ci.org/twosigma/Cook/builds/426823677?utm_source=github_status&utm_medium=notification

integration/tests/cook/util.py
```diff
@@ -1212,7 +1212,8 @@ def hosts_to_consider(cook_url, mesos_url):
     state = get_mesos_state(mesos_url)
     slaves = state['slaves']
     pool = default_pool(cook_url)
-    slaves = [s for s in slaves if s['attributes']['cook-pool'] == pool] if pool else slaves
```
pschorf: what about doing something like `s['attributes'].get('cook-pool', pool) == pool`? That would get you back on one line.

dposada: Good call, I changed it to:

```python
s['attributes'].get('cook-pool', None) == pool
```

because we don't want slaves that don't have the attribute to be considered.

dposada:

PR #961, opened 2018-09-10
Allows for overriding wait interval with environment variable

dposada:

- introducing `COOK_TEST_DEFAULT_WAIT_INTERVAL_MS`, which allows for overriding the default wait interval (i.e. time between attempts) used by most of the `wait_*` functions

In some environments, it's useful to allow the total timeout to be quite high (e.g. 8 minutes), in which case checking the condition once per second can cause the test logs to get way too huge. In these cases, it's also nice to allow for a longer wait between checks.


dposada:
@DaoWen Any feedback on this?

dposada:
@DaoWen Thoughts on this?

dposada:
@DaoWen bump

PR #962, opened 2018-09-10
Don't follow redirects on /queue admin self-impersonation test

DaoWen:

Don't follow the redirect on the `/queue` endpoint in the `test_self_impersonate` integration test.

The redirect was causing problems with Kerberose clients.


DaoWen:
@dposada - GitHub/Travis integration has seemed flaky since yesterday. This build is green. Not sure why it's showing as yellow above.

PR #970, opened 2018-09-11
Avoid sandbox checks if they shouldn't be expected

dposada:

- introducing `util.should_expect_sandbox_directory`, which tells you whether or not you should expect the sandbox directory to get populated
- calling the function in a couple of tests to determine whether or not assertions should be made

In environments where Mesos agents can be garbage collected quickly, and where we're not using the Cook executor, we may never get the sandbox directory populated.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -58,17 +58,18 @@ class CookTest(util.CookTest):
         self.assertEqual(False, job['disable_mea_culpa_retries'])
         self.assertTrue(len(util.wait_for_output_url(self.cook_url, job_uuid)['output_url']) > 0)
 
-        instance = util.wait_for_sandbox_directory(self.cook_url, job_uuid)
-        message = json.dumps(instance, sort_keys=True)
-        self.assertIsNotNone(instance['output_url'], message)
-        self.assertIsNotNone(instance['sandbox_directory'], message)
-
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -244,10 +245,11 @@ def test_failing_submit(self):
         self.assertEqual('failed', instance['status'], message)
```
shamsimam: Which `instance` is this line referring to?

dposada: This is a different test (`test_failing_submit`)

shamsimam: Oops.

integration/tests/cook/util.py
```diff
@@ -1226,3 +1226,19 @@ def num_hosts_to_consider(cook_url, mesos_url):
     num_hosts = len(hosts_to_consider(cook_url, mesos_url))
     logging.info(f'There are {num_hosts} hosts to consider')
     return num_hosts
+
+
+def should_expect_sandbox_directory(instance=None, job=None):
```
shamsimam: Can we have a separate helper function for job that invokes this function which only handles instances?

dposada: Good idea, done

dposada:
@shamsimam Can we merge this?

PR #975, opened 2018-09-14
Capture output from pytest when running integration tests

DaoWen:

- Use `name: ...` instead of `env: NAME=...` to name our Travis jobs.
- Capture integration tests' stdout/stderr in a log file when running in Travis.

The main motivation of this change is that we don't currently get the console output in our Travis log dumps. If we restart a failed job, we still have the rest of the dump, but we lose console, which makes debugging more difficult. Having the console output for the integration test runner improves the log dumps by making them more stand-alone.

And the `name` thing just looks nicer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,45 +22,45 @@ before_install:
 
 matrix:
   allow_failures:
-    - env: NAME='Cook Scheduler benchmark tests'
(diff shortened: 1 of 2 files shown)


dposada:
LGTM

DaoWen:
@dposada Merge?

PR #978, opened 2018-09-18
Use metatransaction for loading unscheduled_jobs

pschorf:
- Use metatransaction for loading jobs in a user's queue for unscheduled_jobs

Fixes a bug where users would see enqueued, uncommitted jobs in unscheduled_jobs.

### DIFF
--- a/scheduler/src/cook/mesos/unscheduled.clj
+++ b/scheduler/src/cook/mesos/unscheduled.clj
@@ -19,8 +19,7 @@
             [cook.mesos.quota :as quota]
             [cook.mesos.share :as share]
             [cook.mesos.util :as util]
-            [clojure.edn :as edn]
-            [metatransaction.core :as mt]))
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/unscheduled.clj
```diff
@@ -107,7 +108,7 @@
   "IFF the job is not first in the user's queue, returns
   [\"You have x other jobs ahead in the queue\", {:jobs [other job uuids]]}]"
   [conn job]
-  (let [db (d/db conn)
+  (let [db (mt/db conn)
```
shamsimam: Nitpick: Should we add a unit tests to verify this behavior change?

pschorf: Done

pschorf:
I felt dirty about adding metatransaction, so I switched to doing an explicit filter on commit latch.

PR #984, opened 2018-10-05
Adds pool to front-of-job-queue metrics

dposada:

- adding pool name to the `front-of-job-queue-mem` and `front-of-job-queue-cpus` metrics

This is another PR in a series to incorporate pool into metrics where it makes sense to separate by pool.


dposada:
This was taken care of in another PR

PR #985, opened 2018-10-08
Adds pool to several scheduler and monitor metrics

dposada:

- instead of statically defining the metrics, fetching them dynamically by title
- adding the pool name to the metric titles via `metric-title`

This is another PR for the effort to "poolify" the necessary metrics.


pschorf:
Moved to #989

PR #987, opened 2018-10-09
Move additional metrics to pool-specific metrics

pschorf:
- Moves additional scheduler metrics to be pool-specific

Provide more accurate metrics for each pool

### DIFF
--- a/scheduler/src/cook/mesos/dru.clj
+++ b/scheduler/src/cook/mesos/dru.clj
@@ -22,7 +22,9 @@
 (defrecord ScoredTask [task dru mem cpus])
 
 (defn metric-title [metric-name pool-name]
-  ["cook-mesos" "dru" metric-name (str "pool-" pool-name)])
+  (if pool-name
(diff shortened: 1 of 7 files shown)


scheduler/src/cook/mesos/dru.clj
```diff
@@ -99,20 +100,18 @@
 
 (defn sorted-task-cumulative-gpu-score-pairs
   "Takes a sorted seq of task entities and the gpu divisor, returns a list of [task cumulative-gpus], preserving the same order of input tasks"
-  [user->dru-divisors user->sorted-running-task-ents]
+  [user->dru-divisors pool-name user->sorted-running-task-ents]
```
shamsimam: Is `pool-name` used here?

pschorf: No, but this is the same "interface" as `sorted-task-scored-task-pairs` so I wanted to keep the arguments the same.

pschorf:
FYI, added another commit with some small fixes.

PR #988, opened 2018-10-09
Update JobClient dependency Apache HttpClient to v4.5.5

DaoWen:

Update JobClient dependency Apache HttpClient to v4.5.5.

A customer has a conflict with the old version, and asked us to upgrade.


DaoWen:
@pschorf - I think this is ready to merge.

PR #989, opened 2018-10-10
Pschorf poolify matched tasks metrics

pschorf:
- instead of statically defining the metrics, fetching them dynamically by title
- adding the pool name to the metric titles via `metric-title` 

This is another PR for the effort to "poolify" the necessary metrics.

Moved from PR #985


pschorf:
Done

PR #992, opened 2018-10-12
[WIP] Optimizer server integration

shamsimam:

- integrate optimizer output into the ranker
- update the simulator to retrieve results from the optimizer server
- include example code for the optimizer server and configuration

Adds support for the group (batch) scheduling optimizer server. The server can act as a 'slow' brain offering hints to the ranker on how to rank pending jobs.


scheduler/src/cook/mesos.clj
```diff
@@ -178,6 +178,7 @@
         datomic-report-chan (async/chan (async/sliding-buffer 4096))
         mesos-heartbeat-chan (async/chan (async/buffer 4096))
         current-driver (atom nil)
+        pool-name->optimizer-schedule-job-ids-atom (atom {})
```
DaoWen: This name is very confusing. We should try to come up with something simpler.
It looks like it's a mapping from pool names onto atoms.
Are the mapped-to elements schedules, or just a collection of job ids, or something else?

scheduler/src/cook/mesos.clj
```diff
@@ -233,16 +235,17 @@
...
+                                      (optimizer/start-optimizer-cycles!
+                                        optimizer-config
+                                        optimizer-trigger-chan
+                                        (fn get-pool-names []
+                                          (keys @pool-name->pending-jobs-atom))
+                                        (fn pool-name->queue [pool-name]
+                                          (get @pool-name->pending-jobs-atom pool-name))
+                                        (fn pool-name->running [pool-name]
+                                          (->> (util/get-running-task-ents (d/db mesos-datomic-conn))
+                                               (filter #(= pool-name (-> % :job/_instance util/job->pool-name)))))
```
DaoWen: Would it make sense to move this filtering logic into the query? E.g., we could add an optional `:pool-name` argument to the `get-running-task-ents` function, when then gets optionally added to the query. 

If we decide to update `get-running-task-ents`, that should probably go into its own PR, and we could rebase this on top.

scheduler/src/cook/mesos/mesos_mock.clj
```diff
@@ -292,9 +292,10 @@
           resources' (subtract-resources available-resources requested-resources)
           tasks (map #(assoc % :launched-time (t/now)) tasks)
           task->runtime-ms (-> state :config :task->runtime-ms)
+          runtime-multiplier (get-in host [:attributes :runtime-multiplier] 1)
           new-time-task-id-pairs (map (fn [{:keys [task-id] :as task}]
-                                        [(-> (t/now) 
-                                             (t/plus (t/millis (task->runtime-ms task)))) 
+                                        [(t/plus (t/now)
+                                                 (-> task task->runtime-ms (* runtime-multiplier) t/millis))
```
DaoWen: Can we move `(t/plus (t/now))` to the end of the threading macro instead of having it outside?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
             [clj-http.client :as http]
             [clj-time.core :as t]
             [clj-time.coerce :as tc]
-            [clj-time.periodic :as periodic]
-            [clojure.core.async :as async]
             [clojure.tools.logging :as log]
-            [cook.util :refer [lazy-load-var PosNum PosInt NonNegInt]]
+            [cook.util :refer [lazy-load-var NonNegInt PosNum PosInt]]
```
DaoWen: 🙄

You didn't move `lazy-load-var` to the end of the list???
`(sort ["lazy" "Non"])` → `("Non" "lazy")`

And you didn't move `cook.util` below `cook.mesos.util`??????

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+;; Elements missing:
+;;   1. Attributes (chip set, az, ..) -- have a combinatorial problem here
+;;   2. Price
+;;   3. Prob preemption in 10 minutes increments (or something like that)
 (def HostInfo
+  "Example:
+   {\"count\" 4
+    \"cpus\" 8
+    \"instance-type\" \"basic\"
+    \"mem\" 240000}"
```
DaoWen: This example use string keys, but the actual definition uses keyword keys.
Using keyword keys in the example would let us get rid of all those nasty backslashes.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+            [schema.core :as s])
+  (:import (java.util UUID)))
 
 (def TimePeriodMs NonNegInt)
 
+
+;; Elements missing:
+;;   1. Attributes (chip set, az, ..) -- have a combinatorial problem here
+;;   2. Price
+;;   3. Prob preemption in 10 minutes increments (or something like that)
```
DaoWen: Should I understand this comment?
Are you going to remove it before merging?
I feel like we need some extra context here, or at very least a `TODO` or `FIXME` to clarify whether this is documentation or some kind of todo notes.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+   (s/optional-key :gpus) PosNum
+   (s/optional-key :time-to-start) TimePeriodMs})
 
 (def Schedule
-  "Schedule returned by optimizer. Each TimePeriodMs is some milliseconds into the future the optimizer
-   recommendation map applies.
-   The optimizer recommendation map currently only has one key, :suggested-matches which has the matches
-   recommended as a value. The matches recommended are structured as a map from the host type in question
-   to a list of job UUIDs the optimizer recommends scheduling
+  "Schedule returned by optimizer.
```
DaoWen: This is a schema? Maybe we should be using [`defschema`](http://plumatic.github.io/schema/schema.core.html#var-defschema) to make that fact more explicit.

Same for the other schema definitions above.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
-    (get-available-host-info [this]
-      [])))
+  (produce-schedule
+    [this queue running]
+    "Returns a schedule of what jobs to place on what machines
+     and what hosts to purchase at different time steps.
+     Conforms to the Schedule schema above
+
+     Parameters:
+     queue -- Ordered list of jobs to run (see cook.mesos.schema job)
```
DaoWen: Is this actually expected to be a list? Or is any sequence (e.g., a vector) ok?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+     running -- Set of tasks running (see cook.mesos.schema instance)
```
DaoWen: Again, is this actually expected to be a set, or is a vector accepted?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+  [instance now]
+  (let [start (:instance/start-time instance)
+        end (:instance/end-time instance now)]
+    (- (.getTime end)
+       (.getTime start))))
+
+(defn calculate-expected-complete-time
+  "Returns the expected time remaining to job completion.
+   Simple model for if run time exceeds expected runtime: the more it exceeds expected, the longer we expect it to take.
+   TODO: get better model!"
```
DaoWen: Is this actually a TODO that you want to complete before merging this PR? If not, maybe it's better to open an issue and reference it from here.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+  [instance expected-duration now]
+  (->> (instance-runtime instance now)
+       (- expected-duration)
+       Math/abs))
```
DaoWen: Why would we want `Math/abs` here? It seems like `(max 0)` would make more sense; i.e., if the time remaining is negative, then it's already done, so no time is left.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+  "Converts a waiting job to an optimizer job representation."
+  [job default-runtime]
+  (let [expected-duration (get job :job/expected-runtime default-runtime)
+        resources (util/job-ent->resources job)
+        batch-uuid (job->batch-uuid job)]
+    (merge {:batch batch-uuid
+            :count 1
+            :dur expected-duration
+            :expected_complete_time -1
+            :running_host_group -1 ; TODO write function to determine this
```
DaoWen: Same note about TODOs: If this comment will stay in the patch, then we should have a corresponding issue and reference it here.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+(defn calculate-expected-complete-time
+  "Returns the expected time remaining to job completion.
+   Simple model for if run time exceeds expected runtime: the more it exceeds expected, the longer we expect it to take.
+   TODO: get better model!"
+  [instance expected-duration now]
+  (->> (instance-runtime instance now)
+       (- expected-duration)
+       Math/abs))
+
+(defn job->batch-uuid
```
DaoWen: Why use `batch` rather than `group`?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+            :running_host_group (-> instance :instance/hostname host-name->host-group)
+            :state "running"
+            :submit_time (-> job :job/submit-time .getTime)
+            :user (:job/user job)
+            :uuid (:job/uuid job)}
+           resources)))
+
+(defn create-optimizer-server
+  [{:keys [cpu-share default-runtime host-info max-waiting mem-share opt-params opt-server step-list]
+    :or {opt-params {}}
```
DaoWen: Do we need a docstring here? E.g., documentation on the `opt-params` might be useful.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+    (reify Optimizer
+      (produce-schedule [_ queued-jobs running-tasks]
+        (let [waiting-jobs (take max-waiting queued-jobs)]
+          (log/info "Optimizer inputs" {:running-tasks (count running-tasks)
+                                        :waiting-jobs (count waiting-jobs)})
+          (if (-> waiting-jobs count pos?)
+            (let [alpha 0.5
+                  now (tc/to-date (t/now))
+                  _ (log/info "Converting waiting jobs to optimizer jobs")
+                  waiting-opt-jobs (timers/time!
```
DaoWen: It's not clear why we want to time this. Shouldn't the `waiting-job->optimizer-job` operation be fast?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+                                         (group-by :user)
+                                         (pc/map-vals #(reduce + 0 (map :mem %)))
+                                         (pc/map-vals #(* (- 1 alpha) %))))
+                  _ (swap! user->ema-mem-usage
+                           (fn [ema-usage]
+                             (->> (pc/map-vals (partial * alpha) ema-usage)
+                                  (merge-with + user->mem-usage))))
+                  ;; Running must be before waiting here because optimizer determines batch order from job order
+                  opt-jobs (concat running-opt-jobs waiting-opt-jobs)
+                  ;; TODO mem-share should be computed per user.
```
DaoWen: Same note as other TODOs.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+                                    (->> running-opt-jobs
+                                         (group-by :user)
+                                         (pc/map-vals #(reduce + 0 (map :mem %)))
+                                         (pc/map-vals #(* (- 1 alpha) %))))
+                  _ (swap! user->ema-mem-usage
+                           (fn [ema-usage]
+                             (->> (pc/map-vals (partial * alpha) ema-usage)
+                                  (merge-with + user->mem-usage))))
+                  ;; Running must be before waiting here because optimizer determines batch order from job order
+                  opt-jobs (concat running-opt-jobs waiting-opt-jobs)
```
DaoWen: Does this also imply that the intra-sequence order in `running-opt-jobs` and `waiting-opt-jobs` affects the results? Are those sorted in some meaningful way?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+         (pos? default-runtime)
+         (s/validate HostInfo host-info)
+         (integer? max-waiting)
+         (pos? max-waiting)
+         (seq step-list)]}
+  (log/info "Optimizer config" config)
+  (let [host-group 0
+        host-name->host-group (constantly host-group)
+        host-infos [host-info]
+        user->ema-mem-usage (atom {})]
```
DaoWen: I think we need a comment somewhere saying what "ema" stands for. Maybe here, maybe elsewhere.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,251 @@
...
+                  request-body (timers/time!
+                                 (timers/timer ["cook-mesos" "scheduler" "optimizer" "prepare-request"])
+                                 (cheshire/generate-string
+                                   {"cpu_share" cpu-share
+                                    "host_groups" host-infos
+                                    "mem_share" mem-share
+                                    "now" (.getTime now)
+                                    "opt_jobs" opt-jobs
+                                    "opt_params" opt-params
+                                    "seed" (.getTime now)
```
DaoWen: Why seed with the current time?
Why do we even need a seed?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -634,12 +633,31 @@
...
-  (->> pending-jobs
-       (filter-based-on-quota user->quota user->usage)
-       (filter (fn [job] (util/job-allowed-to-start? db job)))
-       (take num-considerable)))
+  (let [optimizer-schedule-job-ids-promise (promise)]
+    (swap! pool-name->optimizer-schedule-job-ids-atom
+           (fn [pool-name->optimizer-schedule-job-ids]
+             (->> (pool-name->optimizer-schedule-job-ids pool-name)
+                  (deliver optimizer-schedule-job-ids-promise))
+             ;; TODO should we be clearing out optimizer data in one use?
```
DaoWen: Same note as other TODOs.

scheduler/src/cook/mesos/optimizer.clj
shamsimam: Yes, their orders affect the results. Waiting jobs are sorted by the ranker. Running jobs are not sorted as we expect all running batches to fit in resource constraints and be looked into by the optimizer.

scheduler/src/cook/mesos/optimizer.clj
shamsimam: It allows us to mimic results if the optimizer decides to use a random number generator.

scheduler/src/cook/mesos.clj
DaoWen: Why did we favor filtering outside the query?

scheduler/src/cook/mesos/optimizer.clj
DaoWen: Wouldn't using a constant value as the seed make it easier to mimic the results?

scheduler/src/cook/mesos.clj
shamsimam: I spoke with @pschorf and @dposada and they are okay with moving the pool inside the query. I created #1002

scheduler/src/cook/mesos/optimizer.clj
shamsimam: I was referring to being able to control the randomness in the optimizer from the scheduler. The optimizer prints the seed in its logs and if we decide to run the optimizer independently we can by using the seed.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -634,12 +633,31 @@
...
-       (take num-considerable)))
+  (let [optimizer-schedule-job-ids-promise (promise)]
+    (swap! pool-name->optimizer-suggested-job-ids-atom
+           (fn [pool-name->optimizer-schedule-job-ids]
+             (->> (pool-name->optimizer-schedule-job-ids pool-name)
+                  (deliver optimizer-schedule-job-ids-promise))
+             ;; TODO determine whether we should be clearing out optimizer data after one use?
+             (assoc pool-name->optimizer-schedule-job-ids pool-name [])))
+    (->> (if-let [optimizer-schedule-job-ids (seq @optimizer-schedule-job-ids-promise)]
+           (let [optimizer-schedule-job-ids-set (into #{} optimizer-schedule-job-ids)
```
DaoWen: Nitpick: `(set xs)` vs `(into #{} xs)`

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -634,12 +633,31 @@
...
   "Limit the pending jobs to considerable jobs based on usage and quota.
    Further limit the considerable jobs to a maximum of num-considerable jobs."
-  [db pending-jobs user->quota user->usage num-considerable pool-name]
+  [db pending-jobs user->quota user->usage num-considerable pool-name->optimizer-suggested-job-ids-atom pool-name]
   (log/debug "In" pool-name "pool, there are" (count pending-jobs) "pending jobs:" pending-jobs)
-  (->> pending-jobs
-       (filter-based-on-quota user->quota user->usage)
-       (filter (fn [job] (util/job-allowed-to-start? db job)))
-       (take num-considerable)))
+  (let [optimizer-schedule-job-ids-promise (promise)]
```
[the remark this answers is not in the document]
DaoWen: Also, should this logic be dependent on whether or not the optimizer is enabled? Do we really want to create the atom, do the swap, etc if there's no optimizer?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -634,12 +633,31 @@
...
+    (swap! pool-name->optimizer-suggested-job-ids-atom
+           (fn [pool-name->optimizer-schedule-job-ids]
+             (->> (pool-name->optimizer-schedule-job-ids pool-name)
+                  (deliver optimizer-schedule-job-ids-promise))
+             ;; TODO determine whether we should be clearing out optimizer data after one use?
+             (assoc pool-name->optimizer-schedule-job-ids pool-name [])))
```
DaoWen: We should probably keep the data around for more than one cycle, but not indefinitely.

Since Fenzo isn't guaranteed to see enough resources to match all of the optimizer-recommended jobs, it seems like a terrible waste to pass them to Fenzo just once and then throw them away.

But conversely, the optimizer runs on a different order of magnitude of frequency from the fenzo-matching loop, so we really don't want a bunch of "old news" from the optimizer to stick around until it spits out a new recommended schedule.

We'll need to find a way to keep them around for a while, and then "expire" them or something. Maybe we can put a timestamp in the scheduler output, and ignore it if it's too old?

DaoWen: There's also the issue that we need to remove entries from the optimizer schedule once they get matched by Fenzo, otherwise they continue to take valuable slots from the `num-considerable` limit even after they're running.

scheduler/src/cook/mesos.clj
DaoWen: Works for me

scheduler/src/cook/mesos/optimizer.clj
DaoWen: I think this warrants a comment.

scheduler/src/cook/mesos/optimizer.clj
DaoWen: ok

scheduler/src/cook/mesos/optimizer.clj
DaoWen: ok

scheduler/src/cook/mesos/optimizer.clj
DaoWen: OK, recovering the seed from the logs works too.

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,255 @@
...
+        batch-uuid (job->batch-uuid job)]
+    (merge {:batch batch-uuid
+            :count 1
+            :dur expected-duration
+            :expected_complete_time -1
+            :running_host_group -1 ; we do not support multiple host groups currently
+            :state "waiting"
+            :submit_time (-> job :job/submit-time .getTime)
+            :user (:job/user job)
+            :uuid (:job/uuid job)}
```
dposada: Why are these keys named with underscores instead of dashes?

dposada: Several fields can be destructured from the function arg

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,255 @@
...
+           resources)))
+
+;; TODO Issue #485 remote optimizer should support cpu-share and mem-share computed per user.
+(defn create-optimizer-server
+  "Returns an instance of Optimizer which returns the schedule received from sending the request
+   to a remote server. The input configuration has the following keys:
+   cpu-share: the global cpu share of each user,
+   default-runtime: the default expected runtime of a job,
+   host-info: the homogeneous profile of the cluster we are targeting,
```
dposada: Is "cluster" the right word here?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,255 @@
...
+   max-waiting: the maximum number of waiting jobs to send to the optimizer,
+   mem-share: the global mem share of each user,
+   opt-params: a map with string keys that are passed to the optimizer as additional configuration parameters,
+   opt-server: the url for the remote optimizer server,
+   step-list: the list of future run times (s.g. starting at 0) we want the optimizer to compute for."
```
dposada: What is "s.g."?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,255 @@
...
+        host-group-descriptions [host-info]
+        ;; exponential moving average (ema) of memory usage based on running tasks of users
+        user->ema-mem-usage (atom {})]
+    (reify Optimizer
+      (produce-schedule [_ queued-jobs running-tasks]
+        (let [waiting-jobs (take max-waiting queued-jobs)]
+          (log/info "Optimizer inputs" {:running-tasks (count running-tasks)
+                                        :waiting-jobs (count waiting-jobs)})
+          (if (-> waiting-jobs count pos?)
+            (let [alpha 0.5
```
dposada: What is alpha?

scheduler/src/cook/mesos/optimizer.clj
```diff
@@ -19,122 +19,255 @@
...
+                  _ (log/info "Received optimizer schedule")
+                  scheduled-job-ids (->> (get-in raw-schedule ["schedule" "0" "suggested-matches" (str host-group)] [])
+                                         (map #(UUID/fromString %)))]
+              (log/info "Num jobs scheduled immediately:" (count scheduled-job-ids))
+              {0 {:suggested-matches scheduled-job-ids}})
+            {0 {:suggested-matches []}}))))))
+
+(defn pool-optimizer-cycle
   "Starts a cycle that:
    1. Gets queue, running, offer and purchasable host info
```
dposada: Why "purchasable"?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -207,7 +206,7 @@
 (defn handle-status-update
   "Takes a status update from mesos."
   [conn driver pool->fenzo sync-agent-sandboxes-fn status]
-  (log/info "Mesos status is:" status)
+  (log/debug "Mesos status is:" status)
```
dposada: Separate PR for this?

scheduler/test-resources/optimizer/server/requirements.txt
```diff
@@ -0,0 +1 @@
+Flask==1.0.2
```
dposada: Shouldn't "flask" be lowercase?

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1328,70 +1327,102 @@
...
             num-considerable 1]
         (is (= [job-1]
                (sched/pending-jobs->considerable-jobs
-                 (d/db conn) non-gpu-jobs user->quota user->usage num-considerable nil)))
+                 (d/db conn) non-gpu-jobs user->quota user->usage num-considerable (atom {}) nil)))
         (is (= [job-5]
                (sched/pending-jobs->considerable-jobs
-                 (d/db conn) gpu-jobs user->quota user->usage num-considerable nil)))))
+                 (d/db conn) gpu-jobs user->quota user->usage num-considerable (atom {}) nil)))))
+    (testing "jobs inside usage quota influenced by optimizer"
```
dposada: I would favor a newline before the `testing` clause

scheduler/test/cook/test/mesos/scheduler.clj
```diff
@@ -1328,70 +1327,102 @@
...
+               (sched/pending-jobs->considerable-jobs
+                 (d/db conn) non-gpu-jobs user->quota user->usage num-considerable
+                 pool-name->optimizer-suggested-job-ids-atom "cpu-pool")))
+        (is (= [job-6 job-5]
+               (sched/pending-jobs->considerable-jobs
+                 (d/db conn) gpu-jobs user->quota user->usage num-considerable
+                 pool-name->optimizer-suggested-job-ids-atom "gpu-pool")))
+        (is (= {"cpu-pool" [] "gpu-pool" []}
+               @pool-name->optimizer-suggested-job-ids-atom))))
+    (testing "jobs inside usage quota influenced by optimizer limited by num-considerable of 1"
```
dposada: I would favor a newline before the `testing` clause

scheduler/test/cook/test/simulator.clj
```diff
@@ -560,15 +592,16 @@
                      keywordize-keys
                      ;; This is needed because we want the roles to be strings
                      (transform [ALL :resources MAP-VALS MAP-KEYS] name))
+          _ (println "config file:" config-file)
```
dposada: Should this use `log` instead of `println`?

scheduler/test/cook/test/simulator.clj
```diff
@@ -671,37 +704,142 @@
...
+    (nth (keys m) (count (take-while #(<= % r) w)))))
+
+(defn create-jobs []
+  (let [num-hosts 20
+        num-cpus-per-host 16
+        num-mem-per-host 131072
+        oversubscribe-factor 1.75
+        duration-ms (* 1000 60 60 5)
+        resources (* oversubscribe-factor num-hosts num-cpus-per-host num-mem-per-host duration-ms)
+        _ (println "resources:" resources)
```
dposada: Should this use `log` instead of `println`?

scheduler/test/cook/test/simulator.clj
```diff
@@ -671,37 +704,142 @@
...
+                         json/read-str
+                         first
+                         keywordize-keys)
+        job-id-atom (atom 0)
+        submit-time-atom (atom 0)
+        max-group-size 100
+        group-size-weights (map-from-keys
+                             (fn [k] (-> max-group-size (Math/pow 2) (/ k) (int)))
+                             (map inc (range max-group-size)))
+        _ (println "group-size-weights:" group-size-weights)
```
dposada: Should this use `log` instead of `println`?

scheduler/test/cook/test/simulator.clj
```diff
@@ -671,37 +704,142 @@
...
+                   (recur (reduce conj! jobs group-jobs)
+                          (- resources-left resources-consumed)))))]
+    (println "created" (count jobs) "jobs.")
+    (-> (json/write-str jobs :key-fn #(-> % str (subs 1)))
+        (str/replace "\\/" "/")
+        (->> (spit (str "resources/in-jobs-4K.json"))))))
+
+(comment create-jobs-and-hosts
+  (create-hosts)
+  (create-jobs))
```
dposada: Can we delete the `comment`?

scheduler/src/cook/mesos/optimizer.clj
shamsimam: Done.

scheduler/test/cook/test/mesos/scheduler.clj
shamsimam: Done.

scheduler/test/cook/test/mesos/scheduler.clj
shamsimam: Done.

scheduler/test/cook/test/simulator.clj
shamsimam: I have the comment in there as I do not want the functions to be marked as unused by the IDE (Curisve).

DaoWen: You should definitely add a meta-comment to explain this comment.

PR #994, opened 2018-10-17
Use x-cook-pool header when no pool param is provided

pschorf:
- Supports specifying the header `X-Cook-Pool` to provide the pool parameter on pool-specific requests.

This will make it easier to migrate clusters onto the pool model, by replacing a cluster with a proxy which forwards requests to a different cluster with the `X-Cook-Pool` header set.

### DIFF
--- a/scheduler/config.edn
+++ b/scheduler/config.edn
@@ -5,7 +5,8 @@
                  :one-user #config/env "COOK_ONE_USER_AUTH"}
  :authorization-config {;; These users have admin privileges when using configfile-admins-auth;
                         ;; e.g., they can view and modify other users' jobs.
-                        :admins #{"admin" "root" #config/env "USER"}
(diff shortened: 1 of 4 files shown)


scheduler/config.edn
```diff
@@ -5,7 +5,7 @@
                  :one-user #config/env "COOK_ONE_USER_AUTH"}
  :authorization-config {;; These users have admin privileges when using configfile-admins-auth;
                         ;; e.g., they can view and modify other users' jobs.
-                        :admins #{"admin" "root"}
+                        :admins #{"admin" "root" #config/env "USER"}
```
dposada: Why was this needed?

dposada: Ah, makes sense. This PR is fine, but can we add a comment explaining it?

PR #998, opened 2018-10-23
Use total usage instead of default pool usage at top level

pschorf:
- Return the total usage instead of the default pool usage at the top level in the `/usage` response

It's confusing for the top level `"total_usage"` to not reflect the total usage.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2043,39 +2043,6 @@ class CookTest(util.CookTest):
         finally:
             util.kill_jobs(self.cook_url, job_uuids)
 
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2043,6 +2043,39 @@ def test_user_usage_ungrouped(self):
         finally:
             util.kill_jobs(self.cook_url, job_uuids)
 
+    @unittest.skipUnless(util.are_pools_enabled(), "Requires pools")
+    @pytest.mark.serial
```
dposada: Can we do this with a different user so we don't have to mark it serial?

pschorf: Done

PR #1008, opened 2018-10-31
Fix options to Apache HttpClient SPNego lib to match old behavior

DaoWen:

- Add a static `build` method to `BasicSPNegoSchemeFactory`, which lets us handle all of the `SPNegoSchemeFactory` construction logic for the JobClient in one place.
- Define and document a new static constant flag `USE_CANONICAL_HOSTNAME`.
- Pass the new `USE_CANONICAL_HOSTNAME` into all `SPNegoScheme`-related constructor invocations.

Also fixed a typo: Cre<del>n</del>dential

The new default hostname canonicalization in the Apache HttpClient can cause problems with krb5, so we're disabling it and letting krb5 handle its own hostname conversions.

## Other resources

[Apache HttpClient `SPNegoSchemeFactory` JavaDoc](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/auth/SPNegoSchemeFactory.html)

### DIFF
--- a/jobclient/src/main/java/com/twosigma/cook/jobclient/JobClient.java
+++ b/jobclient/src/main/java/com/twosigma/cook/jobclient/JobClient.java
@@ -199,7 +199,7 @@ public class JobClient implements Closeable, JobClientInterface {
             return setKerberosAuth(null);
         }
 
-        public Builder setKerberosAuth(GSSCredentialProvider gssCrendentialProvider) {
+        public Builder setKerberosAuth(GSSCredentialProvider gssCredentialProvider) {
             Credentials creds = new Credentials() {
                 @Override
                 public String getPassword() {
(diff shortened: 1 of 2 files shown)


WenboZhao:
This change itself alone is good to me. However, it won't work when we import it internally. This is because the change from 4.3.6 to 4.5.5

http://static.javadoc.io/org.apache.httpcomponents/httpclient/4.3.6/org/apache/http/impl/auth/SPNegoSchemeFactory.html
vs
http://static.javadoc.io/org.apache.httpcomponents/httpclient/4.5.5/org/apache/http/impl/auth/SPNegoSchemeFactory.html

DaoWen:
Closing this in favor of #1013.

PR #1011, opened 2018-11-06
Allows ssh-checking instead of sudo-checking for multi-user tests

dposada:

- adding `ssh`-checking code
- adding `COOK_SWITCH_USER_MODE`, which determines which check to perform for multi-user tests

In order to more easily support automated integration test runs, we want to allow an alternative ssh-based approach for the multi-user tests.


pschorf:
Do we need want to provide an alternative command to validate that they can ssh as the given user instead of disabling it altogether?

dposada:
@pschorf Good call, done

dposada:
@DaoWen This is now ready for your review

PR #1012, opened 2018-11-07
removes metatransaction.core and uses datomic.api/db instead

shamsimam:

- removes metatransaction.core and uses datomic.api/db instead

Meta-transactions are not in use and cause a performance penalty.


scrosby:
We might want to stage this change over a few releases. I looked into this change about 4 months ago.

The concern I have is we have a detritus of failed metatransaction jobs that grow in the database. Removing the metatransaction filter exposes those formerly hidden jobs across various stacks where they're currently filtered out. Given that I saw almost two dozen stacks that would be exposed to the formerly hidden data, and I believed I should reason about each individually to determine if the change was safe, I figured the risk wasn't worth it at the time. The core paths were now fast enough, and we could remove it wholesale when cook-sim-v2 came out.

pschorf:
I agree with @scrosby , we can discuss this more in person tomorrow. Thanks for opening the can of worms, though :)

shamsimam:
I figured there was something stopping us making the change, wanted to have this out so we could be constantly reminded of it :)

scrosby:
See also https://github.com/twosigma/Cook/pull/800

pschorf:
Can we close this PR?

dposada:
Should we delete the branch?

PR #1014, opened 2018-11-08
Makes test_group_kill_simple more reliable

dposada:

- allowing reason 99001 on the killed job's instance

We've seen this happen in the wild.


dposada:
@pschorf Can we merge this?

dposada:
@pschorf Can we merge?

PR #1015, opened 2018-11-08
Add pool info to usage command

nsinkov:

show usage broken out by pool

info like share and quota must be broken out by pool to be meaningful


integration/tests/cook/test_cli.py
```diff
@@ -1748,7 +1748,15 @@ def test_usage(self):
         self.assertEqual(0, cp.returncode, cp.stderr)
         self.assertEqual(3, len(uuids))
         uuid_7, uuid_8, uuid_9 = uuids
-        all_uuids = [uuid_1, uuid_2, uuid_3, uuid_4, uuid_5, uuid_6, uuid_7, uuid_8, uuid_9]
+
+        cp, uuids = cli.submit(command, self.cook_url, submit_flags='--cpus 0.2 --mem 32 --pool alpha')
```
dposada: We can't assume here that there's a pool named alpha. We want these tests to work on clusters that don't have pools too. Or that have pools with different names. Search for calls to `util.all_pools` or `util.active_pools` to see examples of existing tests that check whether or not there are pools in the target cluster.

scheduler/bin/run-docker.sh
```diff
@@ -134,9 +134,6 @@ docker create \
     --publish=${COOK_NREPL_PORT}:${COOK_NREPL_PORT} \
     --publish=${COOK_PORT}:${COOK_PORT} \
     --publish=${COOK_SSL_PORT}:${COOK_SSL_PORT} \
-    --publish=4334:4334 \
-    --publish=4335:4335 \
-    --publish=4336:4336 \
```
dposada: I would favor this in a separate PR

cli/cook/subcommands/usage.py
```diff
@@ -127,34 +152,40 @@ def format_percent(n):
...
-                for group, group_usage in application_usage['groups'].items():
-                    usage_map = group_usage['usage']
-                    jobs = group_usage['jobs']
-                    print_info(f'\t- {colors.bold(group if group else "[ungrouped]")}')
-                    print_info(f'\t  {format_usage(usage_map)}')
-                    print_info(f'\t  Jobs: {len(jobs)}')
-                    print_info('')
-            print_info('')
-
+                print_formatted_cluster_usage(cluster, cluster_usage)
```
dposada: I thought we were also going to add a `--pool` flag to the `usage` subcommand

integration/tests/cook/test_cli.py
```diff
@@ -1807,6 +1830,33 @@ def test_usage(self):
...
+                self.assertLessEqual(0, share['gpus'])
+                self.assertLessEqual(round(0.2 * 2, 1), round(cs_usage['cpus'], 1))
+                self.assertLessEqual(round(32 * 2, 1), round(cs_usage['mem'], 1))
+                self.assertLessEqual(0, cs_usage['gpus'])
+                self.assertLessEqual(round(0.2 * 2, 1), round(ungrouped_usage['cpus'], 1))
+                self.assertLessEqual(round(32 * 2, 1), round(ungrouped_usage['mem'], 1))
+                self.assertLessEqual(0, ungrouped_usage['gpus'])
+                self.assertIn(uuid_10, ungrouped_jobs)
+                self.assertIn(uuid_11, ungrouped_jobs)
+
```
dposada: I think we need more integration tests for this, e.g. for some of the cases I mentioned above. Some quick thoughts:

- `--pool` with nothing after
- `--pool` with invalid pool name
- `--pool` with pool name, but cluster has no pools
- `--pool` with pool name that is valid in multiple clusters, currently it shows matching pools in all clusters, is that what we want?
- more?

cli/cook/subcommands/usage.py
```diff
@@ -127,45 +168,56 @@ def format_percent(n):
...
     user = args.get('user')
+    pools = args.get('pools')
 
-    query_result = query(clusters, user)
+    query_result = query(clusters, user, pools)
     if as_json:
         print_as_json(query_result)
     else:
+        if pools and not query_result['clusters'].items():
+            print(f"No usage found for pools {pools}")
```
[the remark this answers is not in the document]
nsinkov: Foo could be a valid pool name on a different cluster

cli/cook/subcommands/usage.py
```diff
@@ -174,6 +226,7 @@ def register(add_parser, add_defaults):
     """Adds this sub-command's parser and returns the action function"""
     parser = add_parser('usage', help='show breakdown of usage by application and group')
     parser.add_argument('--user', '-u', help='show usage for a user')
+    parser.add_argument('--pools', '-p', nargs='*', help='limit results to one or more pools')
```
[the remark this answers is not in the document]
nsinkov: Why do you think that way is better?

cli/cook/subcommands/usage.py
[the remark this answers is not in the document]
nsinkov: How about a summary of which cluster/pool pairings were not shown at the bottom?

cli/cook/subcommands/usage.py
```diff
@@ -127,45 +168,56 @@ def format_percent(n):
...
-            print_info(colors.bold(cluster))
-            print_info(format_share(share_map))
-            print_info(format_usage(usage_map))
-            applications = cluster_usage['applications']
-            if applications:
-                print_info('Applications:')
+        if 'using_pools' in cluster_usage:
+            if cluster_usage['using_pools']:
+                for pool, pool_usage in cluster_usage['pools'].items():
+                    print_formatted_cluster_or_pool_usage(f'{cluster} ({pool} pool)', pool_usage)
```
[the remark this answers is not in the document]
nsinkov: How come?

integration/tests/cook/test_cli.py
nsinkov: We currently don't test any other flags or the non-json output, right?

cli/cook/subcommands/usage.py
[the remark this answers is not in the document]
dposada: I don't understand

cli/cook/subcommands/usage.py
```diff
@@ -127,53 +164,101 @@ def format_percent(n):
...
+
+    missing_pools = pools_set.difference(set(known_pools))
+    if missing_pools:
+        print_error(f"WARNING: " +
+                    (f"'{list(missing_pools)[0]}' is not a valid pool in "
+                     if len(missing_pools) == 1 else
+                     f"{list(missing_pools)} are not valid pools in ") +
+                    (f"cluster '{clusters[0]}'"
+                     if len(clusters) == 1 else
+                     f"clusters {clusters}"))
```
dposada: I think we should make this look more like the error message you get from `jobs` when you don't get any results:

`No matching running jobs for dpo found A / B / C / D / E.`

Specifically:

- remove "WARNING" since it already shows up in red
- remove "clusters"
- remove the brackets and single quotes around the clusters and separate with ` / `
- add a period at the end of the sentence

cli/cook/subcommands/usage.py
```diff
@@ -127,53 +164,101 @@ def format_percent(n):
...
         print_formatted(query_result)
+
     return 0
 
 
 def register(add_parser, add_defaults):
     """Adds this sub-command's parser and returns the action function"""
     parser = add_parser('usage', help='show breakdown of usage by application and group')
     parser.add_argument('--user', '-u', help='show usage for a user')
+    parser.add_argument('--pool', '-p', action='append', help='filter by pool (can be repeated)')
```
dposada: We should bump the version number for this change.

integration/tests/cook/test_cli.py
```diff
@@ -1748,7 +1748,28 @@ def test_usage(self):
...
+                    extra_pool = pool['name']
+                    cp, uuids = cli.submit(command, self.cook_url, submit_flags=f'--cpus 0.2 --mem 32 --pool {extra_pool}')
+                    self.assertEqual(0, cp.returncode, cp.stderr)
+                    uuid_10 = uuids[0]
+                    all_uuids.append(uuid_10)
+                    cp, uuids = cli.submit(command, self.cook_url, submit_flags=f'--cpus 0.2 --mem 32 --pool {extra_pool}')
+                    self.assertEqual(0, cp.returncode, cp.stderr)
+                    uuid_11 = uuids[0]
+                    all_uuids.append(uuid_11)
+                    break
```
dposada: Why `for` then `break`? Couldn't we just do `extra_pool = next(p for p in active_pools if p['name'] != default_pool)`

nsinkov:
poke

PR #1018, opened 2018-11-09
Makes test_hostname_equals_job_constraint work with pools

dposada:

- using `util.hosts_to_consider` to avoid using agents assigned to a pool different from the one we're submitting on
- adding some logs
- fixing the `kill_jobs` call (`'dict_values' object does not support indexing`)

So that we can start running this test again.


dposada:
@DaoWen have you had a chance to look at this?

PR #1022, opened 2018-11-14
Skips test_balanced_host_constraint_cannot_place when there are no agents to consider

dposada:

- changing the existing early `return` to a `pytest.skip` call
- adding a skip when there are 0 hosts to consider

In clusters that dynamically scale up agents in response to demand, this test can encounter the state where there are currently no running agents to consider. In this scenario, we'd rather simply skip the test.


dposada:
@pschorf Thoughts on this?

dposada:
@pschorf Have you had a chance to look at this?

PR #1024, opened 2018-11-15
Adds assertion message to load_resource

dposada:

- adding an assertion message for the status code == 200 assertion in `load_resource`

When this assertion fails, we want more information about the cause.


dposada:
@shamsimam Thoughts on this?

PR #1031, opened 2018-11-20
Outgoing/dockerchanges

scrosby:

- Slight refactor in how we build docker images. 
- Avoids the pip install + requisite CPU time. 

Avoids 200mb of downloads + 200 seconds to generate docker images for
the docker-based integration test workflow.

A docker image build and integration test run in it goes from
350 to 150 seconds.

See https://github.com/twosigma/Cook/pull/954 for the last round of this. I've made the requested review changes.

### DIFF
--- a/integration/Dockerfile
+++ b/integration/Dockerfile
@@ -5,7 +5,6 @@ COPY requirements.txt /opt/cook/integration
 ADD cli.tar.gz /opt/cook/cli/
 RUN pip install -r requirements.txt
 COPY . /opt/cook/integration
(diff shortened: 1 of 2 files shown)


pschorf:
I'll let @DaoWen review this.

DaoWen:
As previously discussed in #954, since we mount the integration tests as a volume when running locally, the rebuild overhead typically isn't much of an issue. But I don't see any harm in merging this, and maybe it'll help with some future workflow.

PR #1032, opened 2018-11-20
Change the data locality service port from 5000 to 35847

scrosby:

- Change the data locality service port from 5000 to 35847

5000 may be used by other services running locally. Use a different,
random port, e.g., 35847 instead to reduce the risk of conflicts.

(I had a conflict with a system program on my laptop that had to be uninstalled temporarily.)


dposada:
Maybe we should allow setting this with an environment variable

scrosby:
Not a big benefit. Most of the uses are inside of docker where they (presumably) run with a clean environment, or at the toplevel. I used environmental variables where feasible.

PR #1033, opened 2018-11-21
Precompile all sources in uberjar

DaoWen:

- Ahead-of-time compile *all* sources when building an uberjar, not just the sources explicitly references through the `cook.components` namespace.

It makes sense to eagerly compile all of our code (including optional plugins), rather than mixing run-time and build-time compilation.

## Other notes

We have a similar PR in the Waiter project: https://github.com/twosigma/waiter/pull/518

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -158,7 +158,7 @@
                    [twosigma/mesomatic "1.0.1-2018.08.28-df2c2ca"]]}
 
    :uberjar
-   {:aot [cook.components]
+   {:aot :all
     :dependencies [[com.datomic/datomic-free "0.9.5206"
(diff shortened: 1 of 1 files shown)


DaoWen:
Closing this as `:aot :all` isn't going to work without some changes.

Error from Travis build:
```
java.lang.RuntimeException: Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay@5811502e, compiling:(schema.clj:16:1)
Exception in thread "main" java.lang.RuntimeException: Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay@5811502e, compiling:(schema.clj:16:1)
```

PR #1035, opened 2018-11-27
Improve error message when cluster argument is not in config

pschorf:
- Improve error message when a user specifies a cluster not in their config

The current error message is confusing.


cli/cook/cli.py
```diff
@@ -47,6 +47,10 @@ def load_target_clusters(config_map, url=None, cluster=None):
     elif config_clusters:
         if cluster:
             clusters = [c for c in config_clusters if c.get('name').lower() == cluster.lower()]
+            if len(clusters) == 0 and len(config_clusters) > 0:
+                config_cluster_names = ', '.join([c.get('name') for c in config_clusters])
+                raise Exception(f'You specified cluster {cluster} which was not present in your config.' +
```
dposada: Should there be a comma before "which"?

```suggestion
                raise Exception(f'You specified cluster {cluster}, which was not present in your config.' +
```

cli/cook/cli.py
```diff
@@ -47,6 +47,10 @@ def load_target_clusters(config_map, url=None, cluster=None):
...
+                                f' You have the following clusters configured: {config_cluster_names}')
```
dposada: Missing period at the end of this sentence.

```suggestion
                                f' You have the following clusters configured: {config_cluster_names}.')
```

PR #1036, opened 2018-11-30
Add additional log message for unexpected response code

pschorf:
- Adds additional log message if we get a non 200 or 401 response to a data query

To improve debugging for requests that don't return any data.


cli/cook/http.py
```diff
@@ -103,6 +103,8 @@ def make_data_request(cluster, make_request_fn):
             return resp.json()
         elif resp.status_code == 401:
             print_error(f'Authentication failed on {cluster["name"]} ({cluster["url"]}).')
+        else:
+            logging.warn(f'Unexpexted response code {resp.status_code} for data request. Response body: {resp.text}')
```
dposada:
```suggestion
            logging.warn(f'Unexpected response code {resp.status_code} for data request. Response body: {resp.text}')
```

pschorf: Fixed

cli/cook/http.py
```diff
@@ -103,6 +103,13 @@ def make_data_request(cluster, make_request_fn):
             return resp.json()
         elif resp.status_code == 401:
             print_error(f'Authentication failed on {cluster["name"]} ({cluster["url"]}).')
+            return []
+        elif resp.status_code == 500:
+            print_error(f'Encountered server error while querying {cluser["name"]}')
```
dposada:
```suggestion
            print_error(f'Encountered server error while querying {cluser["name"]}.')
```

dposada:

PR #1039, opened 2018-11-30
Add additional logging on exceptions

pschorf:
- Wrap all of the default exception handlers with additional logging

Several of the exception types don't log at all, and the default logger doesn't log the request so it's hard to associate the log with the request that caused it.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -2711,6 +2711,11 @@
           (update :body body-matchers)
           (update :response resp-matchers)))))
 
+(defn- logging-exception-handler [base-handler]
+  (fn logging-exception-handler [ex data req]
+    (log/error ex "Error when processing request" (dissoc req ::c-mw/options))
+    (base-handler ex data req)))
+
 ;;
 ;; "main" - the entry point that routes to other handlers
 ;;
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -2711,6 +2711,11 @@
           (update :body body-matchers)
           (update :response resp-matchers)))))
 
+(defn- logging-exception-handler [base-handler]
```
shamsimam: Nitpick: missing docstring

scheduler/src/cook/mesos/api.clj
```diff
@@ -2711,6 +2711,11 @@
...
+  (fn logging-exception-handler [ex data req]
+    (log/error ex "Error when processing request" (dissoc req ::c-mw/options))
```
shamsimam: We should be more restrictive about what we print in the request. Waiter has the following:

```
(-> (dissoc request :body :ctrl :request-time :server-name :server-port :servlet-request :ssl-client-cert :support-info)
      (update-in [:headers] headers/truncate-header-values))
```

pschorf: Done.

scheduler/src/cook/mesos/api.clj
pschorf: Done

pschorf:
Example log:

`2018-11-30 10:40:31,605 ERROR cook.mesos.api [qtp155706761-92] - Error when processing request {:cookies {}, :remote-addr 0:0:0:0:0:0:0:1, :params {}, :headers {accept text/html,application/xhtml+xml,appl
ication/xml;q=0.9,*/*;q=0.8, upgrade-insecure-requests 1, user-agent Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, li..., connection keep-alive, host localhost:12321, accept-l
anguage en-us, accept-encoding gzip, deflate}, :server-port 12321, :content-length nil, :form-params {}, :authorization/user paul, :query-params {}, :content-type nil, :character-encoding nil, :uri /jobs/
8048cfe3-c65b-441f-90e2-7a824d153849, :server-name localhost, :query-string nil, :scheme :http, :request-method :get}
java.lang.Exception: this is a test`

PR #1042, opened 2018-11-30
Limit number of running and waiting jobs loaded by check-queue-position

pschorf:
- Load running job once per invocation of `reasons`
- Limit number of waiting jobs loaded in `check-queue-position`

Improve the performance of the `/unscheduled_jobs` endpoint


scheduler/src/cook/mesos/unscheduled.clj
```diff
@@ -14,13 +14,15 @@
 ;; limitations under the License.
 ;;
 (ns cook.mesos.unscheduled
-  (:require [datomic.api :as d :refer (q)]
```
shamsimam: Thank you!

scheduler/src/cook/mesos/unscheduled.clj
```diff
@@ -160,14 +174,17 @@
...
-                                    db job)
-               (check-exceeds-limit share/get-share
-                                    "The job would cause you to exceed resource shares."
-                                    db job)
-               (check-launch-rate-limit job)
-               (check-queue-position conn job)
-               (check-fenzo-placement conn job)]))))
+      (let [user (:job/user job)
+            pool-name (-> job :job/pool :pool/name)
+            running-jobs (get-jobs-by-user-and-state db user "running" 1000000 pool-name 30)]
```
shamsimam: Nitpick: should we consider sending the `queued-jobs` explicitly too to defend against a performance regression in the future if more of the other `check-*` functions start requiring the waiting jobs?

scheduler/src/cook/mesos/unscheduled.clj
pschorf: You're welcome :)

scheduler/src/cook/mesos/unscheduled.clj
pschorf: Done

PR #1043, opened 2018-12-03
Outgoing/rate limit unscheduled fix

scrosby:

- The gradual rate limit means we're never in debt, so the unscheduled endpoint won't ever report anything anymore. Fix it with new logic. We export when we're hitting a rate limit from the scheduler and expose it to the unscheduled.
- 
- 

The current rate limit unscheduled code is broken. Either remove or fix. This fixes it. :)

### DIFF
--- a/scheduler/src/cook/mesos/unscheduled.clj
+++ b/scheduler/src/cook/mesos/unscheduled.clj
@@ -22,6 +22,7 @@
             [cook.mesos.util :as util]
             [cook.rate-limit :as ratelimit]
             [clojure.edn :as edn]
+            [clojure.tools.logging :as log]
             [datomic.api :as d :refer (q)]))
 
 (defn check-exhausted-retries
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/mesos/unscheduled.clj
```diff
@@ -150,13 +151,15 @@
...
+                             (map #(get % user 0))
+                             (reduce + 0))
+        being-ratelimited? (not (zero? num-ratelimited))
         {:keys [tokens-replenished-per-minute]} ratelimit/job-launch-rate-limiter]
-    (when (and enforcing-job-launch-rate-limit? in-debt?)
-      ["You have exceeded the limit of jobs launched per minute."
-       {:max-jobs-per-minute tokens-replenished-per-minute
-        :seconds-until-can-launch (-> time-until-out-of-debt (/ 1000.0) Math/ceil int)}])))
+    (when (and enforcing-job-launch-rate-limit? being-ratelimited?)
+      ["You are currently limited by the maximum jobs per minute."
```
pschorf: "You are currently rate limited on how many jobs you launch per minute"?

scrosby: Changed.

PR #1044, opened 2018-12-03
Cache refactor and add explicit expiration.

scrosby:

- Refactor the guava cache code into its own toplevel namespace
- Add an explicit :cache-expires-at value to the value dictionary for explicit expiration semantics
- 

- This refactor makes sense.
- The new :cache-expires-at will be used in the future plugin PR.

### DIFF
--- a/scheduler/src/cook/cache.clj
+++ b/scheduler/src/cook/cache.clj
@@ -1,0 +1,52 @@
+(ns cook.cache
+  (:require [clj-time.core :as t])
+  (:import (com.google.common.cache Cache)))
+
+
+(defn lookup-cache!
+  "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+  any misses. Caches only positive hits where both functions return non-nil"
+  [^Cache cache extract-key-fn miss-fn item]
+  (if-let [key (extract-key-fn item)]
+    (if-let [result (.getIfPresent cache key)]
+      result ; we got a hit.
+      (let [new-result (miss-fn item)]
+        ; Only cache non-nil
+        (when new-result
+          (.put cache key new-result))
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -43,25 +44,12 @@
       (.expireAfterAccess 2 TimeUnit/HOURS)
```
pschorf: Should we also move `new-cache`?

scheduler/src/cook/cache.clj
```diff
@@ -0,0 +1,52 @@
...
+
+(defn lookup-cache-with-expiration!
+  "Generic cache. Caches under a key (extracted from the item with extract-key-fn. Uses miss-fn to fill
+  any misses. Caches only positive hits where both functions return non-nil. Also handles expiration if the value
+  is a map with the key :cache-expires-at."
+  [^Cache cache extract-key-fn miss-fn item]
+  (if-let [key (extract-key-fn item)]
+    (locking cache ; TOOD: Consider lock striping based on hashcode of the key to allow concurrent loads.
+      ; If it has a timed expiration, expire it.
+      (when-let [result (.getIfPresent cache key)]
```
pschorf: Can this just be a call to `expire-key!`?

scheduler/src/cook/mesos/util.clj
scrosby: I view cook.cache as 'too global' to put the factory method with a full config. None of the rest of the code uses any configuration and I could envision wanting that to be different in different use-cases.

scheduler/src/cook/cache.clj
scrosby: expire-key! removed

PR #1046, opened 2018-12-04
Removes blessed

dposada:

- switching from using the `blessed` library to simply rolling our own terminal colors, etc.

We are running into issues with `blessed` when using `pyinstaller`. Also, we simply don't use much of the functionality provided by `blessed`, so it's nice to get rid of the dependency.

### DIFF
--- a/cli/cook/progress.py
+++ b/cli/cook/progress.py
@@ -1,10 +1,7 @@
 import threading
 
-from blessed import Terminal
-
 from cook.util import print_info
 
-term = Terminal()
 data = []
 lock = threading.Lock()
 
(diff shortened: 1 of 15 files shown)


cli/cook/progress.py
```diff
@@ -15,7 +12,7 @@ def __print_state(lines_to_move_up):
     lines_to_move_up lines and then printing the current state of the data
     list, which contains [item, status] pairs.
     """
-    print_info(term.move_up * lines_to_move_up, end='')
+    print_info('\033[F' * lines_to_move_up, end='')
```
nsinkov: suggestion:
move magic number to colors.py and rename colors.py to terminal.py

dposada:
@nsinkov This is ready for another pass

PR #1049, opened 2018-12-10
Allows number of hosts to be set with an environment variable

dposada:

- introducing `COOK_TEST_BALANCED_CONSTRAINT_NUM_HOSTS`, which allows the user to set the number of hosts used in `test_balanced_host_constraint_can_place`
- adding logging in `util.__get` when an unexpected status code comes back

In environments which dynamically provision Mesos agents, expecting 30 hosts to come up within the test timeout is not always realistic. The environment variable allows users to tone down this expectation. The logging is just for easier troubleshooting.


dposada:

dposada:
@pschorf bump

PR #1052, opened 2018-12-12
Makes test_memory_limit_exceeded_cook_script more reliable

dposada:

- adding `-9` as an allowable exit code for the python script
- removing the `skipUnless` that keeps the test from running locally

We've seen the exit code `-9` happen in the wild. As for removing the `skipUnless`, this test passes reliably in my local environment, and I think it would be good for devs to be running it.


dposada:
@DaoWen Do you have any feedback on this?

PR #1053, opened 2018-12-12
Marks test_group_failed_only_change_retries_all_active as xfail

dposada:

- marking `test_group_failed_only_change_retries_all_active` as `xfail`

We've seen this test fail because one of the 5 jobs fails with `Invalid task`. Here's an example:

https://travis-ci.org/twosigma/Cook/jobs/467034340#L1465


dposada:
@DaoWen Do you have any feedback on this?

PR #1057, opened 2018-12-14
Improves memory limit tests

dposada:

- changing assumption of exactly 1 instance to at least 1 instance
- using the last instance instead of the first
- adding a log of the chosen instance
- removing `skipUnless(util.continuous_integration(), ...)` from 3 memory limit tests

There can be more than 1 instance, for example, due to a mea-culpa failure like `Invalid Mesos offer`. We want to assert on the last of these instances. As for removing the `skipUnless`, these tests run reliably in my local environment, and I think it's good for devs to be running these.


dposada:
@shamsimam Done

PR #1059, opened 2018-12-18
Add distributionManagement section to jobclient pom

pschorf:
- Add `distributionManagement` section to jobclient pom

Allows releasing on clojars with `mvn deploy`.


pschorf:
@shamsimam ?

PR #1063, opened 2019-01-04
Clean up DockerFile

scrosby:
We don't need to copy the integration test directory (and the virtualenv). We can also have docker cache the transactor untar.

- Some things to reduce docker generation overhead.
- 
-

### DIFF
--- a/scheduler/Dockerfile
+++ b/scheduler/Dockerfile
@@ -35,11 +35,12 @@ WORKDIR /opt/cook
 ## with fetched dependencies as long as project.clj isn't modified
 RUN lein deps
 
+# Datomic setup
+COPY datomic /opt/cook/datomic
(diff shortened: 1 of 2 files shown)


scheduler/Dockerfile
```diff
@@ -35,11 +35,12 @@ WORKDIR /opt/cook
 ## with fetched dependencies as long as project.clj isn't modified
 RUN lein deps
 
-# Copy the whole scheduler into the container
-COPY . /opt/cook/
-
 # Datomic setup
+COPY datomic /opt/cook/datomic
 RUN unzip -uo /opt/cook/datomic/datomic-free-0.9.5394.zip
```
DaoWen: If we converted `datomic-free-0.9.5394.zip` to `datomic-free-0.9.5394.tar.gz`, then we could take advantage of the `ADD` command's ability to auto-extract tarballs.

PR #1065, opened 2019-01-14
Adds plugin mechanism to CLI

dposada:

- adding `plugins` package, which supports config-based plugins
- adding a plugin hook in the `ls` subcommand for retrieving the entries
- adding an integration test for plugging in a dummy `ls` retrieval function

We want to support runtime extensions to the base `cs` subcommands so that environment-specific plugins can be easily configured. For example, some environments may want to specify a different mechanism for browsing files in the sandbox.

### DIFF
--- a/integration/tests/cook/test_cli.py
+++ b/integration/tests/cook/test_cli.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import tempfile
 
 import nbformat
 import pytest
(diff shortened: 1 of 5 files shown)


dposada:
Performance test: I created a copy of the checked-in `.cs.json`, named it `.cs.plugins.json`, and added the following dummy plugin:

```json
  "plugins": {
	"retrieve-ls-entries": {
		"module-name": "does_not_matter",
		"path": "/home/dpo/temp/file.py",
		"function-name": "bar"
	}
  }
```

The plugin code in `file.py` is simply:

```bash
$ cat ~/temp/file.py 
def bar(_, __, ___):
    return [
{'path': 'baz', 'nlink': 1, 'mode': 'drwxr-xr-x'},
{'path': 'qux', 'nlink': 1, 'mode': 'drwxr-xr-x'},
]
```

Then I ran the following:

```bash
$ for i in `seq 1 25`; do /usr/bin/time -f%e -ao ~/cs-ls-no-plugin.txt cs --config .cs.json ls a4bdea13-f45b-4c12-9af4-fd1ac2d1f17a ; done
$ for i in `seq 1 25`; do /usr/bin/time -f%e -ao ~/cs-ls-with-plugin.txt cs --config .cs.plugins.json ls a4bdea13-f45b-4c12-9af4-fd1ac2d1f17a ; done
$ python
Python 3.6.7 (default, Oct 21 2018, 04:56:05) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> data = pd.concat([pd.read_csv('~/cs-ls-with-plugin.txt', header=None).rename(columns={0:'with-plugin'}), pd.read_csv('~/cs-ls-no-plugin.txt', header=None).rename(columns={0:'without-plugin'})], axis=1)
>>> data.describe()
       with-plugin  without-plugin
count        25.00         25.0000
mean          0.18          0.1804
std           0.00          0.0020
min           0.18          0.1800
25%           0.18          0.1800
50%           0.18          0.1800
75%           0.18          0.1800
max           0.18          0.1900
```

integration/tests/cook/test_cli.py
```diff
@@ -1011,6 +1012,51 @@ def test_ls_empty_root_directory(self):
...
+                        "path": temp.name,
+                        "function-name": "dummy_ls_entries"
+                    }
+                }}
+            with cli.temp_config_file(config) as path:
+                flags = f'--config {path} --verbose'
+                cp, entries = cli.ls(uuids[0], self.cook_url, flags=flags)
+                self.assertEqual(0, cp.returncode, cp.stderr)
+                self.assertEqual(2, len(entries))
+                entry1 = entry(path1)
```
yueri: Would it be better to just call `cli.ls_entry_by_name` instead of the one line helper function above to make it clearer that it's calling `ls_entry_by_name` and avoid shadowing `entries`?

dposada: Definitely. Changed, thanks.

dposada:
@pschorf Can we merge this?

PR #1072, opened 2019-01-30
Add support for suitable flag to data local service

pschorf:
- Adds support for a "suitable" flag on data local service responses

This allows more granular control of the scheduling by the data local service.

### DIFF
--- a/scheduler/src/cook/mesos/constraints.clj
+++ b/scheduler/src/cook/mesos/constraints.clj
@@ -143,7 +143,7 @@
           launch-after-age (t/plus (tc/from-date submit-time) (t/seconds launch-wait-seconds))
           launch-without-data? (t/after? (t/now) launch-after-age)]
       (if (contains? data-locality-costs datasets)
-        (let [{:keys [suitable]} (get-in data-locality-costs [datasets (get vm-attributes "HOSTNAME")] {:suitable true})]
+        (let [{:keys [suitable] :or {suitable true}} (get-in data-locality-costs [datasets (get vm-attributes "HOSTNAME")])]
           (if suitable
             [true nil]
             [false "Host is not suitable for datasets"]))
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/mesos/constraints.clj
```diff
@@ -136,18 +136,22 @@
...
           launch-after-age (t/plus (tc/from-date submit-time) (t/seconds launch-wait-seconds))
           launch-without-data? (t/after? (t/now) launch-after-age)]
-      (if (or launch-without-data?
-              (contains? data-locality-costs datasets))
-        [true nil]
-        [false "No data locality costs available"])))
-  (job-constraint-evaluate [this _ _ _]
-    (job-constraint-evaluate this _ _)))
+      (if (contains? data-locality-costs datasets)
+        (let [{:keys [suitable]} (get-in data-locality-costs [datasets (get vm-attributes "HOSTNAME")] {:suitable true})]
```
shamsimam:
```suggestion
        (let [{:keys [suitable] :or {suitable true}} (get-in data-locality-costs [datasets (get vm-attributes "HOSTNAME")])]
```

scheduler/src/cook/mesos/data_locality.clj
```diff
@@ -198,8 +200,10 @@
           {:keys [job/uuid] :as job} (:job task-request)
           datasets (get-dataset-maps job)]
       (if-not (empty? (or datasets []))
-        (let [normalized-fitness (- 1.0
-                                    (get-in (get-data-local-costs) [datasets (.getHostname target-vm)] 1.0))
+        (let [{:keys [cost suitable]}(get-in (get-data-local-costs) [datasets (.getHostname target-vm)] {:cost 1.0
```
shamsimam: Nitpick: missing space.

shamsimam: Can we use `:or` in the destructuring to assign the default values for `cost` and `suitable`?

scheduler/src/cook/mesos/constraints.clj
pschorf: Done

scheduler/src/cook/mesos/data_locality.clj
pschorf: Done

PR #1074, opened 2019-02-05
Introduce a 'central' .cs.json for cs command.

yueri:

- Introduces a central `.cs.json` file for the cs command

We want to have a default `.cs.json` configuration file for all users. This allows users to roll out updates to their configurations (for instance, adding a new cluster) by only updating the default `.cs.json` file rather than the `.cs.json` file of all their end users.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -17,8 +17,12 @@ Custom defaults may be provided for all commands.
 Multiple clusters are supported via configuration.
 
 In order to use the Cook CLI, you’ll need a configuration file. 
-`cs` looks first for a base `.cs.json` file in `${DIR}/.cs.json`, `${DIR}/../.cs.json`, and `${DIR}/../config/.cs.json`, where `${DIR}` is the location of the `cs` executable.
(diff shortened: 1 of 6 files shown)


cli/README.md
```diff
@@ -17,7 +17,8 @@ Custom defaults may be provided for all commands.
 Multiple clusters are supported via configuration.
 
 In order to use the Cook CLI, you’ll need a configuration file. 
-`cs` looks first for a `.cs.json` file in the current directory, and then for a `.cs.json` file in your home directory.
```
yueri: I'm not very happy with the wording here. Can someone think of a better way to phrase this?

dposada: I think this wording is fine. Maybe it would help to add the "why" from your PR:

> ## Why are we making these changes?
> We want to have a default `.cs.json` configuration file for all users. This allows users to roll out updates to their configurations (for instance, adding a new cluster) by only updating the default `.cs.json` file rather than the `.cs.json` file of all their end users.

PR #1075, opened 2019-02-07
Reindent some files.

scrosby:

- Reindent some files.
- 
- 

- Makes the code look nicer.


scheduler/src/cook/mesos/api.clj
```diff
@@ -2205,7 +2205,7 @@
     (if (nil? pool)
       (let [pool-limits (get-pool-limits get-limit-fn db user)]
         (cond-> response
-          (not (empty? pool-limits)) (assoc :pools pool-limits)))
+                (not (empty? pool-limits)) (assoc :pools pool-limits)))
```
shamsimam: `cond->` conditions and clauses should be indented one level. You can change Cursive settings to fix this.

scheduler/src/cook/mesos/api.clj
```diff
@@ -2705,8 +2705,8 @@
                                  (not (string? body)))
                             (coll? body))]
       (cond-> resp
-        json-response (res/content-type "application/json")
-        (and json-response body) (assoc :body (streaming-json-encoder body))))))
+              json-response (res/content-type "application/json")
```
shamsimam: `cond->` conditions and clauses should be indented one level. You can change Cursive settings to fix this.

scheduler/src/cook/mesos/api.clj
```diff
@@ -815,8 +815,8 @@
   (->> datasets
        (map (fn [{:keys [dataset partitions]}]
               (cond-> {:dataset (walk/stringify-keys dataset)}
-                partitions
-                (assoc :partitions (into #{} (walk/stringify-keys partitions))))))
+                      partitions
```
shamsimam: `cond->` conditions and clauses should be indented one level. You can change Cursive settings to fix this.

shamsimam:
Here is a clojure style guide which mentioned how to handle `cond->`: https://github.com/bbatsov/clojure-style-guide#body-indentation

pschorf:
Do we want to merge or close this?

shamsimam:
I would like to merge this after the `cond->` changes.

scheduler/src/cook/mesos/api.clj
scrosby: Change reverted.

scheduler/src/cook/mesos/api.clj
scrosby: Reverted

scheduler/src/cook/mesos/api.clj
scrosby: Reverted.

scrosby:
Reverted the problematic changes and fixed another. Should be good to merge.

PR #1080, opened 2019-02-14
Adding plugins for cat and tail

yueri:

- Adding a HEAD request to the http requests.
- Adding a plugin hook in the `cat` and `tail` subcommands for printing out files
- Adding integration tests for plugging in dummy `cat` and `tail` plugin functions

We want to add runtime support extensions to the `cat` and `tail` subcommands to complement the plugins provided by the `ls` subcommand already. The added HEAD request may be used by plugins to quickly verify they can run successfully, defaulting otherwise to the base command.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -132,6 +132,49 @@ The `usage` command allows you to list cluster share and usage for a particular 
 In addition to totals, it displays a breakdown of usage by application and job group.
 As with `show`, `usage` will list usage across all configured clusters.
 
+### Plugins
+
+You can also specify custom functionality for some subcommands with your own plugin functions.
+You will need to add a `plugins` map to your `.cs.json` file. An example plugin config is shown below.
+
+```json
+{
+  "plugins": {
+    "plugin-name": {
+      "module-name": "module_name",
+      "path": "path_to_plugin_module",
+      "function-name": "function_name"
+    }
+  }
+}
+```
+
+Plugins are available for the following subcommands:
+
+#### `ls`
+
+The `ls` command has the plugin name `retrieve-job-instance-files`.
+The plugin function must take in an instance, sandbox directory, and path,
+and must return an array of dicts containing the file metadata fields
+`path`, `mode`, `nlink`, `size`, `uid`, `gid`, and `mtime`.
+
+#### `cat`
+
+The `cat` command has the plugin name `download-job-instance-file`.
+The plugin function must take in an instance, sandbox directory, and path,
+and must return an object with an `iter_content` method which generates the file contents.
+
+#### `tail`
+
+The `tail` command has the plugin name `read-job-instance-file`.
+The plugin function must take in an instance, sandbox directory, path, offset, and length,
+and must return a dict containing two fields: `data` and `offset`.
+ - When the `offset` param is `None`, the returned `data` is empty and `offset` is the file size.
+ - When the `offset` param is not `None`, the returned `data` is the file contents starting at offset `offset` with length `length`,
+   and the returned `offset` is the `offset` param.
+
+See the integration tests for simple examples of the subcommand plugins.
+
 ### Examples
 
 Simple job creation:
--- a/cli/README.md
+++ b/cli/README.md
@@ -162,7 +162,7 @@ and must return an array of dicts containing the file metadata fields
 
 The `cat` command has the plugin name `download-job-instance-file`.
 The plugin function must take in an instance, sandbox directory, and path,
-and must return an object with an `iter_content` method which generates the file contents.
+and must return an iterator that takes in a chunk size, which generates the file contents.
 
 #### `tail`
 
--- a/cli/cook/http.py
+++ b/cli/cook/http.py
@@ -45,6 +45,12 @@ def configure(config):
             raise Exception(f'Encountered unsupported authentication type "{auth_type}".')
 
 
+def __head(url, params=None, **kwargs):
+    """Sends a HEAD with params to the given url"""
+    logging.info(f'HEAD {url} with {params}')
+    return session.head(url, params=params, timeout=timeouts, **kwargs)
+
+
 def __post(url, json_body):
     """Sends a POST with the json payload to the given url"""
     logging.info(f'POST {url} with body {json_body}')
--- a/cli/cook/subcommands/cat.py
+++ b/cli/cook/subcommands/cat.py
@@ -11,7 +11,7 @@ from cook.util import guard_no_cluster
 
 def cat_for_instance(instance, sandbox_dir, path):
     """Outputs the contents of the Mesos sandbox path for the given instance."""
-    retrieve_fn = plugins.get_fn('cat-plugin', download_file)
+    retrieve_fn = plugins.get_fn('download-job-instance-file', download_file)
     resp = retrieve_fn(instance, sandbox_dir, path)
     try:
         for data in resp.iter_content(chunk_size=4096):
--- a/cli/cook/subcommands/cat.py
+++ b/cli/cook/subcommands/cat.py
@@ -3,6 +3,7 @@ import logging
 import sys
 from functools import partial
 
+from cook import plugins
 from cook.mesos import download_file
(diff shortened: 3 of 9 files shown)


integration/tests/cook/test_cli.py
```diff
@@ -1783,6 +1818,41 @@ def test_cat_binary_file(self):
...
+        with tempfile.NamedTemporaryFile(suffix='.py', delete=True) as temp:
+            plugin_code = """
+class MockResp:
+    def iter_content(self, chunk_size):
+        yield b"Hello"
+        yield b" "
+        yield b"world!"
+
+def dummy_cat_text(_, __, ___):
+    return MockResp()
```
dposada: Why is the `MockResp` with `yield`s needed?

integration/tests/cook/test_cli.py
```diff
@@ -1783,6 +1818,41 @@ def test_cat_binary_file(self):
...
+"""
+            temp.write(plugin_code.encode())
+            temp.flush()
+            config = {
+                'plugins': {
+                    "cat-plugin": {
+                        "module-name": "does_not_matter",
+                        "path": temp.name,
+                        "function-name": "dummy_cat_text"
```
dposada: Nitpick: mix of `"` and `'` above

integration/tests/cook/test_cli.py
```diff
@@ -1783,6 +1818,41 @@ def test_cat_binary_file(self):
...
+                }
+            }
+
+            with cli.temp_config_file(config) as path:
+                flags = f'--config {path} --verbose'
+                cp, uuids = cli.submit('touch file.txt', self.cook_url)
+                self.assertEqual(0, cp.returncode, cp.stderr)
+                cp = cli.wait(uuids, self.cook_url)
+                self.assertEqual(0, cp.returncode, cp.stderr)
+                cp = cli.cat(uuids[0], 'file.txt', self.cook_url, flags=flags)
```
dposada: Do we even need to submit a job to test this?

integration/tests/cook/test_cli.py
```diff
@@ -858,6 +858,41 @@ def test_tail_wait(self):
...
+                }
+            }
+
+            with cli.temp_config_file(config) as path:
+                flags = f'--config {path} --verbose'
+                cp, uuids = cli.submit('touch file.txt', self.cook_url)
+                self.assertEqual(0, cp.returncode, cp.stderr)
+                cp = cli.wait(uuids, self.cook_url)
+                self.assertEqual(0, cp.returncode, cp.stderr)
+                cp = cli.tail(uuids[0], 'file.txt', self.cook_url, flags=flags)
```
dposada: Do we even need to submit a job to test this?

cli/cook/http.py
```diff
@@ -45,6 +45,12 @@ def configure(config):
             raise Exception(f'Encountered unsupported authentication type "{auth_type}".')
 
 
+def __head(url, params=None, **kwargs):
+    """Sends a HEAD with params to the given url"""
+    logging.info(f'HEAD {url} with {params}')
+    return session.head(url, params=params, timeout=timeouts, **kwargs)
+
+
```
dposada: Where is this used?

cli/cook/subcommands/cat.py
```diff
@@ -3,14 +3,16 @@
...
+from cook import plugins
 from cook.mesos import download_file
 from cook.querying import parse_entity_refs, query_unique_and_run, parse_entity_ref
 from cook.util import guard_no_cluster
 
 
 def cat_for_instance(instance, sandbox_dir, path):
     """Outputs the contents of the Mesos sandbox path for the given instance."""
-    resp = download_file(instance, sandbox_dir, path)
+    retrieve_fn = plugins.get_fn('cat-plugin', download_file)
```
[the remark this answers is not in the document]
yueri: That's a good idea. Do you think it would be better to put the documentation here or have a more centralized list of plugin signatures?

integration/tests/cook/test_cli.py
yueri: This is probably where the documentation would help :)

The function `download_file` in mesos.http (which we overwrite) returns a Response object, and we use the `iter_content` function in there to stream the bytes of the file to `cat`. I just wanted to create a quick mock object which had a `iter_content` method to generate "Hello World!"

integration/tests/cook/test_cli.py
yueri: We need to submit a job to get a uuid to pass into `cat`, and `touch` creates a file to pass into `cat` as well.

integration/tests/cook/test_cli.py
yueri: We need to submit a job to get a uuid to pass into `tail`, and `touch` creates a file to pass into `tail` as well.

cli/cook/http.py
yueri: The plugin we're writing uses a HEAD request to check if the data exists, and falls back to the default command if it doesn't. It's not actually used anywhere inside Cook itself.

cli/cook/subcommands/cat.py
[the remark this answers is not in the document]
dposada: Probably somewhere in the README for the CLI would be good

integration/tests/cook/test_cli.py
dposada: Do we have to `wait` for the job?

integration/tests/cook/test_cli.py
dposada: Do we have to `wait` for the job?

yueri: I'm not sure if it's completely necessary, but the other `cat` integration tests `wait` until the job is completed, so I thought I would do the same, even if it is just creating a file.

integration/tests/cook/test_cli.py
yueri: I'm not sure if it's completely necessary, but the other `tail` integration tests `wait` until the job is completed, so I thought I would do the same, even if it is just creating a file.

cli/README.md
```diff
@@ -132,6 +132,49 @@ The `usage` command allows you to list cluster share and usage for a particular
...
+The `ls` command has the plugin name `retrieve-job-instance-files`.
+The plugin function must take in an instance, sandbox directory, and path,
+and must return an array of dicts containing the file metadata fields
+`path`, `mode`, `nlink`, `size`, `uid`, `gid`, and `mtime`.
+
+#### `cat`
+
+The `cat` command has the plugin name `download-job-instance-file`.
+The plugin function must take in an instance, sandbox directory, and path,
+and must return an object with an `iter_content` method which generates the file contents.
```
dposada: This feels a little awkward, since it uses a function from `Response` but doesn't actually require a `Response`. 

We could:

- Make the requirement stronger and say that the function must return a valid `requests` `Response` object.
- Instead of requiring `iter_content` at all (which is a `Response`-specific concept), just require that the function return any valid Python iterator. This might make it feel more like the other two plugin functions, which don't assume you're using a specific HTTP library.

What do you think?

integration/tests/cook/test_cli.py
dposada: If we can skip the `wait` and the test still works, I think it's worth it to make the test faster and potentially more reliable.

integration/tests/cook/test_cli.py
dposada: See question above about making this plugin API clearer

integration/tests/cook/test_cli.py
dposada: Same comment as above

cli/README.md
yueri: Having the function return a valid requests Response object probably makes more sense. The iterator approach will likely be more confusing since your iterator will still have to be wrapped in an `iter_content` method, unless we want to change the API of `read_file`

dposada: I'm fine with changing the API of `read_file` if it makes this cleaner

dposada:

PR #1082, opened 2019-02-20
Copy over the entire :plugins subkey from the config map.

scrosby:

- Copy over the entire :plugin map from the config.
- 
- 

So we have a place to put plugin-relevant arguments.


dposada:
Can you give a specific example of why this is needed?

scrosby:
The configuration map used internally is built from config.edn. The only way for a plugin to get parameters is by examining the config map defstate. WIth this code, plugin parameters can be put in a key (or dictionary) under :plugins in config.edn, and will be copied into the config defstate. Without this code, config.clj would drop the parameters and we don't have a good place for plugin configuration.

PR #1086, opened 2019-02-25
Adding option to load plugins to tests

yueri:

- Adding an environment variable to load plugin json to integration tests.

This change allows us to run the integration tests using our internal plugins by specifying the plugins in an environment variable.

### DIFF
--- a/integration/tests/cook/cli.py
+++ b/integration/tests/cook/cli.py
@@ -58,6 +58,11 @@ def command():
     return os.environ['COOK_CLI_COMMAND'] if 'COOK_CLI_COMMAND' in os.environ else 'cs'
 
 
+def plugins():
+    """If the COOK_PLUGIN_JSON environment variable is set, returns its value, otherwise empty dict"""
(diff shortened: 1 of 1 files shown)


integration/tests/cook/cli.py
```diff
@@ -58,6 +58,18 @@ def command():
     return os.environ['COOK_CLI_COMMAND'] if 'COOK_CLI_COMMAND' in os.environ else 'cs'
 
 
+def plugins():
+    """If the COOK_PLUGIN_JSON environment variable is set, returns its value, otherwise empty dict"""
+    if 'COOK_PLUGIN_JSON' in os.environ:
+        try:
+            return json.loads(os.environ['COOK_PLUGIN_JSON'])
+        except ValueError:
+            logging.warn(f"Plugin json '{os.environ['COOK_PLUGIN_JSON']}' cannot be parsed. Ignoring.")
```
dposada: Maybe we should `raise` here instead of just logging a warning?

PR #1088, opened 2019-02-26
Bumps CLI version

dposada:

- bumping version to `2.14.0`

We've added support to the CLI for specifying plugins for file retrieval operations in `ls`, `cat`, and `tail`.


dposada:
@pschorf Bump

PR #1095, opened 2019-03-04
Catch error in supported isolators

pschorf:
- Add a `try...except` around the check for supported isolators

Improves test reliability

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1312,16 +1312,17 @@ def _supported_isolators():
     mesos_url = retrieve_mesos_url()
     mesos_state = get_mesos_state(mesos_url)
     slave_endpoint = get_agent_endpoint(mesos_state, mesos_state['slaves'][0]['hostname'])
+    slave_response = session.get(slave_endpoint)
     try:
-        slave_state = session.get(slave_endpoint).json()
+        slave_state = slave_response.json()
         if 'flags' in slave_state:
             return set(slave_state['flags']['isolation'].split(','))
         else:
             logger.error(f'Unable to determine flags from slave state at {slave_endpoint}: {slave_state}')
             return []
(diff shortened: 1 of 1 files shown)


integration/tests/cook/util.py
```diff
@@ -1312,11 +1312,15 @@ def _supported_isolators():
     mesos_url = retrieve_mesos_url()
     mesos_state = get_mesos_state(mesos_url)
     slave_endpoint = get_agent_endpoint(mesos_state, mesos_state['slaves'][0]['hostname'])
-    slave_state = session.get(slave_endpoint).json()
-    if 'flags' in slave_state:
-        return set(slave_state['flags']['isolation'].split(','))
-    else:
-        logger.error(f'Unable to determine flags from slave state at {slave_endpoint}: {slave_state}')
+    try:
```
scrosby: Can you put a 'why' here:
""The request can be empty, bad json, etc.""

Two related questions:
  If its OK to return [] on an error, why not return [] all the time?
  If errors aren't good, should we be trying more than once here?

integration/tests/cook/util.py
```diff
@@ -1312,11 +1312,15 @@ def _supported_isolators():
...
+        slave_state = session.get(slave_endpoint).json()
+        if 'flags' in slave_state:
+            return set(slave_state['flags']['isolation'].split(','))
+        else:
+            logger.error(f'Unable to determine flags from slave state at {slave_endpoint}: {slave_state}')
+            return []
+    except Error as e:
+        logger.exception('Unable to parse json response')
```
scrosby: Probably worth printing what the response was here and the error.

scrosby:
Looks good to me. Just want to see a green travis build.

pschorf:
Green!

PR #1096, opened 2019-03-04
Move output url check to separate test

pschorf:
- Move the output URL check into a separate test (from `basic_submit`)

This runs the output URL check in a separate test with a longer-running job. This will ensure that the sandbox syncer has time to run on that task.


dposada:
Will merge when green

pschorf:
I made a couple of changes to this, so take another look before you merge.

pschorf:
Ready to merge now

PR #1102, opened 2019-03-18
Outgoing/auto clean uncommitted

scrosby:

- Runs the clear-uncommitted-jobs on a schedule; 12 hours after starting up and every 24 hours thereafter.
- 
- 

If something goes wrong and a batch of jobs is submitted to Cook, but is never committed, they will slowly build up and clutter the database, causing the scheduler loop to slow down due to get-pending-jobs having to read and skip them.

The cost appears to be around one second every 50k jobs. This is a simple loop that nukes any uncommitted jobs older than a few days.

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -14,9 +14,11 @@
 ;; limitations under the License.
 ;;
 (ns cook.mesos.util
-  (:require [clj-time.coerce :as tc]
+  (:require [chime]
+            [clj-time.coerce :as tc]
             [clj-time.core :as t]
             [clj-time.format :as tf]
+            [clj-time.periodic :as tp]
             [cook.cache :as ccache]
             [cook.config :as config]
             [cook.mesos.pool :as pool]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -690,12 +692,36 @@
...
+
+   If something goes wrong and a batch of jobs is submitted to Cook, but is never committed, they will slowly build up
+   and clutter the database, causing the scheduler loop to slow down due to get-pending-jobs having to read and skip them.
+
+   The cost appears to be around one second every 50k jobs.
+
+   This is a simple loop that nukes any uncommitted jobs older than a few days. It runs every 24 hours. There will be a minor
+   performance hiccough lasting around 1 minute per 1000 jobs or so as this runs and flushes."
+  [conn mesos-leadership-atom]
+  (let [age (tc/to-date (-> -3 t/days t/from-now))
```
pschorf: Shouldn't this be a variable in the chime function?

scrosby: Yes, it should. :+1:

scrosby: Shams suggested 7 days. I'm making that change too._

PR #1103, opened 2019-03-19
Add plugin for pool selection

pschorf:
- Adds a plugin to support selecting the pool for an offer 

Some environments may want to use different logic for selecting the appropriate pool for an offer.

### DIFF
--- a/scheduler/src/cook/plugins/pool.clj
+++ b/scheduler/src/cook/plugins/pool.clj
@@ -11,14 +11,14 @@
     (or (->> offer :attributes (filter #(= attribute-name (:name %))) first :text)
         default-pool)))
 
-(defn create-default-plugin-object
+(defn create-plugin-object
   "Returns the configured PoolSelector, or a no-op if none is defined."
   [config]
   (let [pool-selection (get-in config [:settings :plugins :pool-selection])
         factory-fn (:factory-fn pool-selection)]
     (if factory-fn
       (do
-        (log/info "Creating instance completion plugin with" factory-fn)
+        (log/info "Creating pool selection plugin with" factory-fn)
         (if-let [resolved-fn (cook.plugins.util/resolve-symbol (symbol factory-fn))]
           (resolved-fn config)
           (throw (ex-info (str "Unable to resolve factory fn " factory-fn)))))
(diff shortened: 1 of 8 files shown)


scheduler/src/cook/plugins/pool.clj
```diff
@@ -0,0 +1,29 @@
...
+        default-pool)))
+
+(defn create-default-plugin-object
+  "Returns the configured PoolSelector, or a no-op if none is defined."
+  [config]
+  (let [pool-selection (get-in config [:settings :plugins :pool-selection])
+        factory-fn (:factory-fn pool-selection)]
+    (if factory-fn
+      (do
+        (log/info "Creating instance completion plugin with" factory-fn)
```
dposada:
```suggestion
        (log/info "Creating pool selection plugin with" factory-fn)
```

scheduler/src/cook/plugins/pool.clj
```diff
@@ -0,0 +1,29 @@
...
+            [cook.plugins.util]
+            [mount.core :as mount]))
+
+(defrecord AttributePoolSelector [attribute-name default-pool]
+  PoolSelector
+  (select-pool [this offer]
+    (or (->> offer :attributes (filter #(= attribute-name (:name %))) first :text)
+        default-pool)))
+
+(defn create-default-plugin-object
```
dposada: Why is this "default" instead of just `create-plugin`?

scheduler/src/cook/plugins/pool.clj
pschorf: Done

scheduler/src/cook/plugins/pool.clj
pschorf: Done!

scheduler/src/cook/plugins/pool.clj
```diff
@@ -0,0 +1,29 @@
...
+            [mount.core :as mount]))
+
+(defrecord AttributePoolSelector [attribute-name default-pool]
+  PoolSelector
+  (select-pool [this offer]
+    (or (->> offer :attributes (filter #(= attribute-name (:name %))) first :text)
+        default-pool)))
+
+(defn create-plugin-object
+  "Returns the configured PoolSelector, or a no-op if none is defined."
```
dposada: The no-op part is not true, right?

pschorf: Freaking docstrings. Fixed :)

PR #1108, opened 2019-03-25
Outgoing/cleanup framework

scrosby:

- Cleanup how the framework is propagated throughout the code.

    
    It used to be threaded through the scheduler and api query code, stuffed
    into the job object, then threaded through fenzo, and finally sent as part
    of offer responses. Shortcut all of that indirection and source it
    directly from a global.
 

It gets is closer to letting cook have multiple backends. E.g., manage multiple mesos clusters.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -490,7 +490,6 @@ class CookTest(util.CookTest):
     def memory_limit_exceeded_helper(self, command, executor_type):
         """Given a command that needs more memory than it is allocated, when the command is submitted to cook,
         the job should be killed by Mesos as it exceeds its memory limits."""
-        cook_framework = util.settings(self.cook_url)['mesos-framework-id']
         job_uuid, resp = util.submit_job(self.cook_url, command=command, executor=executor_type, mem=128)
         self.assertEqual(201, resp.status_code, msg=resp.content)
         instance = util.wait_for_instance(self.cook_url, job_uuid)
@@ -522,7 +521,7 @@ class CookTest(util.CookTest):
             else:
                 self.fail('Unknown reason code {}, details {}'.format(instance['reason_code'], instance_details))
         finally:
-            mesos.dump_sandbox_files(util.session, instance, job, cook_framework)
+            mesos.dump_sandbox_files(util.session, instance, job)
             util.kill_jobs(self.cook_url, [job_uuid])
 
     @pytest.mark.memlimit
@@ -1542,7 +1541,6 @@ class CookTest(util.CookTest):
     @pytest.mark.docker
     @unittest.skipUnless(util.has_docker_service(), "Requires `docker inspect`")
     def test_docker_port_mapping(self):
-        cook_framework = util.settings(self.cook_url)['mesos-framework-id']
         job_uuid, resp = util.submit_job(self.cook_url,
                                          command='python -m http.server 8080',
                                          ports=2,
@@ -1606,7 +1604,7 @@ class CookTest(util.CookTest):
             job = util.load_job(self.cook_url, job_uuid)
             self.logger.info(f'Job status is {job["status"]}: {job}')
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
-            mesos.dump_sandbox_files(util.session, instance, cook_framework, cook_framework)
+            mesos.dump_sandbox_files(util.session, instance, job)
 
     def test_unscheduled_jobs(self):
         job_spec = {'command': 'sleep 30',
@@ -2280,7 +2278,6 @@ class CookTest(util.CookTest):
             util.kill_jobs(self.cook_url, job_uuids)
 
     def test_instance_stats_success(self):
-        cook_framework = util.settings(self.cook_url)['mesos-framework-id']
         name = str(uuid.uuid4())
         job_uuid_1, resp = util.submit_job(self.cook_url, command='exit 0', name=name, cpus=0.1, mem=32)
         self.assertEqual(resp.status_code, 201, msg=resp.content)
@@ -2336,7 +2333,7 @@ class CookTest(util.CookTest):
                 self.assertAlmostEqual(sum(mem_times), mem_seconds['total'])
             except:
                 for instance in instances:
-                    mesos.dump_sandbox_files(util.session, instance, instance['parent'], cook_framework)
+                    mesos.dump_sandbox_files(util.session, instance, instance['parent'])
                 raise
         finally:
             util.kill_jobs(self.cook_url, job_uuids)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -565,7 +565,6 @@ class CookTest(util.CookTest):
         # 9223372036854775807 is MAX_LONG(ish), the default value for max_runtime
         self.assertEqual('success', job['state'])
         self.assertTrue('env' in job)
-        self.assertTrue('framework_id' in job)
         self.assertTrue('ports' in job)
         self.assertTrue('instances' in job)
         self.assertEqual('completed', job['status'])
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -490,6 +490,7 @@ class CookTest(util.CookTest):
     def memory_limit_exceeded_helper(self, command, executor_type):
         """Given a command that needs more memory than it is allocated, when the command is submitted to cook,
         the job should be killed by Mesos as it exceeds its memory limits."""
+        cook_framework = util.settings(self.cook_url)['mesos-framework-id']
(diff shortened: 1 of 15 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -140,9 +140,9 @@
 
 (defn TaskAssignmentResult->task-metadata
   "Organizes the info Fenzo has already told us about the task we need to run"
-  [db framework-id mesos-run-as-user ^TaskAssignmentResult task-result]
+  [db mesos-run-as-user ^TaskAssignmentResult task-result]
   (let [{:keys [job task-id] :as task-request} (.getRequest task-result)]
-    (merge (job->task-metadata db framework-id mesos-run-as-user job task-id)
+    (merge (job->task-metadata db @cook.config/framework-id-atom mesos-run-as-user job task-id)
```
scrosby: I kept this here. Long term when cook schedules over multiple domains, it can vary on a per-offer basis in the fenzo cycle. 

Longer term, this will be replaced with a function from the Fenzo result (indicating the scheduling domain matched) that will store the selected scheduling domain into the task structure (and subsequently use that in order to launch it in the right spot.).

Today, however, we have a single scheduling domain --- a single mesos cluster, so just stuff that in here.

we'll encode the domain in fenzo.

integration/tests/cook/test_basic.py
```diff
@@ -565,7 +565,6 @@ def test_get_job(self):
         # 9223372036854775807 is MAX_LONG(ish), the default value for max_runtime
         self.assertEqual('success', job['state'])
         self.assertTrue('env' in job)
-        self.assertTrue('framework_id' in job)
```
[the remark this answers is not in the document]
pschorf: Are we actually removing the field? Or just removing it from the test? Either way, I'd vote to keep it. Even if it's broken, it's part of the API. We can come up with a new field but should probably keep putting something here.

integration/tests/cook/test_basic.py
```diff
@@ -1605,7 +1606,7 @@ def get_docker_info():
             job = util.load_job(self.cook_url, job_uuid)
             self.logger.info(f'Job status is {job["status"]}: {job}')
             util.session.delete('%s/rawscheduler?job=%s' % (self.cook_url, job_uuid))
-            mesos.dump_sandbox_files(util.session, instance, job)
+            mesos.dump_sandbox_files(util.session, instance, cook_framework, cook_framework)
```
pschorf: `util.session, instance, job, cook_framework`

scheduler/src/cook/components.clj
```diff
@@ -338,9 +337,12 @@
     (mount/start-with-args (cook.config/read-config config-file-path))
     (pool/guard-invalid-default-pool (d/db datomic/conn))
     (metrics-jvm/instrument-jvm)
-    (let [server (scheduler-server config)]
+    (let [server (scheduler-server config)
+          framework-id (get-in server [:framework-id])]
       (intern 'user 'main-graph server)
-      (log/info "Started Cook, stored variable in user/main-graph"))
+      (log/info "Started Cook, stored variable in user/main-graph")
+      (log/info "Framework-id " )
```
pschorf:
```suggestion
      (log/info "Framework-id " framework-id)
```

scheduler/src/cook/mesos/api.clj
```diff
@@ -341,8 +341,7 @@
   (-> JobRequestMap
       (dissoc (s/optional-key :group))
       (dissoc (s/optional-key :status))
-      (merge {:framework-id (s/maybe s/Str)
-              :retries-remaining NonNegInt
+      (merge {:retries-remaining NonNegInt
```
pschorf: Discussed above, I don't think we should remove this, even if it's currently broken.

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -1418,21 +1418,15 @@
...
    handle-exit-code handle-progress-message sandbox-syncer-state]
-  (let [sync-agent-sandboxes-fn #(sandbox/sync-agent-sandboxes sandbox-syncer-state configured-framework-id %1 %2)
+  (let [sync-agent-sandboxes-fn #(sandbox/sync-agent-sandboxes sandbox-syncer-state @cook.config/framework-id-atom %1 %2)
         message-handlers {:handle-exit-code handle-exit-code
                           :handle-progress-message handle-progress-message}]
     (mesos/scheduler
       (registered
         [this driver framework-id master-info]
         (log/info "Registered with mesos with framework-id " framework-id)
-        (let [value (-> framework-id mesomatic.types/pb->data :value)]
```
pschorf: Why was this removed?

scheduler/src/cook/mesos/task.clj
pschorf: Can we put a code comment reflecting what you said here?

integration/tests/cook/test_basic.py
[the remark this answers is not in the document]
scrosby: I"m reverting this part of the change. Removing the framework-id out of job is a much bigger change with more communication needed. (I'm starting on that process now.)

FWIW, I'd rather remove it than leave behind something wrong. Regardless, thats going to be on another PR, so we can discuss it then?

scheduler/src/cook/mesos/task.clj
scrosby: Done.

scheduler/src/cook/components.clj
scrosby: I wanted to remove the field from the job structure, but that change needs to be reverted as it turned out to be more invasive than expected. I also reverted the sandbox changes. No reason to change them now and again when we put it into the task structure.

scheduler/src/cook/mesos/scheduler.clj
scrosby: It seemed unnecessary. We get the driver via code around here:

https://github.com/scrosby/Cook/blob/outgoing/cleanup-framework-id/scheduler/src/cook/mesos.clj#L197
create-datomic-scheduler calls create-mesos-scheduler.

Notice that we it what framework id we are. So if we ever get a registration callback for another framework-id, something is really screwed up; its a bug in our implementation/initialization. The driver atom is initialized about 20 lines further down.

We could retain this code and have it refer to the global variable, but I don't see the point, or the value.

scheduler/src/cook/components.clj
scrosby: :) Like that change.

scheduler/src/cook/mesos/scheduler.clj
pschorf: I think the point of this is to make sure there's not something really screwed up :)

It seems pretty harmless to me as a sanity check.

scrosby: Left in.

pschorf:
Build is red, otherwise LGTM

integration/tests/cook/test_basic.py
pschorf: Shouldn't this be re-added?

scheduler/src/cook/components.clj
```diff
@@ -338,9 +337,11 @@
     (mount/start-with-args (cook.config/read-config config-file-path))
     (pool/guard-invalid-default-pool (d/db datomic/conn))
     (metrics-jvm/instrument-jvm)
-    (let [server (scheduler-server config)]
+    (let [server (scheduler-server config)
+          framework-id (get-in server [:framework-id])]
```
pschorf: Can this be removed?

scheduler/test/cook/test/mesos/api.clj
```diff
@@ -1187,19 +1187,18 @@
...
                      disable-mea-culpa-retries executor datasets]
-              :or {disable-mea-culpa-retries false}}
-             framework-id]
+              :or {disable-mea-culpa-retries false}}]
             (cond-> {;; Fields we will fill in from the provided args:
                      :command command
                      :cpus cpus
                      :disable_mea_culpa_retries disable-mea-culpa-retries
-                     :framework_id framework-id
+                     :framework_id nil
```
pschorf: Can this be removed instead of being set to `nil`?

scrosby: No. Unit tests use this to make the pre-canned response they validate against. I either have the field, and set it to nil, or I have to wrap a bunch of unit tests with (with-redefs [framework-id-config (fn .....].

scheduler/src/cook/components.clj
scrosby: Good catch.

integration/tests/cook/test_basic.py
scrosby: Either way.... :)

scheduler/test/cook/test/mesos/api.clj
scrosby: Did the redef as we discussed.

PR #1109, opened 2019-03-26
Make the auto clean uncommitted be a jodatime to match clojure.

scrosby:

- A java.util.Date is the wrong type. Make it JodaTime so that the cleaner works correctly.
- 
-

### DIFF
--- a/scheduler/src/cook/mesos/util.clj
+++ b/scheduler/src/cook/mesos/util.clj
@@ -701,16 +701,15 @@
                                 batch))))
     uncommitted-before))
 
-(defn clear-old-uncommitted-jobs-fn
+(defn clear-old-uncommitted-jobs
   "Returns a function that will, when invoked, clear any uncommitted jobs older than a week if the leadership atom is true."
   [conn mesos-leadership-atom]
-  (fn [_]
     (try
       (when @mesos-leadership-atom
         (let [age (-> -7 t/days t/from-now tc/to-date-time)]
           (clear-uncommitted-jobs conn age false)))
       (catch Exception e
-        (log/error e "Failed to kill the offensive job!")))))
+        (log/error e "Failed to kill the offensive job!"))))
 
 (defn clear-uncommitted-jobs-on-schedule
   "Runs the clear-uncommitted-jobs on a schedule; 12 hours after starting up and every 24 hours thereafter.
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos/util.clj
```diff
@@ -717,7 +717,7 @@
         schedule (tp/periodic-seq start-time frequency)
         chime-fn (fn [_]
                    (when @mesos-leadership-atom
-                     (let [age (-> -7 t/days t/from-now tc/to-date)]
+                     (let [age (-> -7 t/days t/from-now tc/to-date-time)]
```
pschorf: Since we've been burned once by this, can we have a test for it? One approach is to use `util/chime-at-ch` and have a channel passed in so you can manually trigger it in the test.

scheduler/src/cook/mesos/util.clj
```diff
@@ -717,7 +717,7 @@
         schedule (tp/periodic-seq start-time frequency)
         chime-fn (fn [_]
```
pschorf: We should have an error handler here so we can have proper logging.

scheduler/src/cook/mesos/util.clj
scrosby: I didn't see an example of that kind of use, but I tested it this way. Look ok?

scheduler/src/cook/mesos/util.clj

scheduler/src/cook/mesos/util.clj
```diff
@@ -701,6 +701,17 @@
                                 batch))))
     uncommitted-before))
 
+(defn clear-old-uncommitted-jobs-fn
+  "Returns a function that will, when invoked, clear any uncommitted jobs older than a week if the leadership atom is true."
+  [conn mesos-leadership-atom]
+  (fn [_]
```
pschorf: Instead of returning a function here, I think it makes more sense to have this perform the cleanup directly, and have the chime call `(fn [] (clear-old-uncommitted-jobs conn mesos-leadership-atom)`

pschorf:
Looks good to me. Waiting on a green build.

pschorf:
The cook executor tests errored, it looked like due to a problem with the pymesos version change. I cleared the build caches and restarted that stage.

PR #1113, opened 2019-04-01
Don't launch multiple reconnect threads if disconnected is called many times

pschorf:
- Only launch one thread on disconnect until reregistered is called.

It's wasteful to create the extra threads, and it pollutes the logs.

### DIFF
--- a/executor/cook/executor.py
+++ b/executor/cook/executor.py
@@ -418,9 +418,12 @@ class CookExecutor(pm.Executor):
     def reregistered(self, driver, agent_info):
         logging.info('Executor re-registered agent={}'.format(agent_info))
         if self.config.checkpoint:
-            logging.info('Executor checkpointing is enabled. Notifying on reregister_signal')
-            self.reregister_signal.set()
-            self.reregister_signal = None
+            if self.reregister_signal is not None:
+                logging.info('Executor checkpointing is enabled. Notifying on reregister_signal')
+                self.reregister_signal.set()
+                self.reregister_signal = None
+            else:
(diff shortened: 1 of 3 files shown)


executor/cook/executor.py
```diff
@@ -420,13 +420,16 @@ def reregistered(self, driver, agent_info):
         if self.config.checkpoint:
             logging.info('Executor checkpointing is enabled. Notifying on reregister_signal')
             self.reregister_signal.set()
```
shamsimam: We should guard this and the next statement inside a `self.reregister_signal is not None` if clause with a log statement on the else branch.

executor/cook/executor.py
```diff
@@ -420,13 +420,16 @@ def reregistered(self, driver, agent_info):
...
-            self.reregister_signal = Event()
+            self.reregister_signal = None
 
     def disconnected(self, driver):
         logging.info('Mesos requested executor to disconnect')
         if self.config.checkpoint:
-            logging.info('Executor checkpointing is enabled. Waiting for agent recovery.')
-            await_reregister(self.reregister_signal, self.config.recovery_timeout_ms / 1000, self.stop_signal, self.disconnect_signal)
+            if self.reregister_signal is None:
```
shamsimam: We should add a log in the else clause.

executor/cook/executor.py
pschorf: Done

executor/cook/executor.py
pschorf: Done

pschorf:
I still think the current setup is pretty straightforward, so I'd rather not add a new signal to coordinate on. Added the log statements.

PR #1114, opened 2019-04-02
Add RELEASING.md

pschorf:
- Add a RELEASING.md doc for cook executor

Describe the process for releasing a new executor version to PyPI

### DIFF
--- a/executor/RELEASING.md
+++ b/executor/RELEASING.md
@@ -1,0 +1,38 @@
+Releasing Cook Executor
+=======================
+
+Cook Executor is released on [PyPI](https://pypi.org/project/cook-executor/)
+
+Prerequisites
+-------------
+Ensure you can build the executor followng the instructions in README.md
+
+Install `twine`:
+```bash
+pip3 install twine
+```
+
+Test Release
+------------
+Since PyPI does not allow modifying releases, it can be useful to test a release using their test instance.
+```bash
+rm -rf dist/*
+python3 setup.py sdist bdist_wheel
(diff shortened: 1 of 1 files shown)


executor/RELEASING.md
```diff
@@ -0,0 +1,38 @@
...
+------------
+Since PyPI does not allow modifying releases, it can be useful to test a release using their test instance.
+```bash
+rm -rf dist/*
+python3 setup.py sdist bdist_wheel
+python3 -m twine upload --repository-url https://test.pypi.org/legacy/ dist/*
+```
+Then, in a separate virtualenv for testing:
+```bash
+pip install  --index-url https://test.pypi.org/simple/ --no-deps cook-executor==$VERSION
```
shamsimam: `pip3`?

executor/RELEASING.md
```diff
@@ -0,0 +1,38 @@
...
+pip install pymesos==0.3.9 # install any other required dependencies from the main pypi repo
+cook-executor
+```
+If there is an issue with the release, you can just release another version. They are GCed periodically from the test instance.
```
shamsimam:
```suggestion
If there is an issue with the release, you can just release another version. They are GC-ed periodically from the test instance.
```

executor/RELEASING.md
pschorf: Done

executor/RELEASING.md
pschorf: Done

PR #1115, opened 2019-04-03
Change cook executor to be more container-friendly

pschorf:
- Report MESOS_DIRECTORY to cook as the sandbox location instead of MESOS_SANDBOX. This is the root filesystem path instead of the path in the container
- Explicitly write executor.log in the MESOS_SANDBOX, not the current working directory

These make cook executor behave better when in a container.


shamsimam:
Awaiting green build.

PR #1118, opened 2019-04-03
Add support for EXECUTOR_RESET_VARS

pschorf:
- Add support for specifying a list of environment variables to clear in cook executor with `EXECUTOR_RESET_VARS`

Some environment variables may be set as part of starting the cook executor itself, which we would prefer not to propagate to the child process. This environment variable allows setting a list of environment variables to clear.


shamsimam:
Awaiting green build

PR #1120, opened 2019-04-04
Update reset_vars test to work when EXECUTOR_RESET_VARS is configured

pschorf:
- Change the reset_vars test to check the value of `EXECUTOR_RESET_VARS`, and check those variables if configured.

Since a job cannot override `EXECUTOR_RESET_VARS` if it's specified in the config, the test will fail. This allows the test to work in both kinds of environments.


dposada:

pschorf:
Build is ✅

PR #1121, opened 2019-04-04
Allow running cook executor in docker containers

pschorf:
- Allow jobs to run in a docker container using Cook Executor

This will allow jobs using containers to utilize executor-specific features like exit code and progress.

### DIFF
--- a/scheduler/src/cook/mesos/task.clj
+++ b/scheduler/src/cook/mesos/task.clj
@@ -381,8 +381,8 @@
 
             (= executor-key :container-cook-executor)
             (assoc :executor (assoc executor :name cook-executor-name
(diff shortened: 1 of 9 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -378,6 +379,11 @@
             (assoc :executor (assoc executor :name cook-executor-name
                                              :source cook-executor-source))
 
+            (= executor-key :container-cook-executor)
+            (assoc :executor (assoc executor :name cook-executor-name
+                                    :source cook-executor-source
+                                    :container container))
+
```
shamsimam: Nitpick: formatting is off

PR #1122, opened 2019-04-08
Fix bugs when querying for jobs using /jobs

pschorf:
- Use `uuid` instead of `job` as query parameter when hitting `/jobs` endpoint
- Fix group parsing logic when using `/jobs` endpoint
- Release jobclient 0.4.3

The current jobclient code is broken for querying when using the `/jobs` endpoint


dposada:

PR #1125, opened 2019-04-17
Add file_url to job instances

pschorf:
- Adds a new field, `file_url`, on job instances and a plugin definition to generate a file url for an instance.

The Mesos file API can be cumbersome and is Mesos-specific. This allows us to migrate to a simpler API which could be implemented for another scheduler, like k8s.

### DIFF
--- a/scheduler/src/cook/mesos/api.clj
+++ b/scheduler/src/cook/mesos/api.clj
@@ -29,7 +29,6 @@
             [cook.config :as config]
             [cook.cors :as cors]
             [cook.datomic :as datomic]
-            [cook.plugins.submission :as submission-plugin]
             [cook.mesos.data-locality :as dl]
             [cook.mesos.pool :as pool]
             [cook.mesos.quota :as quota]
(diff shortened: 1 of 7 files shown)


scheduler/src/cook/mesos/api.clj
```diff
@@ -947,7 +950,8 @@
         reason (reason/instance-entity->reason-entity db instance)
         exit-code (:instance/exit-code instance)
         progress (:instance/progress instance)
-        progress-message (:instance/progress-message instance)]
+        progress-message (:instance/progress-message instance)
+        file-url (plugins/file-url file-plugin/plugin instance)]
```
scrosby: Why not put this in the cond-> section, and omit it if the result is nil?

pschorf: What do you mean? `file-url` is already in the `cond->`. If you mean removing the variable assignment, I think it's more clear and matches the other fields in `cond->` which all have variables defined.

scheduler/src/cook/plugins/file.clj
```diff
@@ -0,0 +1,27 @@
...
+
+(defrecord NilFileUrlPlugin []
+  FileUrlGenerator
+  (file-url [this instance]
+    nil))
+
+(defn create-plugin-object
+  "Returns the configured FileUrlPlugin, or a NilFileUrlPlugin if none is defined."
+  [config]
+  (let [pool-selection (get-in config [:settings :plugins :file-url])
```
dposada: Naming bug

pschorf: Fixed 😳

PR #1134, opened 2019-05-01
Limit the number of outstanding lookup-key! requests.

scrosby:

- Do the cache lookup inside of a lock
- 
- 

Without this, we can have multiple outstanding requests for the same
key, possibly hitting a backend multiple times.

lookup-cache-with-expiration! has the fix already.


scrosby:
Before committing, I'd like to discuss threadpool concerns....

shamsimam:
Please unwip so that I can merge.

PR #1135, opened 2019-05-07
WIP: Compute cluster refactoring of cook.

scrosby:

- 
- 
-

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -95,15 +95,10 @@
       (or cluster-entity-id (get-mesos-cluster-entity-id (d/db conn) mesos-cluster))
       (atom nil))))
 
+
 ; Internal variable
 (def cluster-name->compute-cluster-atom (atom nil))
-; Internal variable
-(def default-cluster-for-legacy (atom nil))
 
-(defn get-default-cluster-for-legacy
-  "What cluster name to put on for legacy jobs when generating their compute-cluster."
-  []
-  @default-cluster-for-legacy)
 
 (defn- get-mesos-clusters-from-config
   "Get all of the mesos clusters defined in the configuration.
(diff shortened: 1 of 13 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -17,6 +17,25 @@
   (:require [clojure.tools.logging :as log]
             [datomic.api :as d]))
 
+(defprotocol ComputeCluster
+  ; These methods should accept bulk data and process in batches.
+  ;(kill-tasks [this task]
+  ;(launch-tasks [this offers task-metadata-seq]
+  (compute-cluster->db-id [this])
+
+  (get-mesos-driver-atom-hack [this]
```
pschorf: Can we have this return the driver instead of an atom? I think in an ideal world we'd initialize all of the things that need a driver when we take leadership instead of relying on the atom. I'd rather keep those referencing the atom explicitly for now than stick the atom everywhere else.

scheduler/src/cook/mesos/rebalancer.clj
```diff
@@ -485,7 +486,7 @@
             (catch Throwable e
               (log/warn e "Failed to transact preemption")))
           (when-let [task-id (:instance/task-id task-ent)]
-            (mesos/kill-task! driver {:value task-id})))))))
```
pschorf: for instance, this is probably going to break because it's getting an atom now.

scheduler/src/cook/compute_cluster.clj
```diff
@@ -21,6 +21,8 @@
   ; These methods should accept bulk data and process in batches.
   ;(kill-tasks [this task]
   ;(launch-tasks [this offers task-metadata-seq]
+  (ComputeCluster->compute-cluster-name [this])
```
pschorf: `compute-cluster-name [this]`

scheduler/src/cook/compute_cluster.clj
```diff
@@ -21,6 +21,8 @@
...
+
   (compute-cluster->db-id [this])
```
pschorf: This feels like a weird thing to have in the API. Is there any way to avoid this?

scheduler/src/cook/compute_cluster.clj
```diff
@@ -71,7 +75,7 @@
              unfiltered-db compute-cluster-name framework-id)]
     (first query-result)))
 
-(defn get-mesos-ComputeCluster
+(defn- get-mesos-ComputeCluster
```
pschorf: `get-meos-compute-cluster` feels more clojury

scheduler/src/cook/mesos/api.clj
```diff
@@ -962,6 +962,7 @@
     (compute-cluster-entity->map entity)
     (->> (cc/get-default-cluster-for-legacy)  ; Get the default cluster.
```
pschorf: Can we avoid going back to the DB for this?

scheduler/src/cook/mesos/scheduler.clj
```diff
@@ -675,9 +676,9 @@
 (defn- update-match-with-task-metadata-seq
   "Updates the match with an entry for the task metadata for all tasks. A 'match' is a set of jobs that we
   will want to all run on the same host."
-  [{:keys [tasks offer] :as match} db mesos-run-as-user]
-  (log/info "Match: " match)
-  (let [first-offer (-> offer first)
+  [{:keys [tasks leases] :as match} db mesos-run-as-user]
```
pschorf: Why did this change?

scheduler/src/cook/compute_cluster.clj
scrosby: Did this change.

scheduler/src/cook/mesos/rebalancer.clj
scrosby: Yup. Long fixed.

scheduler/src/cook/compute_cluster.clj
scrosby: Done

PR #1136, opened 2019-05-07
Thread compute cluster through Fenzo

scrosby:

- Thread the compute-cluster name through Fenzo from the offer, instead of getting it via the global.
- 
- 

This ensures that the compute cluster is attached to the offer, an enabling refactor for supporting more than one cluster. Now we know to whom to send the response.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -15,6 +15,7 @@
 ;;
 (ns cook.compute-cluster
   (:require [clojure.tools.logging :as log]
+            [cook.config :as config]
             [datomic.api :as d]))
 
 (defn- write-compute-cluster
@@ -62,12 +63,10 @@
      :mesos-framework-id framework-id
      :db-id (or cluster-entity-id (get-mesos-cluster-entity-id (d/db conn) mesos-cluster))}))
 
-(def default-cluster-name-for-legacy (atom nil))
-
 (defn get-default-cluster-name-for-legacy
   "What cluster name to put on for legacy jobs when generating their compute-cluster."
   []
-  @default-cluster-name-for-legacy)
+  (-> config/config :settings :mesos-compute-cluster-name))
 
 ; A hack to store the mesos cluster name, until we refactor the code so that we support multiple clusters. In the long term future
 ; this is probably replaced with a function from driver->cluster-id, or the cluster name is propagated by function arguments and
@@ -96,7 +95,6 @@
   occurs in config.clj. Long term, we need to fix config.clj to not to that, probably
   as part of global cook, at which time, this probably won't need to exist. Until then however....."
   [{:keys [mesos-compute-cluster-name mesos-framework-id]}]
-  (reset! default-cluster-name-for-legacy mesos-compute-cluster-name)
   (reset! mesos-cluster-name-hack mesos-compute-cluster-name)
   [{:compute-cluster-name mesos-compute-cluster-name :framework-id mesos-framework-id}])
 
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -62,6 +62,18 @@
      :mesos-framework-id framework-id
      :db-id (or cluster-entity-id (get-mesos-cluster-entity-id (d/db conn) mesos-cluster))}))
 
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -62,6 +62,18 @@
      :mesos-framework-id framework-id
      :db-id (or cluster-entity-id (get-mesos-cluster-entity-id (d/db conn) mesos-cluster))}))
 
+(def default-cluster-name-for-legacy (atom nil))
+
+(defn get-default-cluster-name-for-legacy
+  "What cluster name to put on for legacy jobs when generating their compute-cluster."
+  []
+  @default-cluster-name-for-legacy)
```
pschorf: Can we get rid of the atom and just have this read the config? This shouldn't change over the lifetime of the process.

scheduler/src/cook/compute_cluster.clj
```diff
@@ -62,6 +62,18 @@
...
+
+; A hack to store the mesos cluster name, until we refactor the code so that we support multiple clusters. In the long term future
+; this is probably replaced with a function from driver->cluster-id, or the cluster name is propagated by function arguments and
+; closed over.
+(def mesos-cluster-name-hack (atom nil))
```
pschorf: Why do we need this in addition to `get-default-cluster-name-for-legacy`?

scheduler/src/cook/mesos/api.clj
```diff
@@ -960,7 +960,7 @@
   [db entity]
   (if entity
     (compute-cluster-entity->map entity)
-    (->> (cc/cluster-name-hack)  ; Get the default cluster.
+    (->> (cc/get-default-cluster-name-for-legacy)  ; Get the default cluster.
```
pschorf: Can we just store the entire compute cluster in the variable instead of going out to the DB?

scheduler/src/cook/compute_cluster.clj
scrosby: mesos-cluster-name-hack is a hack that will go away when we restructure the initialization. get-default-cluster-for-legacy will exist forever in order to back-patch a cluster into legacy task structures.

scheduler/src/cook/compute_cluster.clj
scrosby: Done. I'll make a parallel change in the new patch. :)

scheduler/src/cook/mesos/api.clj
scrosby: Once the patch rolls out, after a week, we'll need to go out to the db every time (and it'll be cached, so that will be cheap), therefore this only optimizes the slow-case.

We still need to map the ComputeCluster to its corresponding datomic entity, because thats what this code expects. Do you want me to add a simple cache here? We can't cache the datomic entity, but we can cache the dictionary generated by compute-cluster-entity->map, and return it every time. Unless you see a nicer way, I don't think that that optimization is worth adding the extra atom and 2 lines of code. 
(when-not @cached-default-compute-cluster-for-legacy-as-a-response-map
    (reset! @cached.... (... the same code as above...))

scheduler/src/cook/compute_cluster.clj
pschorf: Ok, I buy that. Same comment as above, though: can we replace the atom with a string read from config?

scheduler/src/cook/mesos/api.clj
pschorf: If we need the entity, why not just have `get-default-cluster-for-legacy` that returns the entity (cached on startup) and then read the name out of that if needed? It lets us avoid all these gymnastics to get back to the entity.

scrosby: Isn't there a problem when you keep a datomic entity cached in memory; it keeps a lot of extra datomic state cached alive in memory. Thats why we don't do that anymore in Cook?

scheduler/src/cook/compute_cluster.clj
scrosby: Ok.

scheduler/src/cook/mesos/api.clj
pschorf: It could just be a map with the contents of the compute cluster instead of the entity itself to avoid that problem

PR #1140, opened 2019-05-13
Add COOK_MAP_CONTAINER_VOLUME

pschorf:
- Add a new environment variable, `COOK_MAP_CONTAINER_VOLUME` to disable the integration test volume created for the cook executor

If the executor is built into the docker image, you don't want to create the volume mapping.


shamsimam:
Awaiting green build.

PR #1146, opened 2019-05-16
Refactor driver into compute-cluster throughout cook

scrosby:

- Refactor driver propagation into the new compute-cluster.
- 
- 

Preparation for kubernetes support.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -18,6 +18,36 @@
             [cook.config :as config]
             [datomic.api :as d]))
 
+(defprotocol ComputeCluster
+  ; These methods should accept bulk data and process in batches.
+  ;(kill-tasks [this task]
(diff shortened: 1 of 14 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -18,6 +18,36 @@
             [cook.config :as config]
             [datomic.api :as d]))
 
+(defprotocol ComputeCluster
+  ; These methods should accept bulk data and process in batches.
+  ;(kill-tasks [this task]
+  ;(launch-tasks [this offers task-metadata-seq]
+  (compute-cluster-name [this])
```
scrosby: You wanted to know why we had this function in the last review. 

We could propagate the string, the object, or both. I chose to propagate the object, in which case, I'm not sure how I get the string unless I have an explicit function to fetch it.

pschorf: 👍

PR #1147, opened 2019-05-16
Don't set docker volume if executor path is relative

pschorf:
- Don't automatically add a docker volume if the executor path is relative

Mesos won't be able to create the container.


scrosby:
Waiting for green build.

scrosby:
Tried rerunning several times, still red build.

pschorf:
Looks like a bug when the executor is disabled, pushed a new commit with a fix.

pschorf:
The fix made it worse 😞 Running a new build now.

PR #1148, opened 2019-05-22
Outgoing/rename packages

scrosby:

- Rename and reorganize cook's code.
- Move most code out of cook.mesos
- Do some other reorgs while we're at it.


scheduler/src/cook/util2.clj
```diff
@@ -13,16 +13,16 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 ;;
-(ns cook.mesos.util
+(ns cook.util2
```
pschorf: This I don't like...can we roll it into `cook.util`? Or maybe `cook.scheduler.util`?

scrosby: After discussion we renamed this as cook.tools

PR #1151, opened 2019-05-29
Refactor launch task

scrosby:

- Refactor launch task so it occurs behind the ComputeCluster API.
- 
-

### DIFF
--- a/scheduler/src/cook/mesos/task.clj
+++ b/scheduler/src/cook/mesos/task.clj
@@ -63,7 +63,7 @@
            (cook-executor-candidate? job-ent))))
 
 ; TODO: This is mesos specific.
-(defn build-executor-environment
+(defn build-mesos-executor-environment
   "Build the environment for the cook executor."
   [job-ent]
   (let [{:keys [default-progress-regex-string environment log-level max-message-length progress-sample-interval-ms]}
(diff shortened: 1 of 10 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -63,7 +63,7 @@
            (cook-executor-candidate? job-ent))))
 
 ; TODO: This is mesos specific.
-(defn build-executor-environment
```
pschorf: This is probably not actually mesos specific, the kubernetes progress reporter will probably use the same environment variables for backwards compatibility.

scheduler/src/cook/mesos/task.clj
```diff
@@ -78,16 +78,42 @@
...
+  "Extract the executor key value from the job"
+  [db job-ent]
+  (let [container (util/job-ent->container db job-ent)
         ;; If the custom-executor attr isn't set, we default to using a custom
         ;; executor in order to support jobs submitted before we added this field
         custom-executor? (use-custom-executor? job-ent)
+        cook-executor? (use-cook-executor? job-ent)]
+        ;; executor-key configure whether this is a command or custom executor
+        (cond
+                       (and container cook-executor?) :container-cook-executor
```
pschorf: nitpick: indentation looks weird here.

scheduler/src/cook/mesos/task.clj
scrosby: Ok. Name reverted.

scheduler/src/cook/mesos/task.clj
scrosby: Good catch.

pschorf:
It's still marked wip: can you clear that and confirm that the internal build is green?

PR #1153, opened 2019-06-17
Add default container volumes

pschorf:
- Adds support for default container images

Supports adding default container volumes for all jobs.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2838,3 +2838,68 @@ class CookTest(util.CookTest):
                 util.wait_for_instance(self.cook_url, uuid, status='success')
         finally:
             util.kill_jobs(self.cook_url, job_uuids, assert_response=False)
+
+
+    @unittest.skipUnless(util.has_docker_service() and util.docker_tests_enabled(), "Requires `docker inspect`")
+    def test_default_container_volumes(self):
+        settings = util.settings(self.cook_url)
+        default_volumes = util.get_in(settings, 'container-defaults', 'volumes')
+        if default_volumes is None or len(default_volumes) == 0:
+            unittest.skip('Requires a default volume configured')
+        default_volume = default_volumes[0]
+        image = util.docker_image()
+        job_uuid, resp = util.submit_job(self.cook_url,
+                                         command='sleep 300',
+                                         container={'type': 'DOCKER',
+                                                    'docker': {'image': image}})
+        self.assertEqual(resp.status_code, 201, resp.content)
+        instance = util.wait_for_instance(self.cook_url, job_uuid, status='running')
+        self.logger.debug('instance: %s' % instance)
+        try:
+            # Get agent host/port
+            state = util.get_mesos_state(self.mesos_url)
+            agent = [agent for agent in state['slaves']
+                       if agent['hostname'] == instance['hostname']][0]
+
+            # Get container ID from agent
+            def agent_query():
+                return util.session.get(util.get_agent_endpoint(state, instance['hostname']))
+
+            def contains_executor_predicate(agent_response):
+                agent_state = agent_response.json()
+                executor = util.get_executor(agent_state, instance['executor_id'])
+                if executor is None:
+                    self.logger.warning(f"Could not find executor {instance['executor_id']} in agent state")
+                    self.logger.warning(f"agent_state: {agent_state}")
+                return executor is not None
+
+            agent_state = util.wait_until(agent_query, contains_executor_predicate).json()
+            executor = util.get_executor(agent_state, instance['executor_id'])
+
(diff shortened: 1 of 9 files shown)


scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
             (assoc "EXECUTOR_PROGRESS_OUTPUT_FILE_ENV" "EXECUTOR_PROGRESS_OUTPUT_FILE_NAME"
                    "EXECUTOR_PROGRESS_OUTPUT_FILE_NAME" progress-output-file))))
 
+(defn merge-container-defaults
+  "Takes a job container specification and applies any defaults from the config"
+  [container]
+  (when container
+    (let [{:keys [volumes]} (config/container-defaults)
+          get-path (fn [volume] (or (:container-path volume)
+                                    (:host-path volume)))]
```
shamsimam: Use destructuring?

scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
...
+      (cond-> container
+              volumes
+              (update :volumes (fn [container-volumes]
+                                 (let [volumes-to-add (filter (fn [default-volume]
+                                                              (not-any? (fn [container-volume]
```
shamsimam: Formatting seems off

scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
...
+                                                                          (.startsWith (get-path default-volume)
```
shamsimam: Is `get-path` guaranteed to return a non-nil value?

scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
...
+                                                                                       (get-path container-volume))) container-volumes))
```
shamsimam: Nitpick, would have preferred the second argument to the `filter` (i.e. `container-volumes`) in a new line as the filter function spans over multiple lines.`

scheduler/src/cook/tools.clj
```diff
@@ -180,7 +180,7 @@
 
 (defn job-ent->container
   "Take a job entity and return its container"
-  [db job-ent]
```
shamsimam: 👍

scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
             (assoc "EXECUTOR_PROGRESS_OUTPUT_FILE_ENV" "EXECUTOR_PROGRESS_OUTPUT_FILE_NAME"
                    "EXECUTOR_PROGRESS_OUTPUT_FILE_NAME" progress-output-file))))
 
+(defn merge-container-defaults
```
shamsimam: Should this be named `merge-configured-volumes`?

scheduler/src/cook/mesos/task.clj
```diff
@@ -77,10 +77,27 @@
...
+  "Takes a job container specification and applies any defaults from the config"
```
shamsimam: This function appears to only be merging the configured volumes?

scheduler/test/cook/test/mesos/task.clj
```diff
@@ -875,3 +876,46 @@
...
+                                             :volumes [{:host-path "/mnt/app"
+                                                        :mode "rw"}]})))
+      (is (= {:docker {:image "foo"}
+              :volumes [{:host-path "/diff/a"
+                         :container-path "/c/a"}
+                        {:host-path "/mnt/app/data"
+                         :mode "r"}]}
+             (task/merge-container-defaults {:docker {:image "foo"}
+                                             :volumes [{:host-path "/diff/a"
+                                                        :container-path "/c/a"}]}))))))
```
shamsimam: Can we add a case for where `container-path` is `/c`?

integration/tests/cook/test_basic.py
```diff
@@ -2838,3 +2838,68 @@ def test_cook_executor_reset_vars(self):
                 util.wait_for_instance(self.cook_url, uuid, status='success')
         finally:
             util.kill_jobs(self.cook_url, job_uuids, assert_response=False)
+
+
+    @unittest.skipUnless(util.has_docker_service() and util.docker_tests_enabled(), "Requires `docker inspect`")
+    def test_default_container_volumes(self):
```
shamsimam: Can we write this test to have the command read and verify output of files mounted from the defaults and the user's mounted volumes to verify that mounts work as expected?

scheduler/test/cook/test/mesos/task.clj
pschorf: `/c/a/b`? I already have a test above for where the default is inside the user-provided volume. Added the other case.

pschorf:
Pushed some integration test fixes.

PR #1154, opened 2019-06-19
Add python 3 classifier

pschorf:
- Add a python 3 classifier to the executor setup.py

The executor requires python3, so make this clear to potential users.

https://test.pypi.org/project/cook-executor/

### DIFF
--- a/executor/setup.py
+++ b/executor/setup.py
@@ -27,5 +27,8 @@ setup(
         'console_scripts': [
             'cook-executor = cook.__main__:main'
         ]
-    }
+    },
+    classifiers=[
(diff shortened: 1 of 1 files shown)


executor/setup.py
```diff
@@ -27,5 +27,8 @@
         'console_scripts': [
             'cook-executor = cook.__main__:main'
         ]
-    }
+    },
+    classifiers=[
+        "Programming Language :: Python :: 3"
```
shamsimam: Should we be more specific?
```suggestion
        "Programming Language :: Python :: 3.5"
```

PR #1161, opened 2019-06-25
Remove framework id from components

pschorf:
- Removes framework id from components
- Adds support for guessing framework id in responses when feasible

In kubernetes, there won't be a framework id.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -1076,6 +1076,7 @@
                    :cpus (:cpus resources)
                    :disable_mea_culpa_retries (:job/disable-mea-culpa-retries job false)
                    :env (util/job-ent->env job)
+                   ; TODO(pschorf): Remove field
                    :framework_id (guess-framework-id)
                    :gpus (int (:gpus resources 0))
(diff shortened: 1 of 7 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -1067,15 +1076,7 @@
...
-                   ;; The framework information is task-level information, not job-level information.
-                   ;; It should not be here.
-                   ;;
-                   ;; It is not used within the cook scheduler. We need to keep it here temporarily
-                   ;; because direct queries to the mesos agents need the framework-id. In particular, the CLI code and
-                   ;; sandbox-related integration tests are making those queries. They need the framework ID and are
-                   ;; getting it out of the job. Once we put the framework into the task structure and fix those
-                   ;; uses, we can get rid of it.
-                   :framework_id (cook.config/framework-id-config)
+                   :framework_id (guess-framework-id)
```
shamsimam: Nitpick: should we keep a TODO here to remove it when it is safe to do so?

pschorf: Done

PR #1167, opened 2019-07-10
Optimize the job fetching code to not round-trip to UUID.

scrosby:

- Propagate job entities instead of job UUID's when assembling /list API responses. This avoids finding job entities, mapping them to job UUID's, then looking up jobs from their UUID in the /list path.
- 
- 

This is a tech-debt/cleanup. Note that this isn't cleaning up the /jobs endpoint. That code would need more restructuring to fix (it propagates UUID's within the liberator context between the list and the explicit-uuid sub-endpoints).

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -1487,7 +1487,7 @@
 (histograms/defhistogram [cook-mesos api list-request-param-limit])
 (histograms/defhistogram [cook-mesos api list-response-job-count])
 
-(defn list-jobents
+(defn list-job-ents
   "Queries using the params from ctx and returns the jobs that were found as datomic entities."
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -1484,8 +1487,8 @@
 (histograms/defhistogram [cook-mesos api list-request-param-limit])
 (histograms/defhistogram [cook-mesos api list-response-job-count])
 
-(defn list-jobs
-  "Queries using the params from ctx and returns the job uuids that were found"
+(defn list-jobents
```
pschorf:
```suggestion
(defn list-job-ents
```

pschorf:
Still waiting on the build, internal-green, and PR description

PR #1168, opened 2019-07-12
Adds run-description to test metrics collection

dposada:

- adding the `run-description` field

To allow for a description of what the run is doing/testing.

### DIFF
--- a/integration/tests/conftest.py
+++ b/integration/tests/conftest.py
@@ -60,6 +60,7 @@ if 'TEST_METRICS_URL' in os.environ:
                 'host': socket.gethostname(),
                 'user': getpass.getuser(),
                 'run-id': os.getenv('TEST_METRICS_RUN_ID', None),
+                'run-description': os.getenv('TEST_METRICS_RUN_DESCRIPTION', 'open source integration tests'),
                 'build-id': os.getenv('TEST_METRICS_BUILD_ID', None),
                 'result': result,
                 'runtime-milliseconds': (end - start)*1000,


integration/tests/conftest.py
```diff
@@ -60,6 +60,7 @@ def record_test_metric(request):
                 'host': socket.gethostname(),
                 'user': getpass.getuser(),
                 'run-id': os.getenv('TEST_METRICS_RUN_ID', None),
+                'run-description': os.getenv('TEST_METRICS_RUN_DESCRIPTION', 'open source integration tests'),
```
nsinkov: why "open source integration tests" and not just "open source tests"

integration/tests/conftest.py
```diff
@@ -60,6 +60,7 @@ def record_test_metric(request):
                 'host': socket.gethostname(),
                 'user': getpass.getuser(),
                 'run-id': os.getenv('TEST_METRICS_RUN_ID', None),
```
nsinkov: getenv each time or save the values upfront?

integration/tests/conftest.py
dposada: Because this only covers the integration tests (the unit tests are in Clojure)

integration/tests/conftest.py
dposada: I don't feel strongly either way. I think it's fine to leave them as is, but I'm happy to change if you prefer doing it upfront.

dposada:
@nsinkov Thoughts on this?

PR #1169, opened 2019-07-17
Make Cook not fail silently on startup on some startup errors.

scrosby:

- Under some circumstances, Cook can fail to launch without logging anything. At least in this case, make it not fail silently.
- 
- 

- Failing silently with no error shown is irritating to debug.


pschorf:
* PR template
* internal-green
* build

PR #1170, opened 2019-07-17
Light refactor to move things into KubernetesComputeCluster

scrosby:

- These state atoms are compute-cluster specific. Move them within the compute cluster.
- 
- 

- These state atoms are compute-cluster specific, so can't be global.


pschorf:
* build
* pr template
* internal green

scrosby:
Need to rename the branch (because reasons), so need to respin this push request as #1171

PR #1176, opened 2019-07-24
Add maximum command length configuration parameter

pschorf:
- Adds a new configuration parameter to configure maximum command length

Prevents users from submitting arbitrarily long commands.


scheduler/config.edn
```diff
@@ -80,7 +80,8 @@
  :sandbox-syncer {:sync-interval-ms 1000}
  :scheduler {:fenzo-fitness-calculator "cook.scheduler.data-locality/make-data-local-fitness-calculator"
              :offer-incubate-ms 15000
-             :task-constraints {:cpus 10
+             :task-constraints {:command-length-limit 5000
```
shamsimam: 👍 (lexicographic)

shamsimam:
Awaiting green build

PR #1177, opened 2019-07-30
Write Kubernetes state back to Datomic

pschorf:
- Refactor `handle-status-update` to remove sandbox syncing and lingering task killer
- Invoke handle status update from Kubernetes controller

We need to write Kubernetes state back to Datomic.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -176,7 +176,6 @@
     (.addContainersItem pod-spec container)
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
-    (.setTerminationGracePeriodSeconds pod-spec 1) ; TODO(pschorf): make configurable/remove
 
     ; pod
     (.setMetadata pod metadata)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -176,6 +176,7 @@
     (.addContainersItem pod-spec container)
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
+    (.setTerminationGracePeriodSeconds pod-spec 1)
 
     ; pod
     (.setMetadata pod metadata)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -176,7 +176,7 @@
     (.addContainersItem pod-spec container)
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
-    (.setTerminationGracePeriodSeconds pod-spec 1)
+    (.setTerminationGracePeriodSeconds pod-spec 1) ; TODO(pschorf): make configurable/remove
 
     ; pod
     (.setMetadata pod metadata)
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -44,60 +44,20 @@
             :reject-after-match-attempt true})
          node-name->available)))
 
-
-(defn synthesized-pod-state->task-status
-  [pod {:keys [state reason]}]
-  (let [task-state (case state
-                     :pod/waiting :task-staging
-                     :pod/running :task-running
-                     :pod/succeeded :task-finished
-                     :pod/failed :task-failed)
-        task-id (-> pod .getMetadata .getName)]
-
-    {:task-id {:value task-id}
-     :reason reason ; TODO(pschorf): Map as best as possible to mesos reasons
-     :state task-state}))
-
 (defn make-pod-watch-callback
-  [kcc handle-status-update handle-exit-code]
+  [kcc]
   (fn pod-watch-callback
     [prev-pod pod]
     (try
-      (let [prev-state (api/pod->synthesized-pod-state prev-pod)
-            state (api/pod->synthesized-pod-state pod)
-            task-id (-> (or pod prev-pod)
-                        .getMetadata .getName)
-            namespace (-> (or pod prev-pod)
-                          .getMetadata
-                          .getNamespace)]
-        (if (nil? pod)
-          (controller/pod-deleted kcc prev-pod)
-          (controller/pod-update kcc pod))
-
-        ;(if (= namespace "cook")
-        ;  (when (not (= prev-state state))
-        ;    (log/debug "Updating state for" task-id "from" prev-state "to" state)
-        ;    (case (:state state)
-        ;      :pod/running
-        ;      (handle-status-update (synthesized-pod-state->task-status pod state))
-        ;      :pod/waiting
-        ;      (handle-status-update (synthesized-pod-state->task-status pod state))
-        ;      :pod/succeeded
-        ;      (do
-        ;        (handle-status-update (synthesized-pod-state->task-status pod state))
-        ;        (handle-exit-code task-id (:exit state)))
-        ;      :pod/failed
-        ;      (do
-        ;        (handle-status-update (synthesized-pod-state->task-status pod state))
-        ;        (handle-exit-code task-id (:exit state)))
-        ;      :pod/unknown
-        ;      (log/error "Unable to determine pod status in callback")))
-        ;  (log/debug "Skipping state update for" task-id))
-        )
+      (if (nil? pod)
+        (controller/pod-deleted kcc prev-pod)
+        (controller/pod-update kcc pod))
       (catch Exception e
         (log/error e "Error processing status update")))))
 
-(defrecord KubernetesComputeCluster [^ApiClient api-client name entity-id match-trigger-chan exit-code-syncer-state current-pods-atom current-nodes-atom expected-state-map existing-state-map]
+(defrecord KubernetesComputeCluster [^ApiClient api-client name entity-id match-trigger-chan exit-code-syncer-state
+                                     current-pods-atom current-nodes-atom expected-state-map existing-state-map
+                                     pool->fenzo-atom]
   cc/ComputeCluster
   (launch-tasks [this offers task-metadata-seq]
     (let [api (CoreV1Api. api-client)]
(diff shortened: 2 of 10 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -176,6 +176,7 @@
     (.addContainersItem pod-spec container)
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
+    (.setTerminationGracePeriodSeconds pod-spec 1)
```
scrosby: Why is this set, and why is it set so low? (This also feels like maybe it should be a config var)

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -142,7 +94,7 @@
                         (let [nodes @current-nodes-atom
                               pods @current-pods-atom
                               offers (generate-offers nodes pods this)
-                              pool (config/default-pool)
+                              pool (or (config/default-pool) "no-pool")
```
scrosby: This seems like a layering violation. The pool should be defaulted at a higher level. Actually, it should be passed as an argument to the compute cluster factory method. (One pool can have multiple compute clusters.)

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -48,19 +51,42 @@
     (swap! map-atom dissoc key)
     (swap! map-atom assoc key value)))
 
-;; REVIEW: Review handle-status-update
+(defn container-status->failure-reason
+  [^V1PodStatus pod-status ^V1ContainerStatus status]
```
scrosby: Doc comment.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -48,19 +51,42 @@
...
+  ; TODO map kubernetes failure reasons
+  (let [terminated (-> status .getState .getTerminated)]
+    (cond
+      (= "OutOfMemory" (.getReason pod-status)) :reason-container-limitation-memory
+      (= "Error" (.getReason terminated)) :reason-command-executor-failed
+      (= "OOMKilled" (.getReason terminated)) :reason-container-limitation-memory
+      :default :unknown)))
```
scrosby: Log here, then we can grep the log for weird states and add them into the case machine.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -48,19 +51,42 @@
     (swap! map-atom dissoc key)
     (swap! map-atom assoc key value)))
 
-;; REVIEW: Review handle-status-update
+(defn container-status->failure-reason
+  [^V1PodStatus pod-status ^V1ContainerStatus status]
+  ; TODO map kubernetes failure reasons
+  (let [terminated (-> status .getState .getTerminated)]
+    (cond
```
scrosby: Use 'case'?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -48,19 +51,42 @@
...
-    (do
-      ; TODO: Extract failure reason, etc and store in datomic, via refactored handle-status-update.
-      {:expected-state :expected/completed})
-    :pod/succeeded
-    (do
-      ; TODO: Mark job as success in datomic, via refactored handle-status-update.
-      {:expected-state :expected/completed})))
+  [kcc {:keys [synthesized-state pod] :as existing-state-dictionary}]
+  (let [task-id (-> pod .getMetadata .getName)
+        ^V1ContainerStatus job-container-status (->> pod
```
scrosby: This smells like a helper function. Which also helps the indenting here.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -69,12 +95,24 @@
...
+                :state :task-running}]
+    (handle-status-update kcc status)
+    {:expected-state :expected/running}))
+
+(defn pod-was-killed
+  [kcc {:keys [pod] :as existing-state-dictionary}]
+  (let [task-id (-> pod .getMetadata .getName)
+        status {:task-id {:value task-id}
+                :state :task-failed
+                :reason :reason-command-executor-failed}]
```
scrosby: Is this the right reason?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -90,32 +128,42 @@
...
                                  ; TODO: Implement :pod/unknown cases.
                                  ; TODO: Implement :pod/pending cases.
                                  ; TODO: Implement :pod/need-to-fail cases.
                                  [:missing :pod/running] (kill-task api-client expected-state-dict pod)
                                  [:missing :pod/pending] (kill-task api-client expected-state-dict pod)
                                  [:missing :pod/need-to-fail] (kill-task api-client expected-state-dict pod)
                                  [:missing :pod/succeeded] (remove-finalization-if-set-and-delete api-client expected-state-dict pod)
-                                 [:missing :pod/failed] (remove-finalization-if-set-and-delete api-client expected-state-dict pod))]
+                                 [:missing :pod/failed] (remove-finalization-if-set-and-delete api-client expected-state-dict pod)
+                                 [:missing :missing] nil
```
scrosby: Put a comment that this state can be seen because of the recur at the end.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -26,6 +26,7 @@
             [cook.plugins.pool :as pool-plugin]
             [cook.progress :as progress]
             [cook.scheduler.scheduler :as sched]
+            [cook.tools :as util]
```
scrosby: Import this as tools, not util. Should be easy search&replace.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -36,6 +37,39 @@
 (meters/defmeter [cook-mesos scheduler handle-status-update-rate])
 
 
+(defn conditionally-sync-sandbox
+  [conn task-id task-state sync-agent-sandboxes-fn]
```
scrosby: Doc comment.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -36,6 +37,39 @@
...
+  (let [instance-ent (d/entity (d/db conn) [:instance/task-id task-id])]
+    (when (and (#{:task-starting :task-running} task-state)
+               (not= :executor/cook (:instance/executor instance-ent)))
+      ;; cook executor tasks should automatically get sandbox directory updates
```
scrosby: Is there a cook executor for kubernetes??? This comment seems off.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -36,6 +37,39 @@
...
+      (sync-agent-sandboxes-fn (:instance/hostname instance-ent) task-id))))
+
+(defn handle-status-update
```
scrosby: I know for historic reasons, we called this handle-status-update, but maybe we should rename it to 'write-status-update-to-datomic' for consistency with what it actually does?

scrosby: Also, is this copy&paste code from elsewhere? E.g., the predicate at line 55 inside the 'if' has the feel of being copied. Aaah, this was moved from sched/handle-status-update. But if so, why was it removed from the mesos path? Was/is it dead code for mesos?

scheduler/src/cook/kubernetes/api.clj
pschorf: I just added this to speed up the teardown of the pods in k8s for testing. I'll put a TODO on it to revisit later, but for now I think it only helps us.

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: This is deleted in a subsequent PR, can we just wait until then?

scheduler/src/cook/kubernetes/controller.clj
pschorf: Done

scheduler/src/cook/kubernetes/controller.clj
pschorf: Sometimes you want to look at the pod-status instead of the terminated state, so case won't work.

scheduler/src/cook/kubernetes/controller.clj
pschorf: Done

scheduler/src/cook/kubernetes/controller.clj
pschorf: Done

scheduler/src/cook/kubernetes/controller.clj
pschorf: Currently, we just get these back via the normal failed job flow once the command gets killed. We could do something more specific, but I think it's better to be consistent.

scheduler/src/cook/kubernetes/controller.clj
pschorf: Done

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: Done

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: Done

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: This is for mesos.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: This is in the mesos path.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Works for me.

scheduler/src/cook/kubernetes/controller.clj

scheduler/src/cook/kubernetes/controller.clj
scrosby: I don't like, but OK.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
scrosby: Yeah. I was confused.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
scrosby: Yeah. I was confused.

PR #1181, opened 2019-08-05
Initialize kubernetes expected state from datomic

scrosby:

- Initialize the kubernetes expected state from the datomic-stored instance state
- Its been minimally tested by launching a job, killing cook, and then resuming cook.
- Several todo's have been added.

So that cook will reinitialize and resume from a shutdown.

### DIFF
--- a/scheduler/config-k8s.edn
+++ b/scheduler/config-k8s.edn
@@ -46,7 +46,6 @@
            :user-metrics-interval-seconds 60}
  :nrepl {:enabled? true
          :port #config/env-int "COOK_NREPL_PORT"}
- :pools {:default "gamma"}
  :port #config/env-int "COOK_PORT"
  :ssl {:port #config/env-int "COOK_SSL_PORT"
        :keystore-path #config/env "COOK_KEYSTORE_PATH"
--- a/scheduler/config-k8s.edn
+++ b/scheduler/config-k8s.edn
@@ -46,6 +46,7 @@
            :user-metrics-interval-seconds 60}
  :nrepl {:enabled? true
          :port #config/env-int "COOK_NREPL_PORT"}
+ :pools {:default "gamma"}
  :port #config/env-int "COOK_PORT"
  :ssl {:port #config/env-int "COOK_SSL_PORT"
        :keystore-path #config/env "COOK_KEYSTORE_PATH"
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -62,8 +62,9 @@
        (into {})))
 
 (defn task-ent->expected-state
-  [state]
-  (case (:instance/status state)
+  [task-ent]
+  
+  (case (:instance/status task-ent)
     :instance.status/unknown {:expected-state :expected/starting}
     :instance.status/running {:expected-state :expected/running}
     :instance.status/failed {:expected-state :expected/completed}
@@ -74,7 +75,7 @@
   as all pods in kubernetes. We're given an already existing list of all running tasks entities (via (->> (cook.tools/get-running-task-ents)."
   [conn compute-cluster-name running-tasks-ents current-pods-atom]
   (let [db (d/db conn)
-        all-tasks-ids-in-pods (keys @current-pods-atom)
+        all-tasks-ids-in-pods (->> @current-pods-atom keys (into #{}))
         _ (log/info "All tasks in pods: " all-tasks-ids-in-pods)
         running-tasks-in-cc-ents (filter
                                    #(-> % cook.task/task-entity->compute-cluster-name (= compute-cluster-name))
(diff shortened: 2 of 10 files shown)


scheduler/config-k8s.edn
```diff
@@ -46,6 +46,7 @@
            :user-metrics-interval-seconds 60}
  :nrepl {:enabled? true
          :port #config/env-int "COOK_NREPL_PORT"}
+ :pools {:default "gamma"}
```
scrosby: After this change was made, I had problems bringing up cook, so I reverted it. 

Any idea whats going on?

pschorf: Ah, I've been running without pools in the database locally for k8s testing. If you keep it out and skip the seed pools step, it should work.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -55,6 +55,44 @@
...
+  "We need to determine everything we should be tracking when we construct the expected state. We should be tracking all tasks that are in the running state as well
+  as all pods in kubernetes. We're given an already existing list of all running tasks entities (via (->> (cook.tools/get-running-task-ents)."
+  [conn compute-cluster-name running-tasks-ents current-pods-atom]
+  (let [db (d/db conn)
+        all-tasks-ids-in-pods (->> @current-pods-atom keys (into #{}))
+        _ (log/info "All tasks in pods: " all-tasks-ids-in-pods)
+        running-tasks-in-cc-ents (filter
+                                   #(-> % cook.task/task-entity->compute-cluster-name (= compute-cluster-name))
+                                   running-tasks-ents)
+        _ (log/info "Running tasks in compute cluster in datomic: " running-tasks-in-cc-ents)
```
pschorf: Just log the task id here, not the entity. Also, should probably be debug.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -55,6 +55,44 @@
...
+        cc-running-tasks-map (task-ents->map-by-task-id running-tasks-in-cc-ents)
```
pschorf:
```suggestion
        running-task-id->task (task-ents->map-by-task-id running-tasks-in-cc-ents)
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -55,6 +55,44 @@
...
+        cc-running-tasks-ids (->> cc-running-tasks-map keys (into #{}))
+        extra-tasks-map (->> (set/difference all-tasks-ids-in-pods cc-running-tasks-ids)
+                             (map (fn [task-id] [task-id (cook.tools/retrieve-instance db task-id)]))
+                             (filter (fn [[_ task-ent]] (some? task-ent)))
```
pschorf:
```suggestion
                             (filter (fn [[_ task-ent]] (not (nil? task-ent))))
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -55,6 +55,44 @@
...
+                             (into {}))
+        all-tasks-ents-map (set/union extra-tasks-map cc-running-tasks-map)]
```
pschorf: Does `set/union` really do the right thing here? Aren't these both maps? In any case, I think `merge` would be clearer.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -55,6 +55,44 @@
...
+    (doseq [[k v] all-tasks-ents-map]
+      (log/info "Setting expected state for " k " ---> " (task-ent->expected-state v)))
```
pschorf: This feels like a `debug` log. Can we do one info level log at the end with the count or something like that?

scheduler/config-k8s.edn
scrosby: Aaah. Nice.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: https://clojuredocs.org/clojure.core/some_q

Seems better.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Yup. THanks.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Sounds good.

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: I always mix this up with `any?`. Sorry for the mix up.

pschorf:
The last internal build failed, can you take a look? Once it's internal-green I will merge.

PR #1183, opened 2019-08-07
Add support for google credentials

pschorf:
- Add support for google service account credentials for authenticating with GKE

Allows using GKE for managing Kubernetes clusters.


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -290,25 +291,26 @@
...
-    ; failed-to-start pod on a server restart. Thus, if we create a task, store into datomic, but then the cook scheduler
-    ; fails --- before kubernetes creates a pod (either the message isn't sent, or there's a kubernetes problem), we will
-    ; the inability to create a new V1Pod and we can't retry this at the kubernetes level.
-    ;
-    ; Eventually, the stuck pod detector will recognize the stuck pod, kill the task, and a cook-level retry will make
-    ; a new task.
-    ;
-    ; Because the issue is relatively rare and auto-recoverable, we're going to punt on the task-metadata-seq refactor.
-    (log/warn "Unimplemented Operation to launch a pod for becuause we do not reconstruct on startup.")))
+  (let [namespace "cook"]
```
scrosby: Worth a todo (and an item in the punchlist) that the namespace needs to be dynamc.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -1,23 +1,23 @@
...
+  (:import (com.google.auth.oauth2 GoogleCredentials)
+           (io.kubernetes.client ApiClient)
            (io.kubernetes.client.apis CoreV1Api)
-           (io.kubernetes.client.models V1PodStatus V1ContainerState)
            (io.kubernetes.client.util Config)
-           (java.util UUID)))
+           (java.io FileInputStream File)
+           (java.util UUID)
+           (java.util.concurrent Executors ScheduledExecutorService TimeUnit)
+           (com.squareup.okhttp.logging HttpLoggingInterceptor HttpLoggingInterceptor$Level)
```
scrosby: lexicographical order.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -172,12 +172,61 @@
       (cc/write-compute-cluster conn {:compute-cluster/type :compute-cluster.type/kubernetes
                                       :compute-cluster/cluster-name compute-cluster-name}))))
 
+(defn get-bearer-token
```
scrosby: API docs (here and the other functions). Perhaps a quick link to text why this is necessary, and/or a bit of expository text.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -172,12 +172,61 @@
...
+
+(defn make-bearer-token-refresh-task
+  [api-client scoped-credentials]
+  (reify
+    Runnable
+    (run [_]
+      (try
+        (let [bearer-token (get-bearer-token scoped-credentials)]
+          (.setApiKey api-client bearer-token))
+        (catch Exception ex
```
scrosby: Should this retry, or should it suicide cook? If we don't refresh bearer token, in an hour or two, it'll stop functioning. But the critical diagnostic error message will be hours earlier. I want the error message to be fairly close to the end of the log. (So, e.g., setting a flag here and printing it around the authentication failure would also be acceptable.)

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -172,12 +172,61 @@
...
+          (log/error ex "Error refreshing bearer token"))))))
+
+(defn make-api-client
+  [^String config-file base-path ^String google-credentials bearer-token-refresh-seconds verifying-ssl]
```
scrosby: What config file? Would be super useful to document here, (or better yet, the readme) how to get this all working.

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Actually a lot of these were unused from debugging, removed.

scheduler/src/cook/kubernetes/compute_cluster.clj

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Discussed offline

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Added a docstring here, and a readme.

PR #1185, opened 2019-08-12
Added some deftests and other small cleanups.

scrosby:

- Added some deftest stubs for us to fill in.
- 
- 

Indicate the deftests we want to implement as TODO's.


pschorf:
Just waiting on the build

PR #1186, opened 2019-08-13
Merge Kubernetes integration branch to master

scrosby:

- In progress code to add kubernetes support cook.

This code can successfully launch a job on GKE and pass integration tests. It is still a work in in progress and kubernetes support is under active development.

We want to support kubernetes.


scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+## GKE cluster config
+For GKE clusters, OAuth is used to authenticate with the cluster. To connect to a GKE cluster, you need two things:
+- Google JSON credential file: https://cloud.google.com/iam/docs/creating-managing-service-account-keys
+- Cluster master url (this will be the api client base URL)
+Example:
+```clojure
+:compute-clusters [{:factory-fn cook.kubernetes.compute-cluster/factory-fn
+                    :config {:base-url "http://127.0.0.1:8000"
+                             ;; Location of credential file
+                             :google-credentials "/home/myuser/creds.json"}}]
```
dposada: We should `:compute-cluster-name` to this example too

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+# Running with kubernetes has several differences:
+- Use config-k8s.edn instead of config.edn
+- You need to set a path to the kube config with COOK_K8S_CONFIG_FILE
+
+These are done for you in bin/run-kubernetes-local.sh
+
+
+You'll need to setup several vitual environments to use run-local.sh and run-kubernetes-local.sh.
```
dposada:
```suggestion
You'll need to setup several virtual environments to use run-local.sh and run-kubernetes-local.sh.
```

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+* One for building the CLI.
+* One for running run-kubernetes-local.sh
+* One for integration tests
```
dposada: Are these bullets supposed to have instructions? I'm not sure what they're telling me.

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Setup. Detailed command lines.
+
+- Set up minikube
```
dposada: How?

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Run local tidbits
```
dposada: Why `tidbits`?

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+
+scheduler/bin/start-datomic.sh
+
+- Seed pools
+
+lein exec -p scheduler/datomic/data/seed_pools.clj datomic:free://localhost:4334/jobs
+
+- Make namespace
+
+kubectl create -f docs/make-kubernetes-namespace.json
```
dposada: Do I need to run this every time?

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Run local. Its OK to free and restart.
```
dposada:
```suggestion
- Run local. It's OK to free and restart.
```

dposada: What does "to free" mean?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+(defn convert-resource-map
+  [m]
+  {:mem (if (get m "memory")
+          (-> m (get "memory") to-double (/ (* 1024 1024)))
+          0.0)
+   :cpus (if (get m "cpu")
+           (-> m (get "cpu") to-double)
+           0.0)})
+
+(defn get-capacity
```
scrosby: clojuredoc. Also might be worth giving the shape of the input dictionary.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+  [node-name->node]
+  (pc/map-vals (fn [^V1Node node]
+                 (-> node .getStatus .getCapacity convert-resource-map))
+               node-name->node))
+
+(defn get-consumption
```
scrosby: clojuredoc. Also might be worth giving the shape of the input&output dictionary. Also probably a punchlist item to figure out if requests and limit behave as we expect for accounting as far as launching.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+                                                         (apply merge-with + container-requests))))
+                                                (apply merge-with +)))
+                                         node-name->pods)]
+    node-name->requests))
+
+(def cook-container-name-for-job
+  "required-cook-job-container")
+
+(defn task-metadata->pod
+  [{:keys [task-id command container task-request hostname]}]
```
scrosby: clojuredoc.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+    ; pod
+    (.setMetadata pod metadata)
+    (.setSpec pod pod-spec)
+
+    pod))
+;;
+;;  This file contains accessor functions for all kubernetes API. That way we can conveniently mock these functions in clojure.
+;;
+
+(defn TODO
```
scrosby: This dead now?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+  (throw (UnsupportedOperationException. "TODO")))
+
+(defn V1Pod->name
+  "Extract the name of a pod from the pod itself"
+  [^V1Pod pod]
+  (-> pod .getMetadata .getName))
+
+
+(defn pod->synthesized-pod-state
+  [^V1Pod pod]
```
scrosby: clojuredoc

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+      ; Because of the complicated nature of task-metadata-seq --- we can't easily run the V1Pod creation code for a
+      ; failed-to-start pod on a server restart. Thus, if we create a task, store into datomic, but then the cook scheduler
+      ; fails --- before kubernetes creates a pod (either the message isn't sent, or there's a kubernetes problem), we will
+      ; the inability to create a new V1Pod and we can't retry this at the kubernetes level.
+      ;
+      ; Eventually, the stuck pod detector will recognize the stuck pod, kill the task, and a cook-level retry will make
+      ; a new task.
+      ;
+      ; Because the issue is relatively rare and auto-recoverable, we're going to punt on the task-metadata-seq refactor.
+      (log/warn "Unimplemented Operation to launch a pod for becuause we do not reconstruct on startup."))))
```
scrosby: Typo

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+    (cond
+      (= "OutOfMemory" (.getReason pod-status)) :reason-container-limitation-memory
+      (= "Error" (.getReason terminated)) :reason-command-executor-failed
+      (= "OOMKilled" (.getReason terminated)) :reason-container-limitation-memory
+      :default (do
+                 (log/warn "Unable to determine kubernetes state for pod" {:pod-status pod-status
+                                                                           :container-status status})
+                 :unknown))))
+
+(defn handle-status-update
```
pschorf: we should rename this to match `write-status-to-datomic`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -1,147 +1,26 @@
...
-    node-name->requests))
+           (java.util.concurrent Executors ScheduledExecutorService TimeUnit)))
 
 (defn generate-offers
   [node-name->node pod-name->pod compute-cluster]
-  (let [node-name->capacity (get-capacity node-name->node)
-        node-name->consumed (get-consumption pod-name->pod)
+  (let [node-name->capacity (api/get-capacity node-name->node)
+        node-name->consumed (api/get-consumption pod-name->pod)
         ; TODO(pschorf): Should also include recently launched jobs
```
scrosby: This in punchlist?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -163,192 +42,120 @@
...
+(defn task-ent->expected-state
+  [task-ent]
+  
+  (case (:instance/status task-ent)
+    :instance.status/unknown {:expected-state :expected/starting}
+    :instance.status/running {:expected-state :expected/running}
+    :instance.status/failed {:expected-state :expected/completed}
+    :instance.status/success {:expected-state :expected/completed}))
+
+(defn determine-expected-state
```
scrosby: Rename this determine-expected-state-on-startup

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -163,192 +42,120 @@
...
-(defrecord KubernetesComputeCluster [^ApiClient api-client name entity-id match-trigger-chan exit-code-syncer-state current-pods-atom current-nodes-atom]
+(defn task-ents->map-by-task-id
+  [task-ents]
+  (->> task-ents
+       (map (fn [task-ent] [(str (:instance/task-id task-ent)) task-ent]))
+       (into {})))
+
+(defn task-ent->expected-state
+  [task-ent]
+
```
scrosby: Fix the newline. Clojuredoc it too.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+          prev-item (get @state-atom (key-fn item))]
+      (case (.-type update)
+        "ADDED" (swap! state-atom (fn [m] (assoc m (key-fn item) item)))
+        "MODIFIED" (swap! state-atom (fn [m] (assoc m (key-fn item) item)))
+        "DELETED" (swap! state-atom (fn [m] (dissoc m (key-fn item)))))
+      (when callback
+        (try
+          (callback prev-item item)
+          (catch Exception e
+            (log/error e "Error while processing callback")))))))
```
dposada: Should we also log `prev-item` and `item` here?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -163,192 +42,120 @@
...
-                                                                 this
-                                                                 pool->fenzo
-                                                                 (constantly nil) ; no sandboxes to sync
-                                                                 status))
-          pod-callback (make-pod-watch-callback handle-status-update handle-exit-code)]
-      (initialize-pod-watch api-client current-pods-atom pod-callback))
-    (initialize-node-watch api-client current-nodes-atom)
+          pod-callback (make-pod-watch-callback this)]
+      (api/initialize-pod-watch api-client current-pods-atom pod-callback)
+      ; We require that initialize-pod-watch sets current-pods-atom before completing.
```
scrosby: Fix comment. ""Before executing the next line.""

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+
+(defn initialize-pod-watch
+  "Initialize the pod watch. We require that this function set the current-pods-atom before it finishes so that
+  cook.kubernetes.compute-cluster/initialize-cluster works correctly."
```
dposada: This docstring is not clear to me - I would vote to either clarify or remove

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+                                 [:expected/running :pod/running] expected-state-dict
+                                 [:expected/running :pod/succeeded] (pod-has-just-completed kcc existing-state-dict)
+                                 [:expected/completed :pod/succeeded] (remove-finalization-if-set-and-delete api-client expected-state-dict pod)
+                                 [:expected/running :pod/failed] (pod-has-just-completed kcc existing-state-dict)
+                                 [:expected/completed :pod/failed] (remove-finalization-if-set-and-delete api-client expected-state-dict pod)
+                                 [:expected/completed :missing] nil ; Cause it to be deleted.
+                                 [:expected/killed :pod/waiting] (kill-task api-client expected-state-dict pod)
+                                 [:expected/killed :pod/running] (kill-task api-client expected-state-dict pod) ; TODO: Where does the datomic update occur? Do we do it when we do [expected/killed :pod/failed], to be similar to mesos, we update it only when the backend says its dead?
```
pschorf: I think we can remove this TODO

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -163,192 +42,120 @@
...
-                        (let [nodes @current-nodes-atom
-                              pods @current-pods-atom
-                              offers (generate-offers nodes pods this)
-                              pool (config/default-pool)
-                              chan (pool->offers-chan pool)] ; TODO(pschorf): Support pools
-                          (log/info "Processing offers:" offers)
-                          (scheduler/receive-offers chan match-trigger-chan this pool offers))
-                        (catch Exception e
-                          (log/error e "Exception while forwarding offers")))))
     ; TODO(pschorf): Deliver when leadership lost
```
scrosby: This still a TODO?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+
+                                 [:expected/killed :pod/succeeded] (do
```
pschorf: We could change the `do` to just `pod-has-just-completed` and fall back into the normal path.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+                                                                     (pod-has-just-completed kcc existing-state-dict)
+                                                                     (remove-finalization-if-set-and-delete api-client expected-state-dict pod))
+                                 [:expected/killed :pod/failed]
+                                 (do
```
pschorf: Ditto here

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+                                               nil ; watch
+                                               )
+        pod-name->pod (pc/map-from-vals (fn [^V1Pod pod]
+                                          (-> pod
+                                              .getMetadata
+                                              .getName))
+                                        (.getItems current-pods))]
+    (log/info "Updating current-pods-atom with pods" (keys pod-name->pod))
+    (log/error "TODO: We should have two atoms, one for cook pods (that feeds into the controller) and one for
+    all pods (that feeds into machine utilization then into offer generation)")
```
dposada: Seems like this should be a comment rather than an ERROR log

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -363,14 +170,76 @@
...
+           trigger-chans]}]
   (let [conn cook.datomic/conn
         cluster-entity-id (get-or-create-cluster-entity-id conn compute-cluster-name)
-        api-client (Config/fromConfig config-file)
+        api-client (make-api-client config-file base-path google-credentials bearer-token-refresh-seconds verifying-ssl)
         compute-cluster (->KubernetesComputeCluster api-client compute-cluster-name cluster-entity-id
                                                     (:match-trigger-chan trigger-chans)
-                                                    exit-code-syncer-state (atom {}) (atom {}))]
+                                                    exit-code-syncer-state (atom {}) (atom {})
+                                                    (atom {:type :expected-state-map})
```
scrosby: Comment that these initial keys are only to make it easier to trace provenance when debugging.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -363,14 +170,76 @@
...
 (defn factory-fn
-  [{:keys [compute-cluster-name ^String config-file]} {:keys [exit-code-syncer-state
-                                                              trigger-chans]}]
+  [{:keys [compute-cluster-name
+           ^String config-file
+           base-path
+           google-credentials
+           verifying-ssl
+           bearer-token-refresh-seconds]
+    :or {bearer-token-refresh-seconds 300}}
```
scrosby: Don't we normally default config values in cook.config?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+                 :unknown))))
+
+(defn handle-status-update
+  "Helper function for calling scheduler/handle-status-update"
+  [kcc mesos-status]
+  (scheduler/write-status-to-datomic datomic/conn
+                                     @(:pool->fenzo-atom kcc)
+                                     mesos-status))
+
+(defn- get-job-container-status
```
scrosby: Move this into API, and rename 'get-pod-status-for-cook-container-in-pod'. Also, why does this exist? Shouldn't we be using the synthesized-state instead?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+    (reset! current-pods-atom pod-name->pod)
+    (let [watch (WatchHelper/createPodWatch api-client (-> current-pods
+                                                           .getMetadata
+                                                           .getResourceVersion))]
+      (.submit kubernetes-executor ^Callable
+      (fn []
+        (try
+          (handle-watch-updates current-pods-atom watch (fn [p] (-> p .getMetadata .getName)) pod-callback)
+          (catch Exception e
+            (log/error "Error during watch: " (-> e .toString))
```
```suggestion
            (log/error e "Error during watch")
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+  "Extract the constainer status for the api/cook-container-name-for-job container"
+  [^V1PodStatus pod-status]
+  (->> pod-status
+       .getContainerStatuses
+       (filter (fn [container-status] (= api/cook-container-name-for-job
+                                         (.getName container-status))))
+       first))
+
+(defn pod-has-just-completed
+  "A pod has completed."
```
scrosby: Update comment. "So now we need to update the status in datomic and store the exit code."

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+                :state mesos-state
+                :reason (container-status->failure-reason pod-status job-container-status)}
+        exit-code (-> job-container-status .getState .getTerminated .getExitCode)]
+    (handle-status-update kcc status)
+    (sandbox/aggregate-exit-code (:exit-code-syncer-state kcc) task-id exit-code)
+    {:expected-state :expected/completed}))
+
+(defn prepare-expected-state-dict-for-logging
+  [expected-state-dict]
+  ".toString on a pod is incredibly large. Make a version thats been elided."
```
scrosby: Flip this with prior line.

scheduler/src/cook/kubernetes/api.clj
dposada: Should this be `warn` instead of `error`?

scheduler/src/cook/kubernetes/controller.clj
scrosby: Is this TODO now dead with the handle-status-update refactor?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+            (.close watch)
```
dposada: Should this be in a `finally`?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+
+(defn pod-update
+  "Update the existing state for a pod. Include some business logic to e.g., not change a state to the same value more than once.
+  Invoked by callbacks from kubernetes."
+  [{:keys [existing-state-map] :as kcc} ^V1Pod new-pod]
+  (let [pod-name (api/V1Pod->name new-pod)]
+    (locking (calculate-lock pod-name)
+      (let [new-state {:pod new-pod :synthesized-state (api/pod->synthesized-pod-state new-pod)}
+            old-state (get @existing-state-map pod-name)]
+        (swap! existing-state-map assoc pod-name new-state)
```
scrosby: Comment here: 'We always store the updated state, but only reprocess it if it is genuinely different.'

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -0,0 +1,219 @@
...
+  "Update the expected state. Include some business logic to e.g., not change a state to the same value more than once. Marks any state changes Also has a lattice of state. Called externally and from state machine."
+  [{:keys [expected-state-map] :as kcc} pod-name new-expected-state-dict]
+  (locking (calculate-lock [pod-name])
+    (let [old-state (get @expected-state-map pod-name)]
+      (when-not (expected-state-equivalent? new-expected-state-dict old-state)
+        (swap! expected-state-map assoc pod-name new-expected-state-dict)
+        (process kcc pod-name)))))
+
+(defn scan-process
+  "Special verison of process run during scanning."
```
scrosby: Add comment "and holds lock."

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -24,17 +24,56 @@
...
-            [plumbing.core :as pc]))
+            [plumbing.core :as pc]
+            [metrics.counters :as counters]))
 
 (meters/defmeter [cook-mesos scheduler mesos-error])
 (meters/defmeter [cook-mesos scheduler handle-framework-message-rate])
 (meters/defmeter [cook-mesos scheduler handle-status-update-rate])
+(counters/defcounter [cook-mesos scheduler offer-chan-depth])
 
+(def offer-chan-size 100)
```
scrosby: What happens if we overflow this. I'm assumign this is no change from prior behavior?

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -24,17 +24,56 @@
...
+  "For non cook executor tasks, call sync-agent-sandboxes-fn"
+  [conn task-id task-state sync-agent-sandboxes-fn]
+  (let [instance-ent (d/entity (d/db conn) [:instance/task-id task-id])]
+    (when (and (#{:task-starting :task-running} task-state)
+               (not= :executor/cook (:instance/executor instance-ent)))
+      ;; cook executor tasks should automatically get sandbox directory updates
+      (sync-agent-sandboxes-fn (:instance/hostname instance-ent) task-id))))
+
+(defn handle-status-update
+  [conn compute-cluster sync-agent-sandboxes-fn pool->fenzo {:keys [state] :as status}]
```
scrosby: clojuredoc.

integration/tests/cook/util.py
```diff
@@ -1363,3 +1364,25 @@ def _supported_isolators():
...
+
+@functools.lru_cache()
+def _get_compute_cluster_factory_fn():
+    cook_url = retrieve_cook_url()
+    _wait_for_cook(cook_url)
+    init_cook_session(cook_url)
+    compute_clusters = settings(cook_url)['compute-clusters']
+    return compute_clusters[0]['factory-fn']
+
+def using_kubernetes():
```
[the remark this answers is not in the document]
pschorf: 👍 assuming this is just a rename.

scheduler/README-k8s.md
```diff
@@ -0,0 +1,64 @@
...
+                             :config-file "/home/myuser/.kube/config.yml"}}]
+```
+## GKE cluster config
+For GKE clusters, OAuth is used to authenticate with the cluster. To connect to a GKE cluster, you need two things:
+- Google JSON credential file: https://cloud.google.com/iam/docs/creating-managing-service-account-keys
+- Cluster master url (this will be the api client base URL)
+Example:
+```clojure
+:compute-clusters [{:factory-fn cook.kubernetes.compute-cluster/factory-fn
+                    :config {:base-url "http://127.0.0.1:8000"
```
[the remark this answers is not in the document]
pschorf: Why? Should we remove the fake credential path as well? What about the names?

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

scheduler/README-k8s.md
pschorf: 👍

pschorf: 👍 \

scheduler/bin/run-local-kubernetes.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
pschorf: Add an issue for follow up later.

scheduler/config-k8s.edn
```diff
@@ -0,0 +1,82 @@
+{:agent-query-cache {:ttl-ms 1000}
```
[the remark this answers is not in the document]
pschorf: Add an issue for followup

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+                                 V1ObjectMeta V1PodSpec V1PodStatus V1ContainerState V1DeleteOptionsBuilder
+                                 V1DeleteOptions)
+    (io.kubernetes.client.util Watch)
+    (java.util.concurrent Executors ExecutorService)))
+
+
+(def ^ExecutorService kubernetes-executor (Executors/newFixedThreadPool 2))
+
+(defn handle-watch-updates
+  [state-atom ^Watch watch key-fn callback]
```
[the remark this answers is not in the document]
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: They can be noisy, but probably good to at least have something like `(when item (key-fn item))`

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍 to some wordsmithing

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

pschorf: I'd vote to keep it error, but special case the disconnect exceptions to info and no exception stacktrace logging.

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+(defn to-double
+  [^Quantity q]
+  (-> q .getNumber .doubleValue))
+
+(defn convert-resource-map
+  [m]
+  {:mem (if (get m "memory")
+          (-> m (get "memory") to-double (/ (* 1024 1024)))
+          0.0)
+   :cpus (if (get m "cpu")
```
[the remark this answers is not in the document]
pschorf: Not supported yet.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+            (initialize-node-watch api-client current-nodes-atom))
+          (finally
+            (.close watch))))))))
+
+(defn to-double
+  [^Quantity q]
+  (-> q .getNumber .doubleValue))
+
+(defn convert-resource-map
+  [m]
```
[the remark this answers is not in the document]
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍 .  I don't think we need a followup task, it seems pretty clear: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Should be if it's not.

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: 👍

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: 👍

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: I'd vote for "before calling determine-expected-state" as opposed to "next line"

pschorf: Actually, what about just killing the comment and adding a check in `determine-expected-state`?

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Can you replace it with a comment saying that we're keeping leadership indefinitely in Kubernetes?

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: 👍

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: 👍

scheduler/src/cook/kubernetes/controller.clj
pschorf: My vote would be to keep it here since it's used here, and it's private. We use this because we need the actual state object to determine the failure reason.

scheduler/src/cook/kubernetes/controller.clj
pschorf: 👍

scheduler/src/cook/kubernetes/controller.clj
pschorf: Are we even calling this, given that the syntax is bad? Can we just remove it?

scheduler/src/cook/kubernetes/controller.clj
pschorf: 👍

scheduler/src/cook/kubernetes/controller.clj
pschorf: 👍

scheduler/src/cook/kubernetes/controller.clj
pschorf: 👍

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: We just log errors. This was copied out of scheduler.clj

scheduler/src/cook/mesos/mesos_compute_cluster.clj
pschorf: 👍

scheduler/README-k8s.md
scrosby: Revised.

scheduler/README-k8s.md
scrosby: URL added.

scheduler/README-k8s.md
scrosby: Revised.

scheduler/src/cook/kubernetes/api.clj
scrosby: Its information about a hidden dependency between this and another function. Clarified.

scheduler/src/cook/kubernetes/api.clj
scrosby: I want to make sure this isn't forgotten. Its a critical next change. We have 20? 30? TODO's that this could get lost in.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Don't like that. The writeback of expected state is misleading if we do that.

integration/tests/cook/util.py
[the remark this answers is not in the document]
scrosby: More than a rename. This is only true if there is only one compute cluster. (I.e., _get_compute_cluster_factory_fn returns all compute clusters, and the predicate is true if there's only one AND its got the right type)

scheduler/README-k8s.md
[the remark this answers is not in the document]
scrosby: 127.0.0.1 looks like a magic number. If I saw that, I wouldn't (automatically) think I should put in an alternate IP address replacing 127.0.0.1, but look for some kind of proxy that is listening on 8000 and forwarding to google.

I alterered the text here some more with how to get the base URL. Q: Is it always port 8000?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+
+    (.putRequestsItem resources "memory"
+                      (Quantity. (BigDecimal/valueOf ^double (* 1024 1024 mem))
+                                 Quantity$Format/DECIMAL_SI))
+    (.putLimitsItem resources "memory"
+                    (Quantity. (BigDecimal/valueOf ^double (* 1024 1024 mem))
+                               Quantity$Format/DECIMAL_SI))
+    (.putRequestsItem resources "cpu"
+                      (Quantity. (BigDecimal/valueOf ^double cpus)
+                                 Quantity$Format/DECIMAL_SI))
```
dposada: Should we extract a function for 

```clojure
(Quantity. (BigDecimal/valueOf ^double whatever) Quantity$Format/DECIMAL_SI)
```

scheduler/src/cook/kubernetes/api.clj
scrosby: @pschorf Please review.

scheduler/src/cook/kubernetes/api.clj
scrosby: Partially done: @pschorf  can you polish the comment?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+    (.setSpec pod pod-spec)
+
+    pod))
+;;
+;;  This file contains accessor functions for all kubernetes API. That way we can conveniently mock these functions in clojure.
+;;
+
+(defn TODO
+  []
+  (throw (UnsupportedOperationException. "TODO")))
```
dposada: Is this used anywhere? Can we delete it?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+          ;   from the main cook job status and the sidecar statuses, we'll use a naming scheme for sidecars
+          ; * A pod will be considered complete if all containers with name required-* are complete.
+          ; * The main job container will always be named required-cookJobContainer
+          ; * We want a pod to be considered failed if any container with the name required-* fails or any container
+          ;   with the name extra-* fails.
+          ; * A job may have additional containers with the name aux-*
+          job-status (first (filter (fn [c] (= cook-container-name-for-job (.getName c)))
+                                    container-statuses))]
+      (if (some-> pod .getMetadata .getDeletionTimestamp)
+        ; If a pod has been ordered deleted, treat it as if it was gone, Its being async removed.
```
```suggestion
        ; If a pod has been ordered deleted, treat it as if it was gone, it's being async removed.
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+        {:state :missing :reason "Pod was explicitly deleted"}
+        ; If pod isn't being async removed, then look at the containers inside it....
```
```suggestion
        ; If pod isn't being async removed, then look at the containers inside it
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+      (.deleteNamespacedPod
+        api
+        (-> pod .getMetadata .getName)
+        (-> pod .getMetadata .getNamespace)
+        deleteOptions
+        nil
+        nil
+        nil
+        nil
+        nil)
```
dposada: Should we comment these `nil`s like we did the others?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+      (catch JsonSyntaxException e
+        (log/info "Caught the https://github.com/kubernetes-client/java/issues/252 exception.")
```
dposada: Is it worth also logging the pod namespace and name?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+        ; https://github.com/kubernetes-client/java/issues/252
+        ; https://github.com/kubernetes-client/java/issues/86
+        ;
+        ; They actually advise catching the exception and moving on!
+        ;
+        ))))
+
+
+(defn remove-finalization-if-set-and-delete
+  "Remove finalization for a pod if its there. No-op if its not there.
```
```suggestion
  "Remove finalization for a pod if it's there. No-op if it's not there.
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+  We always delete unconditionally, so finalization doesn't matter."
+  [^ApiClient api-client ^V1Pod pod]
+  ; TODO: This likes to noisily throw NotFound multiple times as we delete away from kubernetes.
+  ; I suspect our predicate of existing-states-equivalent needs tweaking.
+  (kill-task api-client pod))
+
+(defn launch-task
+  "Given a pod-name use lookup the associated task, extract the parts needed to syntehsize the kubenretes object and go"
```
dposada: Grammar doesn't make sense in this docstring

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+        (log/info "Launching pod" api launch-pod)
+        (try
+          (-> api
+              (.createNamespacedPod namespace launch-pod nil nil nil))
+          (catch ApiException e
+            (log/error e "Error submitting pod:" (.getResponseBody e)))))
+      ; Because of the complicated nature of task-metadata-seq --- we can't easily run the V1Pod creation code for a
+      ; failed-to-start pod on a server restart. Thus, if we create a task, store into datomic, but then the cook scheduler
+      ; fails --- before kubernetes creates a pod (either the message isn't sent, or there's a kubernetes problem), we will
+      ; the inability to create a new V1Pod and we can't retry this at the kubernetes level.
```
dposada: Grammar doesn't make sense "... we will the inability ..."

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -0,0 +1,316 @@
...
+  (let [namespace "cook"]
+    ;; TODO: IF there's an error, log it and move on. We'll try again later.
+    (if launch-pod
+      (let [api (CoreV1Api. api-client)]
+        (log/info "Launching pod" api launch-pod)
+        (try
+          (-> api
+              (.createNamespacedPod namespace launch-pod nil nil nil))
+          (catch ApiException e
+            (log/error e "Error submitting pod:" (.getResponseBody e)))))
```
dposada: Is there any other information about the pod that we should log here?

scheduler/src/cook/kubernetes/api.clj
dposada: Grammar doesn't make sense "... for because ..."

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -1,147 +1,26 @@
...
-                                                                                           .getRequests
-                                                                                           convert-resource-map))
-                                                                                     containers)]
-                                                         (apply merge-with + container-requests))))
-                                                (apply merge-with +)))
-                                         node-name->pods)]
-    node-name->requests))
+           (java.util.concurrent Executors ScheduledExecutorService TimeUnit)))
 
 (defn generate-offers
```
scrosby: @pschorf  could you please fill in the data shape of what this produces?

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Because the state could legit be empty.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: I did, but I don't see the connection between that and the async chan. Can you revise this? Maybe something like 'because we keep leadership indefinitely, we never send on this channel to indicate leadership lost'?

scheduler/src/cook/kubernetes/controller.clj
scrosby: Pasted into docs. Thank you.

scheduler/src/cook/kubernetes/controller.clj
scrosby: It'll work fine. We just generate the doc string and drop it when executing. (Its dead code, just as if we'd done (defn foo [a b] 1 "2" 3), where 1 and "2" are dead.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
scrosby: Written. @pschorf  can you review?

scheduler/src/cook/kubernetes/api.clj
scrosby: Gone.

scheduler/src/cook/kubernetes/api.clj
scrosby: No point. We know we deleted it (by the state machine logic that triggers deletion.)

scheduler/src/cook/kubernetes/api.clj
scrosby: Reised

scheduler/src/cook/kubernetes/api.clj
scrosby: Changed.

scheduler/src/cook/kubernetes/api.clj
scrosby: For now, no. Best bet is to run it a while in testing, then we can augment logging as we find issues that need more logging.

integration/tests/cook/util.py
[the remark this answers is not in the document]
pschorf: Ok, then let's punt for now.

scheduler/README-k8s.md
pschorf: Done

scheduler/src/cook/kubernetes/api.clj
pschorf: Doesn't really matter to me. Let's just leave it.

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/compute_cluster.clj

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: I don't think we do for the compute cluster configs.

scrosby: Good point. Withdraw the comment.

PR #1189, opened 2019-08-15
Add support for configurable namespaces

pschorf:
- Support configurable namespaces for Kubernetes pods
- Support per-user pod namespaces

In some environments it is desirable to have per-user Kubernetes namespaces, but we want to keep "single namespace" support for local development.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -107,16 +107,26 @@
     (into {}
           (map (fn [[k v]] [k (task-ent->expected-state v)]) all-task-id->task))))
 
+(defn- get-namespace-for-task-metadata
+  [{:keys [kind] :as namespace-config} task-metadata]
+  (case kind
+    :static (:namespace namespace-config)
+    :per-user (-> task-metadata
+                  :command
+                  :user)))
+
 (defrecord KubernetesComputeCluster [^ApiClient api-client name entity-id match-trigger-chan exit-code-syncer-state
                                      current-pods-atom current-nodes-atom expected-state-map existing-state-map
-                                     pool->fenzo-atom]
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -107,16 +107,26 @@
     (into {}
           (map (fn [[k v]] [k (task-ent->expected-state v)]) all-task-id->task))))
 
+(defn- get-namespace-for-task-metadata
```
scrosby: get-namespace-from-task-metadata

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -107,16 +107,26 @@
...
+  [{:keys [kind] :as namespace-config} task-metadata]
+  (case kind
+    :static (:namespace namespace-config)
```
scrosby: Destructure :namespace out.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,8 +241,11 @@
            base-path
            google-credentials
            verifying-ssl
-           bearer-token-refresh-seconds]
-    :or {bearer-token-refresh-seconds 300}}
+           bearer-token-refresh-seconds
+           namespace]
+    :or {bearer-token-refresh-seconds 300
```
scrosby: Can we put documentation for these kubernetes compute cluster options in the config file documentation? 
(I'm not asking you to document for every option, but can you document this option and its possible values?)

scheduler/src/cook/kubernetes/compute_cluster.clj

scheduler/src/cook/kubernetes/compute_cluster.clj

scheduler/src/cook/kubernetes/compute_cluster.clj

PR #1192, opened 2019-08-19
Increase kubernetes timeout from 10 seconds to 120 seconds.

scrosby:

- Increase the timeout
- 
- 

Avoid a lot of log spam


pschorf:
internal green?

PR #1194, opened 2019-08-19
Label pods if they're cook job pods.

scrosby:

- Label pods if they're cook job pods.
- Ignore those pods for cook's kubernetes state management.
- 

This enables soon-to-be-added state scanners which means that we get retries and a proper startup for kubernetes support.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -14,32 +14,59 @@
     (java.util.concurrent Executors ExecutorService)))
 
 
+(def cook-pod-label "twosigma.com/cook-scheduler-job")
+
 (def ^ExecutorService kubernetes-executor (Executors/newFixedThreadPool 2))
+
+(defn is-cook-scheduler-pod
+  "Is this a cook pod? Uses some-> so is null-safe."
+  [^V1Pod pod]
+  (some-> pod .getMetadata .getLabels (.get cook-pod-label)))
+
+(defn make-atom-updater
+  "Given a state atom, returns a callback that updates that state-atom when called with a key, prev item, and item."
+  [state-atom]
+  (fn
+    [key prev-item item]
+    (cond
+      (and (nil? prev-item) (not (nil? item))) (swap! state-atom (fn [m] (assoc m key item)))
+      (and (not (nil? prev-item)) (not (nil? item))) (swap! state-atom (fn [m] (assoc m key item)))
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -14,32 +14,59 @@
...
+
+(defn make-atom-updater
+  "Given a state atom, returns a callback that updates that state-atom when called with a key, prev item, and item."
+  [state-atom]
+  (fn
+    [key prev-item item]
+    (cond
+      (and (nil? prev-item) (not (nil? item))) (swap! state-atom (fn [m] (assoc m key item)))
+      (and (not (nil? prev-item)) (not (nil? item))) (swap! state-atom (fn [m] (assoc m key item)))
+      (and (not (nil? prev-item)) (nil? item)) (swap! state-atom (fn [m] (dissoc m key))))))
```
pschorf: Does this handle `"DELETED"` the right way? I thought the `object` on the watch was still non-nil in that case.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -14,32 +14,59 @@
...
-        "DELETED" (swap! state-atom (fn [m] (dissoc m (key-fn item)))))
-      (when callback
+    (let [watch-response (.next watch)
+          item (.-object watch-response)
+          key (key-fn item)
+          prev-item (get @state-atom key)]
+      (doseq [callback callbacks]
         (try
-          (callback prev-item item)
+          (callback key prev-item item)
```
pschorf: What do you think about just passing `state-atom` and `item` to the callback? Then you can drop the `key-fn` here.

scrosby: It makes this API less confusing and harder to misuse. 
1. There's more than one state atom, so which one gets passed here (and why). E.g., the callback for updating the cook pods atom would also be passed the all-pods atom?
2. The callback that forwards to cook doesn't use the state atom.
3. We move the key lookup into each of the callee's of this function. So we now have mulitple locations that are using it now.

scheduler/src/cook/kubernetes/api.clj
scrosby: No, it doesn't. (the bug is elsewhere, when we're encoding prev-item and item)

scrosby: Fixed now.

pschorf:
Waiting for build.

PR #1195, opened 2019-08-21
Count starting pods against offers

pschorf:
- Counts the starting pods against node utilization when computing offers on kubernetes

Fixes race where resources could be double-booked.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -11,7 +11,6 @@
             [plumbing.core :as pc])
   (:import (com.google.auth.oauth2 GoogleCredentials)
            (io.kubernetes.client ApiClient)
-           (io.kubernetes.client.apis CoreV1Api)
            (io.kubernetes.client.util Config)
            (java.io FileInputStream File)
            (java.util UUID)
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/test/testutil.clj
```diff
@@ -302,6 +305,31 @@
                                  mesos-start-time (assoc :instance/mesos-start-time mesos-start-time))])]
     (d/resolve-tempid (db conn) (:tempids val) id)))
 
+(defn make-task-request [job-ent]
+  (let [considerable->task-id (plumbing.core/map-from-keys (fn [_] (str (d/squuid)))
```
scrosby: clojuredoc this and the other 2 functions please.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -11,18 +11,17 @@
...
-           (io.kubernetes.client.apis CoreV1Api)
            (io.kubernetes.client.util Config)
            (java.io FileInputStream File)
            (java.util UUID)
            (java.util.concurrent Executors ScheduledExecutorService TimeUnit)))
 
 (defn generate-offers
   "Given a compute cluster and maps with node capacity and existing pods, return a list of offers."
-  [node-name->node pod-name->pod compute-cluster]
+  [node-name->node pod-name->pod starting-pod-name->pod compute-cluster]
```
scrosby: Do we want to reorder this to put compute-cluster first? (that seems more 'normal'.

scheduler/src/cook/test/testutil.clj
pschorf: Done

scheduler/src/cook/kubernetes/compute_cluster.clj

PR #1199, opened 2019-08-21
State scanner on startup and on an interval.

scrosby:

- Scan and reprocess all of the task states on startup and on an interval.
- 
- 

- Cook more correctly recovers after a restart. It can also retry failed kubernetes now.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1,5 +1,6 @@
 (ns cook.kubernetes.api
-  (:require [clojure.tools.logging :as log]
+  (:require [clojure.set :as set]
+            [clojure.tools.logging :as log]
             [plumbing.core :as pc])
   (:import
     (com.google.gson JsonSyntaxException)
@@ -69,12 +70,9 @@
   [^V1Pod pod]
   (-> pod .getMetadata .getName))
 
-(defn initialize-pod-watch
-  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes.
-
-  Note that we require that this function set the all-pods-atom before it finishes so that
-  cook.kubernetes.compute-cluster/initialize-cluster works correctly."
-  [^ApiClient api-client all-pods-atom cook-pods-atom cook-pod-callback]
+(defn get-all-pods-in-kubernetes
+  "Get all pods in kubernetes."
+  [api-client]
   (let [api (CoreV1Api. api-client)
         current-pods (.listPodForAllNamespaces api
                                                nil ; continue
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -92,13 +90,28 @@
...
                                               .getName))
                                         (.getItems current-pods))]
-    (log/info "Updating all-pods-atom with pods" (keys pod-name->pod))
-    (reset! all-pods-atom pod-name->pod)
-    (reset! cook-pods-atom
-            (->> pod-name->pod
-                 (filter (fn [[_ pod]] (is-cook-scheduler-pod pod)))
-                 (into {})))
-    (log/info "Updating cook-pods-atom with pods" (keys @cook-pods-atom))
+    [current-pods pod-name->pod]))
```
pschorf: What about just returning `pod-name->pod`? You can get `current-pods` by looking at the values of the map.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -43,6 +46,36 @@
...
+
+(defn scan-states
+  "Scan all states. Note: May block or be slow due to rate limits."
+  [kcc]
+  (log/info "Starting state scan: " )
+  ; TODO Add in rate limits; only visit non-running/running states so fast.
+  ; TODO Add in maximum-visit frequency. Only visit a state once every XX seconds.
+  (let [states-to-scan (states-to-scan kcc)
+        _ (log/info "Doing state scan. Visiting" (count states-to-scan) "states")]
+    (doseq [state states-to-scan]
```
pschorf: Is `state` a task ID? If so, can we rename these variables to make it clear?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -43,6 +46,36 @@
...
+(defn regular-scanner
+  "Trigger a channel that scans all states (shortly) after this function is invoked and on a regular interval."
+  [kcc interval]
+  (let [ch (async/chan (async/sliding-buffer 1))]
+    ; We scan on startup as we load the existing state in, so first scan should be scheduled for a while in the future.
+    (async/pipe (chime-ch (periodic-seq (-> interval time/from-now) interval)) ch)
+    (tools/chime-at-ch ch
+                       (fn scan-states-function []
+                         (scan-states kcc)
+                         {:error-handler (fn [e]
```
pschorf: I think the parens are off here: the function is returning `{:error-handler ...}` but it should be at the same level as the function.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -144,9 +178,13 @@
     ; Initialize the pod watch path.
     (let [conn cook.datomic/conn
           cook-pod-callback (make-cook-pod-watch-callback this)]
+      ; We set expected state first because initialize-pod-watch sets (and invokes callbacks on and reacts to) the expected and the gruadually discovere existing.
+      (reset! expected-state-map (determine-expected-state-on-startup conn api-client name running-task-ents))
+
       (api/initialize-pod-watch api-client all-pods-atom cook-pods-atom cook-pod-callback)
-      ; Before we execute determine-expected-state-on-startup, we need to ensure that initialize-pod-watch sets cook-pods-atom.
-      (reset! expected-state-map (determine-expected-state-on-startup conn name running-task-ents cook-pods-atom)))
+      (if scan-frequency-seconds-config
```
pschorf: Since this has a default value, can it actually be disabled? Should we have some other way to disable it if it's defaulted to every 2 minutes?

scheduler/src/cook/kubernetes/api.clj
scrosby: current-pods is a V1PodList. We need that in order to get its resource version to setup the watch. Everything else just wants the map.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Yup.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Yeah. You're right.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: You're not correct if you don't scan, because without the scan, you'll *never* retry any transient problems. (datomic or otherwise.) That said, just set it to 60 million seconds and it'll only retry every 2 years.

scheduler/src/cook/kubernetes/api.clj
pschorf: 👍

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: I just mean, can we delete the if? Since this will never be non-nil?

scrosby: I expect to set it to null in unit tests.

pschorf: Ah, makes sense.

PR #1200, opened 2019-08-22
Use MB for memory instead of MiB

pschorf:
- Uses MB for memory when interfacing with Kubernetes instead of MiB

The Cook API, Mesos, and Fenzo all assume MB. We need to be consistent between them.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -18,6 +18,10 @@
 
 (def ^ExecutorService kubernetes-executor (Executors/newFixedThreadPool 2))
 
+; Cook, Fenzo, and Mesos use MB for memory. Convert bytes from k8s to MB when passing to fenzo, and MB back to bytes
+; when submitting to k8s.
+(def memory-multiplier 1000000)
+
 (defn is-cook-scheduler-pod
   "Is this a cook pod? Uses some-> so is null-safe."
   [^V1Pod pod]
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -156,7 +156,7 @@
   "Converts a map of Kubernetes resources to a cook resource map {:mem double, :cpus double}"
   [m]
   {:mem (if (get m "memory")
-          (-> m (get "memory") to-double (/ (* 1024 1024)))
+          (-> m (get "memory") to-double (/ 1000000))
```
scrosby: Please put in a comment as to why.
"Cook tracks memory in megabytes (1000*1000) bytes. Convert kubernetes memory into megabytes."

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -233,8 +233,8 @@
     (.setEnv container (into [] env))
     (.setImage container image)
 
-    (.putRequestsItem resources "memory" (double->quantity (* 1024 1024 mem)))
-    (.putLimitsItem resources "memory" (double->quantity (* 1024 1024 mem)))
+    (.putRequestsItem resources "memory" (double->quantity (* 1000000 mem)))
```
scrosby: "Cook tracks memory in megabytes (1000*1000) bytes. Convert its megabytes into bytes."

PR #1201, opened 2019-08-22
Use namespaced pod names for map keys

pschorf:
- Use `{:namespace namespace :name name}` for map keys

Names are not unique globally.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -22,10 +22,10 @@
 
 (defn generate-offers
   "Given a compute cluster and maps with node capacity and existing pods, return a list of offers."
-  [compute-cluster node-name->node pod-name->pod starting-pod-name->pod]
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -159,7 +159,7 @@
         (controller/update-expected-state
           this
           (:task-id task-metadata)
-          {:expected-state :expected/starting :launch-pod {:pod (api/task-metadata->pod task-metadata)
+          {:expected-state :expected/starting :launch-pod {:pod (api/task-metadata->pod pod-namespace task-metadata)
```
scrosby: There's a redundancy here. You're passing now pod-namespace within constructed pod (via the api/task-metadata->pod). However, launch-pod is extracting the namespace via the [:launch-pod :namespace] path.

Suggestion, strip the namespace out of the :launch-pod, and make launch-pod extract the namespace out of the V1Pod. This leaves :launch-pod dictionary with only one key, but I'm fine with that.

PR #1202, opened 2019-08-23
Remove more uses of mesos_url

pschorf:
- Remove more direct uses of `mesos_url`

Makes more tests compatible with Kubernetes.

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1316,7 +1316,9 @@ def mesos_hostnames_to_consider(cook_url, mesos_url):
     return [s['hostname'] for s in slaves]
 
 def kubernetes_hostnames_to_consider():
-    # TODO: support pools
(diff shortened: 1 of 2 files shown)


integration/tests/cook/util.py
```diff
@@ -1305,15 +1313,33 @@ def hosts_to_consider(cook_url, mesos_url):
     slaves = [s for s in slaves if s['attributes'].get(attribute_name, None) == pool] if pool else slaves
     num_to_log = min(len(slaves), 10)
     logging.info(f'First {num_to_log} hosts to consider: {json.dumps(slaves[:num_to_log], indent=2)}')
-    return slaves
+    return [s['hostname'] for s in slaves]
+
+def kubernetes_hostnames_to_consider():
+    # TODO: support pools
```
dposada: Does this not matter yet because none of the tests that use this use pools?

dposada: If so, can we add that context to the comment?

PR #1204, opened 2019-09-04
Use compute-cluster-name in kuberentes pod label

pschorf:
- Use the compute cluster name in Kubernetes cook pod label

Makes it less likely for cook clusters to stomp on each other.


dposada:
@pschorf Don't we need to add `compute-cluster-name` to the call here:

https://github.com/twosigma/Cook/pull/1204/files#diff-bc229d7ae30cc7a52880c221f9438017R137

dposada:
That link didn't really work, it's the call to `initialize-pod-watch` inside the `finally` block

PR #1209, opened 2019-09-11
Reduce scheduler docker image size; cuts docker image growth by 15%

scrosby:

- Reduces the amount we put on top of the existing image by about 140mb.
- 
- 

Smaller dockers build faster and waste less space locally.


scheduler/Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN apt-get -y update
 RUN apt-get -y install software-properties-common
 RUN add-apt-repository ppa:openjdk-r/ppa
 RUN apt-get -y update
-RUN apt-get -y install \
+RUN apt-get --no-install-recommends -y install \
```
DaoWen: Nice catch!

PR #1211, opened 2019-09-11
Improve docs in sorting jobs and for rebalancer.

scrosby:

-  Add a couple of comments that came up in discussion
- 
- 

Explaining why we are doing things and how they work helps people understand the code.

### DIFF
--- a/scheduler/src/cook/rebalancer.clj
+++ b/scheduler/src/cook/rebalancer.clj
@@ -53,6 +53,11 @@
 ;;; Rebalancer runs periodically with a fixed time period. Each period consists of a single cycle, which consists of
 ;;; multiple iterations.
 ;;;
+;;; Within each cycle, the rebalancer takes the first n tasks in a global DRU queue, it then 'tries' to see if they can
+;;; (and should) preempt an existing running task. Only tasks over the share are eligible to be preempted. We only do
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/rebalancer.clj
```diff
@@ -53,6 +53,11 @@
 ;;; Rebalancer runs periodically with a fixed time period. Each period consists of a single cycle, which consists of
 ;;; multiple iterations.
 ;;;
+;;; Within each cycle, the rebalancer takes the first n tasks in a global DRU queue, it then 'tries' to see if they can
+;;; (and should) preempt an existing running task. Only tasks over the share are eligible to be preempted. We only do
```
pschorf:
```suggestion
;;; (and should) preempt any existing running tasks. Only tasks over the share are eligible to be preempted. We only do
```

scheduler/src/cook/rebalancer.clj
```diff
@@ -53,6 +53,11 @@
...
+;;; the preemption if the DRU change is sufficiently large. When we preempt, we determine exactly what tasks are to
```
pschorf:
```suggestion
;;; the preemption if the DRU change is sufficiently large. If there is a set of running tasks that satisfy the criteria, they are killed and the host is reserved to launch the waiting job.
```

pschorf:
Just waiting on a green build

PR #1212, opened 2019-09-16
Introduces x-cook-truncated-results

dposada:

- adding the `x-cook-truncated-results` response header when there are more jobs than the limit you asked for

To let clients know when they need to ask for more with a smaller time window.


dposada:
Putting this on the back burner behind Cook on GKE work.

PR #1213, opened 2019-09-17
Make other non-cli integration tests also use temporal uuid's.

scrosby:

- Make other non-cli integration tests also use temporal uuid's.
- 
- 

Less database clutter.


dposada:
Will merge when green.

PR #1218, opened 2019-09-24
Disable test_cook_instance_num integration test

scrosby:

- Disable the test temporarily
- 
- 

It doesn't work correctly in all environments.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2915,7 +2915,8 @@ class CookTest(util.CookTest):
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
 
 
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2915,7 +2915,8 @@ def test_command_length_limit(self):
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
 
 
-    @unittest.skipUnless(util.supports_exit_code() and not util.has_one_agent(), "Requires exit code support and multiple agents")
+    # Disabled as it doesn't work correctly in all environments.
+    @unittest.skipUnless(False and util.supports_exit_code() and not util.has_one_agent(), "Requires exit code support and multiple agents")
```
dposada: Should we make it `xfail` so it still runs and we can collect the data on it?

PR #1220, opened 2019-09-24
Fix arity bug in initialize-pod-watch

scrosby:

- Fix arity bug in initialize-pod-watch
- 
- 

If we don't, the watch will die and not recover.


dposada:
Will merge when green

PR #1221, opened 2019-09-27
Improvements to k8s volumes

pschorf:
- Use uuids to for Kubernetes volume names  
- Add support for disallowed container mount paths

UUIDs are guaranteed to pass the Kubernetes name validation regex.
Disallowing container paths allows blocking paths that may be added elsewhere in Kubernetes (for instance, in an admission controller.)

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2606,3 +2606,25 @@ class CookTest(util.CookTest):
             self.assertEqual([1, 2], exit_codes, job)
         finally:
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
+
+    @unittest.skipUnless(util.using_kubernetes(), 'Test requires kubernetes')
+    def test_kubernetes_disallowed_volumes(self):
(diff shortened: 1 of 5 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2606,3 +2606,25 @@ def predicate(job):
...
+                                'network': 'HOST',
+                                'force-pull-image': False},
+                     'volumes': [{'host-path': '/tmp',
+                                  'container-path': path}]}
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
+        self.assertEqual(201, resp.status_code)
+        try:
+            util.wait_for_instance(self.cook_url, job_uuid, status='success')
+        finally:
+            util.kill_jobs(self.cook_url, [job_uuid])
```
dposada: This test will fail if/when GKE itself mounts the volume

pschorf: Changed the test per our discussion and added a comment.

PR #1225, opened 2019-10-03
Makes trigger_preemption more reliable

dposada:

- moving the share- and quota-lowering to after the "large job" is running

When the "large job" is subject to the already-lowered share, it's less likely to get scheduled in a congested cluster.

### DIFF
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -395,11 +395,6 @@ class MultiUserCookTest(util.CookTest):
         try:
             large_cpus = util.get_default_cpus()
             small_cpus = large_cpus / 10
-            with admin:
-                # Lower the user's cpu share and quota
-                util.set_limit(self.cook_url, 'share', user.name, cpus=small_cpus, pool=pool)
-                util.set_limit(self.cook_url, 'quota', user.name, cpus=large_cpus, pool=pool)
-
             with user:
                 # Submit a large job that fills up the user's quota
                 base_priority = 99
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_multi_user.py
```diff
@@ -389,18 +389,20 @@ def trigger_preemption(self, pool):
...
             large_cpus = util.get_default_cpus()
             small_cpus = large_cpus / 10
-            with admin:
-                # Lower the user's cpu share and quota
-                util.set_limit(self.cook_url, 'share', user.name, cpus=small_cpus, pool=pool)
-                util.set_limit(self.cook_url, 'quota', user.name, cpus=large_cpus, pool=pool)
+            with self.user_factory.admin():
+                # Reset the user's share and quota
+                util.reset_limit(self.cook_url, 'share', user.name, reason=self.current_name(), pool=pool)
+                util.reset_limit(self.cook_url, 'quota', user.name, reason=self.current_name(), pool=pool)
```
pschorf: Can we assert on these responses as well?

dposada: Good call, done

pschorf:
Feel free to merge when the build is green.

integration/tests/cook/test_multi_user.py
dposada: Turns out that if you try to delete share for a user that doesn't have it defined, you get a 500. Will need to revisit this tomorrow.

pschorf: Sigh

dposada: I changed this to set the limit to the `default` limit instead of deleting it.

dposada: I also ended up marking the preemption tests as `xfail`, see the PR description

PR #1228, opened 2019-10-04
Periodically reset kerberos auth token

pschorf:
- Don't cache the kerberos authentication token between usages of a user object 
- While a user object is active, refresh the kerberos auth token every 60 seconds

This prevents the ticket from expiring and causing spurious test failures.

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -6,6 +6,7 @@ import logging
 import os
 import os.path
 import subprocess
+import threading
 import time
 import unittest
 import uuid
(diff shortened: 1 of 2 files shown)


integration/tests/cook/util.py
```diff
@@ -177,16 +177,31 @@ def _generate_kerberos_ticket_for_user(self, username):
...
+            logger.info(f'Stopping kerberos ticket refresh for {self.name}')
+
     def __enter__(self):
         global session
         super().__enter__()
         assert self.previous_token is None
+        assert self.stop_event is None
         self.previous_token = session.headers.get('Authorization')
-        session.headers['Authorization'] = self.auth_token
+        session.headers['Authorization'] = self._generate_kerberos_ticket_for_user(self.name)
```
dposada: Can't we remove this call since it's happening in `_reset_auth_header`?

PR #1229, opened 2019-10-04
Add support for user, workdir, and env docker parameters

pschorf:
- Add support for `workdir`, `env`, and `docker` parameters

Increase support for the existing mesos API

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -64,12 +64,24 @@ class CookTest(util.CookTest):
         else:
             self.logger.info(f'Exit code not checked because cook executor was not used for {instance}')
 
+    @unittest.skipUnless(util.docker_tests_enabled(), 'requires docker')
     def test_uid(self):
         settings = util.settings(self.cook_url)
+
         username = settings.get('mesos-run-as-user') or os.getenv('USER')
         uid = subprocess.check_output(['/usr/bin/id', '-u', username]).decode('utf-8').strip()
-        command = f'bash -c \'if [[ $UID -eq {uid} ]]; then exit 0; else exit 1; fi\''
-        job_uuid, resp = util.submit_job(self.cook_url, command=command)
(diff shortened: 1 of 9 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2639,6 +2698,32 @@ def test_kubernetes_disallowed_volumes(self):
...
+            self.skipTest(f'Requires {bad_var_name} to be in the disallowed-var-names config')
+        docker_image = util.docker_image()
+
+        container = {'type': 'docker',
+                     'docker': {'image': docker_image,
+                                'network': 'HOST',
+                                'force-pull-image': False,
+                                'parameters': [{'key': 'env',
+                                                'value': f'{bad_var_name}=set'}]}}
+        command = 'bash -c \'if [ -z ${BADVAR+x} ]; then exit 0; else exit 1; fi\''
```
dposada: What does `BADVAR+x` do?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -322,22 +393,29 @@
...
+                           (into env)
+                           (into (param-env-vars parameters))
+                           (into workdir-env-vars)
+                           filter-env-vars))
     (.setImage container image)
 
     (.putRequestsItem resources "memory" (double->quantity (* memory-multiplier mem)))
     (.putLimitsItem resources "memory" (double->quantity (* memory-multiplier mem)))
     (.putRequestsItem resources "cpu" (double->quantity cpus))
+    (.putLimitsItem resources "cpu" (double->quantity cpus))
```
dposada: Should we add a comment about why we're doing this?

integration/tests/cook/test_basic.py
pschorf: That is equivalent to x if `BADVAR` is set to anything (including empty string) and still null if it is not set. It seems like the best way to test for a variable not being set.

dposada:
Will merge when green

PR #1235, opened 2019-10-07
Make user quota calculation more robust to roundoff.

scrosby:

- Make our quota calculation more robust to round-off errors.

Should make the quota comparison more robust, when the new job would put you an epsilon over the quota.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -843,10 +843,9 @@
 
 (defn below-quota?
   "Returns true if the usage is below quota-constraints on all dimensions"
-  [{:keys [count cpus mem] :as quota}
-   {:keys [count cpus mem] :as usage}]
+  [quota usage]
   (every? (fn [[usage-key usage-val]]
-            (<= usage-val (get quota usage-key 0)))
+            (<= usage-val (+ 0.01 (get quota usage-key 0))))
           (seq usage)))
 
 (defn job->usage


dposada:
@scrosby:

This:

> Found this when I was trying to diagnose something else.

Doesn't tell me the "Changes proposed in this PR". Can you please clarify?

scheduler/src/cook/tools.clj
```diff
@@ -843,10 +843,9 @@
 
 (defn below-quota?
   "Returns true if the usage is below quota-constraints on all dimensions"
-  [{:keys [count cpus mem] :as quota}
-   {:keys [count cpus mem] :as usage}]
+  [quota usage]
   (every? (fn [[usage-key usage-val]]
-            (<= usage-val (get quota usage-key 0)))
+            (<= usage-val (+ 0.01 (get quota usage-key 0))))
```
dposada: Can we write a test for this (one that would fail without the added `0.01`)?

scrosby: Not really. I found this when I was trying to diagnose why some job's weren't launching in my GKE efforts. It ended up being unrelated to my problem, but I thought it was worth fixing.

dposada: Why can't we write a test for it?

scrosby: We can contrive a set of jobs that add up and make it overflow. Given current tasking, the mildness of this bug (only important impact is potential on integration test flakiness), I don't want to sink more than minimal effort into this PR.

dposada: Why does it only impact integration tests? Couldn't this affect real users?

scrosby: It could affect real users, but on the margin it only means you may get to launch one job less than you should. On discussion, we decided to defer the fix until we have time to write a unit test.

PR #1236, opened 2019-10-07
Better logging messages for integration test user changing.

scrosby:

- Improve logging around user session changing.

Make diagnosing what actually occurred a little bit easier.

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -117,7 +117,7 @@ class _AuthenticatedUser(object):
 
     def __enter__(self):
         if self.impersonatee:
-            logger.debug(f'Switching to user {self.name} with impersonatee {self.impersonatee}')
+            logger.debug(f'Starting impersonation of {self.impersonatee}')
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
         else:
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -120,8 +120,6 @@ class _AuthenticatedUser(object):
             logger.debug(f'Starting impersonation of {self.impersonatee}')
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
-        else:
-            logger.debug(f"Not switching impersonation header to {self.name} due to None self.impersonatee")
 
     def __exit__(self, ex_type, ex_val, ex_trace):
         if self.impersonatee:
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -117,7 +117,7 @@ class _AuthenticatedUser(object):
 
     def __enter__(self):
         if self.impersonatee:
-            logger.debug(f'Switching to user {self.name} with impersonateee {self.impersonatee}')
+            logger.debug(f'Switching to user {self.name} with impersonatee {self.impersonatee}')
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
         else:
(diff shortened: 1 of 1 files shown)


integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
         return type(self)(self.name, impersonatee=other_username)
 
     def __enter__(self):
-        logger.debug(f'Switching to user {self.name}')
         if self.impersonatee:
+            logger.debug(f'Switching to user {self.name} with impersonateee {self.impersonatee}')
```
dposada:
```suggestion
            logger.debug(f'Switching to user {self.name} with impersonatee {self.impersonatee}')
```

integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
...
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
+        else:
+            logger.debug(f"Not switching impersonation header to {self.name} due to None self.impersonatee")
```
dposada:
```suggestion
            logger.debug(f'Switching to user {self.name} with no impersonation')
```

integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
...
 
     def __exit__(self, ex_type, ex_val, ex_trace):
-        logger.debug(f'Switching back from user {self.name}')
         if self.impersonatee:
+            logger.debug(f'Switching back from user {self.name} for impersonation')
```
dposada: We either need to keep the original log outside the `if`/`else` or we need to add another log in the `else` branch.

I think you may be conflating the idea of switching users with the idea of using impersonation. Those are two different concepts. Many of our tests switch users, very few use impersonation.

dposada: > Note that the code here is completely dead if you're not impersonating

That's not true -- `__exit__` is called by the subclasses of `_AuthenticatedUser`, which is why the log outside the `if` check for impersonation is useful regardless of whether or not impersonation is happening.

integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
...
+            logger.debug(f'Switching to user {self.name} with impersonatee {self.impersonatee}')
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
+        else:
+            logger.debug(f"Not switching impersonation header to {self.name} due to None self.impersonatee")
 
     def __exit__(self, ex_type, ex_val, ex_trace):
-        logger.debug(f'Switching back from user {self.name}')
         if self.impersonatee:
+            logger.debug(f'Switching back from user {self.name} for impersonation')
```
dposada:
```suggestion
            logger.debug(f'Stopping impersonation of {self.impersonatee}')
```

integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
         return type(self)(self.name, impersonatee=other_username)
 
     def __enter__(self):
-        logger.debug(f'Switching to user {self.name}')
         if self.impersonatee:
+            logger.debug(f'Switching to user {self.name} with impersonatee {self.impersonatee}')
```
dposada:
```suggestion
            logger.debug(f'Starting impersonation of {self.impersonatee}')
```

integration/tests/cook/util.py
```diff
@@ -115,14 +115,16 @@ def impersonating(self, other_user):
...
             self.previous_impersonatee = session.headers.get(IMPERSONATION_HEADER)
             session.headers[IMPERSONATION_HEADER] = self.impersonatee
+        else:
+            logger.debug(f"Not switching impersonation header to {self.name} due to None self.impersonatee")
```
dposada: I think we should just remove this `else` block.

scrosby: Ok.

PR #1233, opened 2019-10-07
Fix test_docker_fields

pschorf:
## Changes proposed in this PR
- Use valid docker parameters in `test_docker_fields`

## Why are we making these changes?
Test is currently broken due to parameter validation.

PR #1237, opened 2019-10-07
Pool support

scrosby:

- Support multiple pools with kubernetes, using taints and tolerations.

To make this feature work in kubernetes as well as the existing mesos support.

### DIFF
--- a/integration/tests/cook/test_pools.py
+++ b/integration/tests/cook/test_pools.py
@@ -44,8 +44,6 @@ class PoolsCookTest(util.CookTest):
                     quota_multiplier = 1 if pool_name == default_pool else 2
                     util.set_limit(self.cook_url, 'quota', user.name, cpus=cpus * quota_multiplier, pool=pool_name)
 
-            logging.info(f"User to access: {user}")
-
             with user:
                 util.kill_running_and_waiting_jobs(self.cook_url, user.name)
                 for pool in pools:
--- a/integration/tests/cook/test_pools.py
+++ b/integration/tests/cook/test_pools.py
@@ -21,7 +21,6 @@ class PoolsCookTest(util.CookTest):
 
     def setUp(self):
         self.cook_url = type(self).cook_url
-        self.mesos_url = util.retrieve_mesos_url()
         self.logger = logging.getLogger(__name__)
         self.user_factory = util.UserFactory(self)
 
@@ -45,6 +44,8 @@ class PoolsCookTest(util.CookTest):
                     quota_multiplier = 1 if pool_name == default_pool else 2
                     util.set_limit(self.cook_url, 'quota', user.name, cpus=cpus * quota_multiplier, pool=pool_name)
 
+            logging.info(f"User to access: {user}")
+
             with user:
                 util.kill_running_and_waiting_jobs(self.cook_url, user.name)
                 for pool in pools:
@@ -57,7 +58,7 @@ class PoolsCookTest(util.CookTest):
                                                           command='sleep 600', pool=pool_name)
                     all_job_uuids.append(filling_job_uuid)
                     instance = util.wait_for_running_instance(self.cook_url, filling_job_uuid)
-                    slave_pool = util.slave_pool(self.cook_url, self.mesos_url, instance['hostname'])
+                    slave_pool = util.node_pool(instance['hostname'])
                     self.assertEqual(pool_name, slave_pool)
 
                     # Submit a job that should not get scheduled
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1298,6 +1298,31 @@ def get_kubernetes_nodes():
     logging.info(f'Retrieved kubernetes nodes: {node_json}')
     return node_json['items']
 
+@functools.lru_cache()
+def kubernetes_node_pool(nodename):
+    node = [node for node in get_kubernetes_nodes() if node['metadata']['name'] == nodename]
+    poolname_taint = [taint for taint in node[0]['spec']['taints'] if taint['key'] == 'cook.pool']
+    if len(poolname_taint) == 0:
+        return None
+    poolname = poolname_taint[0].get('value',None)
+    logging.info(f"K8S pool for {nodename} is {poolname}")
+    return poolname
+
+@functools.lru_cache()
+def mesos_node_pool(nodename):
+    cook_url = retrieve_cook_url()
+    mesos_url = retrieve_mesos_url()
+    node_pool = slave_pool(cook_url, mesos_url, nodename)
+    return node_pool
+
+def node_pool(nodename):
+    """Get the pool for a node."""
+    if using_mesos():
+        return mesos_node_pool(nodename)
+    elif using_kubernetes():
+        return kubernetes_node_pool(nodename)
+    else:
+        raise RuntimeError('Unable to determine cluster max CPUs')
 
 def max_kubernetes_node_cpus():
     nodes = get_kubernetes_nodes()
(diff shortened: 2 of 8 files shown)


scrosby:
Integration test suite passes on GKE except for:
[gw1] [  7%] FAILED tests/cook/test_basic.py::CookTest::test_balanced_host_constraint_can_place 
[gw0] [ 34%] FAILED tests/cook/test_basic.py::CookTest::test_docker_fields 

And is internal green!

pschorf:
If you rebase, #1233 should fix `test_docker_fields`. Probably worth looking into what's going on with `test_balanced_host_constraint_can_place`

integration/tests/cook/util.py
```diff
@@ -1298,6 +1298,31 @@ def get_kubernetes_nodes():
...
+    return node_pool
+
+def node_pool(nodename):
+    """Get the pool for a node."""
+    if using_mesos():
+        return mesos_node_pool(nodename)
+    elif using_kubernetes():
+        return kubernetes_node_pool(nodename)
+    else:
+        raise RuntimeError('Unable to determine cluster max CPUs')
```
pschorf: FIx the error message :)

integration/tests/cook/test_pools.py
```diff
@@ -45,6 +44,8 @@ def test_pool_scheduling(self):
                     quota_multiplier = 1 if pool_name == default_pool else 2
                     util.set_limit(self.cook_url, 'quota', user.name, cpus=cpus * quota_multiplier, pool=pool_name)
 
+            logging.info(f"User to access: {user}")
```
pschorf: Can you make the log message more specific? Something like `test_pool_scheduling user: {user}`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -197,12 +200,13 @@
...
-            (= pool-name "no-pool"))
-      (let [nodes @current-nodes-atom
-            pods @all-pods-atom]
-        (generate-offers this nodes pods (controller/starting-namespaced-pod-name->pod this)))
-      []))
+    (let [nodes @current-nodes-atom
+          pods @all-pods-atom
+          starting-instances (controller/starting-namespaced-pod-name->pod this)
+          offers-all-pools (generate-offers this nodes pods starting-instances)]
+      (doseq [[pool offers] offers-all-pools]
```
pschorf: Just log offers for `pool-name`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -197,12 +200,13 @@
...
-    (if (or (= pool-name (config/default-pool)) ; TODO(pschorf): Support pools
-            (= pool-name "no-pool"))
-      (let [nodes @current-nodes-atom
-            pods @all-pods-atom]
-        (generate-offers this nodes pods (controller/starting-namespaced-pod-name->pod this)))
-      []))
+    (let [nodes @current-nodes-atom
+          pods @all-pods-atom
+          starting-instances (controller/starting-namespaced-pod-name->pod this)
+          offers-all-pools (generate-offers this nodes pods starting-instances)]
```
pschorf: Can you add a `TODO` to make this calculation pool-specific instead of generating all of them and throwing them away?
I think it's fine for now, though.

integration/tests/cook/util.py
scrosby: Good find.

integration/tests/cook/test_pools.py
scrosby: Thats a stale debugging line. :( 
Gone

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Bad idea. :) 
If you do that, you get the massive node object in the logged information. That's why I filter out just the size information here and log 1/100th the amount of stuff.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Yeah, and good reminder that we should have a TODO. I'll open a JIRA to track too.

scrosby:
test_balanced_host_constraint_can_place may be unhappy because the cluster is so small. I tried going from 2 to 3 nodes and we'll see how that works. The other bad test is fixed after rebasing.

scrosby:
I disabled the test for kubernetes, explanation is in the diff.

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: I meant you probably don't want to log offers for all pools in each invocation of `pending-offers`. Just log the offers for the pool you're actively looking at. The filtering is fine.

scheduler/src/cook/kubernetes/compute_cluster.clj
pschorf: Can you add the TODO?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -184,6 +184,39 @@
            (-> m (get "cpu") to-double)
            0.0)})
 
+(defn get-taint-key
+  "Get the taint variable out of a taint key out of the V1Taint object."
+  [^V1Taint taint]
+  (.getKey taint))
+
+(defn process-taints
```
pschorf: Taking another look at this, it seems like this would be more clear to have this logic inside `get-node-pool` and `node-schedulable?` instead of building this separate map.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -184,6 +184,39 @@
...
+  If there is no taint of the form cook.pools, return 'no-pool'. Processing a taints involves
+  splitting the list into the cook.pool taint and the other taints."
+  [node-name->node]
+  (pc/map-vals (fn [^V1Node node]
+                 (let [taints-on-node (or (some-> node .getSpec .getTaints) [])]
+                   (group-by #(if (= "cook.pool" (get-taint-key %)) :pool-taint :other-taint)  taints-on-node))) node-name->node))
+
+(defn get-node-pool
+  "Get the pool for a node. In the case of no pool, return 'no-pool"
+  [node->processed-taints node-name]
```
pschorf: What about changing this to just take a Node?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -184,6 +184,39 @@
...
+  (or (some-> node->processed-taints
+               (get node-name)
+               (get :pool-taint)
+               first
+               .getValue)
+      "no-pool"))
+
+(defn node-schedulable?
```
pschorf: Just take a Node

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Its there a couple of lines lower.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Aah. Got it. Ok. Also moved the TODO with this refactor.

scheduler/src/cook/kubernetes/api.clj
scrosby: Then I'd have to reprocess taints twice, with different logic.  Seemed to make more sense to process once and reuse the result, and that way I only have one process-taints logic.

scheduler/src/cook/kubernetes/api.clj
scrosby: Done.

scheduler/src/cook/kubernetes/api.clj
scrosby: Donel

PR #1239, opened 2019-10-09
Add max-ports to task constraints

pschorf:
- Add a `max-ports` config to task constraints

Allows administrators to set a maximum number of ports per job.


integration/tests/cook/test_basic.py
```diff
@@ -2671,3 +2671,15 @@ def test_priority(self):
         _, resp = util.submit_job(self.cook_url, priority=16000001)
         self.assertEqual(400, resp.status_code, msg=resp.content)
         self.assertTrue(f'priority":"(not (between-0-and-16000000' in str(resp.content))
+
+    def test_max_ports(self):
+        settings = util.settings(self.cook_url)
+        max_ports = util.get_in(settings, 'task-constraints', 'max-ports')
```
dposada: Will this always be set?

pschorf: Yes, since we have a default in the config there's no way to disable it.

PR #1242, opened 2019-10-11
Makes test_list_jobs_by_pool more reliable

dposada:

- making the `sleep` match the timeout for the test
- adding some logging in the pool loops

Without the `sleep` adjustment, if the other jobs take a long time to get scheduled, the `sleep` jobs can complete, which makes the listing assertion fail.


pschorf:
Waiting on a green build.

PR #1244, opened 2019-10-14
Refactor handle-resource-offers to not use global compute-cluster

scrosby:
- Refactor handle-resource-offers-refactor to not use global compute-cluster

Intermediate step to supporting multiple compute clusters.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -727,7 +727,7 @@
   "Gets a list of offers from mesos. Decides what to do with them all--they should all
    be accepted or rejected at the end of the function."
   [conn ^TaskScheduler fenzo pool-name->pending-jobs-atom mesos-run-as-user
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -780,7 +780,9 @@
           (meters/mark! handle-resource-offer!-errors)
           (log/error t "In" pool-name "pool, error in match:" (ex-data t))
           (when-let [offers @offer-stash]
-            (cc/restore-offers compute-cluster pool-name offers))
+            ; Group the set of all offers by compute cluster and route them to that compute cluster for restoring.
+            (doseq [[compute-cluster offersubset] (group-by :compute-cluster offers)]
```
dposada:
```suggestion
            (doseq [[compute-cluster offer-subset] (group-by :compute-cluster offers)]
```

scrosby: Changed.

dposada:
Will merge when green

PR #1245, opened 2019-10-14
The cook pool kubernetes taint/toleration is cook-pool, not cook.pool

scrosby:
- The cook pool kubernetes taint/toleration is cook-pool, not cook.pool

This is per-the-spec. It is also more greppable.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -189,7 +189,7 @@
   [^V1Node node]
   ; In the case of nil, we have taints-on-node == [], and we'll map to no-pool.
   (let [taints-on-node (or (some-> node .getSpec .getTaints) [])
-        cook-pool-taint (filter #(= "cook.pool" (.getKey %)) taints-on-node)]
+        cook-pool-taint (filter #(= "cook-pool" (.getKey %)) taints-on-node)]
     (if (= 1 (count cook-pool-taint))
           (-> cook-pool-taint first .getValue)
       "no-pool")))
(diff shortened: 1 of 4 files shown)


scheduler/config-k8s.edn
```diff
@@ -6,7 +6,7 @@
  :authorization-config {;; These users have admin privileges when using configfile-admins-auth;
                         ;; e.g., they can view and modify other users' jobs.
                         ;; Sets "admin" and the user defined in "COOK_ONE_USER_AUTH" to be admins for local development.
-                        :admins #{"admin" #config/env "COOK_ONE_USER_AUTH"}
+                        :admins #{"admin" "root" #config/env "COOK_ONE_USER_AUTH"}
```
dposada: Why is this changing?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -189,7 +189,7 @@
   [^V1Node node]
   ; In the case of nil, we have taints-on-node == [], and we'll map to no-pool.
   (let [taints-on-node (or (some-> node .getSpec .getTaints) [])
-        cook-pool-taint (filter #(= "cook.pool" (.getKey %)) taints-on-node)]
+        cook-pool-taint (filter #(= "cook-pool" (.getKey %)) taints-on-node)]
```
dposada: Should we make this be a constant?

scheduler/config-k8s.edn
scrosby: This is a leak for my run-locally changes that was applied on top of the patch. Removed that commit from the push.

scheduler/src/cook/kubernetes/api.clj
scrosby: Why? Its another level of indirection a constant we will never change because it is API. The constant is also used in python and clojure.

dposada:
Will merge when green

PR #1247, opened 2019-10-14
Allows adjusting jobs on submission via plugin

dposada:

- introducing the `JobAdjuster` plugin
- adding a "demo" implementation that simply adds a label
- adding a test that verifies the "demo" implementation actually modifies the job

We want to allow plugins that can, for example, modify the pool on job submissions.

## What's not in here?

This PR sets up the plumbing for the plugin, and a future PR will introduce the real implementation, which will modify the pool on job submission based on configuration.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -40,6 +40,7 @@
             [cook.unscheduled :as unscheduled]
             [cook.tools :as util]
             [cook.mesos]
+            [cook.plugins.adjustment :as adjustment]
             [cook.plugins.definitions :as plugins]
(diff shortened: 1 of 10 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -1647,8 +1648,10 @@
   ::jobs and ::groups, which specify the jobs and job groups."
   [conn {:keys [::groups ::jobs ::pool] :as ctx}]
   (try
+    (let [])
```
nsinkov: what is this for?

dposada: Removed, thanks for catching

PR #1251, opened 2019-10-16
Refactor straggler handler to not use global compute cluster

scrosby:

- Refactor straggler handler to not use global compute cluster

Prepare for supporting multiple compute clusters

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1053,10 +1053,10 @@
 (defn straggler-handler
   "Periodically checks for running jobs that are in groups and runs the associated
    straggler handler."
-  [conn compute-cluster trigger-chan]
+  [conn trigger-chan]
   (util/chime-at-ch trigger-chan
                     (fn straggler-handler-event []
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1052,10 +1052,10 @@
 (defn straggler-handler
   "Periodically checks for running jobs that are in groups and runs the associated
    straggler handler."
-  [conn compute-cluster trigger-chan]
+  [conn trigger-chan]
   (util/chime-at-ch trigger-chan
                     (fn straggler-handler-event []
-                      (handle-stragglers conn #(cc/kill-task compute-cluster (:instance/task-id %))))
+                      (handle-stragglers conn (fn [task-ent] (cc/kill-task (cook.task/task-ent->ComputeCluster task-ent) (:instance/task-id task-ent)))))
```
dposada: This line seems too long

dposada:
One formatting comment, will merge when green

scheduler/src/cook/scheduler/scheduler.clj
scrosby: I split up the line.

dposada:
Will merge when green

PR #1252, opened 2019-10-16
Refactor fenzo initialization to not use global compute cluster.

scrosby:

- Refactor fenzo initialization to not use global compute cluster.
- Discovered some dead code and deleted it

Prepare for supporting multiple compute clusters

### DIFF
--- a/scheduler/test/cook/test/scheduler/scheduler.clj
+++ b/scheduler/test/cook/test/scheduler/scheduler.clj
@@ -375,7 +375,7 @@
                         :slave-id {:value (str "slave-" (UUID/randomUUID))}
                         :hostname (str "host-" (UUID/randomUUID))}])
         framework-id (str "framework-id-" (UUID/randomUUID))
-        fenzo-maker #(sched/make-fenzo-scheduler nil 100000 nil 1)] ; The params are for offer declining, which should never happen
+        fenzo-maker #(sched/make-fenzo-scheduler 100000 nil 1)] ; The params are for offer declining, which should never happen
     (testing "Consume no schedule cases"
       (are [schedule offers] (= [] (:matches (sched/match-offer-to-schedule (db c) (fenzo-maker) schedule
                                                                             offers (atom {}) nil)))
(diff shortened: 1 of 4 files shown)


scheduler/test/cook/test/scheduler/scheduler.clj
```diff
@@ -929,14 +929,7 @@
   (with-redefs [completion/plugin completion/no-op]
     (let [uri "datomic:mem://test-handle-status-update"
           conn (restore-fresh-database! uri)
-          tasks-killed (atom #{})
```
scrosby: The tasks-killed and driver machinery to update it is dead and unchecked. Shall I open a followup to resurrect it?

scheduler/test/cook/test/scheduler/scheduler.clj
```diff
@@ -929,14 +929,7 @@
...
-          driver (reify msched/SchedulerDriver
-                   (kill-task! [_ task] (swap! tasks-killed conj (:value task)))) ; Conjoin the task-id
-          compute-cluster (testutil/fake-test-compute-cluster-with-driver conn uri driver)
-          fenzo (sched/make-fenzo-scheduler compute-cluster 1500 nil 0.8)
-          synced-agents-atom (atom [])
```
scrosby: The synced-agents-atom and synced-agent-sandboxes-fn machinery to update it is dead and unchecked. Shall I open a followup to resurrect it?

dposada:
Will merge when green

scheduler/test/cook/test/scheduler/scheduler.clj
dposada: I'm guessing you mean the machinery in the test is dead? If so, do we have any other tests for this functionality?

scheduler/test/cook/test/scheduler/scheduler.clj
dposada: I'm guessing you mean the machinery in the test is dead? If so, do we have any other tests for this functionality?

scheduler/test/cook/test/scheduler/scheduler.clj
scrosby: Correct. Don't know. I suspect not. (at least at the unit test level)

scheduler/test/cook/test/scheduler/scheduler.clj
scrosby: Correct. Don't know. I suspect not. (at least at the unit test level)

PR #1260, opened 2019-10-18
Fix Kubernetes Node Watch

scrosby:

- We weren't actually updating our node dictionary when membership changed, so would miss k8s node changes.

So we can see new nodes when they occur.


dposada:
Will merge when green

PR #1264, opened 2019-10-20
Adds leader URL to /info

dposada:

- adding the leader URL to the `/info` endpoint

It's nice to be able to ask who the current leader is.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -139,7 +139,8 @@
   {:authentication-scheme s/Str
    :commit s/Str
    :start-time s/Inst
-   :version s/Str})
+   :version s/Str
+   :leader-url s/Str})
 
 (def PortMapping
   "Schema for Docker Portmapping"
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -139,7 +139,8 @@
   {:authentication-scheme s/Str
    :commit s/Str
    :start-time s/Inst
-   :version s/Str})
+   :version s/Str
+   :leader-url s/Str})
```
shamsimam: Nitpick: lexicographic order

scheduler/src/cook/rest/api.clj
```diff
@@ -1423,7 +1438,8 @@
                     {:authentication-scheme auth-scheme
                      :commit @cook.util/commit
                      :start-time start-up-time
-                     :version @cook.util/version})})))
+                     :version @cook.util/version
+                     :leader-url (leader-url leader-selector)})})))
```
shamsimam: Nitpick: lexicographic order

PR #1276, opened 2019-10-30
Bugfix: If we have problems processing something, don't kill the watch thread.

scrosby:

- If we have problems processing something, don't make the whole thread die

On startup, this processing happens in the watch thread. An exception here causes the watch thread to die, and we don't see new events from kubernetes.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -195,7 +195,10 @@
         ; We always store the updated state, but only reprocess it if it is genuinely different.
         (swap! existing-state-map assoc pod-name new-state)
         (when-not (existing-state-equivalent? old-state new-state)
-          (process kcc pod-name))))))
+          (try
+            (process kcc pod-name)
+            (catch Exception e
+              (log/error e "Error while processing pod-update for" pod-name))))))))
 
 (defn pod-deleted
   "Indicate that kubernetes does not have the pod. Invoked by callbacks from kubernetes."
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -195,15 +195,22 @@
         ; We always store the updated state, but only reprocess it if it is genuinely different.
         (swap! existing-state-map assoc pod-name new-state)
         (when-not (existing-state-equivalent? old-state new-state)
-          (process kcc pod-name))))))
+          (try
+            (process kcc pod-name)
+            (catch Exception e
+              (log/error e "Error while processing pod-update for" pod-name))))))))
```
dposada: Can we include the compute cluster name in this log? To be consistent with other spots, let's do 

```
In compute cluster $name, error while processing pod update for ...
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -195,15 +195,22 @@
...
   "Indicate that kubernetes does not have the pod. Invoked by callbacks from kubernetes."
   [{:keys [existing-state-map] :as kcc} ^V1Pod pod-deleted]
   (let [pod-name (api/V1Pod->name pod-deleted)]
     (locking (calculate-lock pod-name)
       (swap! existing-state-map dissoc pod-name)
-      (process kcc pod-name))))
+      (try
+        (process kcc pod-name)
+        (catch Exception e
+          (log/error e "Error while processing pod-deleting for" pod-name))))))
```
dposada: Can we include the compute cluster name in this log? To be consistent with other spots, let's do 

```
In compute cluster $name, error while processing pod deletion for ...
```

PR #1277, opened 2019-10-30
Add a type annotation to make things more clear.

scrosby:

- Add a type annotation to make things more clear.

Catch some bugs or other weird invocations.


dposada:
Will merge when green

PR #1279, opened 2019-10-30
Report when we detect a pod as deleted

scrosby:

- Explicitly report when we think a pod is deleted

- Help trace down a weird pod lifecycle bug.


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -201,6 +201,7 @@
   "Indicate that kubernetes does not have the pod. Invoked by callbacks from kubernetes."
   [{:keys [existing-state-map] :as kcc} ^V1Pod pod-deleted]
   (let [pod-name (api/V1Pod->name pod-deleted)]
+    (log/info "Detected pod" pod-name "deleted.")
```
dposada: Can we include the compute cluster name in this log? To be consistent with other spots, let's do 

```
In compute cluster $name, detected pod ...
```

Let's also get rid of the period at the end to be consistent with other log lines.

PR #1280, opened 2019-10-30
Better logging about startup/initializing.

scrosby:

- Better contextual logging on startup.

With multiple compute clusters, have the log indicate each one as it starts up.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -123,7 +123,7 @@
         [(make-atom-updater all-pods-atom) ; Update the set of all pods.
          (partial cook-pod-callback-wrap cook-pod-callback compute-cluster-name)] ; Invoke the cook-pod-callback if its a cook pod.
         old-all-pods @all-pods-atom]
-    (log/info "Processing pods" (keys namespaced-pod-name->pod))
+    (log/info "Processing pods for compute cluster " compute-cluster-name " are "(keys namespaced-pod-name->pod))
     ; We want to process all changes through the callback process.
     ; So compute the delta between the old and new and process those via the callbacks.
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -123,7 +123,7 @@
         [(make-atom-updater all-pods-atom) ; Update the set of all pods.
          (partial cook-pod-callback-wrap cook-pod-callback compute-cluster-name)] ; Invoke the cook-pod-callback if its a cook pod.
         old-all-pods @all-pods-atom]
-    (log/info "Processing pods" (keys namespaced-pod-name->pod))
+    (log/info "Processing pods for compute cluster " compute-cluster-name " are "(keys namespaced-pod-name->pod))
```
```suggestion
    (log/info "Processing pods for compute cluster" compute-cluster-name ":" (keys namespaced-pod-name->pod))
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -179,6 +179,7 @@
 
   (initialize-cluster [this pool->fenzo running-task-ents]
     ; Initialize the pod watch path.
+    (log/info "Initializing Kubernetes cluster " name)
```
dposada:
```suggestion
    (log/info "Initializing Kubernetes compute cluster" name)
```

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -239,6 +239,7 @@
     (not (nil? @driver-atom)))
 
   (initialize-cluster [this pool->fenzo _]
+    (log/info "Initializing Mesos cluster " compute-cluster-name)
```
dposada:
```suggestion
    (log/info "Initializing Mesos compute cluster" compute-cluster-name)
```

PR #1282, opened 2019-10-30
Indicate when we're processing a given change.

scrosby:

- Increase visibility when we are processing a pod.

Improve visibility for why we're processing a change. Help diagnose a weird pod lifecycle bug.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -130,6 +130,7 @@
     (doseq [task (set/union (keys namespaced-pod-name->pod) (keys old-all-pods))]
       (doseq [callback callbacks]
         (try
+          (log/info "Doing (startup) callback for " task)
           (callback task (get old-all-pods task) (get namespaced-pod-name->pod task))
           (catch Exception e
             (log/error e "Error while processing callback for" task)))))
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -130,6 +130,7 @@
     (doseq [task (set/union (keys namespaced-pod-name->pod) (keys old-all-pods))]
       (doseq [callback callbacks]
         (try
+          (log/info "Doing (startup) callback for " task)
```
```suggestion
          (log/info "In" compute-cluster-name "compute cluster, doing (startup) callback for" task)
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -65,7 +65,8 @@
   ; TODO Add in maximum-visit frequency. Only visit a task once every XX seconds.
   (let [taskids (taskids-to-scan kcc)
         _ (log/info "Doing taskid scan. Visiting" (count taskids) "taskids")]
-    (doseq [taskid taskids]
+    (doseq [^String taskid taskids]
+      (log/info "Doing scan of " taskid)
```
dposada: Can we include the compute cluster name in this log? To be consistent with other spots, let's do 

```
In compute cluster $name, doing scan of ...
```

dposada:
Will merge when green

PR #1283, opened 2019-10-30
Bug fix: We need a variable sized threadpool for our watcher threads

scrosby:

- Use a cached threadpool for the watches to run in.

If the threadpool is a fixed size, too many compute clusters would cause us to not launch all of out watchers. We need to threads per compute cluster, so the old code broke with only 2 compute clusters.


dposada:
Will merge when green

PR #1284, opened 2019-10-30
Bug fix: We want to restart the k8s node watch even on a clean exit.

scrosby:

- Move the code that relaunches the watcher to the finally block so it always runs.

Without this change, we'd not restart the watch on a clean exit of the watcher, and we'd cease watching for node changes.


dposada:
Will merge when green

PR #1288, opened 2019-10-31
Do not unassign offers to fenzo when we write failed statuses to datomic

scrosby:

- Do not unassign offers to fenzo when we write failed statuses to datomic
- 
- 

In the case of kubernetes, this seems unsafe. we'll mark resources as available when kubernetes
is still deleting the pods and they're not really available. We then overcommit nodes and our pods won't launch.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -205,7 +205,7 @@
 
 (defn write-status-to-datomic
   "Takes a status update from mesos."
-  [conn pool->fenzo status]
+  [conn pool->fenzo status update-fenzo]
   (log/info "Mesos status is:" status)
   (timers/time!
     handle-status-update-duration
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -245,7 +245,10 @@
                job-resources (util/job-ent->resources job-ent)
                pool-name (util/job->pool-name job-ent)
                ^TaskScheduler fenzo (get pool->fenzo pool-name)]
-           (when (#{:instance.status/success :instance.status/failed} instance-status)
+           ; Q: How to disable this for kubernetes only. In the case of kubernetes, this tells k8s that the resources can be used almost immediately.
```
scrosby: I had some odd kubernetes overallocation failures that seemed linked to this code, so I want it disabled for kubernetes.

But, this seems wrong to me for mesos too. It seems we should delete it entirely. However, I don't fully understand why it exists, so I'm opening it up to discussion. If we can figure out what to do next, I'll respin the patch appropriately.

Any idea why this unassign code exists?

scrosby: Paul mentiosn that we probably want this code for kubernetes too. For fenzo constraints to work, fenzo needs to track all running jobs to handlethings like, e.g., constraints, or running a group of jobs on different machines. We need to validate that we load everything into fenzo on startup in this code. We may or may not be doing so.

dposada:
@scrosby Is this ready for review? It still has the wip label

scrosby:
We don't want this change.

PR #1291, opened 2019-11-01
Move some handlers from a compute-cluster/kill to compute-cluster/kill-task-if-possible.

scrosby:
- Change lingering task and straggler handler to use the wrapper.

There are cases, in particular, lingering tasks, stragglers, or cancelled tasks where the task might
outlive the compute cluster it was member of. When this occurs, the looked up compute cluster
is nil and trying to kill via it would cause an NPE. Switch them to using this wrapper function.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -48,7 +48,10 @@
     "Called when offers are not processed to ensure they're still available."))
 
 (defn kill-task-if-possible [compute-cluster task-id]
-  "If compute cluster is nil, print a warning instead of killing the task"
+  "If compute cluster is nil, print a warning instead of killing the task. There are cases, in particular,
+  lingering tasks, stragglers, or cancelled tasks where the task might outlive the compute cluster it was
+  member of. When this occurs, the looked up compute cluster is null and trying to kill via it would cause an NPE,
+  when in really, its relatively innocuous. So, we have this wrapper to use in those circumstances."
   (if compute-cluster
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -48,7 +48,10 @@
     "Called when offers are not processed to ensure they're still available."))
 
 (defn kill-task-if-possible [compute-cluster task-id]
-  "If compute cluster is nil, print a warning instead of killing the task"
+  "If compute cluster is nil, print a warning instead of killing the task. There are cases, in particular,
+  lingering tasks, stragglers, or cancelled tasks where the task might outlive the compute cluster it was
+  member of. When this occurs, the looked up compute cluster is null and trying to kill via it would cause an NPE,
+  when in really, its relatively innocuous. So, we have this wrapper to use in those circumstances."
```
dposada:
```suggestion
  when in reality, it's relatively innocuous. So, we have this wrapper to use in those circumstances."
```

dposada:
Will merge when green

scheduler/src/cook/compute_cluster.clj
dposada: Still seems like we should fix this

scrosby: Odd. How'd it get reverted?

dposada: The apostrophe is still missing

scrosby: Now added.

PR #1293, opened 2019-11-04
Add logs fileserver

nsinkov:

- add a python file server that replicates part of the Mesos `files` endpoint API

We need access to job log files on Kubernetes to replicate the experience on Mesos. The Mesos `files` endpoint API is also used by the `cs` CLI


scheduler/src/cook/rest/api.clj
```diff
@@ -965,20 +966,16 @@
...
-    (when sandbox-directory
-      (str "http://" agent-hostname ":5051" "/files/read.json?path="
-           (URLEncoder/encode sandbox-directory "UTF-8")))
-    (catch Exception e
-      (log/debug e "Unable to retrieve directory path for" task-id "on agent" agent-hostname)
-      nil)))
+   Refer to the 'Using the output_url' section in docs/scheduler-rest-api.adoc for further details.
+   Delegates to the compute cluster implimentation."
+  [instance-entity]
+  (if-let [sandbox-url (:instance/sandbox-url instance-entity)]
```
dposada: Is `:instance/sandbox-url` a new field?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -300,6 +301,25 @@
...
+                                 {:sandbox-url sandbox-url
+                                  :task-id task-id})))
+             [instance] (first (q '[:find ?i
+                                    :in $ ?task-id
+                                    :where
+                                    [?i :instance/task-id ?task-id]]
+                                  db task-id))]
+         (if (nil? instance)
+           (log/error (str "Sandbox file server URL update error. No instance for task-id " task-id))
+           @(d/transact conn [[:db/add instance :instance/sandbox-url sandbox-url]])))
```
dposada: Why do we need `:instance/sandbox-url`? Don't we already have `output-url`?

scheduler/src/cook/rest/api.clj
nsinkov: yes

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: we don't. that's computed on the fly

fileserver/cook/__main__.py
```diff
@@ -0,0 +1,50 @@
...
+    if args is None:
+        args = sys.argv[1:]
+    try:
+        print(sys.version)
+        port, workers = (args + [None] * 2)[0:2]
+        if port is None:
+            logging.error('Must provide port')
+            sys.exit(1)
+        FileServerApplication({
+            'bind': f'0.0.0.0:{port}',
```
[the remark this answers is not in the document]
nsinkov: I don't think so, until we have a reason to think we might leverage this feature

scheduler/src/cook/scheduler/scheduler.clj
dposada: Not sure if I understand, can you point me to the code?

scheduler/src/cook/compute_cluster.clj
```diff
@@ -48,7 +48,12 @@
     "Returns true if this compute cluster makes use of the Cook executor for running tasks")
 
   (container-defaults [this]
-    "Default values to use for containers launched in this compute cluster"))
+    "Default values to use for containers launched in this compute cluster")
+
+  (retrieve-sandbox-url-path [this instance-entity]
```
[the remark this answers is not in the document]
nsinkov: this is a new datomic field and output_url is synthesized from it

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -262,7 +276,7 @@
                                                             path))
                                             true)))
                                       cook-volumes)
-        host-path->name (pc/map-from-keys (fn [_] (str (UUID/randomUUID)))
+        host-path->name (pc/map-from-keys (fn [_] (str "syn-" (d/squuid)))
```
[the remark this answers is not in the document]
nsinkov: debugging, might come in handy some day and no downside

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -407,16 +453,82 @@
     ; "Guaranteed" QoS which requires limits for both memory and cpu.
     (.putLimitsItem resources "cpu" (double->quantity cpus))
     (.setResources container resources)
-    (.setVolumeMounts container (into [] (conj volume-mounts workdir-volume-mount)))
+    (.setVolumeMounts container (into [] (remove nil? (conj volume-mounts workdir-volume-mount init-volume-mount))))
```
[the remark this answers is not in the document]
nsinkov: trying to be more concise vs having more conditional logic just on the mounts I expect to be nil

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -407,16 +453,82 @@
...
+        (.setPort http-get-action (IntOrString. port))
+        (.setPath http-get-action "readiness-probe")
+        (.setHttpGet readiness-probe http-get-action)
+        (.setReadinessProbe container readiness-probe)
+
+        ; resources
+        (.putRequestsItem resources "cpu" (double->quantity cpu-request))
+        (.putLimitsItem resources "cpu" (double->quantity cpu-limit))
+        (.putRequestsItem resources "memory" (double->quantity (* memory-multiplier memory-request)))
+        (.putLimitsItem resources "memory" (double->quantity (* memory-multiplier memory-limit)))
```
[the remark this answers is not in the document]
nsinkov: converts memory to proper units - same multiplier used for the main container

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -538,7 +665,7 @@
           namespace (-> pod .getMetadata .getNamespace)
           ;; TODO: IF there's an error, log it and move on. We'll try again later.
           api (CoreV1Api. api-client)]
-      (log/info "Launching pod with name" pod-name "in namespace" namespace ":" pod)
+      (log/info "Launching pod with name" pod-name "in namespace" namespace ":" (Yaml/dump pod))
```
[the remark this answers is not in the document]
nsinkov: this has been indispensable already. I will fight to the death.

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: https://github.com/twosigma/Cook/pull/1293/files#diff-5dd58675d8adc5809600b0ede0bc4713L1013 - `url-path` IS `output-url`
used to be computed from just hostname on the fly

now the new `instance/sandbox-url` is stored in datomic.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -289,7 +290,22 @@
...
+    ; Hard codes fun stuff like the port we run the agent on.
+    ; Users will need to add the file path & offset to their query.
+    ; Refer to the 'Using the output_url' section in docs/scheduler-rest-api.adoc for further details."
+    [_ {:keys [instance/hostname instance/task-id instance/sandbox-directory]}]
+    (try
+      (when sandbox-directory
+        (str "http://" hostname ":5051" "/files/read.json?path="
+             (URLEncoder/encode sandbox-directory "UTF-8")))
+      (catch Exception e
+        (log/debug e "Unable to retrieve directory path for" task-id "on agent" hostname)
```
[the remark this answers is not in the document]
nsinkov: will make these changes in separate PR

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+#
+#  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+#  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+#  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+#  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+#  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+#  IN THE SOFTWARE.
+#
+"""Module implementing a file server to serve Cook job logs. """
```
DaoWen:
```suggestion
"""Module implementing the Mesos file access REST API to serve Cook job logs. """
```

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+import os
+from operator import itemgetter
+from pathlib import Path
+from stat import *
+
+import gunicorn.app.base
+from flask import Flask, jsonify, request, send_file
+from gunicorn.six import iteritems
```
DaoWen: Since this is Python 3.7 code, it doesn't make sense to import anything from `gunicorn.six`.

(`six` is a Python 2 / 3 compatibility module)

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+class FileServerApplication(gunicorn.app.base.BaseApplication):
+
+    def __init__(self, options=None):
+        self.options = options or {}
+        self.application = app
+        super(FileServerApplication, self).__init__()
+
+    def load_config(self):
+        config = dict([(key, value) for key, value in iteritems(self.options)
```
DaoWen:
```suggestion
        config = dict([(key, value) for key, value in self.options.items()
```

DaoWen: Python 2.7+ supports _dict comprehensions_; i.e., no need to create lists of pairs and convert those to dicts:

```
        config = {key: value for key, value in self.options.items()
                  if key in self.cfg.settings and value is not None}
```

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+                       if key in self.cfg.settings and value is not None])
+        for key, value in iteritems(config):
```
DaoWen:
```suggestion
        for key, value in config.items():
```

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+        return (None, err_message())
+    except Exception as _:
+        return (None, err_message())
+    return (int_val, None)
+
+
+@app.route('/files/read')
+@app.route('/files/read.json')
+def read():
+    path = request.args.get('path')
```
DaoWen: This needs to be sanitized to prevent reading files like `/var/spool/tickets/$USER`

DaoWen: Maybe helpful?
https://docs.python.org/3/library/os.path.html#os.path.normpath
https://docs.python.org/3/library/os.html#os.chroot

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+            self.cfg.set(key.lower(), value)
+
+    def load(self):
+        return self.application
+
+
+@app.route('/files/download')
+@app.route('/files/download.json')
+def download():
+    path = request.args.get('path')
```
DaoWen: Obviously all of these file paths need to be sanitized, not just the one in the `/files/read` endpoint.

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+    '7': 'rwx'
+}
+# maps permission bit masks to a linux permission string. e.g. 775 -> "rwxrwxr-x"
+permissions = {n - 512: ''.join(permission_strings[c] for c in str(oct(n))[-3:]) for n in range(512, 1024)}
+
+
+@app.route('/files/browse')
+@app.route('/files/browse.json')
+def browse():
+    path = request.args.get('path')
```
DaoWen: Obviously all of these file paths need to be sanitized, not just the one in the /files/read endpoint.

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+def try_parse_int(param_name, val):
+    err_message = lambda: f"Failed to parse {param_name}: Failed to convert '{val}' to number.\n"
+    try:
+        int_val = int(val)
+    except ValueError as _:
+        return (None, err_message())
+    except Exception as _:
+        return (None, err_message())
+    return (int_val, None)
```
DaoWen: Nitpick: It's considered more Pythonic to throw the error and catch it rather than returning a pair.

fileserver/cook/file_server.py
DaoWen: Actually, can't we just move this `if` condition into the loop below and get rid of this whole intermediate collection?

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+@app.route('/files/read')
+@app.route('/files/read.json')
+def read():
+    path = request.args.get('path')
+    offset_param = request.args.get('offset')
```
DaoWen:
```suggestion
    offset_param = request.args.get('offset', -1)
```
I'm pretty sure this is equivalent, and simplifies the logic below (no more `None` checks).

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+    if os.path.isdir(path):
+        return "Cannot read a directory.\n", 400
+    if offset is None or offset == -1:
+        return jsonify({
+            "data": "",
+            "offset": os.path.getsize(path),
+        })
+    f = open(path)
+    f.seek(offset)
+    max_length = int(os.getenv('COOK_FILE_SERVER_MAX_READ_LENGTH', '25000000'))
```
DaoWen: Let's promote this to a top-level global value rather than reevaluating it on each request.

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+
+permission_strings = {
+    '0': '---',
+    '1': '--x',
+    '2': '-w-',
+    '3': '-wx',
+    '4': 'r--',
+    '5': 'r-x',
+    '6': 'rw-',
+    '7': 'rwx'
```
DaoWen:
```suggestion
    '7': 'rwx',
```
You used trailing commas above

integration/tests/cook/test_basic.py
```diff
@@ -96,7 +96,6 @@ def test_uid(self):
         finally:
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
 
-    @unittest.skipIf(util.using_kubernetes(), 'Output url is not currently supported on kubernetes')
     def test_output_url(self):
```
DaoWen: Add a test cases for reading/downloading/browsing a well-known path outside the sandbox directory; e.g., `/etc/hosts`.

scheduler/bin/make-gke-test-clusters
```diff
@@ -22,7 +22,7 @@ fi
 
 PROJECT=$1
 ZONE=${2:-us-central1-a}
-CLUSTERNAME=${3:-$USER-test-cluster-$(date '+%Y%m%d-%H%M%S')}
```
DaoWen: Why did you delete the year?

scheduler/config-k8s.edn
```diff
@@ -49,6 +59,9 @@
            :user-metrics-interval-seconds 60}
  :nrepl {:enabled? true
          :port #config/env-int "COOK_NREPL_PORT"}
+ :pools {:default "k8s-alpha"
+         :job-resource-adjustment {:adjust-job-resources-fn cook.kubernetes.api/adjust-job-resources-fn
```
DaoWen: Is this for adding resources for running the filesever sidecar?

nsinkov: yes

scheduler/bin/make-gke-test-clusters
nsinkov: running into name length limit

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+                                   for path in [os.path.join(path, f)
+                                                for f in os.listdir(path)]]
+    ]
+    return jsonify(sorted(retval, key=itemgetter("path")))
+
+
+# This endpoint is used by the kubernetes readiness probe on the fileserver container. Cook will see that the
+# fileserver is ready to serve files and will set the output_url. If we expose the output_url before the server
+# is ready, then someone might use it and get an error.
+@app.route('/readiness-probe')
```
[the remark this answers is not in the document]
nsinkov: only used for readiness

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -342,6 +349,49 @@
...
+    BigDecimal/valueOf
+    (.add (BigDecimal/valueOf b))
+    .doubleValue))
+
+(defn adjust-job-resources-fn
+  "Given the required resources for a job, add to them the resources for any sidecars that will be launched with the job."
+  [{:keys [cpus mem] :as resources}]
+  (if-let [{:keys [cpu-request memory-request]} (-> (config/kubernetes) :sandbox-fileserver :resource-requirements)]
+    (assoc resources
+      :cpus (add-as-decimals cpus cpu-request)
```
DaoWen: I think the lowest precision a pod will accept for CPU is `0.001` (i.e., `1m`). If that's true, it might make more sense to just add the doubles and then round to the thousandths place.

scheduler/config-k8s.edn
```diff
@@ -34,7 +34,17 @@
...
+              :init-container {:command "echo test"
+                               :image "byrnedo/alpine-curl:latest"}
+              :sandbox-fileserver {:command "fileserver"
+                                   :image "twosigma/cook-fileserver:latest"
+                                   :port 23906
+                                   :resource-requirements {:cpu-request 0.1
+                                                           :cpu-limit 0.1
+                                                           :memory-request 128.0
+                                                           :memory-limit 128.0}}
+              :wrapper-script "/bin/sh -c "}
```
DaoWen: Is `:wrapper-script` the right name for this? Maybe `:launch-command` or `:init-command`?

Issues with `wrapper-script`:

1. The command doesn't have to be a script (it can be a binary)
2. It's not really a "wrapper" in the sense that the "real" script isn't wrapped _inside_, but rather passed in as an argument.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -390,8 +441,11 @@
 
     ; container
     (.setName container cook-container-name-for-job)
-    (.setCommand container
-                 ["/bin/sh" "-c" (:value command)])
+    (if wrapper-script
+      (.setCommand container
+                   ["/bin/sh" "-c" (str wrapper-script-workdir "/" wrapper-script " '" (str/escape (:value command) {\' "'\\''"}) "'")])
```
DaoWen: Why do we need `str/escape` here?

Why don't we just assume that the wrapper (launcher?) script is executable, then this becomes:

```clojure
[(str wrapper-script-workdir "/" wrapper-script) (:value command)]
```

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
DaoWen: This is a base10 vs base2 reconciliation, correct? Maybe we should give it a more descriptive name.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -407,16 +461,63 @@
...
     ; pod-spec
     (.addContainersItem pod-spec container)
+
+    ; init container
+    (when-let [{:keys [command image]} init-container]
+      (let [container (V1Container.)]
+        ; container
+        (.setName container cook-init-container-name)
+        (.setImage container image)
+        (.setCommand container ["/bin/sh" "-c" command])
```
DaoWen: Do we default to `/bin/sh` because that's also the default shell currently used on Mesos?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -407,16 +461,63 @@
...
+
+        (.setVolumeMounts container [(workdir-volume-mount-fn true) (fileserver-workdir-volume-mount-fn false)])
+        (.addContainersItem pod-spec container)))
+
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
     (when pool-name
       (.addTolerationsItem pod-spec (toleration-for-pool pool-name)))
-    (.setVolumes pod-spec (into [] (conj volumes workdir-volume)))
+    (.setVolumes pod-spec (into [] (remove nil? (conj volumes workdir-volume wrapper-script-workdir-volume fileserver-workdir-volume))))
```
DaoWen: `(into [] (remove nil? ...` → `(filterv some? ...`

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -490,6 +591,21 @@
...
+  health checks to the server to be more precise."
+  [^V1Pod pod]
+  (when pod
+    (let [^V1PodStatus pod-status (.getStatus pod)
+          container-statuses (.getContainerStatuses pod-status)
+          file-server-status (first (filter (fn [c] (= cook-container-name-for-file-server (.getName c)))
+                                            container-statuses))]
+      (if file-server-status
+        (if (.isReady file-server-status) :running :not-running)
+        :unknown))))
```
DaoWen:
```clojure
(cond
  (nil? file-server-status) :unknown
  (.isReady file-server-status) :running
  :else :not-running)
```

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
DaoWen: Why YAML rather than JSON? It seems like the line breaks in the YAML will destroy our logs.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -140,6 +144,25 @@
...
+        sandbox-fileserver-port (:port sandbox-fileserver)
+        sandbox-url (try
+                      (when (and sandbox-fileserver-port (not (str/blank? pod-ip)))
+                        (str "http://"
+                             (pod-ip->hostname-fn pod-ip)
+                             ":" sandbox-fileserver-port
+                             "/files/read.json?path="
+                             (URLEncoder/encode default-workdir "UTF-8")))
+                      (catch Exception e
+                        (log/debug e "Unable to retrieve directory path for" task-id)
```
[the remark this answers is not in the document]
DaoWen: Why would this throw an exception? Is it being thrown by the `URLEncoder`? Why would that happen?

Like @dposada said, it might be useful to have this logged higher than info-level.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -140,6 +143,25 @@
...
+                        nil))]
+    (scheduler/write-sandbox-url-to-datomic datomic/conn task-id sandbox-url)))
```
DaoWen: Do we actually need to write to datomic when the result was `nil`?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -214,10 +236,16 @@
...
+      (let [new-state {:pod new-pod
+                       :synthesized-state (api/pod->synthesized-pod-state new-pod)
+                       :sandbox-file-server-container-state (api/pod->sandbox-file-server-container-state new-pod)}
             old-state (get @existing-state-map pod-name)]
         ; We always store the updated state, but only reprocess it if it is genuinely different.
         (swap! existing-state-map assoc pod-name new-state)
+        (let [new-file-server-state (:sandbox-file-server-container-state new-state)
+              old-file-server-state (:sandbox-file-server-container-state old-state)]
+          (when (and (= new-file-server-state :running) (not= old-file-server-state :running))
+            (record-sandbox-url new-state)))
```
DaoWen: Is it possible for us to completely miss the `running` state and thus never record the sandbox url? Is there a way we could avoid that by being more defensive here?

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -289,7 +290,22 @@
   (use-cook-executor? [_] true)
 
   (container-defaults [_]
-    container-defaults))
+    container-defaults)
+
+  (retrieve-sandbox-url-path
+    ;"Constructs a URL to query the sandbox directory of the task.
```
DaoWen: This commented-out docstring is weird. Can we move it inside the body and wrap it in a `(comment ...)` instead?

scheduler/src/cook/mesos/mesos_compute_cluster.clj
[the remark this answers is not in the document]
DaoWen: Is it hard to do it in this PR? Where would the compute cluster name come from?

scheduler/bin/make-gke-test-clusters
DaoWen: what name length limit?

scheduler/src/cook/scheduler/scheduler.clj
DaoWen: Why did we switch to storing this in datomic if it can be easily computed from the hostname?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -300,6 +301,25 @@
          (catch Exception e
            (log/error e "Mesos scheduler status update error")))))
 
+(defn write-sandbox-url-to-datomic
+  "Takes a sandbox file server URL from the compute cluster and saves it to datomic."
+  [conn task-id sandbox-url]
+  (try (let [db (db conn)
```
DaoWen: Nitpick: add a newline after `try`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -300,6 +301,25 @@
...
+             _ (when-not task-id
+                 (throw (ex-info "task-id is nil. Something unexpected has happened."
```
DaoWen: Why not lift this outside the `try` block and just log the error rather than `throw`ing?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -300,6 +301,25 @@
...
+                                 {:sandbox-url sandbox-url
+                                  :task-id task-id})))
+             [instance] (first (q '[:find ?i
+                                    :in $ ?task-id
+                                    :where
+                                    [?i :instance/task-id ?task-id]]
+                                  db task-id))]
+         (if (nil? instance)
+           (log/error (str "Sandbox file server URL update error. No instance for task-id " task-id))
```
DaoWen:
```suggestion
           (log/error "Sandbox file server URL update error. No instance for task-id" task-id)
```
`str` isn't necessary since [`log/error`](http://clojure.github.io/tools.logging/#clojure.tools.logging/error) (and the other logging functions) behave like `println` (it joins all of the arguments with spaces).

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -449,34 +469,42 @@
   (getSoftConstraints [_] [])
   (taskGroupName [_] (str (:job/uuid job))))
 
+(def adjust-job-resources-for-pool-fn
```
DaoWen: Even though this isn't a `defn`, I still think it's weird to suffix the name with `-fn` for top-level definitions.

fileserver/cook/file_server.py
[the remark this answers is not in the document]
DaoWen: Why aren't we setting a liveness probe?

scheduler/src/cook/kubernetes/controller.clj
nsinkov: can use the YAML directly to spin up the same pod
When will YAML be an issue? it's fine in elastic

there's another task for the other refactoring

scheduler/src/cook/kubernetes/controller.clj
nsinkov: We'll revisit later

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: hostname is the node hostname and is not usable for the sandbox url

scheduler/src/cook/kubernetes/controller.clj
nsinkov: will not write nil

scheduler/src/cook/kubernetes/api.clj
nsinkov: will verify

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
nsinkov: no, this is just conversion to MB

scheduler/src/cook/kubernetes/api.clj
nsinkov: will call it custom shell and remove the escaping

scheduler/config-k8s.edn
```diff
@@ -34,7 +34,17 @@
...
  :hostname #config/env "COOK_HOSTNAME"
  :kubernetes {:disallowed-container-paths #{"/mnt/bad"}
-              :disallowed-var-names #{"BADVAR"}}
+              :disallowed-var-names #{"BADVAR"}
+              :init-container {:command "echo test"
+                               :image "byrnedo/alpine-curl:latest"}
+              :sandbox-fileserver {:command "fileserver"
+                                   :image "twosigma/cook-fileserver:latest"
+                                   :port 23906
+                                   :resource-requirements {:cpu-request 0.1
```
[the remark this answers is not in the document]
nsinkov: easier for destructuring in clojure and matches the java api

scheduler/bin/make-gke-test-clusters
nsinkov: cluster name length limit

fileserver/cook/file_server.py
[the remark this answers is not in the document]
nsinkov: will harden later

scheduler/config-k8s.edn
DaoWen: settled on `:custom-shell`

scheduler/src/cook/kubernetes/api.clj
DaoWen: Cook Executor is using /bin/sh, so yes, this is good.

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+    '1': '--x',
+    '2': '-w-',
+    '3': '-wx',
+    '4': 'r--',
+    '5': 'r-x',
+    '6': 'rw-',
+    '7': 'rwx',
+}
+# maps permission bit masks to a linux permission string. e.g. 775 -> "rwxrwxr-x"
+permissions = {n - 512: ''.join(permission_strings[c] for c in str(oct(n))[-3:]) for n in range(512, 1024)}
```
scrosby: Can I suggest this function instead:
def perm_bits_to_string(n):
    ''.join([yield ("rwxrwxrwx"[ii] if (n&(1<<(8-ii))!=0) else "-" for ii in range(0,8)])

scheduler/config-k8s.edn
```diff
@@ -49,6 +59,9 @@
            :user-metrics-interval-seconds 60}
  :nrepl {:enabled? true
          :port #config/env-int "COOK_NREPL_PORT"}
+ :pools {:default "k8s-alpha"
```
scrosby: Default is historically gamma, but I'm fine if we only test with pools from now on.

scheduler/src/cook/config.clj
```diff
@@ -429,8 +438,8 @@
                                   :default-pool "no-pool"}
                                  pool-selection)})))
      :kubernetes (fnk [[:config {kubernetes {}}]]
-                   (merge kubernetes
-                          {:default-workdir "/mnt/sandbox"}))}))
+                   (merge {:default-workdir "/mnt/sandbox"}
+                          (update kubernetes :pod-ip->hostname-fn resolve-optional-function identity)))}))
```
scrosby: This seems very weird. Like iits wrapping an inner configuration. I think I know whats going on here, you should deconstruct the kubernetes using fnk, instead of this update.

scheduler/src/cook/config.clj
```diff
@@ -393,7 +397,12 @@
                        ((util/lazy-load-var 'clojure.tools.nrepl.server/start-server) :port port)))
 
      :pools (fnk [[:config {pools nil}]]
-              pools)
+              (cond-> pools
+                (:job-resource-adjustment pools)
+                (update :job-resource-adjustment
```
scrosby: You should deconstruct the pools using fnk, instead of using this update approach.

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
dposada: I agree with @DaoWen

scheduler/src/cook/scheduler/scheduler.clj
dposada: Discussed offline with @DaoWen and @nsinkov, the plan is that we're going to store the `container-hostname` in Datomic and then generate the `output-url` from that.

scheduler/src/cook/config.clj
nsinkov: will refactor later

scheduler/src/cook/config.clj
nsinkov: will refactor later

scheduler/config-k8s.edn
nsinkov: will update

fileserver/cook/file_server.py
nsinkov: will update

fileserver/cook/file_server.py
```diff
@@ -0,0 +1,164 @@
...
+    if path is None:
+        return "Expecting 'path=value' in query.\n", 400
+    if not path_is_valid(path):
+        return "", 404
+    if not os.path.isdir(path):
+        return jsonify([])
+    retval = [
+        {
+            "gid": path_obj.group(),
+            "mode": ('d' if S_ISDIR(st.st_mode) else '-') + permissions[S_IMODE(st.st_mode) % 512],
```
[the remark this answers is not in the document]
nsinkov: will update

fileserver/cook/version.py
```diff
@@ -0,0 +1 @@
+VERSION = '0.0.9'
```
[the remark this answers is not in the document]
nsinkov: that many iterations

fileserver/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.7-alpine
+
+COPY . /app
```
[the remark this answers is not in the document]
nsinkov: will update

fileserver/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.7-alpine
```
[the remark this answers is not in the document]
nsinkov: depend on 3.7 features

PR #1296, opened 2019-11-06
Sends synthetic pods to k8s to trigger GKE scale-up

dposada:
## TL;DR

This PR allows k8s compute clusters in Cook to be configured to send "synthetic" pods in order to trigger the [GKE autoscaler](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler) to scale up [node pools](https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools) for pending Cook jobs that are not getting matched (i.e. all capacity is being consumed).

## Background

Cook matches available capacity with pending jobs and then launches pods corresponding to these matches in k8s. These pod specs have the [node name specified](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#nodename), since Cook is making the bin-packing decisions itself. This makes sense when targeting a static k8s cluster, but we need a different approach when targeting an autoscaling cluster like GKE.

## What’s new in this PR?

The only way to trigger the GKE autoscaler to scale up is by [sending it demand in the form of unassigned pods](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#how_cluster_autoscaler_works). Because we want to maintain control of the bin-packing process in Cook, we need to produce this demand by some means other than sending it the actual pods for our users’ jobs.

This PR introduces:

1. The abstract notion that a `ComputeCluster` can `autoscale!`
1. The `KubernetesComputeCluster` (KCC)-specific implementation, which is to generate pods that correspond to real users’ jobs that aren’t getting matched
1. The hook point in the main scheduler (`scheduler/trigger-autoscaling!`) where we trigger autoscaling based on un-matched (pending) jobs

### Synthetic pods

For each failed task request sent to a KCC, the KCC will generate a synthetic pod that has the exact same resource request as the corresponding real job. 

The differences are:

- It will use a different image
- It will run a different command
- It will run in a different namespace
- It will not have a node specified

The point of these synthetic pods is to trigger the GKE autoscaler to scale up node pools to accommodate the real jobs that are in the queue. Once new capacity opens up, the pending jobs will get matched against real offers and will be scheduled on the cluster.

Example of a synthetic pod triggering GKE scale-up:

```
pod triggered scale-up: [{https://content.googleapis.com/compute/v1/projects/ts-cook-gke-dev/zones/us-central1-a/instanceGroups/gke-dpo-test-cluster-cook-pool-k8s-ga-6b434f5c-grp 3->4 (max: 9)}]
```

Synthetic pods are similar to the [pause pods described here](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler).

### Configuration

The configuration for a KCC to generate synthetic pods lives in a new section of the existing compute cluster configuration:

```
{:factory-fn cook.kubernetes.compute-cluster/factory-fn
 :config {:compute-cluster-name "gke-1"
          :config-file "../scheduler/.cook_kubeconfig_1"
          :synthetic-pods {:command "exit 0"
                           :image "..."
                           :max-pods-outstanding 128
                           :user "..."
                           :pools #{"pool-a" "pool-b" "pool-c"}}}}
```

- `:command` is optional, and defaults to `"exit 0"`
- `:image` is required, specifies the Docker image to use for synthetic pods
- `:max-pods-outstanding` is required, specifies the maximum number of synthetic pods (on this compute cluster) that Cook will allow at a time
- `:user` is required, specifies the user (namespace) to launch the synthetic pods with
- `:pools` is required, and is the set of pool names that this KCC can autoscale for (i.e. a compute cluster can autoscale for a subset of the pools)

### Choosing the compute cluster

When presented with a collection of un-matched task requests for which we want to autoscale, we need to decide how to distribute those requests among the available compute clusters that are configured to autoscale for the pool in question. Currently we choose the compute cluster by taking a hash of the pending job’s UUID and mod-ing that with the number of candidate compute clusters. This is a simple, consistent scheme, but there’s certainly room to improve this distribution (see below). 

The compute cluster will not submit a new synthetic pod for a job which already has a synthetic pod pending or running on the cluster.

Note that if a compute cluster is added or removed from the Cook cluster configuration, the mapping from job to KCC will likely change, which means that we could briefly over-provision the GKE clusters (the KCCs don’t currently know about each others’ pods).

### Removal of the Cook sidecar from the synthetic pods

The synthetic pods are launched via the same code path as real pods, which means they currently get loaded up with a Cook sidecar for serving live logs. We obviously don’t want or need this, so we will have a follow-on to remove this from the synthetic pods.

## Alternatives considered

Another approach we considered was to generate "synthetic offers" inside of Cook. The basic idea is to synthesize offers when the queue of pending jobs grows, and those pending jobs can match against the synthesized offers. Obviously the offers wouldn't be associated with a real node, so the pod submissions would be sent to k8s without a node specified and they would therefore trigger the GKE autoscaler.

The reason we decided not to use this approach are:

1. There are [constraints users can specify on their jobs and groups](https://github.com/twosigma/Cook/blob/master/scheduler/docs/groups.md#host-placement-constraints) which depend on host information in Cook's matching logic in order to be satisfied.
1. As mentioned above, using "pause pods" / "dummy pods" is an existing, common design pattern for triggering the autoscaler.


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
 
   (restore-offers [this pool-name offers])
 
+  (autoscaling? [_]
```
scrosby: Shouldn't this be checked in the factory? If someone puts in a negative number, they're doing something really really weird. That way you just check it once, and return true/false here.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -142,43 +144,51 @@
                                          (.getName container-status))))
        first))
 
+(defn synthetic-pod-job-uuid
```
scrosby: Should be synthetic-pod->job-uuid
Also V1Pod annotation on pod.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+            num-synthetic-pods (count outstanding-synthetic-pods)
+            {:keys [image user command max-pods-outstanding] :or {command "exit 0"}} synthetic-pods-config]
+        (log/info "In" name "compute cluster, there are" num-synthetic-pods
+                  "outstanding synthetic pod(s), and a max of" max-pods-outstanding "are allowed")
+        (if (>= num-synthetic-pods max-pods-outstanding)
+          (log/info "In" name "compute cluster, cannot launch more synthetic pods")
+          (let [using-pools? (config/default-pool)
+                synthetic-task-pool-name (if using-pools? pool-name nil)
+                new-task-requests (remove (fn [{:keys [job]}]
+                                            (some #(= (-> job :job/uuid str) (controller/synthetic-pod-job-uuid %))
```
scrosby: factor out (:job/uuid str)

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+                                                  outstanding-synthetic-pods))
+                                          task-requests)
+                task-metadata-seq (->>
+                                    new-task-requests
+                                    (map (fn [{:keys [job]}]
+                                           {:task-id (str (UUID/randomUUID))
```
scrosby: Don't use UUID/randomUUID

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -10,6 +10,8 @@
            (io.kubernetes.client.models V1Pod V1ContainerStatus V1PodStatus)
            (java.net URLEncoder)))
 
+(def cook-synthetic-pod-job-uuid-label "cook-synthetic-pod-for-job-uuid")
```
scrosby: Kubernetes likes to have a naming scheme of reverse-fqdn, with dot as a separator. We're using cook.FOO elsewhere.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
 
   (restore-offers [this pool-name offers])
 
+  (autoscaling? [_]
+    (and
+      (some-> synthetic-pods-config :image count pos?)
+      (some-> synthetic-pods-config :user count pos?)
+      (some-> synthetic-pods-config :max-pods-outstanding pos?)))
+
+  (autoscale! [this pool-name task-requests]
```
scrosby: As a bigger picture comment, we do not want to name these pods with with their task-id, then we'll find mysterious task-id's in the log. That's goint to be confusing diagnostics. Much better to indicate it in the pod name directly.

E.g., instead of synthesizing labels here, send a 'synthetic pod' flag to task-metadata->pod. Have it syntehsize the labels and have it explicitly name a pod with "syn-UUID", that way we can identify when we see them enter the state machine because of their unique name. That way if I see the name, I know where it comes from.... and not to find it in the database. 


database. My suggestion is when we're

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -764,11 +764,46 @@
                                        {:job-uuid->reserved-host (apply dissoc job-uuid->reserved-host matched-job-uuids)
                                         :launched-job-uuids (into matched-job-uuids launched-job-uuids)})))
 
+(defn distribute-task-requests-to-compute-clusters
```
scrosby: Buggy. This logic assumes that a given pool is in all compute clusters. 

What if a job's UUID happens to map to a compute cluster that cannot launch that pool. We'll auto-scale but be unable to ever launch.

I see a few choices here:
1. Detect if we're stuck and log with an error --- so if this occurs, we have a breadcrumb.
2. Scan all of the compute clusters and their nodes and remember for each pool what compute clusters have nodes in that pool, and distribute offers to only those pools. (this also depends that we have min-nodes-1 for any given pool on a compute cluster)

scheduler/config-k8s.edn
```diff
@@ -16,14 +16,21 @@
...
          ; (like test_default_container_volumes) will fail without this set.
          :run-as-user "root"}
+ :autoscaler {:interval-seconds 180}
  :compute-clusters [{:factory-fn cook.kubernetes.compute-cluster/factory-fn
                      :config {:compute-cluster-name "gke-1"
                               ;; Location of the kubernetes config file. Hardcoded to the location specified by bin/make-gke-test-cluster
-                              :config-file "../scheduler/.cook_kubeconfig_1"}}
+                              :config-file "../scheduler/.cook_kubeconfig_1"
+                              :synthetic-pods {:image "gcr.io/google-containers/alpine-with-bash:1.0"
+                                               :max-pods-outstanding 128
```
shamsimam: In addition to max outstanding pods, should we limit the outstanding cpu/mem so that we do not scale up to too many nodes (GKEs tendency to 'spread' the load across new nodes doesn't help).

scheduler/src/cook/compute_cluster.clj
```diff
@@ -44,6 +44,12 @@
   (restore-offers [this pool-name offers]
     "Called when offers are not processed to ensure they're still available.")
 
+  (autoscaling? [this]
+    "Returns true if this compute cluster should autoscale to satisfy pending jobs")
```
shamsimam: We should have just the one function `autoscale!` which is a no-op when the compute cluster does not support scaling.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -454,7 +454,7 @@
                      env))
                  (:environment command))
         resources (V1ResourceRequirements.)
-        labels {cook-pod-label compute-cluster-name}
+        labels (merge pod-labels {cook-pod-label compute-cluster-name})
```
shamsimam: Nitpick: use `assoc` instead of `merge`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+              (str "In " name " compute cluster, request to autoscale despite invalid / missing config"))
+      (let [outstanding-synthetic-pods (->> @all-pods-atom
+                                            (all-pods this)
+                                            (filter controller/synthetic-pod-job-uuid))
+            num-synthetic-pods (count outstanding-synthetic-pods)
+            {:keys [image user command max-pods-outstanding] :or {command "exit 0"}} synthetic-pods-config]
+        (log/info "In" name "compute cluster, there are" num-synthetic-pods
+                  "outstanding synthetic pod(s), and a max of" max-pods-outstanding "are allowed")
+        (if (>= num-synthetic-pods max-pods-outstanding)
+          (log/info "In" name "compute cluster, cannot launch more synthetic pods")
```
shamsimam: Should we take the top `max-pods-outstanding` from the `outstanding-synthetic-pods` pods to process in this scenario?

scheduler/src/cook/kubernetes/compute_cluster.clj
shamsimam: Nitpick: In addition, we should use a UUID generator that has a well-known prefix/suffix that makes it easy to identify synthetic pods.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -339,7 +389,8 @@
            bearer-token-refresh-seconds
            namespace
            scan-frequency-seconds
-           max-pods-per-node]
+           max-pods-per-node
+           synthetic-pods]
```
shamsimam: Nitpick: rename to `synthetic-pods-config`

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -142,43 +144,51 @@
...
 (defn handle-pod-started
   "A pod has started. So now we need to update the status in datomic."
   [compute-cluster {:keys [pod]}]
-  (let [instance-id (-> pod .getMetadata .getName)
-        ; We leak mesos terminology here ('task') because of backward compatibility.
-        status {:task-id {:value instance-id}
-                :state :task-running}]
-    (write-status-to-datomic compute-cluster status)
-    {:cook-expected-state :cook-expected-state/running}))
+  (when-not (synthetic-pod-job-uuid pod)
```
shamsimam: Do we log anywhere that the synthetic pod started running?

scheduler/src/cook/scheduler/scheduler.clj
shamsimam: Did we do any experiment to validate this distribution function? 

If the function has been validated, as part of future PRs:
- Maybe include smarter heuristics to avoid mapping jobs to compute-clusters we know will never schedule the jobs (e.g. k8s jobs to a mesos cluster). Maybe the compute cluster api needs to support a `pool-supported?` function.
- We should preferentially target pools (known from the job) which are low on resources as those are the ones which most likely need scaling up.

Out of curiosity, what scale's down the resources?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -764,11 +764,46 @@
...
+  "Given a collection of un-matched task requests and a collection of
+  compute clusters, distributes the task requests amongst the compute
+  clusters, using a hash of the pending job's uuid."
```
shamsimam: Docstring should mention it is returning a `compute-cluster->jobs` map. That is the api any future improvements need to stick to.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -764,11 +764,46 @@
...
+  [task-requests compute-clusters]
+  (group-by (fn [{:keys [job]}]
+              (nth compute-clusters
+                   (-> job :job/uuid hash (mod (count compute-clusters)))))
+            task-requests))
+
+(let [pool-name->last-autoscale-atom (atom {})]
+  (defn trigger-autoscaling!
```
shamsimam: This function is low on logs, like when the iteration starts, when autocsaling is invoked on a cluster and when we are done with an iteration. We should also add metrics for these, e.g. time spent by each compute cluster in autoscaling. Number of jobs sent to each compute cluster, etc.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -764,11 +764,46 @@
...
+  "Given a collection of un-matched task requests and a collection of
+  compute clusters, distributes the task requests amongst the compute
+  clusters, using a hash of the pending job's uuid."
+  [task-requests compute-clusters]
+  (group-by (fn [{:keys [job]}]
+              (nth compute-clusters
+                   (-> job :job/uuid hash (mod (count compute-clusters)))))
+            task-requests))
+
+(let [pool-name->last-autoscale-atom (atom {})]
```
shamsimam: Prefer passing `pool-name->last-autoscale-atom` as an argument, make it easier to test.

shamsimam: Nitpick: rename to `pool-name->last-autoscale-time-atom`

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do, thanks

scheduler/src/cook/kubernetes/controller.clj
dposada: Good call, will do, thanks

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do, thanks

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Good idea, will do

scheduler/src/cook/kubernetes/controller.clj
dposada: Ah, thanks, will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Good idea, will do

scheduler/src/cook/scheduler/scheduler.clj
dposada: @scrosby - Thanks! Yep, I agree that the distribution function needs to evolve, as described in the PR description (did you read it?). I would vote for this PR to keep it simple so that we can start to get value from autoscaling, and evolve it as those needs come up.

@shamsimam - You bring up a good point, which is that this currently will scale for Mesos pools in the KCCs, which is a bug. Thanks! I will address this.

scheduler/config-k8s.edn
dposada: @rmanyari Thoughts on this?

scheduler/src/cook/compute_cluster.clj
dposada: Currently using this function to filter down the list of compute clusters to those that can autoscale.

scheduler/src/cook/kubernetes/api.clj
dposada: Thanks, will do

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+      (some-> synthetic-pods-config :user count pos?)
+      (some-> synthetic-pods-config :max-pods-outstanding pos?)))
+
+  (autoscale! [this pool-name task-requests]
+    (try
+      (assert (cc/autoscaling? this)
+              (str "In " name " compute cluster, request to autoscale despite invalid / missing config"))
+      (let [outstanding-synthetic-pods (->> @all-pods-atom
+                                            (all-pods this)
+                                            (filter controller/synthetic-pod-job-uuid))
```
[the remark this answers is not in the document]
dposada: What about it? I am not seeing it.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+      (some-> synthetic-pods-config :image count pos?)
+      (some-> synthetic-pods-config :user count pos?)
+      (some-> synthetic-pods-config :max-pods-outstanding pos?)))
+
+  (autoscale! [this pool-name task-requests]
+    (try
+      (assert (cc/autoscaling? this)
+              (str "In " name " compute cluster, request to autoscale despite invalid / missing config"))
+      (let [outstanding-synthetic-pods (->> @all-pods-atom
+                                            (all-pods this)
```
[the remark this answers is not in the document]
dposada: Good call, will rename and add docstring

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: This logs when the number of outstanding synthetic pods is already >= the max.

We are actually doing what you describe below.

```
(take (- max-pods-outstanding num-synthetic-pods))
```

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do, thanks

scheduler/src/cook/kubernetes/controller.clj
dposada: Yes, it will get logged by the state machine when the state transition happens

scheduler/src/cook/scheduler/scheduler.clj
dposada: Good call, thanks

scheduler/src/cook/scheduler/scheduler.clj
dposada: Good call, will do

scheduler/src/cook/scheduler/scheduler.clj
dposada: Will do, thanks

scheduler/config-k8s.edn
rmanyari: I think number of pods is fine for now. For a v2, I'd consider breaking down this value by pool. Essentially, you want a higher value for pools that grow large (say spot) and smaller for others, to avoid too much over-provisioning.

scheduler/src/cook/kubernetes/compute_cluster.clj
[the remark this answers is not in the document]
shamsimam:
```
(->> first-line
  second-line-two-spaces-in)
```

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Use d/squuid, as is used for the other tasks.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Still worth a TODO here. Is there an open ticket.

scheduler/src/cook/kubernetes/compute_cluster.clj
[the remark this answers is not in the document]
dposada: Ah, OK, I think I need to configure Cursive to do that :)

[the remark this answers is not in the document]
dposada: This style guide disagrees:

https://github.com/bbatsov/clojure-style-guide#threading-macros

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Actually, I think we want to keep this one as is. This is the key from the config, so it's nice for it to just be `:synthetic-pods`.

scheduler/src/cook/kubernetes/compute_cluster.clj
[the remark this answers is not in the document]
shamsimam: Oh! I'm sure we saw the other variant somewhere and started using it :'(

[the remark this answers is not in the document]
dposada: Interesting. I'm fine either way, just haven't yet found a way to change this in my Cursive config -- do you know how?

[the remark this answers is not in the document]
shamsimam: Using `Alt+Enter` on the form, instructions - https://cursive-ide.com/userguide/formatting.html
I can also show you tomorrow morning.

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Moved this to the factory, thanks

scheduler/src/cook/kubernetes/controller.clj
dposada: Renamed and added type annotation

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done, we're now using `synthetic-{pool}-{d/squuid}`

scheduler/src/cook/kubernetes/controller.clj
dposada: Done, now using `twosigma.com/cook-scheduler-synthetic-pod-job-uuid` to match the existing `twosigma.com/cook-scheduler-job`

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: We're now using `synthetic-{pool}-{d/squuid}`

scheduler/src/cook/scheduler/scheduler.clj
dposada: Both issues mentioned above are now fixed -- the KCC's `:synthetic-pods` config now includes a `:pools` set, which specifies the set of pools for which the KCC should autoscale.

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -242,6 +243,55 @@
...
+  (autoscaling? [_]
+    (and
+      (some-> synthetic-pods-config :image count pos?)
+      (some-> synthetic-pods-config :user count pos?)
+      (some-> synthetic-pods-config :max-pods-outstanding pos?)))
+
+  (autoscale! [this pool-name task-requests]
+    (try
+      (assert (cc/autoscaling? this)
+              (str "In " name " compute cluster, request to autoscale despite invalid / missing config"))
```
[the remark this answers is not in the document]
dposada: Using the `autoscaling?` function to filter down the list of compute clusters to those that can autoscale in the given pool.

scheduler/src/cook/kubernetes/compute_cluster.clj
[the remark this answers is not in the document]
dposada: Renamed to `add-starting-pods`

scheduler/src/cook/scheduler/scheduler.clj
dposada: Done

scheduler/src/cook/scheduler/scheduler.clj
dposada: Added logs and metrics

scheduler/src/cook/scheduler/scheduler.clj
dposada: Ended up removing this. Since each compute cluster has `:max-pods-outstanding`, and we won't submit more than 1 synthetic pod for each un-matched task, I don't think we need to additionally throttle this based on time.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -414,6 +418,13 @@
     (.setEffect toleration "NoSchedule")
     toleration))
 
+(def toleration-for-deletion-candidate-of-autoscaler
+  (let [^V1Toleration toleration (V1Toleration.)]
+    (.setKey toleration "DeletionCandidateOfClusterAutoscaler")
+    (.setOperator toleration "Exists")
+    (.setEffect toleration "PreferNoSchedule")
+    toleration))
```
shamsimam: Nitpick: a good place to use [doto](https://clojuredocs.org/clojure.core/doto)

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -593,6 +604,13 @@
 
     (.setNodeName pod-spec hostname)
     (.setRestartPolicy pod-spec "Never")
+    
+    (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
```
shamsimam: Some comments on what this achieves would be appreciated.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
...
+                                         synthetic-pods)
+          waiting-synthetic-pods-in-pool (filter (fn [^V1Pod pod]
+                                                   (let [pod-spec (.getSpec pod)
+                                                         tolerations (.getTolerations pod-spec)]
+                                                     (some (fn [^V1Toleration toleration]
+                                                             (and (= (.getKey toleration) "cook-pool")
+                                                                  (= (.getValue toleration) pool-name)))
+                                                           tolerations)))
+                                                 waiting-synthetic-pods)]
+      (count waiting-synthetic-pods-in-pool))
```
shamsimam: Nitpick: I like naming the anonymous functions so that it is easy to locate during debugging.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
...
+    0))
+
+(defn total-resource
+  "Given a map from node-name->resource->amount and a resource,
+  returns the total amount of that resource for all nodes."
+  [node-name->resource-map resource]
```
shamsimam: Nitpick: rename `resource` to `resource-keyword` to make it clear what the function is doing in the `map` below

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
...
+  (->> node-name->resource-map vals (map resource) (reduce +)))
+
+(defn counter
```
shamsimam: Nitpick: rename to `retrieve-compute-cluster-counter`?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
                  "compute cluster, unable to get node from node name" node-name)
       false)))
 
+(defn num-waiting-synthetic-pods
```
shamsimam: Nitpick: rename to `compute-num-waiting-synthetic-pods`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
-      (log/info "Generated" (count offers-this-pool) "offers for pool" pool-name "in compute cluster" name
-                     (into [] (map #(into {} (select-keys % [:hostname :resources])) offers-this-pool)))
-      offers-this-pool))
+    (let [pods (add-starting-pods this @all-pods-atom)
+          num-waiting-synthetic-pods (num-waiting-synthetic-pods this pods pool-name)
+          nodes @current-nodes-atom]
+      (log/info "In" name "compute cluster, got asked for pending offers for pool" pool-name
+                {:total-pods (count pods)
+                 :synthetic-pods num-waiting-synthetic-pods
+                 :nodes (count nodes)})
```
shamsimam: Nitpick: lexicographic order :)

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
+      (if (pos? num-waiting-synthetic-pods)
+        (do
```
shamsimam: We should limit how long we avoid generating offers rather than being open to the possibility of waiting indefinitely because synthetic pods didn't get scheduled.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
+                                            (add-starting-pods this)
+                                            (filter controller/synthetic-pod->job-uuid))
+            num-synthetic-pods (count outstanding-synthetic-pods)
+            {:keys [image user command max-pods-outstanding] :or {command "exit 0"}} synthetic-pods-config]
+        (log/info "In" name "compute cluster, there are" num-synthetic-pods
+                  "outstanding synthetic pod(s), and a max of" max-pods-outstanding "are allowed")
+        (if (>= num-synthetic-pods max-pods-outstanding)
+          (log/info "In" name "compute cluster, cannot launch more synthetic pods")
+          (let [using-pools? (config/default-pool)
+                synthetic-task-pool-name (if using-pools? pool-name nil)
```
shamsimam:
```suggestion
                synthetic-task-pool-name (when using-pools? pool-name)
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
+                new-task-requests (remove (fn [{{:keys [job/uuid]} :job}]
+                                            (some #(= (str uuid) (controller/synthetic-pod->job-uuid %))
+                                                  outstanding-synthetic-pods))
+                                          task-requests)
+                task-metadata-seq (->>
+                                    new-task-requests
+                                    (map (fn [{{:keys [job/uuid] :as job} :job}]
+                                           {:task-id (str "synthetic-" pool-name "-" (d/squuid))
+                                            :command {:user user :value command}
```
shamsimam: As mentioned earlier it would be great if the command output the id of the job for which this synthetic pod is created. We can include an environment variable for that job id and then configure the command in the setting to be an `echo` message of the environment variable.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -331,6 +445,19 @@
...
+  [compute-cluster-name synthetic-pods-config]
+  (when synthetic-pods-config
+    (when-not (and
+                (-> synthetic-pods-config :image count pos?)
+                (-> synthetic-pods-config :user count pos?)
+                (-> synthetic-pods-config :max-pods-outstanding pos?)
+                (-> synthetic-pods-config :pools set?)
+                (-> synthetic-pods-config :pools count pos?))
+      (throw (ex-info (str "In " compute-cluster-name " compute cluster, invalid synthetic pods config")
+                      synthetic-pods-config)))))
```
shamsimam: Can we use `schema` for the validation or pre-conditions? It will make it easier to detect which validation failed.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -284,6 +284,10 @@
     (async/go
       (async/>! (pool->offers-chan pool-name) offers)))
 
+  (autoscaling? [_ _] false)
+
+  (autoscale! [_ _ _])
```
shamsimam: We should include a warning or error log here that this function was invoked.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -767,11 +770,54 @@
...
+                    "un-matched task(s), distributed to compute clusters as follows:"
+                    (->> compute-cluster->task-requests
+                         (pc/map-keys cc/compute-cluster-name)
+                         (pc/map-vals count)))
+          (doseq [[compute-cluster requests-for-cluster] compute-cluster->task-requests]
+            (histograms/update! (histograms/histogram
+                                  (metric-title "autoscale!-tasks"
+                                                pool-name
+                                                compute-cluster))
+                                (count requests-for-cluster))
```
shamsimam: Nitpick:
```suggestion
(histograms/update!
  (histograms/histogram (metric-title "autoscale!-tasks" pool-name compute-cluster))
  (count requests-for-cluster))
```

scheduler/config-k8s.edn
```diff
@@ -20,11 +20,21 @@
                      :config {:compute-cluster-name "gke-1"
                               ;; Location of the kubernetes config file. Hardcoded to the location specified by bin/make-gke-test-cluster
                               :config-file "../scheduler/.cook_kubeconfig_1"
+                              :synthetic-pods {:image "gcr.io/google-containers/alpine-with-bash:1.0"
+                                               :max-pods-outstanding 128
+                                               :user #config/env "USER"
+                                               :command "exit 0"
```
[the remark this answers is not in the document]
dposada: The pod is labeled with the job UUID that it corresponds to

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -414,6 +418,13 @@
     (.setEffect toleration "NoSchedule")
     toleration))
 
+(def toleration-for-deletion-candidate-of-autoscaler
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/kubernetes/api.clj
dposada: Good call, will do

scheduler/src/cook/kubernetes/api.clj
dposada: Thanks, will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Why `retrieve`? I'm not sure if it's helpful. And we're already in the `compute-cluster` namespace.

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Good call, will do

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: We should be able to map these back to their jobs with the pod label

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Will do

scheduler/src/cook/mesos/mesos_compute_cluster.clj
dposada: Will do

scheduler/src/cook/scheduler/scheduler.clj
dposada: Is this just for the formatting? I think the line was too long the way you suggested.

shamsimam: Yes, it will be formatted similar to the timer metric call below.

scheduler/src/cook/kubernetes/compute_cluster.clj
shamsimam: It'd be nice if the function name began with a verb.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -767,11 +770,54 @@
...
+                   (-> job :job/uuid hash (mod (count compute-clusters)))))
+            task-requests))
+
+(defn trigger-autoscaling!
+  "Autoscales the given pool to satisfy the given match failures (i.e. pending jobs), if:
+  - There is at least one failure
+  - There is at least one compute cluster configured to do autoscaling"
+  [failures pool-name compute-clusters]
+  (try
+    (let [task-requests (map #(.. (first %) (getRequest)) failures)
```
scrosby: I"m not sure if the logic belongs here or above (see comments in pending-offers, but:

1. We should probably autoscale at a longer granularity than every offer cycle. If it takes 3 minutes to launch a node, there's not much point autoscaling much more rapidly than that. The system (k8s) can't respond that fast. Thus, autoscale maybe every 20 offer cycles. this lessens the impact of waiting for our synthetic pods to be launched by k8s (See comment in pending-offers). (I prefer to go by offer cycle, that way the overhead is constant  fraction of the work done.) If offer cycle is every 10 seconds, 20 cycles is 3 minutes, comparable to node launch time.

2. We don't generate a synthetic pod for each unmatched job. We give ourselves some hysteresis. E.g., We only generate synthetic pods for the first half of the outstanding queue. E.g., min(max-pods-outstanding, count(failures)/2),

3. If autoscaling cycle is ~3 minutes (the time I assume to launch a node), then we can autoscale by a max of 200 pods every 3 minutes. If we had 5k jobs suddenly added, it would take 25 cycles, or about 120 minutes to scale up to hit it.  We might want to consider increasing maximum of max-pods-oustanding to 500 or 1000.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -213,25 +221,27 @@
...
-  [compute-cluster pod-name k8s-actual-state & {:keys [reason]}]
-  (let [{:keys [status exit-code]} (calculate-pod-status compute-cluster pod-name k8s-actual-state :reason reason)]
-    (write-status-to-datomic compute-cluster status)
-    (when exit-code
-      (sandbox/aggregate-exit-code (:exit-code-syncer-state compute-cluster) pod-name exit-code))
-    ; Must never return nil, we want it to return non-nil so that we will retry with writing the state to datomic in case we lose a race.
-    ; Being in the (completed,*) state, will cause us to delete the pod, transitioning to (completed,missing), and thence
-    ; to deleting from the map, into (missing,missing) state.
-    {:cook-expected-state :cook-expected-state/completed}))
+  [compute-cluster pod-name {:keys [pod] :as k8s-actual-state} & {:keys [reason]}]
```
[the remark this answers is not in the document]
dposada: Good ideas. Will do, thanks.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -484,7 +495,7 @@
 
 (defn ^V1Pod task-metadata->pod
   "Given a task-request and other data generate the kubernetes V1Pod to launch that task."
-  [namespace compute-cluster-name {:keys [task-id command container task-request hostname]}]
+  [namespace compute-cluster-name {:keys [task-id command container task-request hostname pod-labels]}]
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
...
+                                           (or (-> pod
+                                                   .getStatus
+                                                   nil?)
+                                               (-> pod
+                                                   api/pod->synthesized-pod-state
+                                                   :state
+                                                   (= :pod/waiting))))
+                                         synthetic-pods)
+          waiting-synthetic-pods-in-pool (filter (fn [^V1Pod pod]
+                                                   (let [pod-spec (.getSpec pod)
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -31,18 +34,65 @@
...
+    0))
+
+(defn total-resource
+  "Given a map from node-name->resource->amount and a resource,
+  returns the total amount of that resource for all nodes."
+  [node-name->resource-map resource]
+  (->> node-name->resource-map vals (map resource) (reduce +)))
+
+(defn counter
+  "Given a metric name and a compute cluster name, returns a counter metric."
```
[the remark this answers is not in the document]
dposada: Good call, will do

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
+                                            (filter controller/synthetic-pod->job-uuid))
+            num-synthetic-pods (count outstanding-synthetic-pods)
+            {:keys [image user command max-pods-outstanding] :or {command "exit 0"}} synthetic-pods-config]
+        (log/info "In" name "compute cluster, there are" num-synthetic-pods
+                  "outstanding synthetic pod(s), and a max of" max-pods-outstanding "are allowed")
+        (if (>= num-synthetic-pods max-pods-outstanding)
+          (log/info "In" name "compute cluster, cannot launch more synthetic pods")
+          (let [using-pools? (config/default-pool)
+                synthetic-task-pool-name (when using-pools? pool-name)
+                new-task-requests (remove (fn [{{:keys [job/uuid]} :job}]
```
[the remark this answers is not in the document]
dposada: Good point, will do

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -96,17 +96,24 @@
...
-  [{:keys [name] :as compute-cluster} pod-name]
+  [{:keys [name] :as compute-cluster} pod-name ^V1Pod pod]
   (log/info "In compute cluster" name ", pod" pod-name "submission failed")
-  (let [instance-id pod-name
-        status {:reason :reason-task-invalid
-                :state :task-failed
-                :task-id {:value instance-id}}]
-    (write-status-to-datomic compute-cluster status)
-    {:cook-expected-state :cook-expected-state/completed}))
+  (when-not (synthetic-pod->job-uuid pod)
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -96,17 +96,24 @@
...
+  "If the given pod is a synthetic pod for autoscaling, returns the job uuid
+  that the pod corresponds to (stored in a pod label). Otherwise, returns nil."
+  [^V1Pod pod]
+  (some-> pod .getMetadata .getLabels (.get api/cook-synthetic-pod-job-uuid-label)))
+
 (defn handle-pod-submission-failed
   "Marks the corresponding job instance as failed in the database and
   returns the `completed` cook expected state"
-  [{:keys [name] :as compute-cluster} pod-name]
+  [{:keys [name] :as compute-cluster} pod-name ^V1Pod pod]
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -767,11 +770,54 @@
...
+(defn distribute-task-requests-to-compute-clusters
+  "Given a collection of un-matched task requests and a collection of
+  compute clusters, distributes the task requests amongst the compute
+  clusters, using a hash of the pending job's uuid. Returns a
+  compute-cluster->task-request map. That is the API any future
+  improvements need to stick to."
+  [task-requests compute-clusters]
+  (group-by (fn [{:keys [job]}]
+              (nth compute-clusters
+                   (-> job :job/uuid hash (mod (count compute-clusters)))))
```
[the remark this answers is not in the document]
dposada: What do you mean by active compute clusters?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -767,11 +770,54 @@
...
+            task-requests))
+
+(defn trigger-autoscaling!
+  "Autoscales the given pool to satisfy the given match failures (i.e. pending jobs), if:
+  - There is at least one failure
```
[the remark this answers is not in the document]
dposada: Good point. I would vote to make this a TODO and address in a subsequent PR. If you feel strongly about doing it in this PR, let me know.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -231,18 +284,79 @@
...
+                                            (some #(= (str uuid) (controller/synthetic-pod->job-uuid %))
+                                                  outstanding-synthetic-pods))
+                                          task-requests)
+                task-metadata-seq (->>
+                                    new-task-requests
+                                    (map (fn [{{:keys [job/uuid] :as job} :job}]
+                                           {:task-id (str "synthetic-" pool-name "-" (d/squuid))
```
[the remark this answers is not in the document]
dposada: Sounds good to me! Will do.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -103,8 +103,11 @@
 (timers/deftimer [cook-mesos scheduler generate-user-usage-map-duration])
 
 (defn metric-title
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/scheduler/scheduler.clj
dposada: @rmanyari - Thoughts on this?

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: In general, I agree. I think in this case it's nice that the name matches that of the underlying function.

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
dposada: Added a `def` and a docstring, hopefully this helps explain.

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/api.clj
dposada: Added the comment to the `def` for the string constant

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Changed it so that only "recent" synthetic pods count, the idea being that if a synthetic pod was created > X seconds ago (e.g. X might be 180), then we stop considering it here as a "waiting" synthetic pod, which means it will no longer hold up offer generation.

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
dposada: It's actually part of the k8s autoscaler, not GKE-specific:

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/deletetaint/delete.go#L38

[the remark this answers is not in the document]
scrosby: GOt it. LGTM. I'll let @shamsimam  mark this as resolved if he's happy with it.

PR #1297, opened 2019-11-06
Add more logging when we join on the mesos driver

scrosby:

- Improve the logging so that we unambiguously log when the driver join! unblocks.

Help debug/understand this code for our leadership changes/refactors.

### DIFF
--- a/scheduler/src/cook/mesos/mesos_compute_cluster.clj
+++ b/scheduler/src/cook/mesos/mesos_compute_cluster.clj
@@ -273,7 +273,7 @@
           (mesomatic.scheduler/join! driver)
           (log/info "In" compute-cluster-name "compute cluster, mesomatic scheduler join unblocked; we appear to have lost driver connection")
           (catch Exception e
-            (log/warn e "Lost mesos driver with exception")
+            (log/warn e "In" compute-cluster-name "compute cluster, lost mesos driver with exception")
             e)
           (finally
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -271,7 +271,9 @@
           ; scheduler/join! is a blocking call which returns or throws when the driver loses it's connection to mesos.
           ; Run this in an async thread which will deliver either the Status on a normal exit, or an exception if thrown.
           (mesomatic.scheduler/join! driver)
+          (log/info "Join unblocked. We appear to have lost driver connection")
```
dposada:
```suggestion
          (log/info "In" compute-cluster-name "compute cluster, mesomatic scheduler join unblocked; we appear to have lost driver connection")
```

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -271,7 +271,9 @@
...
           (catch Exception e
+            (log/warn e "Lost mesos driver with exception")
```
dposada:
```suggestion
            (log/warn e "In" compute-cluster-name "compute cluster, lost mesos driver with exception")
```

dposada:
@scrosby Should I go ahead and merge or do you want to wait for the build?

PR #1299, opened 2019-11-07
Adds `COOK_TEST_DEFAULT_SUBMIT_POOL` env var

dposada:

- allowing `COOK_TEST_DEFAULT_SUBMIT_POOL` to dictate the pool to use on submission, if the calling path does not set one
- ensuring that test users have the default share and quota 

We don't always want to use Cook Scheduler's configured default pool for our test runs.

The share and quota fixes are needed because some tests can leave share/quota at small values if they terminate unexpectedly.

### DIFF
--- a/integration/tests/cook/test_impersonation.py
+++ b/integration/tests/cook/test_impersonation.py
@@ -119,6 +119,11 @@ class ImpersonationCookTest(util.CookTest):
         user1 = self.user_factory.new_user()
         job_uuids = []
         try:
+            with self.user_factory.admin():
+                # Reset the user's share and quota
+                util.set_limit_to_default(self.cook_url, 'share', user1.name)
+                util.set_limit_to_default(self.cook_url, 'quota', user1.name)
+            
             # normal user can self-impersonate
             with user1.impersonating(user1):
                 job_uuid, resp = util.submit_job(self.cook_url, command='sleep 1')
--- a/integration/tests/cook/test_impersonation.py
+++ b/integration/tests/cook/test_impersonation.py
@@ -119,9 +119,6 @@ class ImpersonationCookTest(util.CookTest):
         user1 = self.user_factory.new_user()
         job_uuids = []
         try:
-            with self.user_factory.admin():
-                util.reset_share_and_quota(self.cook_url, user1.name)
-            
             # normal user can self-impersonate
             with user1.impersonating(user1):
                 job_uuid, resp = util.submit_job(self.cook_url, command='sleep 1')
--- a/integration/tests/cook/test_impersonation.py
+++ b/integration/tests/cook/test_impersonation.py
@@ -121,8 +121,8 @@ class ImpersonationCookTest(util.CookTest):
         try:
             with self.user_factory.admin():
                 # Reset the user's share and quota
-                util.set_limit_to_default(self.cook_url, 'share', user1.name)
-                util.set_limit_to_default(self.cook_url, 'quota', user1.name)
+                util.set_limit_to_default(self.cook_url, 'share', user1.name, util.default_submit_pool())
+                util.set_limit_to_default(self.cook_url, 'quota', user1.name, util.default_submit_pool())
             
             # normal user can self-impersonate
             with user1.impersonating(user1):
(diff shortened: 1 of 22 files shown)


integration/tests/cook/test_impersonation.py
```diff
@@ -119,6 +119,9 @@ def test_self_impersonate(self):
         user1 = self.user_factory.new_user()
         job_uuids = []
         try:
+            with self.user_factory.admin():
```
scrosby: This change seems unrelated to the PR. Split off?

integration/tests/cook/test_multi_user.py
```diff
@@ -45,6 +45,9 @@ def test_job_delete_permission(self):
 
     def test_group_delete_permission(self):
         user1, user2 = self.user_factory.new_users(2)
+        with self.user_factory.admin():
```
scrosby: This change seems unrelated to the PR. Split off?

integration/tests/cook/util.py
```diff
@@ -490,6 +490,10 @@ def minimal_group(**kwargs):
     return dict(uuid=str(make_temporal_uuid()), **kwargs)
 
 
+def default_submit_pool():
```
scrosby: Should have a comment here. Including saying why its OK for response to be None. Also consider inlining this into the caller to make the Noneness of a potential response more visible.

integration/tests/cook/test_impersonation.py
scrosby: Why is this code in this specific location, in this test? This looks to be setup/reset code that belongs somewhere else.

integration/tests/cook/test_multi_user.py
scrosby: Why is this code in this specific location, in this test? This looks to be setup/reset code that belongs somewhere else.

integration/tests/cook/util.py
dposada: Good call, added a docstring.

integration/tests/cook/test_multi_user.py
dposada: Yep, this is unrelated to the default submit pool change. That being said, I saw failures for these tests on this branch, so it seems appropriate to fix them on this branch.

As for why it's here, the resetting needs to happen for the specific user the test is using, which is determined in the test.

integration/tests/cook/test_impersonation.py
dposada: Yep, this is unrelated to the default submit pool change. That being said, I saw failures for these tests on this branch, so it seems appropriate to fix them on this branch.

As for why it's here, the resetting needs to happen for the specific user the test is using, which is determined in the test.

scrosby: Thanks for explanation. 

Are the test failures caused by the intended code changes in this branch?

integration/tests/cook/test_multi_user.py
scrosby: Thanks for explanation.

Are the test failures caused by the intended code changes in this branch?

integration/tests/cook/test_impersonation.py
dposada: Nope, they were caused by a test run that was aborted and left a test user's quota in a too-low state.

integration/tests/cook/test_multi_user.py
dposada: Nope, they were caused by a test run that was aborted and left a test user's quota in a too-low state.

integration/tests/cook/test_impersonation.py
scrosby: *hmms* this still smells a bit off. Shouldn't we have something that resets the whole cluster to a known good state, if this can go wrong?

dposada: Good call, moved these calls to the `UserFactory`

integration/tests/cook/test_multi_user.py
dposada: Moved these calls to the `UserFactory`

dposada:
@scrosby Resolved, thanks

PR #1300, opened 2019-11-07
Workaround bug in running with minimesos.

scrosby:
- Hardcode an earlier python image into the executor build

We do testing in minimesos, which is unmaintained and 2 years old.
Our cook-executor docker image is based on python. As of 6 days ago,
its been updated. It now requires a version of glibc that is too new
and unsupported in minimesos's images. Temporary fix to force a specific
python image.

### DIFF
--- a/executor/Dockerfile.build
+++ b/executor/Dockerfile.build
@@ -1,4 +1,4 @@
-FROM python:3.5
+FROM python:3.5.9-stretch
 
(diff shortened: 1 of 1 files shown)


dposada:
Will merge when green

scrosby:
Internalgreen on mesos. GKE seems flaky.

PR #1301, opened 2019-11-08
Tweaks GKE setup scripts

dposada:

- adding project arg to `help-delete-temporary-clusters`
- adding `--project` to all `glcoud` invocations
- adding a comment with prerequisites to the `make-` scripts
- removing the data local server from the k8s config

The GKE setup script changes are so that the scripts respect the project passed in at the top level.

The data local server removal is because we don't run with the data local server configured in any environment, so there's no need to configure it here.

### DIFF
--- a/scheduler/bin/help-delete-temporary-clusters
+++ b/scheduler/bin/help-delete-temporary-clusters
@@ -1,19 +1,23 @@
 #!/usr/bin/env bash
 
-# Usage: ./bin/help-delete-temporary-clusters <zone>
+# Usage: ./bin/help-delete-temporary-clusters <project> <zone>
 #   Delete all temporary clusters within a zone. The sibling scripts here mark clusters they create.
 #   This is intended to be used by other scripts and not directly.
+#   <project> is a gcloud project.
 #   <zone> can be a zone. E.g., us-central1-a
 
 
 set -e
 
-ZONE=$1
+PROJECT=$1
+ZONE=$2
+
+gcloud="gcloud --project $PROJECT"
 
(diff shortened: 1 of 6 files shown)


scheduler/bin/help-delete-temporary-clusters
```diff
@@ -1,19 +1,23 @@
...
+PROJECT=$1
+ZONE=$2
+
+gcloud="gcloud --project $PROJECT"
 
 # Nuke all existing temporary clusters; don't want to keep on making more idle clusters each time you invoke this.
 echo "---- Deleting any existing temporary clusters."
-gcloud container clusters list --filter 'resourceLabels.longevity=temporary'
-for i in `gcloud container clusters list --filter 'resourceLabels.longevity=temporary' --format="value(name)"`
+$gcloud container clusters list --filter 'resourceLabels.longevity=temporary'
```
scrosby: If we're setting this up with automatic names and expecting a project to be shared by multiple users, shouldn't this be changed to
#1: Label the cluster with the user in question.
#2: Only delete clusters by the same user
#3: Consider prompting a user if they want to delete the clusters before deleting them. (with a -f option for automated purposes)
#4: Delete in parallel. Its pretty slow.

dposada: I agree, those all sound like good ideas.

My goal for this PR was just to get the scripts working, without any functionality improvements. As is, they're not respecting the project being passed in at the top level, so many of the gcloud commands fail. This PR gets them to where they run successfully using the project provided.

PR #1302, opened 2019-11-08
Additional file to .dockerignore.

scrosby:

- Ignore a jar when building docker images.

The cook jar is built within the docker image, so there's no need to copy
the for-local-invocation cook jar into the docker image. It just bloats it.

### DIFF
--- a/scheduler/.dockerignore
+++ b/scheduler/.dockerignore
@@ -6,6 +6,7 @@ bin
 classes
 datomic/datomic*/data
 datomic/datomic*/log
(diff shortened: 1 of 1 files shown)


scrosby:
Yes. However, that file is created when we uberjar in the Dockerfile, and is copied into place afterwards. https://github.com/twosigma/Cook/blob/master/scheduler/Dockerfile#L49

shamsimam:
Awaiting green build and appropriate label.

PR #1303, opened 2019-11-08
Remove long-dead dependency.

scrosby:

- Remove a long-dead dependency

This dependency has been dead since 2016 when federation was removed. Its serialization code.


shamsimam:
Waiting on green build and appropriate label.

PR #1308, opened 2019-11-12
WIP: Test running mesos + kubernetes compute clusters

scrosby:

- Get our tests to pass with mesos or kubernetes within one scheduler invocation.

Test kubernetes + mesos within one cook. This is planned as a throwaway PR, to test the integration test infrastructure to determine what changes are needed, there, to support kubernetes+mesos.

The good news: This version does pass integration testing, running locally (with integration testing on both k8s and mesos, run within a single launch.)

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -68,6 +68,10 @@ class CookTest(util.CookTest):
             self.logger.info(f'Exit code not checked because cook executor was not used for {instance}')
 
     @unittest.skipUnless(util.docker_tests_enabled(), 'requires docker')
+    @pytest.mark.scheduler_not_in_docker
+    # If the cook scheduler is running in a docker container, it won't be able to lookup UID's or GID's. Under those circumstances,
+    # the cook scheduler won't be able to validate that the docker container is running as the right UID/GID and it will fail.
+    # Thus, in such an integration testing environment, we need to disable this test.
     def test_uid(self):
         settings = util.settings(self.cook_url)
 
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -724,7 +724,7 @@ class CookTest(util.CookTest):
     def test_get_job(self):
         # schedule a job
         job_spec = util.minimal_job()
-        resp = util.session.post('%s/rawscheduler' % self.cook_url, json={'jobs': [job_spec]})
+        _, resp = util.submit_jobs(self.cook_url, [job_spec])
         self.assertEqual(201, resp.status_code, msg=resp.content)
 
         # query for the same job & ensure the response has what it's supposed to have
@@ -1973,6 +1973,7 @@ class CookTest(util.CookTest):
         finally:
             util.kill_jobs(self.cook_url, uuids)
 
+    @unittest.skipIf(util.has_ephemeral_hosts(), util.EPHEMERAL_HOSTS_SKIP_REASON)
     def test_attribute_equals_hostname_constraint(self):
         max_slave_cpus = util.max_node_cpus()
         task_constraint_cpus = util.task_constraint_cpus(self.cook_url)
@@ -2142,7 +2143,7 @@ class CookTest(util.CookTest):
                 pool_usage = usage_data['pools'][pool['name']]
                 self.assertEqual(set(pool_usage.keys()), {'total_usage', 'grouped', 'ungrouped'}, pool_usage)
                 self.assertEqual(set(pool_usage['ungrouped'].keys()), {'running_jobs', 'usage'}, pool_usage)
-            default_pool = util.default_pool(self.cook_url)
+            default_pool = util.default_submit_pool() or util.default_pool(self.cook_url)
             if default_pool:
                 # If there is a default pool configured, make sure that our jobs,
                 # which don't have a pool, are counted as being in the default pool
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -800,11 +800,11 @@ class MultiUserCookTest(util.CookTest):
             resp = util.reset_limit(self.cook_url, 'share', user, reason=None)
             self.assertEqual(resp.status_code, 400, resp.text)
 
-            default_pool = util.default_pool(self.cook_url)
+            default_pool = util.default_submit_pool() or util.default_pool(self.cook_url)
             if default_pool is not None:
                 for limit in ['quota', 'share']:
                     # Get the default cpus limit
-                    resp = util.get_limit(self.cook_url, limit, "default")
+                    resp = util.get_limit(self.cook_url, limit, "default", pool=default_pool)
                     self.assertEqual(200, resp.status_code, resp.text)
                     self.logger.info(f'The default limit in the {default_pool} pool is {resp.json()}')
                     default_cpus = resp.json()['cpus']
@@ -814,7 +814,7 @@ class MultiUserCookTest(util.CookTest):
                     self.assertEqual(resp.status_code, 201, resp.text)
 
                     # Check that the limit is returned for no pool
-                    resp = util.get_limit(self.cook_url, limit, user)
+                    resp = util.get_limit(self.cook_url, limit, user, pool=default_pool)
                     self.assertEqual(resp.status_code, 200, resp.text)
                     self.assertEqual(100, resp.json()['cpus'], resp.text)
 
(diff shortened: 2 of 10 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -720,7 +724,7 @@ def test_memory_limit_exceeded_mesos_script(self):
     def test_get_job(self):
         # schedule a job
         job_spec = util.minimal_job()
-        resp = util.session.post('%s/rawscheduler' % self.cook_url, json={'jobs': [job_spec]})
+        resp = util.session.post('%s/jobs' % self.cook_url, json={'jobs': [job_spec], 'pool' : util.default_submit_pool()})
```
dposada: We should just do:

```suggestion
        _, resp = util.submit_job(self.cook_url)
```

`submit_job` takes care of using the correct endpoint, submitting a `minimal_job()`, and using the `default_submit_pool()`.

integration/tests/cook/test_multi_user.py
```diff
@@ -116,14 +116,14 @@ def test_job_cpu_quota(self):
...
-                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=0)
+                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=0, pool=util.default_submit_pool())
                 self.assertEqual(resp.status_code, 201, resp.text)
             with user:
                 _, resp = util.submit_job(self.cook_url)
                 self.assertEqual(resp.status_code, 422, msg=resp.text)
             # User with tiny quota can't submit bigger jobs, but can submit tiny jobs
             with admin:
-                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=0.25)
+                resp = util.set_limit(self.cook_url, 'quota', user.name, cpus=0.25, pool=util.default_submit_pool())
```
dposada: Seems like we should just make the default for all these limit-setting calls when no pool is specified be `default_submit_pool()`

integration/tests/cook/test_multi_user.py
```diff
@@ -883,10 +883,10 @@ def test_queue_endpoint(self):
         job_spec = {'group': group['uuid'],
                     'command': 'sleep 30',
                     'cpus': util.max_cpus()}
-        uuids, resp = util.submit_jobs(self.cook_url, job_spec, clones=100, groups=[group])
+        uuids, resp = util.submit_jobs(self.cook_url, job_spec, clones=100, groups=[group], pool=util.default_submit_pool())
```
dposada: This is the default behavior

integration/tests/cook/util.py
```diff
@@ -1635,16 +1640,17 @@ def _get_compute_cluster_factory_fn():
...
-    return _get_compute_cluster_factory_fn() == 'cook.mesos.mesos-compute-cluster/factory-fn'
-
+    return os.getenv("COOK_TEST_COMPUTE_CLUSTER_TYPE","mesos") == "mesos"
 
 def has_one_agent():
-    return node_count() == 1
-
+    if using_kubernetes():
+        # Kubernetes always has more than one agent.
+        return False
```
dposada: Why?

integration/tests/cook/util.py
```diff
@@ -1252,18 +1260,20 @@ def default_pool(cook_url):
 def all_pools(cook_url):
     """Returns the list of all pools that exist"""
     resp = session.get(f'{cook_url}/pools')
-    return resp.json(), resp
+    disallow_pools_regex = os.getenv('COOK_TEST_DISALLOW_POOLS_REGEX')
+    all_pools = resp.json() if disallow_pools_regex is None else \
+        [p for p in resp.json() if not re.match(disallow_pools_regex, p['name'])]
+    if disallow_pools_regex is not None:
+        for ii in resp.json():
+            logger.info(f"For Pool {ii['name']} we have {re.match(disallow_pools_regex, ii['name'])}")
```
dposada: I would vote to either remove this log or make it a little more clear, e.g.:

```suggestion
            logger.info(f"For pool {ii['name']}, matching against the disallowed pools regex returns {re.match(disallow_pools_regex, ii['name'])}")
```

integration/tests/cook/util.py
```diff
@@ -1635,12 +1645,11 @@ def _get_compute_cluster_factory_fn():
...
 
 def using_kubernetes():
-    return get_kubernetes_compute_cluster() is not None
+    return os.getenv("COOK_TEST_COMPUTE_CLUSTER_TYPE", "mesos") == "kubernetes"
 
 
 def using_mesos():
-    return _get_compute_cluster_factory_fn() == 'cook.mesos.mesos-compute-cluster/factory-fn'
-
+    return os.getenv("COOK_TEST_COMPUTE_CLUSTER_TYPE", "mesos") == "mesos"
```
dposada: Seems like we should extract this into a function:

```
def compute_cluster_type_under_test():
    return os.getenv("COOK_TEST_COMPUTE_CLUSTER_TYPE", "mesos")
````

File: scheduler/Dockerfile [hunk 7]
```diff
@@ -45,8 +44,13 @@ COPY config* /opt/cook/
 COPY resources /opt/cook/resources
 COPY java /opt/cook/java
 COPY src /opt/cook/src
+
 RUN lein uberjar
 RUN cp "target/cook-$(lein print :version | tr -d '"').jar" datomic-free-0.9.5394/lib/cook-$(lein print :version | tr -d '"').jar
+COPY .cook_kubeconfig_* /opt/cook/
+
+# Ugly hack. Our .cook_kubeconfig lookup assumes it can be found in ../scheduler/ so make a symlink
+RUN ln -s /opt/cook /opt/scheduler
```
dposada: Aren't these the same two lines as above?

integration/tests/cook/conftest.py
```diff
@@ -35,6 +35,7 @@ def _ssh_check(user):
     """
     hostname = socket.gethostname()
     logging.info(f'Checking ssh as {user} to {hostname}')
+    print(f'Checking ssh as {user} to {hostname}')
```
dposada: Remove?

scheduler/bin/run-local-composite.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
+
+# Usage: ./bin/run-local.sh
```
dposada: This doesn't match the filename

scheduler/bin/run-local-composite.sh
```diff
@@ -0,0 +1,82 @@
...
+    keytool -genkeypair -keystore ${COOK_KEYSTORE_PATH} -storetype PKCS12 -storepass cookstore -dname "CN=cook, OU=Cook Developers, O=Two Sigma Investments, L=New York, ST=New York, C=US" -keyalg RSA -keysize 2048
+fi
+
+if ! [ -x "$(command -v flask)" ]; then
+    echo "Please install flask"
+    exit 1
+fi
+
+INTEGRATION_DIR="$(dirname ${SCHEDULER_DIR})/integration"
+FLASK_APP=${INTEGRATION_DIR}/src/data_locality/service.py flask run -p 35847 &
```
dposada: Can we ditch the data locality service? It's not used in any real environment.

scheduler/config-composite.edn
```diff
@@ -0,0 +1,100 @@
...
+                              :config-file "../scheduler/.cook_kubeconfig_2"}}
+                    {:factory-fn cook.mesos.mesos-compute-cluster/factory-fn
+                     :config {:failover-timeout-ms nil ; When we close the instance of Cook, all its tasks are killed by Mesos
+                              :master #config/env "MESOS_MASTER"
+                              :framework-id #config/env "COOK_FRAMEWORK_ID"
+                              :compute-cluster-name "local-mesos"}}]
+ :cors-origins ["https?://cors.example.com"]
+ :data-local {:fitness-calculator {:cache-ttl-ms 60000
+                                   :cost-endpoint #config/env "DATA_LOCAL_ENDPOINT"
+                                   :update-interval-ms 1000}}
```
dposada: Again, I would vote to ditch this.

integration/tests/cook/test_basic.py
scrosby: Nice.

integration/tests/cook/test_multi_user.py
scrosby: Done.

integration/tests/cook/test_multi_user.py
scrosby: Yup. Nuked.

integration/tests/cook/util.py
scrosby: Simplifies this code. I hate that we run these tests in so many environments and randomly disable subsets of tests. No production environment will have only a single node.

integration/tests/cook/util.py
scrosby: Good idea.

scheduler/bin/run-local-composite.sh
scrosby: FIxed.

scheduler/bin/run-local-composite.sh
scrosby: Yes, but lets do it as another PR; maybe something for next week?

PR #1321, opened 2019-11-27
Makes several tests more reliable

dposada:

See comments inline with changes.

See comments inline with changes.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1690,7 +1690,8 @@ class CookTest(util.CookTest):
             self.cook_url,
             command='cat /.dockerenv',
             container={'type': 'DOCKER',
-                       'docker': {'image': image}})
+                       'docker': {'image': image}},
+            max_retries=5)
         self.assertEqual(resp.status_code, 201)
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertIn('success', [i['status'] for i in job['instances']])
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -675,7 +675,7 @@ class CookTest(util.CookTest):
             job_details = f"Job details: {json.dumps(job, sort_keys=True)}"
             self.assertEqual('failed', job['state'], job_details)
             self.assertLessEqual(1, len(job['instances']), job_details)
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -675,7 +675,7 @@ def memory_limit_exceeded_helper(self, command, executor_type, mem=128):
             job_details = f"Job details: {json.dumps(job, sort_keys=True)}"
             self.assertEqual('failed', job['state'], job_details)
             self.assertLessEqual(1, len(job['instances']), job_details)
-            instance = job['instances'][-1]
+            instance = next(i for i in job['instances'] if i['reason_code'] in [2002, 99003])
```
dposada: Instead of grabbing the "last" instance, grab the instance with the expected instance code.

We should never assume that a particular instance (by index) of a job is the one we're interested in.

integration/tests/cook/test_basic.py
```diff
@@ -1064,6 +1064,11 @@ def test_list_jobs_by_pool(self):
             jobs.append(util.load_job(self.cook_url, job_uuid))
             job_uuid, resp = util.submit_job(self.cook_url, pool=pool_name, name=name, command='exit 0')
             self.assertEqual(201, resp.status_code)
+            # We wait for the job to start running, and only then start waiting for it to
+            # complete; otherwise we could get a false-negative on wait_for_job 'completed'
+            # because of a scheduling delay. Having two separate waits also disambiguates
+            # the root cause of a wait-timeout failure.
+            util.wait_for_job_in_statuses(self.cook_url, job_uuid, ['completed', 'running'])
```
dposada: Wait for the job to start running, and only then start waiting for it to complete; otherwise we could get a false-negative on wait_for_job 'completed' because of a scheduling delay.

integration/tests/cook/test_basic.py
```diff
@@ -1690,7 +1695,8 @@ def test_basic_docker_job(self):
             self.cook_url,
             command='cat /.dockerenv',
             container={'type': 'DOCKER',
-                       'docker': {'image': image}})
+                       'docker': {'image': image}},
+            max_retries=5)
```
dposada: Let the job retry, so that a single failed instance does not cause the test to fail.

PR #1322, opened 2019-11-27
Prevents errors in launch-tasks from one compute cluster affecting others

dposada:

- catching inside the `doseq` that iterates over compute clusters calling `launch-tasks`

We don't want an error launching tasks in one compute cluster to cause other compute clusters to not launch tasks.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -674,7 +674,7 @@
   [matches conn db fenzo mesos-run-as-user pool-name]
   (let [matches (map #(update-match-with-task-metadata-seq % db mesos-run-as-user) matches)
         task-txns (matches->task-txns matches)]
-    (log/info "Writing tasks" task-txns)
+    (log/info "In" pool-name "pool, writing tasks" task-txns)
     ;; Note that this transaction can fail if a job was scheduled
     ;; during a race. If that happens, then other jobs that should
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -686,13 +686,12 @@
...
           (log/warn e
-                    "Transaction timed out, so these tasks might be present"
+                    "In" pool-name "pool, transaction timed out, so these tasks might be present"
                     "in Datomic without actually having been launched in Mesos"
                     matches)
           (throw e))))
-    (log/info "Launching" (count task-txns) "tasks")
+    (log/info "In" pool-name "pool, launching" (count task-txns) "tasks")
     (ratelimit/spend! ratelimit/global-job-launch-rate-limiter ratelimit/global-job-launch-rate-limiter-key (count task-txns))
-    (log/debug "Matched tasks" task-txns)
```
nsinkov: why don't you want this logged?

dposada: Because it's already being logged before this:

https://github.com/twosigma/Cook/pull/1322/files#diff-6c16cf7360b42776defb5ddcfa4d6f95R677

dposada: Not sure if that link works well, but it's being logged with this line:

```
(log/info "In" pool-name "pool, writing tasks" task-txns)
```

on line 677 of the patched file.

PR #1323, opened 2019-12-03
improves time to build user->quota-fn by aggregating queries

shamsimam:

- improves time to build `user->quota-fn` by aggregating queries (gives around **4X** improvement in my tests)
- improves time to build user->quota-fn by aggregating count quota query (gives another **3X to 5X** improvement in my tests)

`create-user->quota-fn` (used in `rank-jobs`) makes lots of individual queries against datomic to build the `user->quota-cache`. Improve the execution time of `create-user->quota-fn` ends up improving the time spent ranking jobs.

### DIFF
--- a/scheduler/src/cook/quota.clj
+++ b/scheduler/src/cook/quota.clj
@@ -14,7 +14,8 @@
 ;; limitations under the License.
 ;;
 (ns cook.quota
-  (:require [cook.pool :as pool]
+  (:require [clojure.set :as set]
+            [cook.pool :as pool]
             [cook.schema]
             [cook.tools :as util]
             [datomic.api :as d]
@@ -159,6 +160,33 @@
     (->> (d/q query db type pool-name' requesting-default-pool)
       (into {}))))
 
+(defn- retrieve-count-quota-from-user-entities
+  "Returns the user->count-quota for all count quota specified on the user entity."
+  [db users]
+  (let [query '[:find ?u ?c
+                :in $ [?u ...]
+                :where
+                [?e :quota/user ?u]
+                [?e :quota/count ?c]]]
+    (->> (d/q query db users)
+      (into {}))))
+
+(defn- retrieve-user->count-quota
+  "Returns the map for user to count-quota for all the provided users."
+  [db pool-name all-users default-quota]
+  (let [user->resource-count-quota (pool-name->type->user->quota db pool-name :count)
+        remaining-users (vec (set/difference (set all-users) (set (keys user->resource-count-quota))))
+        user->entity-count-quota (when (seq remaining-users)
+                                   (retrieve-count-quota-from-user-entities db remaining-users))]
+    (pc/map-from-keys
+      (fn [user]
+        (int
+          ; prefer resource over the field on the user for count quota
+          (or (get user->resource-count-quota user)
+              (get user->entity-count-quota user)
+              default-quota)))
+      all-users)))
+
 (defn create-user->quota-fn
   "Returns a function which will return the quota same as `(get-quota db user)`
    snapshotted to the db passed in. However, it queries for all users with quota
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
                           count-field
                           (conj [:db/retract [:quota/user user] :quota/count count-field]))))))
 
+(defn- pool-name->type->user->quota
```
shamsimam: @scrosby I didn't see much performance benefit from the aggregating quota types change and also didn;t like how that that code ended up. So leaving the resource type aggregation out.

scrosby: Sounds good. We don't want to make this code harder to understand for only a little benefit.

scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
...
+        remaining-users (vec (set/difference (set all-users) (set (keys user->resource-count-quota))))
+        user->entity-count-quota (when (seq remaining-users)
+                                   (retrieve-count-quota-from-user-entities db remaining-users))]
+    (pc/map-from-keys
+      (fn [user]
+        (int
+          ; prefer resource over the field on the user for count quota
+          (or (get user->resource-count-quota user)
+              (get user->entity-count-quota user)
+              default-quota)))
```
dposada: Let's either duplicate the comment from `get-quota` or refer to it

scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
...
+                [?e :quota/user ?u]
+                [?e :quota/resource ?r]
+                [?r :resource/type ?t]
+                [?r :resource/amount ?a]
+                [(cook.pool/check-pool $ ?r :resource/pool ?pool-name ?requesting-default-pool)]]]
+    (->> (d/q query db type pool-name' requesting-default-pool)
+      (into {}))))
+
+(defn- retrieve-count-quota-from-user-entities
+  "Returns the user->count-quota for all count quota specified on the user entity."
```
scrosby: What is the user->count-quota? Is this the legacy implementation for representing a quota of running jobs? I.e., distinguish this from the next function? Why is this function?
(This is a documentation/clojuredoc improvement request.)

scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
...
+  [db users]
+  (let [query '[:find ?u ?c
+                :in $ [?u ...]
+                :where
+                [?e :quota/user ?u]
+                [?e :quota/count ?c]]]
+    (->> (d/q query db users)
+      (into {}))))
+
+(defn- retrieve-user->count-quota
```
scrosby: Inconsistent naming with prior function?

scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
...
-                               (map (fn [user]
-                                      [user (get-quota db user pool-name)]))
-                               ;; In case default-user doesn't have an explicit quota
-                               (cons [default-user (get-quota db default-user pool-name)])
-                               (into {}))]
+  (let [default-type->quota (get-quota db default-user pool-name)
+        default-count-quota (get default-type->quota :count)
+        all-resource-types (util/get-all-resource-types db)
+        type->user->quota (pc/map-from-keys #(pool-name->type->user->quota db pool-name %) all-resource-types)
+        all-quota-users (d/q '[:find [?user ...] :where [?q :quota/user ?user]] db)
```
scrosby: Probably worth a clojuredoc comment that there are no duplicates in this list.

scheduler/src/cook/quota.clj
```diff
@@ -142,22 +143,70 @@
...
+        user->count-quota (retrieve-user->count-quota db pool-name all-quota-users default-count-quota)
+        user->quota-cache (-> (pc/map-from-keys
+                                (fn [user]
+                                  (-> (pc/map-from-keys
+                                        #(get-in type->user->quota [% user] (get default-type->quota %))
+                                        all-resource-types)
+                                    (assoc :count (user->count-quota user))))
+                                all-quota-users)
+                            (assoc default-user default-type->quota))]
```
scrosby: What about associating a :count quota to the default-user? Is this handled by get-quota for :count correctly?

shamsimam: `default-type->quota` computed by invoking `get-quota` already accounts for it.

scheduler/src/cook/quota.clj
shamsimam: Added comment.

scheduler/src/cook/quota.clj
shamsimam: Renamed all the functions to `retrieve-user->*` and updated the docstrings.

scheduler/src/cook/quota.clj
shamsimam: Updated the docstring.

shamsimam:
@scrosby addressed your feedback, please review when you get a chance.

PR #1334, opened 2019-12-10
Some more tweaks to de-layer docker image

scrosby:

- Reduce the layers in the docker image.

Smaller docker images and faster to build


scrosby:

PR #1336, opened 2019-12-10
Make composite mesos+k8s edn for docker+gke

scrosby:

- Make a config.edn that activates mesos + GKE

Make it easier for others to replicate my test environment.


dposada:
Will merge when green

scrosby:

PR #1337, opened 2019-12-10
Tweak dockerignore for integration tests + scheduler

scrosby:

- Ignore some virtualenv environments

We tend to use these names for virtualenv environments. Lets .dockerignore them so that we get smaller docker images that build faster.


scrosby:

PR #1338, opened 2019-12-10
Makes container defaults be compute-cluster specific

dposada:

- moving the container defaults configuration from the top-level to the compute-cluster

Not all compute clusters can use the same container defaults. For example, you might need to configure Mesos with one set of default volumes that don't exist and therefore don't work in k8s.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -224,7 +224,10 @@
 
   (restore-offers [this pool-name offers])
 
-  (container-defaults [_] {}))
+  (container-defaults [_]
+    ; We don't currently support specifying
+    ; container defaults for k8s compute clusters
+    {}))
 
 (defn get-or-create-cluster-entity-id
   [conn compute-cluster-name]
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -222,7 +222,9 @@
                      (into [] (map #(into {} (select-keys % [:hostname :resources])) offers-this-pool)))
       offers-this-pool))
 
-  (restore-offers [this pool-name offers]))
+  (restore-offers [this pool-name offers])
+
+  (container-defaults [_] {}))
 
 (defn get-or-create-cluster-entity-id
   [conn compute-cluster-name]
--- a/scheduler/src/cook/mesos/task.clj
+++ b/scheduler/src/cook/mesos/task.clj
@@ -80,9 +80,9 @@
 (defn merge-container-defaults
   "Takes a job container specification and applies any defaults from the config.
    Currently only supports volumes."
-  [container]
+  [container compute-cluster]
   (when container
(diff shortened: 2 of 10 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -222,7 +222,9 @@
                      (into [] (map #(into {} (select-keys % [:hostname :resources])) offers-this-pool)))
       offers-this-pool))
 
-  (restore-offers [this pool-name offers]))
+  (restore-offers [this pool-name offers])
+
+  (container-defaults [_] {}))
```
shamsimam: Nitcpick: Should we add a TODO that the k8s container defaults will be implemented later?

scheduler/src/cook/mesos/task.clj
```diff
@@ -127,10 +127,10 @@
 
 (defn job->task-metadata
   "Takes a job entity, returns task metadata"
-  [db mesos-run-as-user job-ent task-id]
+  [mesos-run-as-user job-ent task-id compute-cluster]
```
shamsimam: Nitpick: can we make `compute-cluster` the first argument? Keeping the more 'static' value as the first arguments helps with currying. I feel it also helps with readability in `a->b` named functions when `a` is moved as far to the right as possible in parameter order.

scheduler/src/cook/mesos/task.clj
```diff
@@ -80,9 +80,9 @@
 (defn merge-container-defaults
   "Takes a job container specification and applies any defaults from the config.
    Currently only supports volumes."
-  [container]
+  [container compute-cluster]
```
shamsimam: Nitpick: can we make `compute-cluster` the first argument? Keeping the more 'static' value as the first arguments helps with currying.

scheduler/test/cook/test/mesos/task.clj
```diff
@@ -346,7 +347,8 @@
                         :extract false
                         :value "file:///path/to/cook-executor"}}
         mesos-run-as-user nil]
-    (with-redefs [cook.config/executor-config (constantly executor)]
+    (with-redefs [cook.config/executor-config (constantly executor)
+                  cc/container-defaults (constantly {})]
```
shamsimam: Nitpick: Should we define `compute-cluster (Object.)` and validate it is the argument received in the `cc/container-defaults` mocked functions?

dposada: Seems like overkill to me

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Added a comment saying that we currently don't support container defaults for k8s

scheduler/src/cook/mesos/task.clj
dposada: Good call, done

scheduler/src/cook/mesos/task.clj
dposada: Good call, done

PR #1347, opened 2019-12-20
Fixes bugs in util.wait_for_instance, test_change_failed_retries

dposada:

1. in `wait_for_instance`, instead of returning the 0th instance, return the 0th instance that has the desired status
1. in `test_change_failed_retries`, don't assume that there was a non-mea-culpa failure on the job

1. Despite the `wait_until` filtering on status, this function is always blindly returning the 0th instance, which can have a different status, causing tests to fail.
1. If the job fails with a mea-culpa reason, e.g. `Agent removed`, then it will have the full 4 retries instead of only 3.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1242,12 +1242,23 @@ class CookTest(util.CookTest):
             resp = util.retry_jobs(self.cook_url, retries=4, failed_only=True, jobs=jobs)
             self.assertEqual(201, resp.status_code, resp.text)
             jobs = util.query_jobs(self.cook_url, True, uuid=jobs).json()
-            # We expect both jobs to be running now.
-            # The first job (which we killed and retried) should have 3 retries remaining
-            # (the attempt before resetting the total retries count is still included).
-            job_details = f"Job details: {json.dumps(jobs[0], sort_keys=True)}"
+            # We expect both jobs to be waiting or running now.
+            job_details = f"Job details: {json.dumps(jobs[0], sort_keys=True, indent=2)}"
+            self.logger.info(job_details)
             self.assertIn(jobs[0]['status'], ['waiting', 'running'], job_details)
-            self.assertEqual(3, jobs[0]['retries_remaining'], job_details)
+            non_mea_culpa_failure_exists = any(not i['reason_mea_culpa'] for i in jobs[0]['instances'])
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -1242,12 +1242,23 @@ def instance_query():
...
-            # We expect both jobs to be running now.
-            # The first job (which we killed and retried) should have 3 retries remaining
-            # (the attempt before resetting the total retries count is still included).
-            job_details = f"Job details: {json.dumps(jobs[0], sort_keys=True)}"
+            # We expect both jobs to be waiting or running now.
+            job_details = f"Job details: {json.dumps(jobs[0], sort_keys=True, indent=2)}"
+            self.logger.info(job_details)
             self.assertIn(jobs[0]['status'], ['waiting', 'running'], job_details)
-            self.assertEqual(3, jobs[0]['retries_remaining'], job_details)
+            non_mea_culpa_failure_exists = any(not i['reason_mea_culpa'] for i in jobs[0]['instances'])
```
nsinkov: add `self.assertEqual(4, jobs[0]['max_retries'], job_details)`

integration/tests/cook/test_basic.py
```diff
@@ -1242,12 +1242,23 @@ def instance_query():
...
+                self.assertEqual(3, jobs[0]['retries_remaining'], job_details)
+            else:
+                self.logger.info(f'Job {job_uuid} does not have a non-mea-culpa failure')
+                # There was no non-mea-culpa failure on the first job, so:
+                # We should now have the full 4 retries that we asked for.
+                self.assertEqual(4, jobs[0]['retries_remaining'], job_details)
+
             # The second job (which started with the default 1 retries)
             # should have 1 remaining since the failed_only flag was set.
             job_details = f"Job details: {json.dumps(jobs[1], sort_keys=True)}"
```
nsinkov: add `self.assertEqual(1, jobs[1]['max_retries'], job_details)`

integration/tests/cook/test_basic.py
nsinkov: nitpick: I would just add some more utility functions to get a failed non-mea culpa job and not have the branch

integration/tests/cook/test_basic.py

integration/tests/cook/test_basic.py

PR #1350, opened 2019-12-20
Make reconciler only reconcile mesos tasks.

scrosby:

- Make reconciler only reconcile mesos tasks.

Right now if you have kubernetes + mesos enabled, the mesos reconciler sees all these kubernetes jobs not running in mesos and kills them off. Make the mesos reconciler only look at jobs in the relevant compute cluster instead.

### DIFF
--- a/scheduler/src/cook/mesos/mesos_compute_cluster.clj
+++ b/scheduler/src/cook/mesos/mesos_compute_cluster.clj
@@ -105,7 +105,7 @@
         (future
           (try
             (sched/reconcile-jobs conn)
-            (sched/reconcile-tasks (d/db conn) driver pool->fenzo)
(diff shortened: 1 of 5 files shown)


dposada:
Will merge when green

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -292,7 +292,7 @@
     container-defaults))
 
 ; Internal method
```
shamsimam: This comment appears to be outdated now?

scrosby: It is still an internal method. Only had to do this for testing purposes.

PR #1352, opened 2019-12-26
WIP: Change executor http timeout and increase logging.

scrosby:

- Change executor HTTP timeout
- Increase logging about executor
- 

We are finding some executor failures in some cases. We think associated with this timeout. Decrease the timeout to 2 minutes and log additional information.

### DIFF
--- a/executor/Dockerfile.build
+++ b/executor/Dockerfile.build
@@ -1,6 +1,7 @@
 FROM python:3.5.9-stretch
 
 RUN pip install pyinstaller==3.3
+RUN apt-get update && apt-get --no-install-recommends -y install libzookeeper-mt-dev && apt-get clean && rm -Rf /var/lib/apt/lists/*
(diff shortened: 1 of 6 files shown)


executor/cook/executor.py
```diff
@@ -475,6 +484,7 @@ def killTask(self, driver, task_id):
         task_id_str = task_id['value'] if 'value' in task_id else task_id
         grace_period = os.environ.get('MESOS_EXECUTOR_SHUTDOWN_GRACE_PERIOD', '')
         cio.print_and_log('Received kill for task {} with grace period of {}'.format(task_id_str, grace_period))
+        cu.log_thread_stack_traces()
```
shamsimam: Can you include sample output?

scrosby: Output is too verbose to include.

PR #1355, opened 2020-01-06
Log executor fails

scrosby:

- Log whenever the executor is asked to kill

We sometimes see the executor get wedged. This is an effort to figure out how and where it gets wedged.


dposada:
Will merge when green.

PR #1360, opened 2020-01-14
fixes the compute-cluster argument in mesos-scheduler call to receive-offers

shamsimam:

- fixes the compute-cluster argument in mesos-scheduler call to receive-offers

Passing incorrect argument causes the receive offer code to fail which eventually leads to no scheduling of jobs.

### DIFF
--- a/scheduler/src/cook/mesos/mesos_compute_cluster.clj
+++ b/scheduler/src/cook/mesos/mesos_compute_cluster.clj
@@ -158,14 +158,14 @@
                   (if-let [offers-chan (get pool->offers-chan pool-name)]
                     (do
                       (log/info "Processing" offer-count "offer(s) for known pool" pool-name)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -158,14 +158,14 @@
                   (if-let [offers-chan (get pool->offers-chan pool-name)]
                     (do
                       (log/info "Processing" offer-count "offer(s) for known pool" pool-name)
-                      (sched/receive-offers offers-chan match-trigger-chan this pool-name offers))
+                      (sched/receive-offers offers-chan match-trigger-chan compute-cluster pool-name offers))
```
dposada: Per @scrosby, this argument is only used for declining offers.

PR #1362, opened 2020-01-16
populates the user parameter in docker container parameters

shamsimam:

- populates the user parameter in docker container parameters when not provided explicitly in the job
- removes dependency on me.raynes/conch

When docker user parameters are not provided, the command runs as root inside the docker container. Instead, we would like to limit the container to always run as the user on the job.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -368,7 +368,7 @@ class CookTest(util.CookTest):
         self.assertEqual(False, docker['force-pull-image'])
         # the user parameter is added when missing
         self.assertEqual(3, len(docker['parameters']))
-        self.assertTrue(next(True for p in docker['parameters'] if p['key'] == 'env'))
+        self.assertTrue(any(p['key'] == 'user' for p in docker['parameters']))
         self.assertEqual('FOO=bar', next(p['value'] for p in docker['parameters'] if p['key'] == 'env'))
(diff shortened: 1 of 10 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -230,7 +230,9 @@ def test_docker_fields(self):
         self.assertEqual(docker_image, docker['image'])
         self.assertEqual('HOST', docker['network'])
         self.assertEqual(False, docker['force-pull-image'])
-        self.assertEqual(2, len(docker['parameters']))
+        # the user parameter is added when missing
+        self.assertEqual(3, len(docker['parameters']))
+        self.assertTrue(next(True for p in docker['parameters'] if p['key'] == 'env'))
```
dposada: Should this be:

```suggestion
        self.assertTrue(any(p['key'] == 'user' for p in docker['parameters']))
```

(using `any` and also testing for the `'user'` key?)

scheduler/src/cook/rest/api.clj
```diff
@@ -552,20 +542,34 @@
                           ports)]
       {:docker/port-mapping port-maps})))
 
+(defn- ensure-user-parameter
+  "Ensures the presence of the user parameter by attaching it when missing in params."
+  [user id params]
+  (let [user-id (util/user->user-id user)
+        group-id (util/user->group-id user)
```
dposada: If these return `nil`, we should throw an error.

shamsimam: added exception handling

PR #1365, opened 2020-01-17
ssh cli support for kubernetes

nsinkov:

use plugin to ssh to kubernetes

we can't use ssh directly when a job is scheduled on kubernetes - we need to use kubectl

### DIFF
--- a/cli/cook/subcommands/ssh.py
+++ b/cli/cook/subcommands/ssh.py
@@ -1,13 +1,12 @@
 import logging
 import os
 
-from cook import plugins
-from cook import terminal
+from cook import http, plugins, terminal
 from cook.querying import query_unique_and_run, parse_entity_refs
 from cook.util import print_info, guard_no_cluster
 
 
-def kubectl_exec_to_instance(instance_uuid):
+def kubectl_exec_to_instance(instance_uuid, _):
     os.execlp('kubectl', 'kubectl',
               'exec',
               '-c', os.getenv('COOK_CONTAINER_NAME_FOR_JOB', 'required-cook-job-container'),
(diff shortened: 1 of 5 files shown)


dposada:
Will merge when green

cli/cook/subcommands/ssh.py
```diff
@@ -1,19 +1,38 @@
...
-    os.execlp(command, 'ssh', '-t', hostname, f'cd "{sandbox_dir}" ; bash')
+    compute_cluster = instance["compute-cluster"]
+    compute_cluster_type = compute_cluster["type"]
+    compute_cluster_name = compute_cluster["name"]
+    if compute_cluster_type == "kubernetes":
+        kubectl_exec_to_instance_fn = plugins.get_fn('kubectl-exec-to-instance', kubectl_exec_to_instance)
+        cook_cluster_settings = http.get(cluster, 'settings', params={}).json()
+        compute_cluster_config = next(c for c in (s['config'] for s in cook_cluster_settings['compute-clusters']) if
+                                      c['compute-cluster-name'] == compute_cluster_name)
+        kubectl_exec_to_instance_fn(instance["task_id"], compute_cluster_config)
```
dposada: Can we write an integration test for this?

cli/cook/subcommands/ssh.py
```diff
@@ -1,19 +1,38 @@
...
 
 
-def ssh_to_instance(instance, sandbox_dir):
-    """Attempts to ssh (using os.execlp) to the Mesos agent corresponding to the given instance."""
+def kubectl_exec_to_instance(instance_uuid, _):
+    os.execlp('kubectl', 'kubectl',
+              'exec',
+              '-c', os.getenv('COOK_CONTAINER_NAME_FOR_JOB', 'required-cook-job-container'),
+              '-it', instance_uuid,
+              '--', '/bin/bash')
```
dposada: Have we tested that this works?

cli/cook/subcommands/ssh.py
nsinkov: what about the cli tests?

dposada: Ah, forgot the CLI tests don't currently support specifying a docker image to submit with. Until we solve that, not sure how much sense it makes to write a test for this.

dposada:
Will merge when green

cli/cook/subcommands/ssh.py
nsinkov: yes

PR #1368, opened 2020-01-21
logs absolute path of the tracked progress file

shamsimam:

- logs absolute path of the tracked progress file

We would like to unambiguously and easily determine the progress file being tracked by the executor.


shamsimam:
@DaoWen can you please include this change in your version of the progress updater sidecar?

dposada:
Will merge when green

shamsimam:
@dposada My bad, green now :)

PR #1372, opened 2020-01-22
disable test that can't run in travis

nsinkov:

- add functionality to disable integration tests in travis using travis_skip test marker

Some integration tests require too much modification to the travis pipeline (e.g. upgrading mesos) to work. Add the travis_skip test marker to those tests to skip


dposada:
Will merge when green

PR #1376, opened 2020-01-23
Improve our node and pod watches so that they retry forever instead of exiting on certain errors

scrosby:

- Improve our node and pod watches so that they retry forever instead of exiting on certain errors. 
- Also improve logging.

Fix a bug where we lose track of whats going on.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -147,7 +147,7 @@
                   (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
                   (catch Exception e
                     (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
-                    (Thread/sleep 60000)
+                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
                     nil)))
         ^Callable first-success (->> tmpfn repeatedly (some identity))]
     (.submit kubernetes-executor ^Callable first-success)))
@@ -192,7 +192,7 @@
                   (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
                   (catch Exception e
                     (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
-                    (Thread/sleep 60000)
+                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
                     nil)))
         ^Callable first-success (->> tmpfn repeatedly (some identity))]
     (.submit kubernetes-executor ^Callable first-success)))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -102,8 +102,10 @@
                                                    (.getItems current-pods))]
     [current-pods namespaced-pod-name->pod]))
 
-(defn initialize-pod-watch
-  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
+
+(declare initialize-pod-watch)
+(defn ^Callable initialize-pod-watch-helper
+  "Help creating pod watch. Returns a new watch Callable"
   [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
   (let [[current-pods namespaced-pod-name->pod] (get-all-pods-in-kubernetes api-client)
         ; 2 callbacks;
@@ -116,9 +118,9 @@
     ; So compute the delta between the old and new and process those via the callbacks.
     ; Note as a side effect, the callbacks mutate all-pods-atom
     (doseq [task (set/union (keys namespaced-pod-name->pod) (keys old-all-pods))]
+      (log/info "In" compute-cluster-name "compute cluster, doing (startup) callback for" task)
       (doseq [callback callbacks]
         (try
-          (log/info "In" compute-cluster-name "compute cluster, doing (startup) callback for" task)
           (callback task (get old-all-pods task) (get namespaced-pod-name->pod task))
           (catch Exception e
             (log/error e "Error while processing callback for" task)))))
@@ -126,20 +128,33 @@
     (let [watch (WatchHelper/createPodWatch api-client (-> current-pods
                                                            .getMetadata
                                                            .getResourceVersion))]
-      (.submit kubernetes-executor ^Callable
       (fn []
         (try
           (handle-watch-updates all-pods-atom watch get-pod-namespaced-key
                                 callbacks)
           (catch Exception e
-            (log/error e "Error during watch"))
+            (log/error e "Error during pod watch for compute cluster" compute-cluster-name))
           (finally
             (.close watch)
-            (initialize-pod-watch api-client compute-cluster-name all-pods-atom cook-pod-callback))))))))
+            (initialize-pod-watch api-client compute-cluster-name all-pods-atom cook-pod-callback)))))))
 
-(defn initialize-node-watch
-  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
-  [^ApiClient api-client current-nodes-atom]
+(defn initialize-pod-watch
+  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
+  ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
+  (let [tmpfn (fn [] (try
+                       (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
+                       (catch Exception e
+                         (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
+                         (Thread/sleep 60000)
+                         nil)))
+        ^Callable first-success (->> tmpfn repeatedly (some identity))]
+    (.submit kubernetes-executor ^Callable first-success)))
+
+(declare initialize-node-watch)
+(defn initialize-node-watch-helper
+  "Help creating node watch. Returns a new watch Callable"
+  [^ApiClient api-client compute-cluster-name current-nodes-atom]
   (let [api (CoreV1Api. api-client)
         current-nodes (.listNode api
                                  nil ; includeUninitialized
@@ -157,16 +172,28 @@
                                           (.getItems current-nodes))]
     (reset! current-nodes-atom node-name->node)
     (let [watch (WatchHelper/createNodeWatch api-client (-> current-nodes .getMetadata .getResourceVersion))]
-      (.submit kubernetes-executor ^Callable
       (fn []
         (try
           (handle-watch-updates current-nodes-atom watch (fn [n] (-> n .getMetadata .getName))
                                 [(make-atom-updater current-nodes-atom)]) ; Update the set of all nodes.
           (catch Exception e
-            (log/warn e "Error during node watch"))
+            (log/warn e "Error during node watch for compute cluster" compute-cluster-name))
           (finally
             (.close watch)
-            (initialize-node-watch api-client current-nodes-atom))))))))
+            (initialize-node-watch api-client compute-cluster-name current-nodes-atom)))))))
+
+(defn initialize-node-watch
+  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name current-nodes-atom]
+    ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
+  (let [tmpfn (fn [] (try
+                       (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
+                       (catch Exception e
+                         (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
+                         (Thread/sleep 60000)
+                         nil)))
+        ^Callable first-success (->> tmpfn repeatedly (some identity))]
+    (.submit kubernetes-executor ^Callable first-success)))
 
 (defn to-double
   "Map a quantity to a double, whether integer, double, or float."
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -147,7 +147,7 @@
                   (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
                   (catch Exception e
                     (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
-                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
+                    (Thread/sleep (:reconnect-delay-ms config/kubernetes))
                     nil)))
         ^Callable first-success (->> tmpfn repeatedly (some identity))]
     (.submit kubernetes-executor ^Callable first-success)))
@@ -192,7 +192,7 @@
                   (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
                   (catch Exception e
                     (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
-                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
+                    (Thread/sleep (:reconnect-delay-ms config/kubernetes))
                     nil)))
         ^Callable first-success (->> tmpfn repeatedly (some identity))]
     (.submit kubernetes-executor ^Callable first-success)))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -141,13 +141,14 @@
 (defn initialize-pod-watch
   "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
   [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
-  ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
-  (let [tmpfn (fn [] (try
-                       (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
-                       (catch Exception e
-                         (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
-                         (Thread/sleep 60000)
-                         nil)))
+  ; We'll iterate trying to connect to k8s until the initialize-pod-watch-helper returns a watch function.
+  (let [tmpfn (fn []
+                (try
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
     ; So compute the delta between the old and new and process those via the callbacks.
     ; Note as a side effect, the callbacks mutate all-pods-atom
     (doseq [task (set/union (keys namespaced-pod-name->pod) (keys old-all-pods))]
+      (log/info "In" compute-cluster-name "compute cluster, doing (startup) callback for" task)
```
dposada: 👍

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
...
-  [^ApiClient api-client current-nodes-atom]
+(defn initialize-pod-watch
+  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
+  ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
+  (let [tmpfn (fn [] (try
+                       (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
+                       (catch Exception e
+                         (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
+                         (Thread/sleep 60000)
```
dposada: Can we make the sleep amount be configurable?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
...
-            (initialize-pod-watch api-client compute-cluster-name all-pods-atom cook-pod-callback))))))))
+            (initialize-pod-watch api-client compute-cluster-name all-pods-atom cook-pod-callback)))))))
 
-(defn initialize-node-watch
-  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
-  [^ApiClient api-client current-nodes-atom]
+(defn initialize-pod-watch
+  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
+  ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
```
```suggestion
  ; We'll iterate trying to connect to k8s until the initialize-pod-watch-helper returns a watch function.
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
...
+  (let [tmpfn (fn [] (try
```
dposada: Can we move the `(try` to the next line?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -201,23 +201,26 @@
...
-        (regular-scanner this (time/seconds scan-frequency-seconds-config))
-        (log/info "State scan disabled because no interval has been set")))
-
-    ; Initialize the node watch path.
-    (api/initialize-node-watch api-client current-nodes-atom)
-
-    (reset! pool->fenzo-atom pool->fenzo)
+    ; We may iterate forever trying to bring up kubernetes. However, our caller expects us to eventually return,
+    ; so we launch within a future so that our caller can continue initializing other clusters.
+    (future
```
dposada: Shouldn't we wrap everything inside this `future` in a `try`/`catch`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
...
+                       (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
+                       (catch Exception e
+                         (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
+                         (Thread/sleep 60000)
+                         nil)))
+        ^Callable first-success (->> tmpfn repeatedly (some identity))]
```
dposada: Maybe `some?` instead of `identity`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,43 @@
...
+    (.submit kubernetes-executor ^Callable first-success)))
```
dposada: What happens if this line throws? Wouldn't we stop trying to re-start the watch?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +172,28 @@
...
+            (log/warn e "Error during node watch for compute cluster" compute-cluster-name))
           (finally
             (.close watch)
-            (initialize-node-watch api-client current-nodes-atom))))))))
+            (initialize-node-watch api-client compute-cluster-name current-nodes-atom)))))))
+
+(defn initialize-node-watch
+  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name current-nodes-atom]
+    ; We'll iterate trying to connect to k8s until the initilzie-pod-watch-helper returns a watch function.
```
```suggestion
  ; We'll iterate trying to connect to k8s until the initialize-node-watch-helper returns a watch function.
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +172,28 @@
...
+  (let [tmpfn (fn [] (try
```
dposada: Can we move the `(try` to the next line?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +172,28 @@
...
+                       (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
+                       (catch Exception e
+                         (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
+                         (Thread/sleep 60000)
```
dposada: Can we make the sleep amount configurable?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +172,28 @@
...
+                         nil)))
+        ^Callable first-success (->> tmpfn repeatedly (some identity))]
```
dposada: Maybe `some?` instead of `identity`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +172,28 @@
...
+    (.submit kubernetes-executor ^Callable first-success)))
```
dposada: What happens if this line throws? Wouldn't we stop trying to re-start the watch?

scheduler/src/cook/kubernetes/api.clj
scrosby: Why? This smells like a set once and leave it forever. I'd rather remove it than do the extra wiring to make it configurable.

scheduler/src/cook/kubernetes/api.clj
scrosby: Fixed.

scheduler/src/cook/kubernetes/api.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: And do what within the catch handler? Given the current body, the only thing to go wrong is initializing the startup state. If that fails, then I'm fine with the whole CC being dead. restart and retry later.

scheduler/src/cook/kubernetes/api.clj
scrosby: (some some?) seems weird. Don't see anything wrong with the code as-is now.

scheduler/src/cook/kubernetes/api.clj
scrosby: Correct. If the ExecutorService can't accept it, we die. If our runtime is that broken, the only sane error recovery is fail all of Cook, with System. exit().

scheduler/src/cook/kubernetes/api.clj
scrosby: Fixed.

scheduler/src/cook/kubernetes/api.clj
scrosby: ok

scheduler/src/cook/kubernetes/api.clj
scrosby: Same response as earlier.

scheduler/src/cook/kubernetes/api.clj
scrosby: Same response as earlier.

scheduler/src/cook/kubernetes/api.clj
scrosby: Same response as earlier.

scheduler/src/cook/kubernetes/api.clj
dposada: I'm fine with removing it.

If we think we need it, then I suspect we're going to want control over it.

scheduler/src/cook/kubernetes/api.clj
dposada: Agreed, there is nothing wrong with the current code, but `some?` seems more clear to me because it makes it obvious that we're looking for non-nil values. What is weird about `(some some?)`?

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Agreed, but it would be nice to have an ERROR log when this happens

scheduler/src/cook/kubernetes/api.clj
dposada: Ah, got it, I had misunderstood what `kubernetes-executor` was.

scheduler/src/cook/kubernetes/api.clj
dposada: Same response as above.

scheduler/src/cook/kubernetes/api.clj
dposada: Same response as above.

scheduler/src/cook/kubernetes/api.clj
scrosby: Nuking it.

scheduler/src/cook/kubernetes/api.clj
scrosby: Nuking it.

scheduler/src/cook/kubernetes/api.clj
scrosby: Changed.

scheduler/src/cook/kubernetes/api.clj
scrosby: Changed.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/api.clj
scrosby: ok.

scheduler/src/cook/kubernetes/api.clj
dposada: I thought we were going to make it configurable?

scheduler/src/cook/kubernetes/api.clj
dposada: I thought we were making it configurable?

scheduler/src/cook/kubernetes/api.clj
dposada: I still see `identity`

scheduler/src/cook/kubernetes/api.clj
dposada: I still see `identity`

scheduler/src/cook/kubernetes/api.clj
scrosby: Making it configurable. (comment pre-dated our discussion)

scheduler/src/cook/kubernetes/api.clj
scrosby: Pushing the changes raced with testing them. I've pushed now.

scheduler/src/cook/kubernetes/api.clj
scrosby: This comment pre-dates our conversation.

scheduler/src/cook/kubernetes/api.clj
scrosby: Pushing the changes raced this comment. I've pushed now.

scheduler/src/cook/kubernetes/api.clj
scrosby: I tested this change and now I remember why I didn't use some? Some? returns a boolean. You need identify because you need the first non-nil --- you need the value to pass to the executor.

scheduler/src/cook/kubernetes/api.clj
scrosby: I tested this change and now I remember why I didn't use some? Some? returns a boolean. You need identify because you need the first non-nil --- you need the value to pass to the executor.

scheduler/src/cook/kubernetes/api.clj
dposada: Ah, got it, makes sense.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -116,30 +118,44 @@
...
+(defn initialize-pod-watch
+  "Initialize the pod watch. This fills all-pods-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name all-pods-atom cook-pod-callback]
+  ; We'll iterate trying to connect to k8s until the initialize-pod-watch-helper returns a watch function.
+  (let [tmpfn (fn []
+                (try
+                  (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
+                  (catch Exception e
+                    (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
+                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
```
dposada: We already have a `config/kubernetes` function we can use here.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +173,29 @@
...
+(defn initialize-node-watch
+  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name current-nodes-atom]
+  ; We'll iterate trying to connect to k8s until the initialize-node-watch-helper returns a watch function.
+  (let [tmpfn (fn []
+                (try
+                  (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
+                  (catch Exception e
+                    (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
+                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
```
dposada: We already have a `config/kubernetes` function we can use here.

scheduler/src/cook/kubernetes/api.clj
dposada: Got it, thanks.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -201,23 +201,29 @@
...
+          (api/initialize-pod-watch api-client name all-pods-atom cook-pod-callback)
+          (if scan-frequency-seconds-config
+            (regular-scanner this (time/seconds scan-frequency-seconds-config))
+            (log/info "State scan disabled because no interval has been set")))
+
+        ; Initialize the node watch path.
+        (api/initialize-node-watch api-client name current-nodes-atom)
+
+        (reset! pool->fenzo-atom pool->fenzo)
+        (catch Exception e
```
dposada: Should this be `Throwable` instead of `Exception`?

scrosby: No. Only reason to catch throwable but not exception is to suicide.

dposada: Seems like we would want the log line below if a `Throwable` is thrown.

scheduler/src/cook/kubernetes/api.clj
scrosby: Changed.

scheduler/src/cook/kubernetes/api.clj
scrosby: Changed.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -142,12 +142,14 @@
...
   ; We'll iterate trying to connect to k8s until the initialize-pod-watch-helper returns a watch function.
-  (let [tmpfn (fn []
+  (let [{:keys [reconnect-delay-ms]} (config/kubernetes)
+        tmpfn (fn []
                 (try
                   (initialize-pod-watch-helper api-client compute-cluster-name all-pods-atom cook-pod-callback)
                   (catch Exception e
                     (log/error e "Error during initial setup of looking at pods for" compute-cluster-name)
-                    (Thread/sleep (get-in config/config [:settings :kubernetes :reconnect-delay-ms]))
+                    (log/info "Sleeping" reconnect-delay-ms "milliseconds before reconnect")
```
dposada: Can we add the compute cluster name to this log?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -157,16 +175,31 @@
...
+  "Initialize the node watch. This fills current-nodes-atom with data and invokes the callback on pod changes."
+  [^ApiClient api-client compute-cluster-name current-nodes-atom]
+  ; We'll iterate trying to connect to k8s until the initialize-node-watch-helper returns a watch function.
+  (let [{:keys [reconnect-delay-ms]} (config/kubernetes)
+        tmpfn (fn []
+                (try
+                  (initialize-node-watch-helper api-client compute-cluster-name current-nodes-atom)
+                  (catch Exception e
+                    (log/error e "Error during initial setup of looking at nodes for" compute-cluster-name)
+                    (log/info "Sleeping" reconnect-delay-ms "milliseconds before reconnect")
```
dposada: Can we add the compute cluster name to this log?

scrosby: Ok.

scheduler/src/cook/kubernetes/api.clj
scrosby: Ok.

dposada:
Will merge when green

PR #1377, opened 2020-01-23
Add REST endpoint for posting job progress updates

DaoWen:

Add REST endpoint `/progress/<instance-uuid>` for POSTing job instance progress updates.

Job instance progress updates are currently communicated via Mesos framework messages. We need a new mechanism for reporting progress to use on Kubernetes.

We plan to have our Kubernetes log sidecar duplicate some of the current Cook Executor functionality to monitor log files for progress messages, but then POST to this new endpoint rather than sending Mesos framework messages.

This PR only implements the endpoint. The new progress reporting functionality for the log sidecar on Kubernetes will come as a separate patch.

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1732,28 +1732,16 @@ def rebalancer_interval_seconds():
     return interval_seconds
 
 
-__auto_sequence_lock = threading.Lock()
-__auto_sequence_counter = 0
-
-
 def send_progress_update(cook_url, instance, assert_response=True, allow_redirects=True,
-                         sequence=None, percent=None, message=None, json=None):
+                         sequence=None, percent=None, message=None):
     """Submit a job instance progress update via the rest api"""
-    global __auto_sequence_counter
-    if json is not None:
-        payload = json
-    else:
-        payload = {}
-        if sequence is not None:
-            payload['progress_sequence'] = sequence
-        elif sequence is not False:
-            with __auto_sequence_lock:
-                __auto_sequence_counter += 1
-                payload['progress_sequence'] = __auto_sequence_counter
-        if percent is not None:
-            payload['progress_percent'] = percent
-        if message is not None:
-            payload['progress_message'] = message
+    payload = {}
+    if sequence is not None:
+        payload['progress_sequence'] = sequence
+    if percent is not None:
+        payload['progress_percent'] = percent
+    if message is not None:
+        payload['progress_message'] = message
     # NOTE: using `requests` rather than `session` here because this endpoint should not require authentication
     response = requests.post(f'{cook_url}/progress/{instance}', allow_redirects=allow_redirects, json=payload)
     if assert_response:
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1732,6 +1732,39 @@ def rebalancer_interval_seconds():
     return interval_seconds
 
 
+__auto_sequence_lock = threading.Lock()
+__auto_sequence_counter = 0
+
+
+def send_progress_update(cook_url, instance, assert_response=True, allow_redirects=True,
+                         sequence=None, percent=None, message=None, json=None):
+    """Submit a job instance progress update via the rest api"""
+    global __auto_sequence_counter
+    if json is not None:
+        payload = json
+    else:
+        payload = {}
+        if sequence is not None:
+            payload['progress_sequence'] = sequence
+        elif sequence is not False:
+            with __auto_sequence_lock:
+                __auto_sequence_counter += 1
+                payload['progress_sequence'] = __auto_sequence_counter
+        if percent is not None:
+            payload['progress_percent'] = percent
+        if message is not None:
+            payload['progress_message'] = message
+    # NOTE: using `requests` rather than `session` here because this endpoint should not require authentication
+    response = requests.post(f'{cook_url}/progress/{instance}', allow_redirects=allow_redirects, json=payload)
+    if assert_response:
+        response_info = {'code': response.status_code, 'msg': response.content}
+        assert response.status_code == 202, response_info
+        response_json = response.json()
+        assert response_json['instance'] == instance, response_info
+        assert response_json['message'] == 'progress update accepted', response_info
+    return response
+
+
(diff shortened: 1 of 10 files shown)


integration/tests/cook/util.py
```diff
@@ -1732,6 +1732,39 @@ def rebalancer_interval_seconds():
...
 
 
+__auto_sequence_lock = threading.Lock()
+__auto_sequence_counter = 0
+
+
+def send_progress_update(cook_url, instance, assert_response=True, allow_redirects=True,
+                         sequence=None, percent=None, message=None, json=None):
+    """Submit a job instance progress update via the rest api"""
+    global __auto_sequence_counter
```
shamsimam: Tests should explicitly provide their sequence number instead of us incrementing a global variable. In such a scenario `None` can be used to signal that we do not wish to send the sequence in the json payload like `percent` and `message`..

scheduler/src/cook/rest/api.clj
```diff
@@ -1639,6 +1652,38 @@
...
+     :exists? (constantly false)  ;; triggers path for moved-temporarily?
+     :existed? instance-request-exists?
+     :can-post-to-missing? (constantly false)
+     :moved-temporarily? (fn [ctx]
+                           ;; only the leader handles progress updates
+                           ;; the client is expected to cache the redirect location
+                           (let [request-path (get-in ctx [:request :uri])]
+                             (if @leadership-atom
+                               [false {}]
+                               [true {:location (str (leader-url leader-selector) request-path)}])))
```
shamsimam: Nitpick: should we use a namespaced keyword here?

scheduler/src/cook/rest/api.clj
```diff
@@ -1420,9 +1429,13 @@
 
 (def leader-hostname-regex #"^([^#]*)#([0-9]*)#([a-z]*)#.*")
 
+(defn leader-selector->leader-id
```
shamsimam: Missing docstring.

scheduler/src/cook/rest/api.clj
```diff
@@ -1420,9 +1429,13 @@
...
+  [leader-selector]
+  (-> leader-selector .getLeader .getId))
```
shamsimam: Is there always guaranteed to be a known leader? If not, should we generate a `503` error?

scheduler/src/cook/rest/api.clj
DaoWen: `:location` is something that Liberator recognizes for doing redirects, not something I picked.

scheduler/src/cook/rest/api.clj
DaoWen: We looked at the code and `.getLeader` returns a _dummy participant_ if there's no leader available; however, that dummy participant will return false on `.isLeader`, so we can use that to detect this scenario.

integration/tests/cook/util.py
DaoWen: Done

scheduler/src/cook/rest/api.clj
DaoWen: Refactored to throw `IllegalStateException`, which is caught in the handler and results in a 503.

scheduler/src/cook/rest/api.clj
DaoWen: Added

scheduler/src/cook/rest/api.clj
```diff
@@ -68,6 +69,7 @@
            (java.util Date UUID)
            javax.servlet.ServletResponse
            org.apache.curator.test.TestingServer
+           (org.apache.curator.framework.recipes.leader LeaderSelector)
```
shamsimam: Nitpick: lexicographic order

scheduler/src/cook/rest/api.clj
```diff
@@ -3148,7 +3215,25 @@
...
+            "/progress/:uuid" [uuid]
+            :path-params [uuid :- s/Uuid]
+            (c-api/resource
+              ;; NOTE: The authentication for this endpoint is disabled via cook.components/conditional-auth-bypass
+              {:post {:summary "Update the progress of a Job Instance"
+                      :parameters {:body-params JobInstanceProgressRequest}
+                      :responses {202 {:description "The progress update was accepted."}
+                                  307 {:description "Redirecting request to leader node."}
+                                  400 {:description "Invalid request format."}
+                                  404 {:description "The supplied UUID doesn't correspond to a valid job instance."}}
```
shamsimam: Nitpick: 503 is not documented

scheduler/src/cook/rest/api.clj
```diff
@@ -1420,9 +1430,22 @@
...
-(defn leader-url
+(defn leader-selector->leader-id
+  "Get the current leader node's id from a leader-selector object.
+   Throws if there is currently no leader available."
+  [^LeaderSelector leader-selector]
+  (let [leader (.getLeader leader-selector)]
+    ;; NOTE: .getLeader returns a dummy object when no leader is available,
+    ;; but the dummy object always returns false for the .isLeader predicate.
+    (when-not (.isLeader leader)
+      (throw (IllegalStateException. "Leader is temporarily unavailable.")))
```
shamsimam: Are previous consumers of `leader-selector->leader-url` affected by this new exception?

DaoWen: No, the original function that calls this function (i.e., `leader-selector->leader-url`) already throws `IllegalStateException` when the leader id is in an invalid format, which is what happens when there's no leader and the id comes back blank. The only thing that changed is that we get a more specific error message now.

scheduler/src/cook/rest/api.clj
DaoWen: Good point! Added.

scheduler/src/cook/rest/api.clj
DaoWen: Fixed

DaoWen:
@shamsimam - Feedback addressed, and green build!

DaoWen:
Cleaned up the new integration test a bit, and then rebased onto current master.

PR #1379, opened 2020-01-24
Avoid over-verbose logging of compute cluster object

scrosby:

- Do not log the compute cluster object, it can be very big. Just log the name.

Avoid log bloat --- found when looking for issues similar to https://github.com/twosigma/Cook/pull/1375


scrosby:
Change rejected.

PR #1384, opened 2020-01-26
Makes pod-watch exception log INFO level

dposada:

- changing the pod-watch exception log from `ERROR` level to `INFO` level

It's expected and frequent for the pod-watch to throw when it times out.

`ERROR` makes it harder to distinguish these from unexpected errors, and `INFO` will still allow us to view these in the normal places.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -133,7 +133,7 @@
           (handle-watch-updates all-pods-atom watch get-pod-namespaced-key
                                 callbacks)
           (catch Exception e
-            (log/error e "Error during pod watch for compute cluster" compute-cluster-name))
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -133,7 +133,7 @@
           (handle-watch-updates all-pods-atom watch get-pod-namespaced-key
                                 callbacks)
           (catch Exception e
-            (log/error e "Error during pod watch for compute cluster" compute-cluster-name))
```
sradack: Should we have a different clause for the expected exception and leave this as is?

dposada: Good call, changed this to check the cause and only switch to `INFO` for `SocketTimeoutException`.

PR #1389, opened 2020-01-27
Kill flask server on exit in run-local.sh

DaoWen:

- Kill flask server on exit in run-local.sh
- Allow setting `COOK_CONFIG` to override the config file path

- Since we start a background progress in this script, we should at least try to clean it up on exit.
- The run-docker.sh script allows using `COOK_CONFIG` like this, so it seems reasonable to add it here as well. This way I can override the default when running my test setup.


DaoWen:
@dposada said he'd rather just rip out the data-local stuff altogether, so I'm closing this for now.

PR #1390, opened 2020-01-27
Various small tweaks to the state machine

scrosby:

- Fix a small bug in state machine.
- A variety of weird states aren't as weird as we thought and/or are symptomatic of deeper bugs.
- Improve the logging.

In the kubernetes tracking state machine, the old bad code would go:
(completed,succeeded) -> (missing,succeeded) -> (missing,missing), inadvertently deleting its local state first and keeping the kubernetes state.
The correct transitions should go:
(completed, succeeded) -> (completed,missing), (missing,missing).
so we're always tracking the local state.

Weird states are supposed to be things that should never occur. So we want less false positives in the log.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -308,7 +308,11 @@
                                         ; normal occurrence if someone submits a kill request on a job that fails to launch.
                                         :missing (do
                                                    ; TODO: We will review these and may downgrade this to an info later.
-                                                   (log/warn "In compute cluster" name ", pod" pod-name "in a weird cook expected state:" cook-expected-state-dict "and k8s actual state" k8s-actual-state-dict)
+                                                   (log/warn "In compute cluster" name ", pod" pod-name
+                                                             "in a weird cook expected state:"
+                                                             (prepare-cook-expected-state-dict-for-logging cook-expected-state-dict)
+                                                             "and k8s actual state"
+                                                             (prepare-k8s-actual-state-dict-for-logging k8s-actual-state-dict))
                                                    ; TODO: Avoid a race. Say a launch occurs, followed by a kill, followed by a watch update.
                                                    ; Before the watch update, we'll think the k8s-actual-state is missing.
                                                    ; We see (:killed,:missing), with nothing to do, so we move to (;missing,:missing)
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -39,7 +39,7 @@
   "Kill pod is the same as deleting a pod, we semantically distinguish the two operations.
   Delete is used for completed pods that we're done with.
   Kill is used for possibly running pods we want to kill so that they fail.
-  Returns the cook-expected-state-dict of nil."
+  Returns the cook-expected-state-dict passed in."
   [api-client cook-expected-state-dict pod]
   (api/delete-pod api-client pod)
   cook-expected-state-dict)
@@ -367,7 +367,7 @@
                                         ; We shouldn't hit these unless we get a database rollback.
                                         :pod/failed (kill-pod-in-weird-state compute-cluster pod-name
                                                                              nil k8s-actual-state-dict)
-                                        ; This can only occur in testing when you're e.g., blowing away the database.
+                                        ; This can occur in testing when you're e.g., blowing away the database.
                                         ; It will go through :missing,:missing and then be deleted from the map.
                                         ; TODO: May be evidence of a bug where we process pod changes when we're starting up.
                                         ; Currently occurs because kill's can race ahead of launches, we kill something that has
@@ -379,7 +379,7 @@
                                         ; Unlike the other :pod/unknown states, no datomic state to update.
                                         :pod/unknown (kill-pod-in-weird-state compute-cluster pod-name
                                                                               nil k8s-actual-state-dict)
-                                        ; This can only occur in testing when you're e.g., blowing away the database.
+                                        ; This can occur in testing when you're e.g., blowing away the database.
                                         ; It will go through :missing,:missing and then be deleted from the map.
                                         ; TODO: May be evidence of a bug where we process pod changes when we're starting up.
                                         ; Currently occurs because kill's can race ahead of launches, we kill something that has
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -40,9 +40,9 @@
   Delete is used for completed pods that we're done with.
   Kill is used for possibly running pods we want to kill so that they fail.
   Returns the cook-expected-state-dict of nil."
-  [api-client pod]
+  [api-client cook-expected-state-dict pod]
   (api/delete-pod api-client pod)
-  nil)
+  cook-expected-state-dict)
 
 (defn kill-pod
   "Kill pod is the same as deleting a pod, we semantically distinguish the two operations.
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -40,9 +40,9 @@
   Delete is used for completed pods that we're done with.
   Kill is used for possibly running pods we want to kill so that they fail.
   Returns the cook-expected-state-dict of nil."
```
dposada: Isn't this no longer true?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -304,11 +304,18 @@
...
                                       :cook-expected-state/killed
                                       (case pod-synthesized-state-modified
-                                        ; This is interesting. This indicates that something deleted it behind our back!
-                                        ; Weird. We always update datomic first. Could happen if someone manually removed stuff from kubernetes.
+                                        ; TODO: This can also legitimately occur in a
+                                        ; normal occurrence if someone submits a kill request on a job that fails to launch.
                                         :missing (do
-                                                   (log-weird-state compute-cluster pod-name
-                                                                    cook-expected-state-dict k8s-actual-state-dict)
+                                                   (log/warn "In compute cluster" name ", pod" pod-name "in a weird cook expected state:" cook-expected-state-dict "and k8s actual state" k8s-actual-state-dict)
```
dposada: Why aren't we using `log-weird-state`?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -363,8 +370,9 @@
                                         ; This can only occur in testing when you're e.g., blowing away the database.
                                         ; It will go through :missing,:missing and then be deleted from the map.
                                         ; TODO: May be evidence of a bug where we process pod changes when we're starting up.
-                                        :pod/running (kill-pod-in-weird-state compute-cluster pod-name
-                                                                              nil k8s-actual-state-dict)
+                                        ; Currently occurs because kill's can race ahead of launches, we kill something that has
+                                        ; been added to datomic, but hasn't been submitted to k8s yet.
```
dposada: Above we say:

> This can only occur in testing when you're e.g., blowing away the database.

That seems to be in conflict with this.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -374,8 +382,9 @@
                                         ; This can only occur in testing when you're e.g., blowing away the database.
                                         ; It will go through :missing,:missing and then be deleted from the map.
                                         ; TODO: May be evidence of a bug where we process pod changes when we're starting up.
-                                        :pod/waiting (kill-pod-in-weird-state compute-cluster pod-name
-                                                                              nil k8s-actual-state-dict)))]
+                                        ; Currently occurs because kill's can race ahead of launches, we kill something that has
+                                        ; been added to datomic, but hasn't been submitted to k8s yet.
```
dposada: Above we say:

> This can only occur in testing when you're e.g., blowing away the database.

That seems to be in conflict with this.

scheduler/src/cook/kubernetes/controller.clj
scrosby: This case is slightly strange, but expected under some race conditions. :( 
Its too useful to ignore; its found several bugs, but can't be an error.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Fair. 'only' isn't applicable anymore.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Fair. 'only' isn't applicable anymore.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Good catch. Fixed.

scheduler/src/cook/kubernetes/controller.clj
dposada: Ah, got it. If it's expected, then shouldn't it be `INFO` level? Seems like things that we expect to happen on a regular basis should be `INFO`. 

Maybe we should create a `log-state` function for cases like this where it's expected but we don't want to ignore it.

dposada:
@scrosby It's not clear to me what bug is being fixed with this change -- can you elaborate?

scheduler/src/cook/kubernetes/controller.clj
scrosby: As discussed, we'll do it if we have a second instance of downgraded warnings. For now, keep it this way.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -286,14 +286,14 @@
...
-                                        :pod/failed (delete-pod api-client pod)
+                                        :pod/failed (delete-pod api-client cook-expected-state-dict pod)
                                         ; Who resurrected this pod? Where did it come from? Do we have two instances of cook?
                                         :pod/running (kill-pod-in-weird-state compute-cluster pod-name
                                                                               cook-expected-state-dict
                                                                               k8s-actual-state-dict)
                                         ; The writeback to datomic has occurred, so there's nothing to do except to delete the pod from kubernetes
                                         ; and remove it from our tracking.
-                                        :pod/succeeded (delete-pod api-client pod)
+                                        :pod/succeeded (delete-pod api-client cook-expected-state-dict pod)
```
scrosby: This is the bug fix. The old bad code would go:
(completed,succeeded) -> (missing,succeeded) -> (missing,missing), inadvertently deleting its local state first and keeping the kubernetes state.
The correct transitions should go:
(completed, succeeded) -> (completed,missing), (missing,missing).
so we're always tracking the local state.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -307,6 +307,7 @@
                                         ; TODO: This can also legitimately occur in a
                                         ; normal occurrence if someone submits a kill request on a job that fails to launch.
                                         :missing (do
+                                                   ; TODO: We will review these and may downgrade this to an info later.
                                                    (log/warn "In compute cluster" name ", pod" pod-name "in a weird cook expected state:" cook-expected-state-dict "and k8s actual state" k8s-actual-state-dict)
```
dposada: This should be using the `prepare-for-logging` functions.

scheduler/src/cook/kubernetes/controller.clj
dposada: Thanks, this is really helpful.

I think this is context that is super useful to put in the PR description.

dposada: @scrosby:

Seems like we should add a test to make sure we're now handling this case correctly. 

Specifically:

- assert that we only attempt to delete the pod once (without this, we're attempting 2 deletes)
- assert that we cycle through (completed, missing) when the pod deleted watch message comes in
- assert that we don't cycle through any other state pairs than the correct ones above

scrosby: I can do the first one, I can somewhat do the second, but can't do the last. We recur at the end of processing so some of the intermediate states  aren't visible outside the function. (They can still be visited if we have a failure.)

scheduler/src/cook/kubernetes/controller.clj

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -38,39 +38,62 @@
...
+                          :k8s-actual-state-map (atom {name {:synthesized-state {:state k8s-actual-state} :pod nil}})}
+                         name)
+                       (:cook-expected-state (get @cook-expected-state-map name {}))))
+        count-delete-pod (atom 0)]
+    (with-redefs [controller/delete-pod  (fn [_ cook-expected-state-dict _] (swap! count-delete-pod inc) cook-expected-state-dict)
+                  controller/log-weird-state (fn [_ _ _ _] :illegal_return_value_should_be_unused)
+                  controller/handle-pod-completed (fn [_ _] {:cook-expected-state :cook-expected-state/completed})
+                  controller/write-status-to-datomic (fn [_] :illegal_return_value_should_be_unused)]
+
+      (is (= :cook-expected-state/completed (do-process :cook-expected-state/killed :pod/succeeded)))
```
dposada: I thought our "starting" state pair for this bug fix is (completed, succeeded), not (killed, succeeded)?

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -38,39 +38,62 @@
...
+                           (atom {name {:cook-expected-state cook-expected-state}})]
+                       (controller/process
+                         {:api-client nil
+                          :cook-expected-state-map cook-expected-state-map
+                          :k8s-actual-state-map (atom {name {:synthesized-state {:state k8s-actual-state} :pod nil}})}
+                         name)
+                       (:cook-expected-state (get @cook-expected-state-map name {}))))
+        count-delete-pod (atom 0)]
+    (with-redefs [controller/delete-pod  (fn [_ cook-expected-state-dict _] (swap! count-delete-pod inc) cook-expected-state-dict)
+                  controller/log-weird-state (fn [_ _ _ _] :illegal_return_value_should_be_unused)
```
dposada: Why do we need to `with-redefs` the `log-weird-state` call?

scrosby: Here it doesn't make any sense, it just avoids spamming the log.

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Yeah. Good catch.

dposada: How was this test passing when the starting state pair was wrong? Wouldn't we want the test to fail if we accidentally start it with the wrong state pair?

scrosby: When you're in killed,succeeded you transition to (completed,succeeded), and do the deletion protocol, which is what we measure. That intermediate state was invisible because of the recur.

dposada: OK

dposada:
Will merge when green

shamsimam:
We should update the state transition diagram document too.

scrosby:
State diagram is correct.

PR #1394, opened 2020-01-30
Fix a race where cook scheduler can kill a task then later launch a t…

scrosby:

- We add a lock around launch-task in the scheduler and kill-task in compute cluster to avoid cook sending a nonsensical kill-task followed by launch task on the same task. 

This bug means that Cook can leave behind an orphaned k8s-pod (or, very likely, mesos task) running to completion, unmonitored.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -20,7 +20,8 @@
 
 ; There's an ugly race where the core cook scheduler can kill a job before it tries to launch it.
 ; What happens is:
-;   1. In launch-matched-tasks, we write instance objects to datomic for everything that matches, we have not submitted these to the compute cluster backends yet.
+;   1. In launch-matched-tasks, we write instance objects to datomic for everything that matches,
+;      we have not submitted these to the compute cluster backends yet.
 ;   2. A kill command arrives to kill the job. The job is put into completed.
 ;   3. The monitor-tx-queue happens to notice the job just completed. It sees the instance written in step 1.
 ;   4. We submit a kill-task to the compute cluster backend.
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -18,6 +18,28 @@
             [cook.config :as config]
             [datomic.api :as d]))
 
+; There's an ugly race where the core cook scheduler can kill a job before it tries to launch it.
+; What happens is:
+;   1. In launch-matched-tasks, we write instance objects to datomic for everything that matches, we have not submitted these to the compute cluster backends yet.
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -18,6 +18,28 @@
             [cook.config :as config]
             [datomic.api :as d]))
 
+; There's an ugly race where the core cook scheduler can kill a job before it tries to launch it.
+; What happens is:
+;   1. In launch-matched-tasks, we write instance objects to datomic for everything that matches, we have not submitted these to the compute cluster backends yet.
```
dposada: Nitpick: this line seems too long

scheduler/src/cook/compute_cluster.clj
```diff
@@ -18,6 +18,28 @@
...
+;   2. A kill command arrives to kill the job. THe job is put into completed.
```
dposada:
```suggestion
;   2. A kill command arrives to kill the job. The job is put into completed.
```

scheduler/src/cook/compute_cluster.clj
```diff
@@ -18,6 +18,28 @@
...
+;   3. This monitor-tx-queue happens to notice the job just completed. It sees the instance written in step 1.
```
dposada:
```suggestion
;   3. The monitor-tx-queue happens to notice the job just completed. It sees the instance written in step 1.
```

scheduler/src/cook/compute_cluster.clj
```diff
@@ -64,10 +86,11 @@
 (defn safe-kill-task
   "A safe version of kill task that never throws. This reduces the risk that errors in one compute cluster propagate and cause problems in another compute cluster."
   [{:keys [name] :as compute-cluster} task-id]
-  (try
-    (kill-task compute-cluster task-id)
-    (catch Throwable t
-      (log/error t "In compute cluster" name ", error killing task" task-id))))
+  (locking kill-lock-object
```
dposada: Why not put the `locking` inside the `try`?

scheduler/src/cook/compute_cluster.clj
scrosby: Ok. Wrapped it.

scheduler/src/cook/compute_cluster.clj
scrosby: It can go either way in the current code. However, this is safer; as I expect us to add additional logic to the catch handler per some queued tickets.

dposada: OK, sounds good

dposada:
Will merge when green

PR #1395, opened 2020-01-30
Fix a race where we don't kill a pod thats been sent to kubernetes because the watch hasn't returned yet.

scrosby:

- When we're in the state where we have a starting pod, and kill it, we may think the pod's not on kubernetes because the watch is stale, so we'll not kill it. In this change, we kill the pod. 
- We also log when remaining weirder corner cases occur.

Makes the cook log look less scary and suppresses it mentioning about weird states.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -326,12 +326,17 @@
                                         ; So, a better approach. If we detect a :missing,:killed, then we opportunistically
                                         ; try to kill the pod. This is why update-cook-expected-state saves :launch-pod,
                                         ; so its available here.
-                                        (if-let [pod (:launch-pod cook-expected-state-dict)]
+                                        (if-let [pod (some-> cook-expected-state-dict :launch-pod :pod)]
                                           ; This is interesting. This indicates that something deleted it behind our back!
                                           ; Weird. We always update datomic first. Could happen if someone manually removed stuff from kubernetes.
-                                          (do (log/info "In compute cluster" name ", opportunistically killing" name "because of potential race where kill arrives before the watch responds to the launch")
+                                          (do (log/info "In compute cluster" name ", opportunistically killing" pod-name "because of potential race where kill arrives before the watch responds to the launch")
                                               (kill-pod api-client :ignored pod))
                                           (do
+                                            (log/info "In compute cluster" name ", pod" pod-name
+                                                      "was killed with cook expected state"
+                                                      (prepare-cook-expected-state-dict-for-logging cook-expected-state-dict)
+                                                      "and k8s actual state"
+                                                      (prepare-k8s-actual-state-dict-for-logging k8s-actual-state-dict))
                                             (log-weird-state compute-cluster pod-name
                                                              cook-expected-state-dict k8s-actual-state-dict)
                                             (handle-pod-killed compute-cluster pod-name)))
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -323,7 +323,7 @@
                                         ; If we did nothing, we'd log this as a weird state, then when the watch
                                         ; shows up, we'd see (:missing, :starting) and log that as a weird state too.
                                         ;
-                                        ; So, a better approach. If we detect a :missing,:killed, then we opportunistically
+                                        ; So, a better approach. If we detect a (:killed, :missing), then we opportunistically
                                         ; try to kill the pod. This is why update-cook-expected-state saves :launch-pod,
                                         ; so its available here.
                                         (if-let [pod (some-> cook-expected-state-dict :launch-pod :pod)]
@@ -333,7 +333,7 @@
                                             ; We treat a deleting pod in kubernetes the same as a missing pod when coming up with a synthesized state.
                                             ; That's good for (almost) all parts of the system. However,
                                             ; If it is legitimately missing, then something weird is going on. If it is
-                                            ; deleting, that's an expected state. So, lets be selective with our logging.
+                                            ; deleting, that's an expected state. So, let's be selective with our logging.
                                             (if (= (:state synthesized-state) :missing)
                                               (log/info "In compute cluster" name ", pod" pod-name
                                                         "was killed with cook expected state"
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -330,7 +330,10 @@
                                           (do
                                             (log/info "In compute cluster" name ", opportunistically killing" pod-name
                                                       "because of potential race where kill arrives before the watch responds to the launch")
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -304,24 +304,34 @@
...
-                                                   (handle-pod-killed compute-cluster pod-name))
+                                        :missing
+                                        ; There's a race where we can launch then kill, when the kill arrives after
+                                        ; the launch, but before the watch updates k8s-actual-state.
+                                        ; (k8s-actual-state isn't the actual state, because of this watch lag)
+                                        ; When that happens, we will have an actual state of :missing.
+                                        ; If we did nothing, we'd log this as a weird state, then when the watch
+                                        ; shows up, we'd see (:missing, :starting) and log that as a weird state too.
+                                        ;
+                                        ; So, a better approach. If we detect a :missing,:killed, then we opportunistically
```
dposada:
```suggestion
                                        ; So, a better approach. If we detect a (:killed, :missing), then we opportunistically
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -304,24 +304,34 @@
...
+                                        ; try to kill the pod. This is why update-cook-expected-state saves :launch-pod,
+                                        ; so its available here.
+                                        (if-let [pod (some-> cook-expected-state-dict :launch-pod :pod)]
+                                          (do (log/info "In compute cluster" name ", opportunistically killing" pod-name "because of potential race where kill arrives before the watch responds to the launch")
```
dposada: The `log/info` should start on the next line.

Also, the line seems too long.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -304,24 +304,34 @@
...
+                                              (kill-pod api-client :ignored pod))
+                                          (do
+                                            ; We treat a deleting pod in kubernetes the same as a missing pod when coming up with a synthesized state.
+                                            ; That's good for (almost) all parts of the system. However,
+                                            ; If it is legitimately missing, then something weird is going on. If it is
+                                            ; deleting, that's an expected state. So, lets be selective with our logging.
```
dposada:
```suggestion
                                            ; deleting, that's an expected state. So, let's be selective with our logging.
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -304,24 +304,34 @@
...
+                                            (if (= (:state synthesized-state) :missing)
+                                              (log/info "In compute cluster" name ", pod" pod-name
+                                                        "was killed with cook expected state"
```
dposada: I don't think we need this log. We already have the same thing logged at the same level at the top of the `process` function.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -434,9 +444,12 @@
   "Update the cook expected state. Include some business logic to e.g., not change a state to the same value more than once. Marks any state changes Also has a lattice of state. Called externally and from state machine."
   [{:keys [cook-expected-state-map] :as compute-cluster} pod-name new-cook-expected-state-dict]
   (locking (calculate-lock [pod-name])
-    (let [old-state (get @cook-expected-state-map pod-name)]
-      (when-not (cook-expected-state-equivalent? new-cook-expected-state-dict old-state)
-        (swap! cook-expected-state-map assoc pod-name new-cook-expected-state-dict)
+    (let [old-state (get @cook-expected-state-map pod-name)
+          ; Save the launch pod. We may need it in order to kill it. See note under (:killed, :missing) in process, above.
+          old-pod (:launch-pod old-state)
+          new-expected-state-dict-merged (merge {:launch-pod old-pod} new-cook-expected-state-dict)]
```
dposada: Can we `assoc` here instead of `merge`-ing?

scheduler/src/cook/kubernetes/controller.clj
scrosby: Ok. Fixed.

scheduler/src/cook/kubernetes/controller.clj
scrosby: This log is here to distinguish this branch from the other branch. Each of the three subcases in this state should be distinguished clearly.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Not obviously. What I want here is to use the entire new-cook-expected-state-dict, except that if it lacks :launch-pod, we should copy from the prior state, if it exists, without having to enumerate the specific keys in new-cook-expected-state-dict. I can't think of a nice idiom to do this with assoc as it overwrites whats already there.

scheduler/src/cook/kubernetes/controller.clj
dposada: OK sounds good

scheduler/src/cook/kubernetes/controller.clj
dposada: Ah, makes sense, thanks

dposada:
@scrosby Did you see my question above?

> Shouldn't we add a test for this?

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -72,6 +72,46 @@
...
+        k8s-actual-state-map (atom {})
+        mock-cc {:api-client nil
+                 :cook-expected-state-map cook-expected-state-map
+                 :k8s-actual-state-map k8s-actual-state-map}
+        extract-cook-expected-state (fn []
+                                      (:cook-expected-state (get @cook-expected-state-map pod-name {})))
+        count-kill-pod (atom 0)]
+    (with-redefs [controller/kill-pod  (fn [_ cook-expected-state-dict _] (swap! count-kill-pod inc) cook-expected-state-dict)
+                  controller/launch-pod (fn [_ cook-expected-state-dict] cook-expected-state-dict)
+                  controller/log-weird-state (fn [_ _ _ _] :illegal_return_value_should_be_unused)
```
dposada: Why do we need to `with-redefs` `log-weird-state`?

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -72,6 +72,46 @@
...
+                  controller/handle-pod-completed (fn [_ _] {:cook-expected-state :cook-expected-state/completed})
+                  ;controller/handle-pod-killed (fn [_ _] {:cook-expected-state :cook-expected-state/completed})
```
dposada: Delete?

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Nuked.

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Nuked.

PR #1397, opened 2020-01-31
Makes test_progress_update_submit more reliable

dposada:

We need to wait for the job to complete in this test.

If we only wait until Running, the job might start running, then the instance might fail with e.g. "Agent removed", and then we'd have to wait for it to get scheduled again but we'd already have moved past this wait, which can cause the test to flake.


DaoWen:
I don't think this is the right fix for this test, but it's better to apply this patch now, and I'll remove all of the unnecessary assertions (e.g., wait for sandbox) later.

PR #1399, opened 2020-01-31
Add support to blacklist nodes in kubernetes that have certain labels

scrosby:

- Add support to blacklist nodes in kubernetes that have certain labels. Nodes with these labels will not be used for scheduling. These labels can be dynamically added and removed.

So we can defer using nodes until they're fully initialized.

### DIFF
--- a/scheduler/config-k8s.edn
+++ b/scheduler/config-k8s.edn
@@ -19,11 +19,13 @@
  :compute-clusters [{:factory-fn cook.kubernetes.compute-cluster/factory-fn
                      :config {:compute-cluster-name "gke-1"
                               ;; Location of the kubernetes config file. Hardcoded to the location specified by bin/make-gke-test-cluster
-                              :config-file "../scheduler/.cook_kubeconfig_1"}}
+                              :config-file "../scheduler/.cook_kubeconfig_1"
+                              :blacklist-labels ["blacklist-nodes-with-this-label-key"]}}
                     {:factory-fn cook.kubernetes.compute-cluster/factory-fn
                      :config {:compute-cluster-name "gke-2"
                               ;; Location of the kubernetes config file. Hardcoded to the location specified by bin/make-gke-test-cluster
-                              :config-file "../scheduler/.cook_kubeconfig_2"}}]
+                              :config-file "../scheduler/.cook_kubeconfig_2"
+                              :blacklist-labels ["blacklist-nodes-with-this-label-key-1" "blacklist-nodes-with-this-label-key-2"]}}]
  :cors-origins ["https?://cors.example.com"]
  :database {:datomic-uri #config/env "COOK_DATOMIC_URI"}
  :plugins {:job-submission-validator {:batch-timeout-seconds 40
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -241,22 +241,28 @@
...
-      (when-not below-pod-count-capacity
-        (log/info "Filtering out" node-name "because it is at or above its pod count capacity of"
-                  pod-count-capacity "(" pods-on-node ")"))
-      (and schedulable below-pod-count-capacity))))
+          labels-on-node (or (some-> node .getMetadata .getLabels) {})
+          matching-blacklist-labels (some #(get labels-on-node %) blacklist-labels)]
+      (cond  (seq other-taints) (do
+                                  (log/info "Filtering out" node-name "because it has taints" other-taints)
+                                  false)
+             (>= pods-on-node pod-count-capacity) (do
```
scrosby: FYI: I've been unable to get this integration test to pass in open source, so I cannot test if I've broken this functionality. Testing suggestions?

scheduler/config-k8s.edn
```diff
@@ -19,11 +19,13 @@
  :compute-clusters [{:factory-fn cook.kubernetes.compute-cluster/factory-fn
                      :config {:compute-cluster-name "gke-1"
                               ;; Location of the kubernetes config file. Hardcoded to the location specified by bin/make-gke-test-cluster
-                              :config-file "../scheduler/.cook_kubeconfig_1"}}
+                              :config-file "../scheduler/.cook_kubeconfig_1"
+                              :blacklist-labels ["blacklist-nodes-with-this-label-key"]}}
```
dposada: I think we should avoid "blacklist". How about "blocklist" or "deny-list" or "ignore"?

Also, maybe the word "node" should be in the config key, e.g. `:node-ignore-list`.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -241,22 +241,28 @@
     (-> node-name->pods (get node-name []) count)))
 
 (defn node-schedulable?
-  "Can we schedule on a node. For now, yes, unless there are other taints on it. TODO: Incorporate other node-health measures here."
-  [^V1Node node pod-count-capacity pods]
+  "Can we schedule on a node. For now, yes, unless there are other taints on it. TODO: Incorporate other node-health measures here.
```
dposada: > For now, yes, unless there are other taints on it.

Isn't this no longer true?

scheduler/config-k8s.edn
scrosby: Agreed. Changed.

scheduler/src/cook/kubernetes/api.clj
scrosby: Edited the sentences together.

dposada:
@scrosby Did you see my question above?

> Shouldn't we add a test for this?

scrosby:
Yes. We discussed this in person on Friday. I answered, that question only has one answer. We should always add a test. 
As to whether it was worth the time/effort, we did not come to a conclusion. We didn't come to a resolution on whether we should do this particular test or not. We'll have to manufacture up an appropriate V1Pod to run such a test; we can't mock the accessors via with-redefs.

PR #1402, opened 2020-02-01
Adds test and logging for default pool

dposada:

- adding an integration test that explicitly submits with no pool specified
- commenting out the sidecar configuration
- adding some logs for settings and for `(config/default-pool)`

For some reason, when the `:job-resource-adjustment` config is present, `(config/default-pool)` returns `nil`, which is bad. The test will help us catch issues like this in the future.

Fixing the issue will come in a later patch (@nsinkov has identified the cause).

### DIFF
--- a/scheduler/config-k8s.edn
+++ b/scheduler/config-k8s.edn
@@ -35,16 +35,19 @@
  :hostname #config/env "COOK_HOSTNAME"
  :kubernetes {:disallowed-container-paths #{"/mnt/bad"}
               :disallowed-var-names #{"BADVAR"}
-              :init-container {:command ["/bin/sh" "-c" "echo sample init container running"]
-                               :image "byrnedo/alpine-curl:latest"}
-              :sandbox-fileserver {:command ["fileserver"]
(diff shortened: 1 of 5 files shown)


scheduler/config-k8s.edn
```diff
@@ -35,16 +35,19 @@
...
+              ;                 :image "byrnedo/alpine-curl:latest"}
+              ;:sandbox-fileserver {:command ["fileserver"]
+              ;                     :image "twosigma/cook-fileserver:latest"
+              ;                     :port 23906
+              ;                     :resource-requirements {:cpu-request 0.1
+              ;                                             :cpu-limit 0.1
+              ;                                             :memory-request 128.0
+              ;                                             :memory-limit 128.0}}
+              ;:custom-shell ["/bin/sh" "-c"]
+              }
```
shamsimam: We should put the comment in a TODO if we choose to leave the code commented out.

Else I would vote to remove the code altogether.
Same for the commented code below.

dposada: Added `TODO`s, thanks

PR #1403, opened 2020-02-04
Simplify progress reporting integration test logic

DaoWen:

- Replace the wait_for_sandbox helper with wait_for_output_url
- Don't require sandbox directory or output url for progress tests
- Remove assertions requiring executor to be "cook"
- Add is_job_progress_supported predicate
- Refactor to use new wait_for_instance_with_progress helper
- Fix default progress values used in integration code
- Fix some warnings caused by invalid string escape sequences

Making these integration tests compatible with Kubernetes scheduler. These changes enable us to use these tests with the progress reporter in the [Kubernetes cook sidecar process](https://github.com/twosigma/Cook/pull/1404) (still WIP).


integration/tests/cook/test_basic.py
```diff
@@ -625,27 +612,23 @@ def test_multiple_progress_updates_submit(self):
...
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         message = json.dumps(job, sort_keys=True)
         self.assertEqual('success', job['state'], message)
-
-        instance = util.wait_for_sandbox_directory(self.cook_url, job_uuid, 'success')
-        message = json.dumps(instance, sort_keys=True)
-        self.assertIsNotNone(instance['output_url'], message)
-        self.assertIsNotNone(instance['sandbox_directory'], message)
-        self.assertEqual('cook', instance['executor'])
-        util.sleep_for_publish_interval(self.cook_url)
```
shamsimam: Is the exit code going to be populated after all the progress reporting has completed? If not, we might still need the sleep.

integration/tests/cook/test_basic.py
```diff
@@ -661,20 +644,15 @@ def test_multiple_progress_updates_submit_stdout(self):
...
```
shamsimam: Is the exit code going to be populated after all the progress reporting has completed? If not, we might still need the sleep.

integration/tests/cook/util.py
```diff
@@ -1330,6 +1315,14 @@ def is_cook_executor_in_use():
         return False
 
 
+@functools.lru_cache()
+def is_job_progress_supported():
+    """Returns true if the current job execution environment supports progress reporting"""
+    # Mesos supports progress reporting only with Cook Executor,
+    # but our progress reporter sidecar is always enabled on Kubernetes.
+    return is_cook_executor_in_use() # or using_kubernetes()
```
shamsimam: Please add a TODO here

integration/tests/cook/test_basic.py
DaoWen: I don't understand what `sleep_for_publish_interval` is doing here.

We sleep/retry until the exit code shows up. That part's unchanged. Now we _also_ sleep/retry until the progress percentage we want shows up—which I thought would be redundant with this call to `sleep_for_publish_interval`.

Do we need both?

integration/tests/cook/test_basic.py
DaoWen: Same as above.

integration/tests/cook/util.py
DaoWen: 👍

integration/tests/cook/test_basic.py
DaoWen: We discussed this offline and we both concluded that the extra sleep here is not necessary.

However, I did add a new `time.sleep` call to a couple of the tests. Since the jobs they run include sleep commands too, it seems like the test should sleep for the same amount of time before it starts polling for a successful instance.

integration/tests/cook/util.py
DaoWen: Done

DaoWen:
Rebased onto master to resolve a merge conflict.

PR #1406, opened 2020-02-04
Fix overlapping k8s mounts

scrosby:

- Fix overlapping k8s mounts. When the workdir has the same path as a mount, we make a bad pod configuration that's rejected by k8s.

We have a failing integration test and this is a legit configuration.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -357,30 +357,34 @@ class CookTest(util.CookTest):
                                   'host-path': '/var/lib/mno',
                                   'container-path': '/var/lib/pqr'}]}
         job_uuid, resp = util.submit_job(self.cook_url, container=container)
-        self.assertEqual(resp.status_code, 201, msg=resp.content)
-        self.assertEqual(resp.content, str.encode(f"submitted jobs {job_uuid}"))
-        job = util.load_job(self.cook_url, job_uuid)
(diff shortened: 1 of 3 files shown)


scrosby:
Good suggestion. Done.

scrosby:
Internal green except for test_progress_update_rest failed internally. That's new functionality so I suspect flakiness. Job succeeded.

integration/tests/cook/test_basic.py
```diff
@@ -357,30 +357,34 @@ def test_docker_fields(self):
...
+            self.assertLessEqual(4, len(volumes))
+            self.assertIn({'host-path': '/var/lib/abc'}, volumes)
+            self.assertIn({'mode': 'RW',
+                           'host-path': '/var/lib/def'}, volumes)
+            self.assertIn({'host-path': '/var/lib/ghi',
+                           'container-path': '/var/lib/jkl'}, volumes)
+            self.assertIn({'mode': 'RW',
+                           'host-path': '/var/lib/mno',
+                           'container-path': '/var/lib/pqr'}, volumes)
+            util.wait_for_job_in_statuses(self.cook_url, job_uuid, ['running', 'completed'])
```
dposada: Sorry, I should have thought of this before:

I think we should assert that the job succeeded, because it could fail and still pass this assertion.

PR #1407, opened 2020-02-04
Suppress errors deleting already deleted pods

scrosby:

- Suppress raising an error when we try to delete a pod thats not in k8s.

Avoids log spam and makes it easier to followup on real problems.


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -671,7 +672,13 @@
         ;
         ; They actually advise catching the exception and moving on!
         ;
-        ))))
+        (log/info "Caught the https://github.com/kubernetes-client/java/issues/252 exception."))
```
dposada: Can we add the pod name to this log?

scrosby: Done.

dposada:
Will merge when green

PR #1409, opened 2020-02-05
Fix nil default pool

nsinkov:

- don't try to resolve function names when processing the config file

The act of resolving function names when processing the config file has side effects. Some of the side effects can be the calling of functions that depend on the config itself. Until we figure out a way to resolve function names without side effects we need to go back to resolving functions at runtime


dposada:
Will merge when green

PR #1416, opened 2020-02-07
Avoid NPE when a running job goes completely missing

scrosby:

- Write new logic for the :running,:missing case.

In cases where a pod disappears on us completely, we'd get a NPE. This can occur in some cases when GKE preempts a node. This changes the logic into a path that no longer NPE's.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -1,5 +1,6 @@
 (ns cook.kubernetes.controller
-  (:require [cook.config :as config]
+  (:require [cook.compute-cluster :as cc]
+            [cook.config :as config]
             [cook.datomic :as datomic]
             [cook.kubernetes.api :as api]
             [cook.mesos.sandbox :as sandbox]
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -368,8 +380,13 @@
                                         :missing (do
                                                    (log/error "In compute cluster" name ", something deleted"
                                                               pod-name "behind our back")
-                                                   ; TODO: Should mark mea culpa retry
-                                                   (handle-pod-completed compute-cluster k8s-actual-state-dict))
+                                                     ; A :cook-expected-state/running job suddenly disappearing in k8s is
+                                                     ; a sign of a preemption, so treat this case as a missed preemption.
+                                                     ; TODO: When we have a better story for preemption, we should switch this to
+                                                     ; go through the handle-pod-completed stack. THat needs to be guarded with a null
```
dposada:
```suggestion
                                                     ; go through the handle-pod-completed stack. That needs to be guarded with a null
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -368,8 +380,13 @@
                                         :missing (do
                                                    (log/error "In compute cluster" name ", something deleted"
```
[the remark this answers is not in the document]
scrosby: Yes and no. As soon as we get real preemption detector, these should go away except for genuine errors.

scrosby:
I had the nil check in handle-pod-completed in an earlier version, but the only path where it could be nil was the one case that's now removed as we're not calling it in that context where it can be nil.

As far as adding an exception handler, what should the exception handler do. My vote is let the code blow up, then we'll fix it. I don't want to be in the game of predicting how kubernetes will misbehave and guessing on what the right response is.

scheduler/src/cook/kubernetes/controller.clj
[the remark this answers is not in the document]
dposada: Makes sense. How about if we change to `info` now and change back to `error` when we get the preemption detector?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -107,13 +107,13 @@
     (write-status-to-datomic compute-cluster status)
     {:cook-expected-state :cook-expected-state/completed}))
 
-(defn handle-pod-preemption;
+(defn handle-pod-preemption
```
dposada: Should we change the (running, waiting) case to use this new function as well?

dposada:
> the only path where it could be nil was the one case that's now removed as we're not calling it in that context where it can be nil

Sounds good.

> what should the exception handler do

It should catch, log an `ERROR`, and write the instance status as Failed to Datomic. The point is to lower the likelihood of dead pods leaving `Running` instances behind.

> I don't want to be in the game of predicting how kubernetes will misbehave and guessing on what the right response is

Totally agree. We don't need to predict how k8s will behave. We just need to do our best to not let errors leave behind forever-`Running` instances.

scheduler/src/cook/kubernetes/controller.clj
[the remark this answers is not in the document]

scheduler/src/cook/kubernetes/controller.clj

scrosby:
Added the extra logic to handle-completed-pod and the resulting refactor to support it.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -167,28 +168,54 @@
                                          (.getName container-status))))
        first))
 
+(defn make-failed-unknown-task-result
```
dposada: Maybe this should just be `make-failed-task-result`, since it's not always using the `unknown` reason.

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -2,7 +2,8 @@
   (:require [clojure.test :refer :all]
             [cook.kubernetes.api :as api]
             [cook.kubernetes.controller :as controller]
-            [cook.test.testutil :as tu])
+            [cook.test.testutil :as tu]
+            [cook.compute-cluster :as cc])
```
dposada: Lexicographic

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -165,6 +166,38 @@
...
+        (is (= :reason-task-unknown @reason)))))
+
+  (testing "pod mismatch"
+    (let [pod (tu/pod-helper "podA" "hostA" {})
+          reason (atom nil)]
+      (.setStatus pod (V1PodStatus.))
+      (with-redefs [cc/compute-cluster-name (constantly "compute-cluster-name")
+                    controller/write-status-to-datomic (fn [_ status] (reset! reason (:reason status)))]
+        (try
+        (controller/handle-pod-completed nil "podB" {:pod pod :synthesized-state {:state :pod/failed}})
```
dposada: We can use `thrown?` here: https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/is

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -165,6 +166,38 @@
...
+        ; We expect an AssertionError
+        (is false) ; This line shouldn't execute.
+        (catch AssertionError e)))))
+  (testing "pod is nil"
```
dposada: For consistency, newline before the new `testing`

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -167,28 +168,54 @@
...
+            :state :task-failed
+            :reason (or reason :reason-task-unknown)}})
+
+(defn calculate-pod-status
+  "Calculate the pod status of a completed pod. Factored out so that we can wrap it in an exception handler."
+  [compute-cluster pod-name {:keys [synthesized-state pod]} & {:keys [reason]}]
+  (let [instance-id (some-> pod .getMetadata .getName)]
+    (when instance-id
+      (assert (= instance-id pod-name)))
+    ; If we have an actual pod here....
```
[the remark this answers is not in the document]
scrosby: Changed.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -167,28 +168,54 @@
...
+                                                                    pod-status job-container-status))}
+              exit-code (some-> job-container-status .getState .getTerminated .getExitCode)]
+          {:status status :exit-code exit-code})
+        (do
+          (log/error "In compute cluster" (cc/compute-cluster-name compute-cluster) ", pod" pod-name
+                     "had a null pod")
+          (make-failed-unknown-task-result pod-name reason)))
+      (catch Throwable t
+        (log/error t "In compute cluster" (cc/compute-cluster-name compute-cluster) ", pod" pod-name
+                   "threw an error computing calculate pod status for write to datomic.")
```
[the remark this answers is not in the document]

scheduler/src/cook/kubernetes/controller.clj
scrosby: Renamed.

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Changed.

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Ok.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -179,10 +179,11 @@
   "Calculate the pod status of a completed pod. Factored out so that we can wrap it in an exception handler."
   [compute-cluster pod-name {:keys [synthesized-state pod]} & {:keys [reason]}]
   (let [instance-id (some-> pod .getMetadata .getName)]
+    ; If we have an actual pod here, make sure its got the same name as pod-name.
```
dposada:
```suggestion
    ; If we have an actual pod here, make sure it has the same name as pod-name.
```

scheduler/test/cook/test/kubernetes/controller.clj
scrosby: Ah. Was not aware of it. Thanks!

PR #1417, opened 2020-02-07
Downgrade this ERROR to an INFO. This exclusively is 'job can't start.

scrosby:

- Downgrade from ERROR to INFO.

Its a symptom of an innocous error, a kill/launch race where a user kills a job shortly before it launches. Doing this cuts the number of logged errors by 50x.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -822,7 +822,7 @@
               matched-considerable-jobs-head?)))
         (catch Throwable t
           (meters/mark! handle-resource-offer!-errors)
-          (log/error t "In" pool-name "pool, error in match:" (ex-data t))
+          (log/info t "In" pool-name "pool, error in match:" (ex-data t))
           (when-let [offers @offer-stash]
             ; Group the set of all offers by compute cluster and route them to that compute cluster for restoring.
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -822,7 +822,7 @@
               matched-considerable-jobs-head?)))
         (catch Throwable t
           (meters/mark! handle-resource-offer!-errors)
-          (log/error t "In" pool-name "pool, error in match:" (ex-data t))
+          (log/info t "In" pool-name "pool, error in match:" (ex-data t))
```
dposada: Can we change to INFO for only the "The job can't start now" exception?

Seems like other exceptions should still be ERROR.

scrosby: There are no other exception subtypes in the last 30 days thrown. This code is not executed in our integration tests. I'm reluctant to make any change more invasive than this without testing it, and testing it pushes this change over the quicky level of effort.

dposada: We control the "The job can't start now" exception and we catch it here. It seems easy to add a check for that exception (e.g. with a field in the exception info map) and only log `info` for that case.

PR #1418, opened 2020-02-07
don't resolve the sandbox directory when ssh'ing kubernetes

nsinkov:

- don't look up sandbox directory when ssh'ing kubernetes

looking up the sandbox directory when ssh'ing kubernetes is not needed and will fail when there is no ouput_url. There will not be an ouput_url when the fileserver sidecar is not enabled. But a user should still be able to ssh into their pod

### DIFF
--- a/cli/cook/querying.py
+++ b/cli/cook/querying.py
@@ -243,7 +243,7 @@ def __get_latest_instance(job):
     raise CookRetriableException(f'Job {job["uuid"]} currently has no instances.')
 
 
-def query_unique_and_run(clusters, entity_ref, command_fn, wait=False):
+def query_unique_and_run(clusters, entity_ref, command_fn, wait=False, do_not_resove_sandbox_directory=False):
     """Calls query_unique and then calls the given command_fn on the resulting job instance"""
 
     def query_unique_and_run():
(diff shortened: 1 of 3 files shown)


cli/cook/querying.py
```diff
@@ -243,19 +243,21 @@ def __get_latest_instance(job):
     raise CookRetriableException(f'Job {job["uuid"]} currently has no instances.')
 
 
-def query_unique_and_run(clusters, entity_ref, command_fn, wait=False):
+def query_unique_and_run(clusters, entity_ref, command_fn, wait=False, do_not_resove_sandbox_directory=False):
```
dposada: Seems easier to understand/read to have this be inverted - `resolve_sandbox_directory=True`

cli/cook/querying.py
```diff
@@ -243,19 +243,21 @@ def __get_latest_instance(job):
...
             instance = __get_latest_instance(job)
-            directory = mesos.retrieve_instance_sandbox_directory(instance, job)
+            directory = partial(mesos.retrieve_instance_sandbox_directory, instance=instance, job=job)\
+                if do_not_resove_sandbox_directory else mesos.retrieve_instance_sandbox_directory(instance, job)
             command_fn(instance, directory, query_result['cluster'])
         elif query_result['type'] == Types.INSTANCE:
             instance, job = query_result['data']
-            directory = mesos.retrieve_instance_sandbox_directory(instance, job)
+            directory = partial(mesos.retrieve_instance_sandbox_directory, instance=instance, job=job) \
+                if do_not_resove_sandbox_directory else mesos.retrieve_instance_sandbox_directory(instance, job)
```
dposada: Seems confusing to me that it sometimes returns a function and sometimes returns a value

cli/cook/querying.py
nsinkov: ok

cli/cook/querying.py
nsinkov: that's the tradeoff for otherwise not refactoring much. do you have a preferred alternative?

cli/cook/querying.py
```diff
@@ -243,19 +243,21 @@ def __get_latest_instance(job):
     raise CookRetriableException(f'Job {job["uuid"]} currently has no instances.')
 
 
-def query_unique_and_run(clusters, entity_ref, command_fn, wait=False):
+def query_unique_and_run(clusters, entity_ref, command_fn, wait=False, resove_sandbox_directory=True):
```
dposada:
```suggestion
def query_unique_and_run(clusters, entity_ref, command_fn, wait=False, resolve_sandbox_directory=True):
```

cli/cook/querying.py
dposada: OK sounds good

cli/cook/querying.py
dposada: @nsinkov ☝️

dposada:
Will merge when green

PR #1419, opened 2020-02-10
Rename fileserver to more generic cook.sidecar

DaoWen:

Rename fileserver to more generic cook.sidecar

We want to add more functionality (e.g., progress reporting) to the sidecar.


DaoWen:
Self-merging since Shams already approved (pending internal-green).

PR #1420, opened 2020-02-10
Added new integration test for kubernetes volume=workdir mount successfully launches.

scrosby:

- Fix integration test broken internally in PR #1406.

- The change to the integration test to validate the instance successfully launched succeeded in minimesos and k8s, but failed in mesos. So, revert that change and make a new integration test for overlapping workdir and volume mount.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -408,6 +408,7 @@ class CookTest(util.CookTest):
             self.assertEqual(resp.content, str.encode(f"submitted jobs {job_uuid}"))
             job = util.load_job(self.cook_url, job_uuid)
             util.wait_for_job(self.cook_url, job_uuid, 'completed')
+            util.wait_for_instance(self.cook_url, job_uuid, status='success')
         finally:
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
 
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -385,8 +385,29 @@ def test_docker_fields(self):
...
+                                               {'key': 'workdir', 'value': '/var/lib/pqr'}]},
+                     'volumes': [{'mode': 'RW',
+                                  'host-path': '/var/lib/mno',
+                                  'container-path': '/var/lib/pqr'}]}
+        job_uuid, resp = util.submit_job(self.cook_url, container=container)
+        try:
+            self.assertEqual(resp.status_code, 201, msg=resp.content)
+            self.assertEqual(resp.content, str.encode(f"submitted jobs {job_uuid}"))
+            job = util.load_job(self.cook_url, job_uuid)
+            util.wait_for_job(self.cook_url, job_uuid, 'completed')
```
dposada: Don't we want:

```
util.wait_for_instance(self.cook_url, job_uuid, status='success')
```

?

scrosby: arrrgghhh! Good catch.

dposada:
Will merge when green

PR #1422, opened 2020-02-11
Retry k8s startup forever

scrosby:

- Retry k8s startup forever.
- Some minor logging improvements. 

The kubernetes watches used to retry forever, but the initial cluster setup would not retry. It needs to 
enumerate the pods in k8s in order to determine what state to load from datomic.


dposada:
Will merge when green

PR #1423, opened 2020-02-11
Fast fail when we get a bad request response from k8s

scrosby:

- Fast fail when we get a bad request response from k8s

If we submit a broken configuration, no need to let it get stuck.


dposada:
Will merge when green

PR #1425, opened 2020-02-12
Uses 'Normal exit' reason for successful k8s instances

dposada:

- adding the "Normal exit" reason (documented but never implemented until now)
- using the "Normal exit" reason when a k8s instance succeeds
- adding the `task-id` to the log when an unknown reason is passed to `mesos-reason->cook-reason-entity-id`

Without the first two changes, we're doing the following on every successful pod:

1. Calling `container-status->failure-reason` for no good reason
1. Logging warnings unnecessarily

The addition of `task-id` to the log is to make the log more useful for troubleshooting.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -193,10 +193,9 @@
                            :task-finished
                            :task-failed)
               reason (or reason 
-                         (if succeeded?
-                           :reason-normal-exit
-                           (container-status->failure-reason compute-cluster pod-name
-                                                             pod-status job-container-status)))
+                         (when succeeded? :reason-normal-exit)
+                         (container-status->failure-reason compute-cluster pod-name
+                                                           pod-status job-container-status))
               status {:task-id {:value pod-name}
                       :state task-state
                       :reason reason}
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -188,14 +188,18 @@
...
-              task-state (if (= (:state synthesized-state) :pod/succeeded)
+              succeeded? (= (:state synthesized-state) :pod/succeeded)
+              task-state (if succeeded?
                            :task-finished
                            :task-failed)
+              reason (if succeeded?
+                       :reason-normal-exit
+                       (or reason
+                           (container-status->failure-reason compute-cluster pod-name
+                                                             pod-status job-container-status)))
```
shamsimam: Should we avoid overriding reason even when it is present?
```
(or reason
    (when succeeded?
      :reason-normal-exit)
    (container-status->failure-reason compute-cluster pod-name pod-status job-container-status))
```

scheduler/src/cook/schema.clj
```diff
@@ -1196,6 +1196,12 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
 
 (def reason-entities
   [{:db/id (d/tempid :db.part/user)
+    :reason/code 1000
+    :reason/string "Normal exit"
+    :reason/name :normal-exit
+    :reason/mea-culpa? false
+    :reason/mesos-reason :reason-normal-exit}
```
shamsimam: Unfortunate naming :'(

scheduler/src/cook/kubernetes/controller.clj
dposada: Good call, done

scheduler/src/cook/schema.clj
dposada: Yeah :(

PR #1426, opened 2020-02-12
Decrease default integration test job size.

scrosby:

- Reduce the memory used by 8x forintegration test test jobs.
- Reduce the CPU used by >8x for integration test test jobs.

Increase parallelism and decrease integration test runtimes to give us higher testing throughput.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -915,26 +915,6 @@ class CookTest(util.CookTest):
         self.assertEqual(resp.status_code, 201)
         return util.get_user(self.cook_url, job_uuid)
 
-    def test_list_jobs_by_state(self):
-        # schedule a bunch of jobs in hopes of getting jobs into different statuses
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -917,7 +917,7 @@ def determine_user(self):
 
     def test_list_jobs_by_state(self):
         # schedule a bunch of jobs in hopes of getting jobs into different statuses
-        job_specs = [util.minimal_job(command=f"sleep {i}") for i in range(1, 20)]
+        job_specs = [util.minimal_job(command=f"sleep {i}", cpus=1) for i in range(1, 20)]
```
dposada: Why do these need 1 cpu?

scrosby: Because the test expects some of the jobs to wait in the queue. If you make them all very small, then all of them will execute and the check for Waiting (lower down) will break.

dposada: Discussed offline, we're going to delete this test

PR #1427, opened 2020-02-12
Name test jobs with their caller function to better identify them.

scrosby:

- Most jobs are annotated with which integration test created them. 

Many/most of our test jobs have the same name, this can make it hard to trace where a test job came from (when the underlying integration test passes). Try to auto-pick a name variant to make it easier to do this tracing.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2370,12 +2370,12 @@ class CookTest(util.CookTest):
             util.kill_jobs(self.cook_url, job_uuids)
 
     def test_submit_with_no_name(self):
-        # We need to manually set the 'uuid' to avoid having the
-        # job name automatically set by the submit_job function
+        # The code specialcases name=None and removes it before submitting.
         job_with_no_name = {'uuid': str(util.make_temporal_uuid()),
                             'command': 'ls',
                             'cpus': 0.1,
                             'mem': 16,
+                            'name' : None,
                             'max-retries': 1}
         job_uuid, resp = util.submit_job(self.cook_url, **job_with_no_name)
         self.assertEqual(resp.status_code, 201, msg=resp.content)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2370,7 +2370,7 @@ class CookTest(util.CookTest):
             util.kill_jobs(self.cook_url, job_uuids)
 
     def test_submit_with_no_name(self):
-        # The code specialcases name=None and removes it before submitting.
+        # The job submission code special-cases name=None and removes it before submitting.
         job_with_no_name = {'uuid': str(util.make_temporal_uuid()),
                             'command': 'ls',
                             'cpus': 0.1,
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -7,6 +7,7 @@ import os
 import os.path
 import re
 import subprocess
+import sys
 import threading
 import time
 import unittest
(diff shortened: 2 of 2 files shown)


integration/tests/cook/util.py
```diff
@@ -455,14 +456,25 @@ def make_temporal_uuid():
...
+    return ""
 
 def minimal_job(**kwargs):
     job = {
         'command': 'echo Default Test Command',
         'cpus': get_default_cpus(),
         'max_retries': 1,
         'mem': int(os.getenv('COOK_DEFAULT_JOB_MEM_MB', 256)),
-        'name': 'default_test_job',
+        'name': ('default_test_job-'+get_caller()),
```
dposada: What about the cases where our caller is not the test function? For example, all of the usages of `util.submit_jobs` which end up calling this function. Seems like instead of relying on the stack frames, we should get the name of the currently running test from pytest.

integration/tests/cook/test_basic.py
```diff
@@ -2381,12 +2381,12 @@ def test_user_usage_ungrouped(self):
             util.kill_jobs(self.cook_url, job_uuids)
 
     def test_submit_with_no_name(self):
-        # We need to manually set the 'uuid' to avoid having the
-        # job name automatically set by the submit_job function
+        # The code specialcases name=None and removes it before submitting.
```
dposada:
```suggestion
        # The code special-cases name=None and removes it before submitting.
```

integration/tests/cook/util.py
```diff
@@ -538,11 +553,15 @@ def submit_jobs(cook_url, job_specs, clones=1, pool=None, headers=None, log_requ
         headers = {}
     if isinstance(job_specs, dict):
         job_specs = [job_specs] * clones
-
+    caller = get_caller()
     def full_spec(spec):
+        if 'name' not in spec:
+            spec['name']="default_job-"+caller
```
dposada: Can you run this through the Python code formatter?

dposada: Why are we using `default_job-` here but `default_test_job-` above?

integration/tests/cook/util.py
```diff
@@ -615,6 +634,8 @@ def submit_job(cook_url, pool=None, headers=None, **kwargs):
     """Create and submit a single job"""
     if headers is None:
         headers = {}
+    if 'name' not in kwargs:
+        kwargs['name']="default_job-"+get_caller()
```
dposada: Since we've got this in three places, can we make it a helper function, e.g. `make_default_job_name`?

integration/tests/cook/util.py
scrosby: Yup. That's why I have similar code in the other main codepaths, e.g., submit_jobs and full_spec.

integration/tests/cook/util.py
scrosby: Adding another call into the stack means get_caller would now see another frame on the stack and get the wrong result, making that function more.... special. That said, it makes sense to refactor out the prefix.

integration/tests/cook/test_basic.py
scrosby: Done

integration/tests/cook/util.py
scrosby: Done.

scrosby: Made consistent.

integration/tests/cook/util.py
dposada: :+1:

dposada:
Will merge when green

PR #1435, opened 2020-02-14
Gate the port-related tests to actually having ports enabled.

scrosby:

- Have two integration tests properly skip themselves when max-ports is too low.

- When we set max-ports to zero, these tests fail when they should be skipping themselves.


dposada:
Will merge when green

PR #1437, opened 2020-02-14
Adds mapping for REASON_TASK_KILLED_DURING_LAUNCH

dposada:
## Changes proposed in this PR

- adding a mapping for this status reason from Mesos

## Why are we making these changes?

Mesomatic doesn't have a mapping for [`REASON_TASK_KILLED_DURING_LAUNCH`](http://mesos.apache.org/documentation/latest/task-state-reasons/#for-state-task_killed), so we're rolling our own mapping for it here.

This avoids tons of "Unknown mesos reason" warning logs.

### Related Issue: https://github.com/clojusc/mesomatic/issues/53

PR #1438, opened 2020-02-14
Adds running reason for k8s instances

dposada:

- adding a "running" reason
- setting the `:reason` to this new reason for k8s instances

The `write-status-to-datomic` code in the scheduler expects a non-`nil` reason, and when it gets `nil` (without this patch), it logs a warning and uses the `:mesos-unknown` reason, which is not what we want.

### DIFF
--- a/scheduler/docs/reason-code
+++ b/scheduler/docs/reason-code
@@ -3,6 +3,7 @@
 01001: Killed by user (Not Implemented)
 01002: Preempted by rebalancer
 01003: REASON_CONTAINER_PREEMPTED
+01005: Running
 
 02xxx: Job Misconfiguration
 02000: REASON_CONTAINER_LIMITATION
--- a/scheduler/test/cook/test/rest/api.clj
+++ b/scheduler/test/cook/test/rest/api.clj
@@ -1479,10 +1479,12 @@
                   (is (= (assoc (expected-job-map job)
                            :container (assoc-in docker-container
                                                 [:docker :parameters]
-                                                [{:key "tee" :value "tie"}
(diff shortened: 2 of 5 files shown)


scheduler/docs/reason-code
```diff
@@ -3,6 +3,7 @@
 01001: Killed by user (Not Implemented)
 01002: Preempted by rebalancer
 01003: REASON_CONTAINER_PREEMPTED
+01005: Running
```
dposada: Why `01005` instead of `01004`?: PR #1437 is introducing `01004`.

scheduler/test/cook/test/rest/api.clj
```diff
@@ -1492,9 +1492,9 @@
                   (is (= (assoc (expected-job-map job)
                            :container (assoc-in docker-container
                                                 [:docker :parameters]
-                                                [{:key "user" :value "1234:2345"}
-                                                 {:key "tee" :value "tie"}
-                                                 {:key "fee" :value "fie"}]))
+                                                [{:key "tee" :value "tie"}
+                                                 {:key "fee" :value "fie"}
+                                                 {:key "user" :value "1234:2345"}]))
```
dposada: The changes to this test are unrelated to the code changes in this PR. The test was failing, though, so I fixed it to get the build green.

nsinkov: just change to make it a set and update-in set on api/fetch-job-map

dposada: Changed the test to use `frequencies`

PR #1440, opened 2020-02-14
don't fail test when init script is used

nsinkov:

- Don't fail test when pod exits non-zero in the test_workdir_volume_overlap test and the cook init script is used

The test_workdir_volume_overlap test uses a non-writable workdir which causes the cook init script to exit non-zero


dposada:
Can we just use a writable workdir instead? Seems much simpler.

nsinkov:
This is a temporary change until we separate the sandbox and workdir and then I'll undo this

dposada:
OK, can we add a comment with what you said above? It wasn't clear this was temporary.

dposada:
Will merge when green

PR #1442, opened 2020-02-18
Change the log line for blocklist to log the key and value.

scrosby:

- Change the log line for blocklist to log the key and value.

Nicer logging when we block a node from being used.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -290,8 +290,7 @@
           node-name (some-> node .getMetadata .getName)
           pods-on-node (num-pods-on-node node-name pods)
           labels-on-node (or (some-> node .getMetadata .getLabels) {})
-          matching-node-blocklist-labels (filter #(contains? labels-on-node %) node-blocklist-labels)
-          matching-node-blocklist-keyvals (map (fn [key] [key (get labels-on-node key)]) matching-node-blocklist-labels)]
+          matching-node-blocklist-keyvals (select-keys labels-on-node node-blocklist-labels)]
       (cond  (seq other-taints) (do
                                   (log/info "Filtering out" node-name "because it has taints" other-taints)
                                   false)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -275,7 +275,9 @@
           node-name (some-> node .getMetadata .getName)
           pods-on-node (num-pods-on-node node-name pods)
           labels-on-node (or (some-> node .getMetadata .getLabels) {})
-          matching-node-blocklist-labels (some #(get labels-on-node %) node-blocklist-labels)]
+          label-names-on-node (into #{} (keys labels-on-node))
```
dposada: I don't see where this variable is being consumed.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -275,7 +275,8 @@
           node-name (some-> node .getMetadata .getName)
           pods-on-node (num-pods-on-node node-name pods)
           labels-on-node (or (some-> node .getMetadata .getLabels) {})
-          matching-node-blocklist-labels (some #(get labels-on-node %) node-blocklist-labels)]
+          matching-node-blocklist-labels (filter #(contains? labels-on-node %) node-blocklist-labels)
+          matching-node-blocklist-keyvals (map (fn [key] [key (get labels-on-node key)]) matching-node-blocklist-labels)]
```
dposada: Seems like we can simplify this by using [`select-keys`](https://clojuredocs.org/clojure.core/select-keys).

scrosby: Thats a nice function I was previously unaware! Thanks!

scheduler/src/cook/kubernetes/api.clj
scrosby: Good catch. My IDE flaked.

dposada:
Will merge when green

PR #1443, opened 2020-02-18
Fix integration test that attempts to submit a bad job with no image.

scrosby:

- Fix integration test that attempts to submit a bad job with no image.

Avoid a false positive error log.


dposada:
Will merge when green

PR #1444, opened 2020-02-18
Avoid NPE in sandbox calculation when compute cluster is not found.

scrosby:

- Avoid NPE in sandbox calculation when compute cluster is not found.
- Fix a couple of comments.

NPE is bad.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -986,11 +986,11 @@
   "Gets a URL to query the sandbox directory of the task.
    Users will need to add the file path & offset to their query.
    Refer to the 'Using the output_url' section in docs/scheduler-rest-api.adoc for further details.
-   Delegates to the compute cluster implimentation."
+   Delegates to the compute cluster implementation."
   [instance-entity]
   (if-let [sandbox-url (:instance/sandbox-url instance-entity)]
     sandbox-url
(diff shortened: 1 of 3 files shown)


nsinkov:
when is it normal for compute-cluster-name->ComputeCluster to return nil?

scheduler/src/cook/rest/api.clj
```diff
@@ -986,11 +986,11 @@
   "Gets a URL to query the sandbox directory of the task.
    Users will need to add the file path & offset to their query.
    Refer to the 'Using the output_url' section in docs/scheduler-rest-api.adoc for further details.
-   Delegates to the compute cluster implimentation."
+   Delegates to the compute cluster implementation."
   [instance-entity]
   (if-let [sandbox-url (:instance/sandbox-url instance-entity)]
     sandbox-url
-    (let [compute-cluster (task/task-ent->ComputeCluster instance-entity)]
+    (when-let [compute-cluster (task/task-ent->ComputeCluster instance-entity)]
```
nsinkov: make this if-let and log an error with the entity ID

scrosby: Done.

scrosby:
It can be nil when the compute cluster referrred to in the entity isn't in config.edn anymore.

nsinkov:
will merge when green

PR #1447, opened 2020-02-19
Improves synthetic pods

dposada:

There are a few different changes all related to improving the way synthetic pods work. The main change is introducing priority classes for Cook job pods and synthetic pods.

The priority classes allow the job pods to preempt the synthetic pods and avoid instance failures due to pending synthetic pods racing with the job pods when new capacity becomes available.

I've marked significant changes in this PR with a comment explaining the "why".

### DIFF
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1450,7 +1450,8 @@ def node_pool(nodename):
 
 def max_kubernetes_node_cpus():
     nodes = get_kubernetes_nodes()
-    return max([float(n['status']['capacity']['cpu'])
+    k8s_node_overhead_cpus = 1
+    return max([float(n['status']['capacity']['cpu']) - k8s_node_overhead_cpus
                 for n in nodes])
 
 
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -1450,7 +1450,9 @@ def node_pool(nodename):
 
 def max_kubernetes_node_cpus():
     nodes = get_kubernetes_nodes()
-    k8s_node_overhead_cpus = 1
+    # We need to account for per-node overhead
+    # (capacity not usable by our Cook pods)
+    k8s_node_overhead_cpus = int(os.getenv("COOK_TEST_K8S_NODE_OVERHEAD_CPUS", 0))
     return max([float(n['status']['capacity']['cpu']) - k8s_node_overhead_cpus
                 for n in nodes])
 
--- a/scheduler/bin/help-make-cluster
+++ b/scheduler/bin/help-make-cluster
@@ -36,12 +36,26 @@ $gcloud container clusters get-credentials "$CLUSTERNAME" --zone "$ZONE"
 
 # Make some node-pools --- this takes a while, but it helps guarantee that we have nodes with the appropriate cook-pool taints.
 echo "---- Making extra k8s-alpha and k8s-gamma nodepools for cook pools (please wait 5-10 minutes)"
-$gcloud container node-pools create cook-pool-k8s-gamma --zone "$ZONE" --cluster="$CLUSTERNAME" --disk-size=20gb --machine-type=g1-small \
+$gcloud container node-pools create cook-pool-k8s-gamma \
+       --zone "$ZONE" \
+       --cluster="$CLUSTERNAME" \
+       --disk-size=20gb \
+       --machine-type=g1-small \
        --node-taints=cook-pool=k8s-gamma:NoSchedule \
-       --enable-autoscaling --min-nodes=3 --max-nodes=6
-$gcloud container node-pools create cook-pool-k8s-alpha --zone "$ZONE" --cluster="$CLUSTERNAME" --disk-size=20gb --machine-type=g1-small \
+       --enable-autoscaling \
+       --min-nodes=3 \
+       --max-nodes=6 \
+       --node-labels=cook_pool=k8s-gamma
+$gcloud container node-pools create cook-pool-k8s-alpha \
+       --zone "$ZONE" \
+       --cluster="$CLUSTERNAME" \
+       --disk-size=20gb \
+       --machine-type=g1-small \
        --node-taints=cook-pool=k8s-alpha:NoSchedule \
-       --enable-autoscaling --min-nodes=2 --max-nodes=4
+       --enable-autoscaling \
+       --min-nodes=2 \
+       --max-nodes=4 \
+       --node-labels=cook_pool=k8s-alpha
 
 echo "---- Setting up cook namespace in kubernetes"
 KUBECONFIG=${COOK_KUBECONFIG} kubectl create -f docs/make-kubernetes-namespace.json
--- a/scheduler/bin/help-make-cluster
+++ b/scheduler/bin/help-make-cluster
@@ -10,10 +10,14 @@
 
 set -e
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
 PROJECT=$1
 ZONE=$2
 CLUSTERNAME=$3
 COOK_KUBECONFIG=$4
+
+VERSION=1.15.9-gke.9
 
 gcloud="gcloud --project $PROJECT"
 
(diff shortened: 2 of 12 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -68,7 +68,8 @@
     (-> k8s-actual-state-dict
         (update-in [:synthesized-state :state] #(or % :missing))
         (dissoc :pod)
-        (assoc :pod-status (some-> pod .getStatus)))
+        (assoc :pod-status (some-> pod .getStatus))
+        (assoc :node-name (some-> pod .getSpec .getNodeName)))
```
dposada: This should be using `pod->node-name`.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -461,11 +456,15 @@
 
                                       :cook-expected-state/starting
                                       (case pod-synthesized-state-modified
-                                        :missing (launch-pod compute-cluster
-                                                             cook-expected-state-dict pod-name)
+                                        :missing (if (:pod-deleted? synthesized-state)
+                                                   (do
+                                                     (log/info "In compute cluster" name ", pod" pod-name
+                                                               "was deleted while it was expected starting")
+                                                     (handle-pod-killed compute-cluster pod-name))
```
dposada: Need to update the state diagram to capture this.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -851,28 +852,41 @@
...
+                    (swap! pool-name->pending-jobs-atom
+                           remove-matched-jobs-from-pending-jobs
+                           matched-job-uuids pool-name)
+                    (log/debug "In" pool-name
+                               "pool, updated pool-name->pending-jobs-atom:"
+                               @pool-name->pending-jobs-atom)
+                    (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
+                    (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
+                    matched-considerable-jobs-head?))]
+            (trigger-autoscaling! failures pool-name compute-clusters)
```
dposada: Would be good to add a comment about why this needs to be here.

scheduler/src/cook/kubernetes/controller.clj
dposada: Scratch that, the state diagram was already capturing this due to failed pod submissions:

```
        // (Starting, missing) -> Completed happens for some failed pod submissions
        Starting -> Completed [label=":succeeded\n:failed\n:unknown\n:missing"]
```

scheduler/src/cook/scheduler/scheduler.clj
dposada: Done

scheduler/src/cook/kubernetes/controller.clj

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -68,7 +68,8 @@
     (-> k8s-actual-state-dict
         (update-in [:synthesized-state :state] #(or % :missing))
         (dissoc :pod)
-        (assoc :pod-status (some-> pod .getStatus)))
+        (assoc :pod-status (some-> pod .getStatus))
+        (assoc :node-name (api/pod->node-name pod)))
```
dposada: ### Why?

Adding the node name here makes it possible to include these logs when searching the logs for a particular node name, which is a common practice.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -96,18 +97,17 @@
...
 
-(defn synthetic-pod->job-uuid
-  "If the given pod is a synthetic pod for autoscaling, returns the job uuid
-  that the pod corresponds to (stored in a pod label). Otherwise, returns nil."
-  [^V1Pod pod]
-  (some-> pod .getMetadata .getLabels (.get api/cook-synthetic-pod-job-uuid-label)))
+(defn synthetic-pod?
+  "Given a pod name, returns true if it has the synthetic pod prefix"
+  [pod-name]
+  (str/starts-with? pod-name api/cook-synthetic-pod-name-prefix))
```
dposada: ### Why?

This allows us to identify synthetic pods even when we don't a `V1Pod` object to inspect.

See [this comment](https://github.com/twosigma/Cook/pull/1296#discussion_r379062248) for more context.

scheduler/src/cook/kubernetes/controller.clj
dposada: ### Why?

There's a legitimate scenario where we launch a pod and then it gets deleted before it started running. In this case, we need to go down the `handle-pod-killed` path.

scheduler/bin/help-make-cluster
```diff
@@ -34,14 +38,35 @@ KUBECONFIG=${COOK_KUBECONFIG} $gcloud container clusters get-credentials "$CLUST
 # Add credentials to default kubeconfig for convenience purposes.
 $gcloud container clusters get-credentials "$CLUSTERNAME" --zone "$ZONE"
 
+echo "---- Upgrading the cluster master version to $VERSION"
+$gcloud container clusters upgrade "$CLUSTERNAME" --cluster-version $VERSION --zone "$ZONE" --quiet --master
```
dposada: ### Why?

k8s 1.15.x has fixes to pod preemption support which we need.

scheduler/bin/help-make-cluster
```diff
@@ -34,14 +38,35 @@ KUBECONFIG=${COOK_KUBECONFIG} $gcloud container clusters get-credentials "$CLUST
...
+
+echo "---- Setting up cook priority classes in kubernetes"
+KUBECONFIG=${COOK_KUBECONFIG} kubectl create -f "${DIR}/priority-class-cook-workload.yaml"
+KUBECONFIG=${COOK_KUBECONFIG} kubectl create -f "${DIR}/priority-class-synthetic-pod.yaml"
```
dposada: ### Why?

These priority classes allow real Cook job pods to preempt Cook synthetic pods. This is important because otherwise, when new nodes become available, pending synthetic pods can race against matched Cook jobs and cause the matched Cook jobs to get failures.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -34,44 +34,6 @@
...
-                                                   (let [pod-spec (.getSpec pod)
-                                                         tolerations (.getTolerations pod-spec)]
-                                                     (some (fn [^V1Toleration toleration]
-                                                             (and (= (.getKey toleration) "cook-pool")
-                                                                  (= (.getValue toleration) pool-name)))
-                                                           tolerations)))
-                                                 waiting-synthetic-pods)]
-      (count waiting-synthetic-pods-in-pool))
-    0))
-
```
dposada: ### Why?

We no longer need skip offer generation due to pending synthetic pods. Instead, we're using pod preemption and allowing job pods to preempt synthetic pods in order to avoid races for scheduling on new capacity between them. This function was only being used to determine whether or not skip offer generation.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -240,6 +202,12 @@
   (let [starting-pods (controller/starting-namespaced-pod-name->pod compute-cluster)]
     (-> pods (merge starting-pods) vals)))
 
+(defn synthetic-pod->job-uuid
+  "If the given pod is a synthetic pod for autoscaling, returns the job uuid
+  that the pod corresponds to (stored in a pod label). Otherwise, returns nil."
+  [^V1Pod pod]
+  (some-> pod .getMetadata .getLabels (.get api/cook-synthetic-pod-job-uuid-label)))
```
dposada: This was just moved here from a different file unchanged.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -349,23 +305,41 @@
...
-                                            ; We need to label the synthetic pods so that we
-                                            ; can opt them out of some of the normal plumbing,
-                                            ; like mapping status back to a job instance
-                                            :pod-labels {api/cook-synthetic-pod-job-uuid-label (str uuid)}}))
-                                    (take (- max-pods-outstanding num-synthetic-pods)))]
+                sidecar-resource-requirements (-> (config/kubernetes) :sidecar :resource-requirements)
+                task-metadata-seq
+                (->> new-task-requests
+                     (map (fn [{{:keys [job/uuid] :as job} :job}]
+                            {:task-id (str api/cook-synthetic-pod-name-prefix "-" pool-name "-" uuid)
```
dposada: ### Why?

Using the job UUID in the synthetic pod name makes tracing these much nicer than having to inspect the pod label.

scheduler/src/cook/kubernetes/controller.clj
shamsimam: Nitpick: We can merge the two `assoc`s

integration/tests/cook/util.py
```diff
@@ -1450,7 +1450,10 @@ def node_pool(nodename):
 
 def max_kubernetes_node_cpus():
     nodes = get_kubernetes_nodes()
-    return max([float(n['status']['capacity']['cpu'])
+    # We need to account for per-node overhead
+    # (capacity not usable by our Cook pods)
```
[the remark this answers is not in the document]
dposada: In our internal environments, it's constant.

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Yes, we're checking for that a few lines above:

```
                new-task-requests (remove (fn [{{:keys [job/uuid]} :job}]
                                            (some #(= (str uuid) (synthetic-pod->job-uuid %))
                                                  outstanding-synthetic-pods))
                                          task-requests)
```

If for some reason this check fails, it's actually a feature that this will prevent us from over-submitting synthetic pods.

scheduler/src/cook/kubernetes/controller.clj
dposada: Good call, thanks

scheduler/bin/help-make-cluster
dposada: It's possible, but in our internal environments, we actually want this to be handled at the cluster creation level by the "infrastructure" side.

scheduler/src/cook/kubernetes/controller.clj
dposada: Done!

PR #1453, opened 2020-02-29
Makes "Container launch failed" mea culpa

dposada:

- making "Container launch failed" mea culpa

In the wild, these errors are almost always caused by platform issues rather than issues with the user's job.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -384,7 +384,6 @@ class CookTest(util.CookTest):
             self.assertIn({'mode': 'RW',
                            'host-path': '/var/lib/mno',
                            'container-path': '/var/lib/pqr'}, volumes)
-            util.wait_for_job(self.cook_url, job_uuid, 'completed')
         finally:
             util.kill_jobs(self.cook_url, [job_uuid], assert_response=False)
 
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -362,7 +362,7 @@
       (dissoc (s/optional-key :group))
       (dissoc (s/optional-key :status))
       (merge {:framework-id (s/maybe s/Str)
-              :retries-remaining NonNegInt
+              :retries-remaining s/Int
               :status s/Str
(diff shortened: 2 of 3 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -362,7 +362,7 @@
       (dissoc (s/optional-key :group))
       (dissoc (s/optional-key :status))
       (merge {:framework-id (s/maybe s/Str)
-              :retries-remaining NonNegInt
+              :retries-remaining s/Int
```
dposada: There's a bug in the retries remaining logic that causes this number to sometimes be negative. It's not clear yet where/what the bug is, but when it happens, this schema is causing us to not be able to query the job through the API, which is frustrating for users and makes it very difficult to track down the bug. So, while we're not yet fixing the bug, we're making it easier to troubleshoot by making the response schema less strict.

shamsimam: Let's create a separate PR for this. Also, we should add a TODO and surrounding logs to detect when the retries remaining becomes negative.

integration/tests/cook/test_basic.py
```diff
@@ -384,7 +384,6 @@ def test_docker_fields(self):
             self.assertIn({'mode': 'RW',
                            'host-path': '/var/lib/mno',
                            'container-path': '/var/lib/pqr'}, volumes)
-            util.wait_for_job(self.cook_url, job_uuid, 'completed')
```
shamsimam: Maybe we should have a wait for job that doesn't assert on the status, just the the job stopped running. Removing the wait may otherwise have implications on the number of jobs running concurrently from the tests.

dposada: This `wait_for_job` was recently added, but then we ended up deciding to create a completely separate test for making sure that the job completes with some of these parameters. This test is just to check that the fields got set correctly.

scheduler/src/cook/rest/api.clj
dposada: Created #1455 for this.

Issue #200, opened 2016-10-25
Track current fair share as a metric


wyegelwel:
Users would like to know "how many resources should I expect to be able to get _right now_". This is the case for users already running and a new user with nothing yet running.

This can approximately be computed by taking the output of rank-jobs (including running and pending) and run it through a second instance of fenzo supplied with enough hosts to mirror the resources in the cluster.

There are two problems with that:
1. It doesn't tell how many resources a new user can expect to get
2. If a running user has requested less than the amount of resources a new user can expect to get, it will just return the resources they are currently consuming instead of what they can expect to get.

wyegelwel:
duplicate of #157

PR #1456, opened 2020-03-03
Add SSL verification between k8s at cook

scrosby:

- Add SSL verification between the cook and k8s.

To prevent man-in-the-middle with someone pretending to be the backend.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -430,7 +430,7 @@
     (when (some? verifying-ssl)
       (.setVerifyingSsl api-client verifying-ssl))
     ; Loading ssl-cert-path must be last SSL operation we do in setting up API Client. API bug.
-    ; See explanation in comments in https://github.com/kubernetes-client/java/pull/200 and issue #199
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -429,6 +429,11 @@
       (.setBasePath api-client base-path))
     (when (some? verifying-ssl)
       (.setVerifyingSsl api-client verifying-ssl))
+    ; Loading ssl-cert-path must be last SSL operation we do in setting up API Client. API bug.
+    ; See explanation in comments in https://github.com/kubernetes-client/java/pull/200 and issue #199
```
dposada: Don't think you meant for this to link to

https://github.com/twosigma/Cook/issues/199

scrosby: Its related to both. Technically, we're imitating the same fix as #200, for the bug in #199.

dposada: Are you sure it's related to:

`https://github.com/twosigma/Cook/pull/199`?

That's what this links to:

https://github.com/twosigma/Cook/pull/199

PR #1464, opened 2020-03-05
Autoscales based on pending jobs instead of match failures

dposada:

- instead of using match failures, using `pending-jobs` post-match to trigger autoscaling

Let's say we have 1000 jobs in the queue, we send the top 200 to Fenzo, and 100 of those get matched. Instead of autoscaling for only the remaining 100, we want to autoscale for the 900 still-pending jobs. Of course, the KCC `autoscale!` implementation still has a configurable cap on the number of outstanding synthetic pods per compute cluster.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -783,7 +783,7 @@
   compute-cluster->task-request map. That is the API any future
   improvements need to stick to."
   [task-requests compute-clusters]
-  (group-by (fn [{:keys [job]}]
+  (group-by (fn [job]
               (nth compute-clusters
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -794,7 +794,7 @@
   - There is at least one compute cluster configured to do autoscaling"
   [failures pool-name compute-clusters]
   (try
-    (let [task-requests (map #(.. (first %) (getRequest)) failures)
+    (let [task-requests failures
```
nsinkov: these are no longer task-requests, they are jobs

dposada: Renamed, thanks

PR #1467, opened 2020-03-09
Rename metrics to be consistent with prior metric naming

scrosby:

- Rename metrics
- Avoid using compute-cluster-* in a metric name. 

Make our new metrics more consistent, in naming, with existing metrics. Also, avoid using compute-cluster-* in metric names so that we can pattern-match this substring in logstash.


dposada:
Will merge when green

PR #1468, opened 2020-03-09
add tty support

nsinkov:

- enable tty in kubernetes containers

allocate a TTY to support tools that need to read from stdin


dposada:
Will merge when green

PR #1472, opened 2020-03-11
Separate k8s job pods' workdir and sandbox

DaoWen:

Separate the concepts of the working directory and sandbox directory for jobs in k8s pods.

1. The working directory (which may be specified by the user) is not guaranteed to be writable.
2. This separation better models the current behavior of Cook on Mesos.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -29,7 +29,7 @@
 (def cook-pod-label "twosigma.com/cook-scheduler-job")
 (def cook-synthetic-pod-job-uuid-label "twosigma.com/cook-scheduler-synthetic-pod-job-uuid")
 (def cook-pool-label "cook-pool")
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -382,57 +382,33 @@
...
-                                                            volume-mount (V1VolumeMount.)
-                                                            read-only (not (= mode "RW"))]
-                                                        (.setName volume-mount (host-path->name host-path))
-                                                        (.setMountPath volume-mount container-path)
-                                                        (.setReadOnly volume-mount read-only)
-                                                        [container-path volume-mount])))
-                                            filtered-cook-volumes)
-        volume-mounts (vals container-path->volume-mounts)
-        ; Workdir volume can overlap the main docker volumes.
-        ; So we either need to map the the apropriate volume or make a new one.
```
DaoWen: Since we no longer create a workdir volume, this isn't a problem anymore.

PR #1473, opened 2020-03-11
cs ssh to $HOME in k8s job pods

DaoWen:

Make `cs ssh` utility `cd` to the sandbox directory when it connects to a Kubernetes pod. The current behavior puts the user in the default directory used by `kubectl exec`, which is the `workDir` set on the pod's container spec.

For our Kubernetes implementation, we want to mimic the behavior of `cs ssh` when targeting Mesos (which has always gone directly to the sandbox directory).


DaoWen:
This has been manually tested against #1472. I should probably add an integration test for this.

PR #1477, opened 2020-03-13
Adds synthetic pod anti-affinity to previous hosts

dposada:

- adding node "anti"-affinity to synthetic pods, disallowing them from being scheduled on nodes where their corresponding job has already run

Cook has a "novel host constraint", which disallows a job from running on the same host twice. So, we need to avoid running a synthetic pod on any of the hosts that the real job won't be able to run on. Otherwise, the synthetic pod won't trigger the cluster autoscaler.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -655,6 +655,8 @@
           (.setVolumeMounts container [(workdir-volume-mount-fn true) (sidecar-workdir-volume-mount-fn false)])
           (.addContainersItem pod-spec container))))
 
+    ; We're either using the hostname (in the case of users' job pods)
+    ; or pod-hostnames-to-avoid (in the case of synthetic pods), but not both.
     (if hostname
       ; Why not just set the node name?
       ; The reason is that setting the node name prevents pod preemption
@@ -665,6 +667,7 @@
       (when (seq pod-hostnames-to-avoid)
         ; Use node "anti"-affinity to disallow scheduling
         ; on hostnames we're being told to avoid
+        ; (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity)
         (let [affinity (V1Affinity.)
               node-affinity (V1NodeAffinity.)
               node-selector (V1NodeSelector.)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -525,7 +525,7 @@
 (defn ^V1Pod task-metadata->pod
   "Given a task-request and other data generate the kubernetes V1Pod to launch that task."
   [namespace compute-cluster-name
-   {:keys [task-id command container task-request hostname pod-labels
+   {:keys [task-id command container task-request hostname pod-labels pod-hostnames-to-avoid
            pod-priority-class pod-supports-cook-init? pod-supports-cook-sidecar?]
     :or {pod-priority-class cook-job-pod-priority-class
          pod-supports-cook-init? true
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -356,7 +353,17 @@
...
+                             :task-id (str api/cook-synthetic-pod-name-prefix "-" pool-name "-" uuid)
+                             :task-request {:scalar-requests
+                                            (cond->> (walk/stringify-keys (tools/job-ent->resources job))
+                                                     ; We need to account for any sidecar resource requirements that the
+                                                     ; job we're launching this synthetic pod for will need to use.
+                                                     sidecar-resource-requirements
+                                                     (merge-with +
+                                                                 {"cpus" (:cpu-request sidecar-resource-requirements)
+                                                                  "mem" (:memory-request sidecar-resource-requirements)}))
+                                            :job {:job/pool {:pool/name synthetic-task-pool-name}}}}))
```
dposada: These were moved down here just to make the map keys be lexicographically sorted.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -525,7 +525,7 @@
 (defn ^V1Pod task-metadata->pod
   "Given a task-request and other data generate the kubernetes V1Pod to launch that task."
   [namespace compute-cluster-name
-   {:keys [task-id command container task-request hostname pod-labels
```
shamsimam: Where do we handle the scenario where `hostname` is not present in `pod-hostnames-to-avoid`?

scheduler/src/cook/kubernetes/compute_cluster.clj
shamsimam: Thank you!

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -89,13 +89,19 @@
     [this _ vm-attributes _]
     (job-constraint-evaluate this _ vm-attributes)))
 
+(defn job->previous-hosts-to-avoid
+  "Given a job, returns the set of hostnames the
+  job's instances have run on (except preemptions)."
+  [job]
+  (->> (:job/instance job)
+       (remove #(true? (:instance/preempted? %)))
+       (mapv :instance/hostname)))
```
shamsimam: We need to make this a set before returning.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -655,13 +655,29 @@
...
       ; The reason is that setting the node name prevents pod preemption
       ; (https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/)
       ; from happening. We want this pod to preempt lower priority pods
       ; (e.g. synthetic pods).
-      (add-node-selector pod-spec k8s-hostname-label hostname))
+      (add-node-selector pod-spec k8s-hostname-label hostname)
+      (when (seq pod-hostnames-to-avoid)
+        ; Use node "anti"-affinity to disallow scheduling
+        ; on hostnames we're being told to avoid
+        (let [affinity (V1Affinity.)
```
shamsimam: Nitpick: Is there a link to some documentation explaining what we're setting up below?
It'd be nice to include in a comment here.

scheduler/src/cook/kubernetes/api.clj
dposada: We're either using `hostname` (in the case of users' job pods) or `pod-hostnames-to-avoid` (in the case of synthetic pods), but not both.

dposada: Added a comment with the same info.

scheduler/src/cook/scheduler/constraints.clj
dposada: Added call to `distinct`

scheduler/src/cook/kubernetes/api.clj
dposada: Good call, added a link to [this](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity).

scheduler/test/cook/test/scheduler/constraints.clj
```diff
@@ -347,3 +347,14 @@
           (is (= [true nil] (constraints/job-constraint-evaluate constraint nil {"HOSTNAME" "hostC"})))
           ;; unsuitable
           (is (= [false "Host is not suitable for datasets"] (constraints/job-constraint-evaluate constraint nil {"HOSTNAME" "hostB"}))))))))
+
+(deftest test-job->previous-hosts-to-avoid
```
shamsimam: Nitpick: The assertion also needs to validate on length to confirm duplicates are not returned.

dposada: :man_facepalming: Good call, adding this in PR #1479.

PR #1484, opened 2020-03-23
support cs cli commands on kubernetes

nsinkov:

- change cs cli commands to leverage kubectl

the cs cli commands rely on Mesos APIs. kubectl can be used to implement the same functionality on Kubernetes

### DIFF
--- a/cli/cook/subcommands/cat.py
+++ b/cli/cook/subcommands/cat.py
@@ -4,9 +4,9 @@ import os
 import sys
 from functools import partial
 
-from cook import http, plugins
+from cook import plugins
 from cook.mesos import download_file
-from cook.querying import parse_entity_refs, query_unique_and_run, parse_entity_ref
+from cook.querying import get_compute_cluster_config, parse_entity_refs, query_unique_and_run, parse_entity_ref
 from cook.util import guard_no_cluster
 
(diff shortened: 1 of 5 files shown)


cli/cook/subcommands/cat.py
```diff
@@ -21,6 +20,30 @@ def cat_for_instance(instance, sandbox_dir, _, path):
...
+    When using Kubernetes, calls the exec command of the kubectl cli.
+    """
+    compute_cluster = instance["compute-cluster"]
+    compute_cluster_type = compute_cluster["type"]
+    compute_cluster_name = compute_cluster["name"]
+    if compute_cluster_type == "kubernetes":
+        kubectl_cat_instance_file_fn = plugins.get_fn('kubectl-cat-for-instance', kubectl_cat_instance_file)
+        cook_cluster_settings = http.get(cluster, 'settings', params={}).json()
+        compute_cluster_config = next(c for c in (s['config'] for s in cook_cluster_settings['compute-clusters']) if
+                                      c['compute-cluster-name'] == compute_cluster_name)
```
dposada: Can we extract this into a function since we do it in a few places?

nsinkov: i'll update

PR #1486, opened 2020-03-25
Fix help text

nsinkov:

- Remove Mesos-only help text

can schedule on kubernetes also now

### DIFF
--- a/cli/cook/subcommands/ssh.py
+++ b/cli/cook/subcommands/ssh.py
@@ -49,6 +49,6 @@ def ssh(clusters, args, _):
 
 def register(add_parser, _):
     """Adds this sub-command's parser and returns the action function"""
-    parser = add_parser('ssh', help='ssh to Mesos agent by job or instance uuid')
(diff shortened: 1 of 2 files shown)


cli/cook/subcommands/ssh.py
```diff
@@ -49,6 +49,6 @@ def ssh(clusters, args, _):
 
 def register(add_parser, _):
     """Adds this sub-command's parser and returns the action function"""
-    parser = add_parser('ssh', help='ssh to Mesos agent by job or instance uuid')
+    parser = add_parser('ssh', help='ssh to Mesos agent or container by job or instance uuid')
```
dposada: Why are we leaving the word Mesos here but removing it from `ls`?

nsinkov: Technically a Mesos agent is not the same as a container. But I'm ok just having container only

dposada: Yeah, that sounds good

PR #1489, opened 2020-03-28
Marks pods with un-initialized containers as failed

dposada:

- adding a new failure reason "Container initialization timed out"
- fast-failing jobs using this new reason when they get into the `ContainersNotInitialized` condition for a prolonged (configurable) amount of time
- making `scan-process` re-synthesize the state so that time-dependent conditions (like the one being introduced here) will get calculated even when no new message has come through on the pod watch

Without this, pods can remain in the `ContainersNotInitialized` state forever, and their jobs stay in the `Running` state forever, despite not actually having started running the user's workload.


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -601,9 +613,10 @@
...
-    (timers/time! (metrics/timer "process-lock" name)
-      (locking (calculate-lock pod-name)
-        (process compute-cluster pod-name)))))
+  "Given a pod-name, looks up the pod and delegates
+  to synthesize-state-and-process-pod-if-changed."
+  [{:keys [k8s-actual-state-map name] :as compute-cluster} pod-name]
+  (timers/time!
+    (metrics/timer "scan-process" name)
+    (let [{:keys [pod]} (get @k8s-actual-state-map pod-name)]
+      (synthesize-state-and-process-pod-if-changed compute-cluster pod))))
```
dposada, line 622: ## Why?

We need `scan-process` to re-synthesize the state so that time-dependent conditions (like the one being introduced in this PR) will get calculated even when no new message has come through on the pod watch.

scheduler/docs/reason-code
```diff
@@ -12,6 +12,8 @@
 02001: REASON_CONTAINER_LIMITATION_DISK
 02002: REASON_CONTAINER_LIMITATION_MEMORY
 02003: Max runtime exceeded
+02004: Task was a straggler
+02005: Container initialization timed out
```
sradack, line 16: "Container initialization timed out" feels closer to 01xxx (e.g. "Scheduling failed on host") than it does 02xxx.  What do you think?

scheduler/src/cook/config.clj
```diff
@@ -445,6 +445,7 @@
                                  pool-selection)})))
      :kubernetes (fnk [[:config {kubernetes {}}]]
                    (merge {:default-workdir "/mnt/sandbox"
+                           :pod-condition-containers-not-initialized-seconds 60
```
sradack, line 448: How long do containers normally take to initialize?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -601,9 +613,10 @@
...
-  (timers/time! (metrics/timer "scan-process" name)
-    (timers/time! (metrics/timer "process-lock" name)
-      (locking (calculate-lock pod-name)
-        (process compute-cluster pod-name)))))
+  "Given a pod-name, looks up the pod and delegates
+  to synthesize-state-and-process-pod-if-changed."
+  [{:keys [k8s-actual-state-map name] :as compute-cluster} pod-name]
+  (timers/time!
+    (metrics/timer "scan-process" name)
+    (let [{:keys [pod]} (get @k8s-actual-state-map pod-name)]
```
sradack, line 621: Reading `k8s-actual-state-map` used to occur inside the lock.  Is there an issue reading from `k8s-actual-state-map`  before acquiring the lock?

scheduler/docs/reason-code
dposada: Good call, will do

scheduler/src/cook/config.clj
dposada: I posted an example timeline internally.

This number of seconds starts from when we've already seen the `ContainersNotInitialized` reason. 

Here's the relevant k8s code that adds this condition:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/generate.go#L172-L187

We can make the default higher if we think this is overly sensitive, what do you think? 

Alternatively, we can see what happens in the wild and adjust this via config.edn, since this is just the default.

scheduler/docs/reason-code
dposada: Done

scheduler/src/cook/kubernetes/controller.clj
dposada: Good call. I moved the lock back out to callers of `synthesize-state-and-process-pod-if-changed` and put this `get` inside the lock.

scheduler/src/cook/config.clj
dposada: Changed the default from 60 to 120

scheduler/src/cook/kubernetes/controller.clj
dposada: Added this info to the PR description

PR #1492, opened 2020-03-30
Support for specifying the default container on a per-pool basis

scrosby:

- Support for specifying the default container on a per-pool basis.

User convenience. We can specify a default docker image that works for the majority of cases.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1855,8 +1855,19 @@ class CookTest(util.CookTest):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertIn('success', [i['status'] for i in job['instances']])
 
-    @unittest.skipUnless(util.docker_tests_enabled(), "Requires a test docker image")
+    @unittest.skipUnless(util.docker_tests_enabled(), "Requires we're in an environment that requires docker images.")
+    @unittest.skipUnless(util.default_submit_pool() is not None, "Test requires a default test pool")
     def test_default_container_for_pool(self):
+        default_pool = util.default_submit_pool()
+        settings_dict = util.settings(self.cook_url)
+        list_of_pool_match_rules = settings_dict.get("pools", {}).get("default-container-for-pool", [])
+        if len(list_of_pool_match_rules) == 0:
+            self.skipTest("No pool match rules defined")
+        matching_containers = [ii["container"] for ii in list_of_pool_match_rules if
+                               re.match(ii["pool-regex"], default_pool)]
+        if len(matching_containers) == 0:
+            self.skipTest("No rule matches the default submit pool")
+        expected_container = matching_containers[0]
         # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
         job_uuid, resp = util.submit_job(
             self.cook_url,
@@ -1865,6 +1876,7 @@ class CookTest(util.CookTest):
             max_retries=5)
         self.assertEqual(resp.status_code, 201)
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        self.assertEquals(job["container"]["docker"]["image"], expected_container["docker"]["image"])
         self.assertIn('success', [i['status'] for i in job['instances']])
 
     @unittest.skipUnless(util.has_docker_service() and not util.using_kubernetes(),
@@ -3120,8 +3132,6 @@ class CookTest(util.CookTest):
         self.assertEqual('failed', job['instances'][0]['status'], job)
         self.assertEqual('Invalid task', job['instances'][0]['reason_string'], job)
 
-    @unittest.skipUnless(util.docker_image_for_default_pool() and not util.docker_image(),
-                         'We need a docker image to run this test, but dont have a pool from which to learn it.')
     def test_submit_pool_unspecified(self):
         job_uuid, resp = util.submit_job(self.cook_url, pool=util.POOL_UNSPECIFIED)
         self.assertEqual(resp.status_code, 201, resp.content)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1870,7 +1870,7 @@ class CookTest(util.CookTest):
         expected_container = matching_containers[0]
         # Make sure we have a different configured docker image and default image, we fail the test to make sure that we
         # don't inadvertently skip it.
-        self.failIfEqual(expected_container["docker"]["image"],util.docker_image())
+        self.failIfEqual(expected_container["docker"]["image"], util.docker_image())
 
         # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
         job_uuid, resp = util.submit_job(
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1855,6 +1855,18 @@ class CookTest(util.CookTest):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertIn('success', [i['status'] for i in job['instances']])
 
+    @unittest.skipUnless(util.docker_tests_enabled(), "Requires a test docker image")
+    def test_default_container_for_pool(self):
+        # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
+        job_uuid, resp = util.submit_job(
+            self.cook_url,
+            command='cat /.dockerenv',
+            container=None,
+            max_retries=5)
+        self.assertEqual(resp.status_code, 201)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        self.assertIn('success', [i['status'] for i in job['instances']])
+
     @unittest.skipUnless(util.has_docker_service() and not util.using_kubernetes(),
                          "Requires `docker inspect`. On kubernetes, need to add support and write a separate test.")
     def test_docker_port_mapping(self):
@@ -3108,6 +3120,8 @@ class CookTest(util.CookTest):
         self.assertEqual('failed', job['instances'][0]['status'], job)
         self.assertEqual('Invalid task', job['instances'][0]['reason_string'], job)
 
+    @unittest.skipUnless(util.docker_image_for_default_pool() and not util.docker_image(),
+                         'We need a docker image to run this test, but dont have a pool from which to learn it.')
     def test_submit_pool_unspecified(self):
         job_uuid, resp = util.submit_job(self.cook_url, pool=util.POOL_UNSPECIFIED)
         self.assertEqual(resp.status_code, 201, resp.content)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1860,7 +1860,7 @@ class CookTest(util.CookTest):
     def test_default_container_for_pool(self):
         default_pool = util.default_submit_pool()
         settings_dict = util.settings(self.cook_url)
-        list_of_pool_match_rules = settings_dict.get("pools", {}).get("default-container-for-pool", [])
+        list_of_pool_match_rules = settings_dict.get("pools", {}).get("default-containers", [])
         if len(list_of_pool_match_rules) == 0:
             self.skipTest("No pool match rules defined")
         matching_containers = [ii["container"] for ii in list_of_pool_match_rules if
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1868,6 +1868,10 @@ class CookTest(util.CookTest):
         if len(matching_containers) == 0:
             self.skipTest("No rule matches the default submit pool")
         expected_container = matching_containers[0]
+        # Make sure we have a different configured docker image and default image, we fail the test to make sure that we
+        # don't inadvertently skip it.
+        self.failIfEqual(expected_container["docker"]["image"],util.docker_image())
+
         # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
         job_uuid, resp = util.submit_job(
             self.cook_url,
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -1880,7 +1880,7 @@ class CookTest(util.CookTest):
             max_retries=5)
         self.assertEqual(resp.status_code, 201)
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
-        self.assertEquals(job["container"]["docker"]["image"], expected_container["docker"]["image"])
+        self.assertEqual(job["container"]["docker"]["image"], expected_container["docker"]["image"])
         self.assertIn('success', [i['status'] for i in job['instances']])
 
     @unittest.skipUnless(util.has_docker_service() and not util.using_kubernetes(),
--- a/integration/tests/cook/util.py
+++ b/integration/tests/cook/util.py
@@ -444,32 +444,9 @@ def scheduler_info(cook_url):
     assert resp.status_code == 200, response_info
     return resp.json()
 
-@functools.lru_cache()
-def docker_image_for_default_pool():
-    """What is the docker image for the default pool."""
-    cook_url = retrieve_cook_url()
-    settings_dict = settings(cook_url)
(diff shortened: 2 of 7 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -1859,6 +1859,18 @@ def test_basic_docker_job(self):
...
+    def test_default_container_for_pool(self):
+        # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
+        job_uuid, resp = util.submit_job(
+            self.cook_url,
+            command='cat /.dockerenv',
+            container=None,
+            max_retries=5)
+        self.assertEqual(resp.status_code, 201)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        self.assertIn('success', [i['status'] for i in job['instances']])
```
dposada, line 1872: Should we also assert that the container the job ended up getting was the configured one?

integration/tests/cook/test_basic.py
```diff
@@ -3075,6 +3087,8 @@ def test_pod_submission_failed(self):
         self.assertEqual('failed', job['instances'][0]['status'], job)
         self.assertEqual('Invalid task', job['instances'][0]['reason_string'], job)
 
+    @unittest.skipUnless(util.docker_image_for_default_pool() and not util.docker_image(),
+                         'We need a docker image to run this test, but dont have a pool from which to learn it.')
```
dposada, line 3091: > We need a docker image to run this test

Why?

scheduler/src/cook/rest/api.clj
```diff
@@ -699,7 +699,13 @@
...
                                   :constraint/pattern pattern}]))
                             constraints)
-        container (if (nil? container) [] (build-container user db-id container))
+        guessed-pool (or (:pool/name pool) (config/default-pool))
+        default-container-for-pool (get-in config/config [:settings :default-container-for-pool guessed-pool])
+        container (if (nil? container)
+                    (if (and guessed-pool default-container-for-pool)
+                      (build-container user db-id default-container-for-pool)
+                      [])
+                    (build-container user db-id container))
```
dposada, line 708:
```suggestion
        pool-name (or (:pool/name pool) (config/default-pool))
        default-container-for-pool (get-in config/config [:settings :default-container-for-pool pool-name])
        container (if (nil? container)
                    (if (and pool-name default-container-for-pool)
                      (build-container user db-id default-container-for-pool)
                      [])
                    (build-container user db-id container))
```

scheduler/src/cook/config.clj
```diff
@@ -412,6 +412,8 @@
 
      :api-only? (fnk [[:config {api-only? false}]]
                   api-only?)
+     :default-container-for-pool (fnk [[:config {default-container-for-pool {}}]]
+                                   default-container-for-pool)
```
dposada, line 416: A few thoughts:

1. Maybe "image" instead of "container"
1. Maybe this should live under the existing `:pools` section of the config
1. Maybe this should be a regex for the pool name so that we don't have to have an entry for every single pool and remember to add/remove entries when we add/remove pools

integration/tests/cook/util.py
```diff
@@ -430,10 +430,33 @@ def scheduler_info(cook_url):
...
+def docker_image_explicit():
+    """Was a docker image set explicitly via an environmental variable? Used to gate tests that depend on default image."""
     return os.getenv('COOK_TEST_DOCKER_IMAGE')
 
+@functools.lru_cache()
+def docker_image():
+    env_image = docker_image_explicit()
+    if env_image is not None:
+        return env_image
+    return docker_image_for_default_pool()
```
dposada, line 458: I don't get why we want this to start returning `docker_image_for_default_pool()`

integration/tests/cook/test_basic.py
scrosby: We have logic in the integration tests that tries to fill in a docker image onto everything. However, that logic assumes we have a pool specified (so that it can lookup the right docker image). When there's no pool, it fails to attach an image, and this test fails

scheduler/src/cook/config.clj
scrosby: 1. I chose container, because you give the entire container spec, that way we're robust to future image and image schema changes ---- you can specify parameters if you want to. (I borrowed the approach from how the integration tests name it and name the subfield. Its 'container', not 'image'.
2. I like that idea.
3. Regexp could be problematic. What if multiple match? I'm also thinking of test scenario's where we want to use a default everywhere except for a test or snowflake pools and use a different container for that one. (E.g., imaging a GPU pool with a special blessed GPU image that's different from the default everywhere else.)

integration/tests/cook/test_basic.py
dposada: Shouldn't this test run just fine in environments where we're not using Docker? It currently does.

scheduler/src/cook/config.clj
dposada: Ah, makes sense on "container". 

For regex, I think you'd want a list of (regex, container) pairs, and you take the first one that matches (order matters). That would allow for the flexibility you're describing without having to have an entry for every single pool and remember to add/remove entries when we add/remove pools.

integration/tests/cook/test_basic.py
scrosby: Done

integration/tests/cook/test_basic.py
scrosby: Restructured how we test due to our conversation. Moot.

scheduler/src/cook/rest/api.clj
scrosby: Rename done.

scheduler/src/cook/config.clj
scrosby: Ok.

integration/tests/cook/util.py
scrosby: Like other comment. Moot.

integration/tests/cook/test_basic.py
dposada: Since we're defaulting more than just the image, should we assert on everything we're defaulting being correct?

scheduler/config-composite.edn
```diff
@@ -61,7 +61,22 @@
...
+                                                        :image "python:3.6",
+                                                        :network "HOST"
+                                                        :force-pull-image false
+                                                        :parameters []}}}
+          {:pool-regex "k8s-gamma" :container {:type "docker"
+                                               :docker {
+                                                        :image "python:3.6",
+                                                        :network "HOST"
+                                                        :force-pull-image false
+                                                        :parameters []}}}]}
```
dposada, line 79: Shouldn't we combine these into one entry?

scheduler/config-k8s.edn
```diff
@@ -75,7 +75,20 @@
...
+                                                        :image "python:3.5.9-stretch",
+                                                        :network "HOST"
+                                                        :force-pull-image false
+                                                        :parameters []}}}
+          {:pool-regex "k8s-gamma" :container {:type "docker"
+                                               :docker {
+                                                        :image "python:3.5.9-stretch",
+                                                        :network "HOST"
+                                                        :force-pull-image false
+                                                        :parameters []}}}]}
```
dposada, line 91: Shouldn't we combine these into one entry?

scheduler/src/cook/rest/api.clj
```diff
@@ -730,7 +738,14 @@
                                                                 (str/lower-case operator))
                                   :constraint/pattern pattern}]))
                             constraints)
-        container (if (nil? container) [] (build-container user db-id container))
+        pool-name (or (:pool/name pool) (config/default-pool))
+        default-container-for-pool (get-in config/config [:settings :pools :default-container-for-pool])
```
dposada, line 742: Given the new location and structure, maybe `:default-container-for-pool` isn't the best name -- maybe just `:default-containers`, since there can be multiple and we're already under the `:pools` key.

scheduler/src/cook/rest/api.clj
```diff
@@ -730,7 +738,14 @@
...
+        container (if (nil? container)
+                    (let [guessed-container (match-default-container-for-pool default-container-for-pool pool-name)]
+                      (if (and pool-name guessed-container)
```
dposada, line 745: Why are we checking `pool-name` here?

scheduler/src/cook/rest/api.clj
```diff
@@ -730,7 +738,14 @@
                                                                 (str/lower-case operator))
                                   :constraint/pattern pattern}]))
                             constraints)
-        container (if (nil? container) [] (build-container user db-id container))
+        pool-name (or (:pool/name pool) (config/default-pool))
+        default-container-for-pool (get-in config/config [:settings :pools :default-container-for-pool])
+        container (if (nil? container)
+                    (let [guessed-container (match-default-container-for-pool default-container-for-pool pool-name)]
```
dposada, line 744: Maybe `default-container` instead of `guessed-container`?

integration/tests/cook/test_basic.py
scrosby: Which other fields should we check? Image is the only one we set by default in the integration test suite (via our ENV variable.)

scheduler/src/cook/rest/api.clj
scrosby: If we're not running with a pools, then there's no pool to match guessed containers. This makes the logic explicit, instead of implicit, that without pools, we are not using [:config :pools :default-containers] and keeps nil's out of that code.

scheduler/src/cook/rest/api.clj
scrosby: Changed.

scheduler/config-composite.edn
scrosby: Merged

scheduler/config-k8s.edn
scrosby: Merged

scheduler/src/cook/rest/api.clj
scrosby: Changed.

integration/tests/cook/test_basic.py
dposada: Oh I thought we were defaulting more than just that, e.g.:

```
                                           {:type "docker"
                                            :docker {
                                                     :image "python:3.6",
                                                     :network "HOST"
                                                     :force-pull-image false
                                                     :parameters []}}
```

scheduler/config-composite.edn
```diff
@@ -64,19 +64,13 @@
...
-          {:pool-regex "k8s-gamma" :container {:type "docker"
-                                               :docker {
-                                                        :image "python:3.6",
-                                                        :network "HOST"
-                                                        :force-pull-image false
-                                                        :parameters []}}}]}
+         :default-containers
+         [{:pool-regex "k8s-.*" :container {:type "docker"
+                                            :docker {
+                                                     :image "python:3.6",
```
dposada, line 70: Move to previous line

scheduler/config-k8s.edn
```diff
@@ -76,19 +76,13 @@
...
-          {:pool-regex "k8s-gamma" :container {:type "docker"
-                                               :docker {
-                                                        :image "python:3.5.9-stretch",
-                                                        :network "HOST"
-                                                        :force-pull-image false
-                                                        :parameters []}}}]}
+         :default-containers
+         [{:pool-regex "k8s-.*" :container {:type "docker"
+                                            :docker {
+                                                     :image "python:3.5.9-stretch",
```
dposada, line 82: Move to previous line

scheduler/src/cook/rest/api.clj
dposada: Ah, got it, makes sense. In this case (`pool-name` is `nil`), should we just avoid making the call to `match-default-container-for-pool`?

integration/tests/cook/test_basic.py
scrosby: I pulled that out of what the integration tests normally submit, except that integration tests submit a different image name (python:3.5.9-stretch) that works with minimesos.

scheduler/src/cook/rest/api.clj
scrosby: Ok.

scheduler/config-composite.edn
scrosby: Ok

scheduler/config-k8s.edn
scrosby: Ok

integration/tests/cook/test_basic.py
dposada: Sounds good

integration/tests/cook/test_basic.py
```diff
@@ -1868,6 +1868,10 @@ def test_default_container_for_pool(self):
         if len(matching_containers) == 0:
             self.skipTest("No rule matches the default submit pool")
         expected_container = matching_containers[0]
+        # Make sure we have a different configured docker image and default image, we fail the test to make sure that we
+        # don't inadvertently skip it.
+        self.failIfEqual(expected_container["docker"]["image"],util.docker_image())
```
dposada, line 1873: Can we run this through the Python code formatter?

integration/tests/cook/test_basic.py
```diff
@@ -1855,6 +1855,34 @@ def test_basic_docker_job(self):
...
+
+        # Special logic in util.submit_jobs.full_spec maps container=None and removes it from the submitted job spec.
+        job_uuid, resp = util.submit_job(
+            self.cook_url,
+            command='cat /.dockerenv',
+            container=None,
+            max_retries=5)
+        self.assertEqual(resp.status_code, 201)
+        job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
+        self.assertEquals(job["container"]["docker"]["image"], expected_container["docker"]["image"])
```
dposada, line 1883: `assertEquals` is deprecated in Python 3:

https://docs.python.org/3/library/unittest.html#deprecated-aliases

scheduler/src/cook/rest/api.clj
```diff
@@ -679,7 +679,15 @@
...
-(s/defn make-job-txn
+(defn get-default-container-for-pool
+  "Given a pool name, determine a default container that should be run on it."
+  [default-containers effective-pool-name]
+  (->> default-containers
+       (filter (fn [{:keys [pool-regex]}] (re-find (re-pattern pool-regex) effective-pool-name)))
+       first
+       :container))
+
+  (s/defn make-job-txn
```
dposada, line 690: Is the indentation off on this line?

scheduler/test/cook/test/rest/api.clj
```diff
@@ -2315,3 +2315,13 @@
...
   (is (not (api/valid-date-str? "20180132"))))
+
+(deftest test-match-default-containers
+  (let [default-containers
+        [{:pool-regex "^foo$" :container {:foo 1}}
+         {:pool-regex ".*" :container {:bar 2}}
+         {:pool-regex "^baz$" :container {:baz 3}}]]
+    (is (api/get-default-container-for-pool default-containers "foo") {:foo 1})
+    (is (api/get-default-container-for-pool default-containers "bar") {:bar 2})
+    (is (api/get-default-container-for-pool default-containers "baz") {:bar 2}))
```
dposada, line 2326: I think these are all missing a call to `=`

integration/tests/cook/test_basic.py
scrosby: Good catch.

scheduler/src/cook/rest/api.clj
scrosby: How did that happen?!?

scheduler/test/cook/test/rest/api.clj
scrosby: I agree. Good catch.

integration/tests/cook/test_basic.py
scrosby: Done

scheduler/test/cook/test/rest/api.clj
dposada: @scrosby I still don't see the calls to `=`

scrosby: I did it. Not sure where the commit got lost at. In any case. I've validated its done.

PR #1511, opened 2020-04-16
Refactor CLI to support plugin overrides from wrapper

DaoWen:

- Add an optional `plugins` argument to `main`. This will be the new (and only) way to configure site-wide plugins rather than using the config file.
- Refactor modules/plugins loading logic to read from the `plugins` map that was passed to `main`. The `.cs.json` config file is no longer consulted for `plugins` or `module` entries.

Making site-wide changes to plugged-in logic is difficult/impossible if various users have old plugin overrides copied in `.cs.json` files in home and working directories. The new `plugins` argument to `main` allows the `cs` package maintainer to manage all side-specific changes centrally and not worry about stale configuration files.

Users that still want custom behavior can create their own wrapper that invokes either the site-specific wrapper's `main` or bypasses it to use the vanilla `main`.

## Other notes

Since this is a breaking change, we plan to bump the `cs` major version number in the next release.


cli/cook/http.py
```diff
@@ -12,18 +12,24 @@
...
     """Configures HTTP timeouts and retries to be used"""
     global session
     global timeouts
     http_config = config.get('http')
     modules_config = http_config.get('modules')
-    session_module_name = modules_config.get('session-module')
-    adapters_module_name = modules_config.get('adapters-module')
-    session_module = importlib.import_module(session_module_name)
-    adapters_module = importlib.import_module(adapters_module_name)
-    logging.getLogger(session_module_name).setLevel(logging.DEBUG)
```
DaoWen, line 25: Dropped this line since the `requests` library doesn't actually have its own logger. The `requests` documentation says that you should configure the `urllib3` logger if you want debug output:
https://2.python-requests.org/en/master/api/#api-changes

DaoWen:
@dposada recommended that we just drop the current configuration-based plugins support altogether to simplify the code and minimize tech debt. Reworking that now.

PR #1515, opened 2020-04-20
Allows removing cpu limit in k8s

dposada:

- introducing config field `:set-container-cpu-limit?`, which defaults to `true` (for backwards compatibility)
- if set to `false`, we don't set the cpu limit on the job container

To allow pods to get the [Burstable QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-burstable).

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -624,8 +624,11 @@
 
     (.putRequestsItem resources "memory" (double->quantity (* memory-multiplier mem)))
     (.putLimitsItem resources "memory" (double->quantity (* memory-multiplier mem)))
-    ; We only set a request for cpu, and not a limit, in order to make this burstable
     (.putRequestsItem resources "cpu" (double->quantity cpus))
+    (when (:set-container-cpu-limit? (config/kubernetes))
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -625,9 +625,10 @@
     (.putRequestsItem resources "memory" (double->quantity (* memory-multiplier mem)))
     (.putLimitsItem resources "memory" (double->quantity (* memory-multiplier mem)))
     (.putRequestsItem resources "cpu" (double->quantity cpus))
-    ; We would like to make this burstable, but for various reasons currently need pods to run in the
-    ; "Guaranteed" QoS which requires limits for both memory and cpu.
-    (.putLimitsItem resources "cpu" (double->quantity cpus))
+    (when (:set-container-cpu-limit? (config/kubernetes))
```
DaoWen, line 628: Can we destructure this on line 574, in the same place as all the other values we pull from `(config/kubernetes)`?

dposada: Good call, done

DaoWen:
Unit tests passed, so this should still be safe to merge based on the previous green run.

PR #1518, opened 2020-04-22
Handle Killed by user reason code

nsinkov:

- Handle Killed by user reason code

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -425,7 +425,9 @@
                                                (if-let [compute-cluster (cc/compute-cluster-name->ComputeCluster compute-cluster-name)]
                                                  (do (log/info "Attempting to kill task" task-id "in" compute-cluster-name "due to job completion")
                                                      (meters/mark! tx-report-queue-tasks-killed)
-                                                     (cc/safe-kill-task compute-cluster task-id))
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -425,7 +425,9 @@
                                                (if-let [compute-cluster (cc/compute-cluster-name->ComputeCluster compute-cluster-name)]
                                                  (do (log/info "Attempting to kill task" task-id "in" compute-cluster-name "due to job completion")
                                                      (meters/mark! tx-report-queue-tasks-killed)
-                                                     (cc/safe-kill-task compute-cluster task-id))
+                                                     (cc/safe-kill-task compute-cluster task-id)
```
scrosby, line 428: Sorry I didn't catch this earlier, but shouldn't we do the db update first, then kill. Then we're consistent with how it works in cancelled tasks?

nsinkov: OK. but why do we do it in that order for the cancelled-task-killer
maybe we should flip  the order there

scrosby: We'd be churning the reason twice with the race. We'd see the executor die, put that reason in, then we'd be replacing it with killed-by-user.

PR #1522, opened 2020-04-22
Change output of errors from cs submit from stdout to stderr

scrosby:

- Change output of errors from cs submit from stdout to stderr

If we have an error, we want it printed to stderr to maximize the changes it is seen by users instead of gobbled by a stdout redirection.

### DIFF
--- a/integration/tests/cook/cli.py
+++ b/integration/tests/cook/cli.py
@@ -31,6 +31,11 @@ def stdout(cp):
     return decode(cp.stdout).strip()
 
 
+def stderr(cp):
+    """Returns the UTF-8 decoded and stripped stdout of the given CompletedProcess"""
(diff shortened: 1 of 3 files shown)


integration/tests/cook/cli.py
```diff
@@ -31,6 +31,11 @@ def stdout(cp):
     return decode(cp.stdout).strip()
 
 
+def stderr(cp):
+    """Returns the UTF-8 decoded and stripped stdout of the given CompletedProcess"""
```
DaoWen, line 35:
```suggestion
    """Returns the UTF-8 decoded and stripped stderr of the given CompletedProcess"""
```

DaoWen:
Merging based on previous green run (prior to comment typo fix).

PR #1523, opened 2020-04-23
Add memory-overhead to checkpointing config

nsinkov:

- add memory-overhead to checkpointing config
- increase memory required for main container by memory-overhead when checkpointing is enabled

The process of checkpointing might use extra memory. A job will get an out of memory error if we don't account for the extra memory used

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -504,7 +504,8 @@
 
 (defn adjust-job-resources
   "Adjust required resources for a job, adding any additional resources. Additional resources might be needed for
-   sidecars or extra processes that will be launched with the job."
+   sidecars or extra processes that will be launched with the job. This is called from make-task-request in the scheduler.
+   The adjusted resource value is used for bin-packing when scheduling."
   [{:keys [job/checkpoint] :as job} {:keys [cpus mem] :as resources}]
   (let [{:keys [cpu-request memory-request] :as resource-requirements}
         (-> (config/kubernetes) :sidecar :resource-requirements)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -505,18 +505,19 @@
 (defn adjust-job-resources
   "Adjust required resources for a job, adding any additional resources. Additional resources might be needed for
    sidecars or extra processes that will be launched with the job."
-  [job {:keys [cpus mem] :as resources}]
-  (let [adjusted-resources
-        (if-let [{:keys [cpu-request memory-request]} (-> (config/kubernetes) :sidecar :resource-requirements)]
-          (assoc resources
-            :cpus (add-as-decimals cpus cpu-request)
-            :mem (add-as-decimals mem memory-request))
-          resources)
-        adjusted-resources
-        (if-let [{:keys [memory-overhead]} (util/job-ent->checkpoint job)]
-          (update adjusted-resources :mem #(add-as-decimals % memory-overhead))
-          adjusted-resources)]
-    adjusted-resources))
+  [{:keys [job/checkpoint] :as job} {:keys [cpus mem] :as resources}]
+  (let [{:keys [cpu-request memory-request] :as resource-requirements}
+        (-> (config/kubernetes) :sidecar :resource-requirements)
+        checkpoint-memory-overhead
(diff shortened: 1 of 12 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -503,13 +503,21 @@
...
 (defn adjust-job-resources
-  "Given the required resources for a job, add to them the resources for any sidecars that will be launched with the job."
-  [{:keys [cpus mem] :as resources}]
-  (if-let [{:keys [cpu-request memory-request]} (-> (config/kubernetes) :sidecar :resource-requirements)]
-    (assoc resources
-      :cpus (add-as-decimals cpus cpu-request)
-      :mem (add-as-decimals mem memory-request))
-    resources))
+  "Adjust required resources for a job, adding any additional resources. Additional resources might be needed for
+   sidecars or extra processes that will be launched with the job."
```
dposada, line 507: I think we should add a line to the docstring about how/where this function gets used.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -574,6 +582,7 @@
         {:keys [custom-shell default-checkpoint-config init-container set-container-cpu-limit? sidecar]}
         (config/kubernetes)
         checkpoint (when checkpoint (merge default-checkpoint-config (util/job-ent->checkpoint job)))
+        checkpoint-memory-overhead (:memory-overhead checkpoint)
```
dposada, line 585: Do we need to account for this when sending synthetic pods too?

https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/kubernetes/compute_cluster.clj#L359-L367

scheduler/src/cook/rest/api.clj
```diff
@@ -182,7 +182,8 @@
   ; preemption - checkpoint is created on preemption before the VM is stopped
   {:mode (s/enum "auto" "periodic" "preemption")
    (s/optional-key :options) CheckpointOptions
-   (s/optional-key :periodic-options) PeriodicCheckpointOptions})
+   (s/optional-key :periodic-options) PeriodicCheckpointOptions
+   (s/optional-key :memory-overhead) s/Num})
```
dposada, line 186: Why are we exposing this in the API?

scheduler/src/cook/kubernetes/api.clj
nsinkov: No. The synthetic pod is not enabling checkpointing.

scheduler/src/cook/rest/api.clj
nsinkov: I was thinking to allow for maximum flexibility in case the checkpointing behavior is highly variable.

If this ends up not being used and/or is removed later, would you rather not add it in at all?

scheduler/src/cook/kubernetes/api.clj
dposada: But doesn't the synthetic pod need to ask for the "extra" memory? Just like it asks for the extra resources for the sidecar even though it doesn't fire up the sidecar.

scheduler/src/cook/rest/api.clj
dposada: Yeah, I don't think this is appropriate for users to set.

scheduler/src/cook/kubernetes/api.clj
nsinkov: oops. good point. we shouldn't have added the sidecar-resource-requirements for the synthetic pod. It should have used the same codepath as the scheduler. I'll fix

dposada: Thanks, sounds good

dposada: Is your fix going to make #1528 moot?

nsinkov: yes. updated

scheduler/src/cook/rest/api.clj
dposada: Even if the memory used for checkpointing is a function of the job, shouldn't we be able to define that function rather than asking the user for the answer?

nsinkov: removed

dposada:
Will merge when green

PR #1526, opened 2020-04-24
Adds lock-sharding to k8s controller

dposada:

- adding configuration field `:kubernetes :controller-lock-num-shards`
- creating the corresponding number of lock objects
- introducing `with-process-lock`, a macro that captures the timers and locking logic for anything that makes a call to `process`

We want to introduce parallelism for pod submission API calls, and this is a prerequisite. Without this, any parallelism will get bottle-necked on the current single lock acquisition.

### DIFF
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -31,6 +31,7 @@
            (com.netflix.fenzo VMTaskFitnessCalculator)
            (java.io File)
            (java.net InetAddress)
+           (java.util.concurrent.locks ReentrantLock)
            (org.apache.curator.test TestingServer)
            (org.apache.log4j DailyRollingFileAppender Logger PatternLayout)))
 
@@ -456,8 +457,9 @@
                          lock-objects
                          (repeatedly
                            controller-lock-num-shards
-                           #(Object.))]
-                     (merge {:controller-lock-objects (with-meta
+                           #(ReentrantLock.))]
+                     (merge {:controller-lock-num-shards controller-lock-num-shards
+                             :controller-lock-objects (with-meta
                                                         lock-objects
                                                         {:json-value (str lock-objects)})
                              :default-workdir "/mnt/sandbox"
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -444,13 +444,23 @@
                                   :default-pool "no-pool"}
                                  pool-selection)})))
      :kubernetes (fnk [[:config {kubernetes {}}]]
-                   (merge {:controller-lock-num-shards 32
-                           :default-workdir "/mnt/sandbox"
-                           :pod-condition-containers-not-initialized-seconds 120
-                           :pod-condition-unschedulable-seconds 60
-                           :reconnect-delay-ms 60000
-                           :set-container-cpu-limit? true}
-                          kubernetes))}))
+                   (let [{:keys [controller-lock-num-shards]
+                          :or {controller-lock-num-shards 32}}
+                         kubernetes
+                         lock-objects
+                         (repeatedly
+                           controller-lock-num-shards
+                           #(Object.))]
+                     (log/info "Doing" controller-lock-num-shards "way lock-sharding in k8s")
+                     (merge {:controller-lock-objects (with-meta
+                                                        lock-objects
+                                                        {:json-value (str lock-objects)})
+                             :default-workdir "/mnt/sandbox"
+                             :pod-condition-containers-not-initialized-seconds 120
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/config.clj
```diff
@@ -444,12 +444,23 @@
...
-                           :set-container-cpu-limit? true}
-                          kubernetes))}))
+                   (let [{:keys [controller-lock-num-shards]
+                          :or {controller-lock-num-shards 32}}
+                         kubernetes
+                         lock-objects
+                         (repeatedly
+                           controller-lock-num-shards
+                           #(Object.))]
+                     (log/info "Doing" controller-lock-num-shards "way lock-sharding in k8s")
```
scrosby, line 454: Why are we logging this and not other parameters? Isn't it all set in :settings anyways. We can also copy the value over, explicitly, so its in :settings instead of requiring the user to count the number of Object's in the array?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -9,22 +9,32 @@
...
+           (nth lock-objects#
+                (-> ~pod-name
+                    hash
+                    (mod (count lock-objects#))))
+           timer-context#
+           (timers/start
+             (metrics/timer
+               "process-lock-acquire"
+               ~compute-cluster-name))]
+       (locking lock-object#
```
scrosby, line 35: If this is going to be a macro, switch to https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html do the try/finally. Locking time should be microseconds on an empty queue.

scheduler/src/cook/config.clj
```diff
@@ -444,12 +444,23 @@
...
                                  pool-selection)})))
      :kubernetes (fnk [[:config {kubernetes {}}]]
-                   (merge {:default-workdir "/mnt/sandbox"
-                           :pod-condition-containers-not-initialized-seconds 120
-                           :pod-condition-unschedulable-seconds 60
-                           :reconnect-delay-ms 60000
-                           :set-container-cpu-limit? true}
-                          kubernetes))}))
+                   (let [{:keys [controller-lock-num-shards]
+                          :or {controller-lock-num-shards 32}}
```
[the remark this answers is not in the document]
dposada: Good call, will add

scheduler/src/cook/config.clj
dposada: Good points, will do

scheduler/src/cook/kubernetes/controller.clj
dposada: I don't understand the "If this is going to be a macro" part of your comment -- isn't the decision to use `ReentrantLock` orthogonal to whether or not this is a macro?

scrosby: Yes and no. Part of the convenience of using a monitor-style lock is conciser syntax and harder to misuse (because you're forced syntatically to free the lock at the end.) However, those advantages are eliminated when using a macro. ReentrantLock is faster, the modern choice,  and has more metrics --- in particular, queue size.

scheduler/src/cook/config.clj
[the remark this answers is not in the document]
dposada: Done

scheduler/src/cook/config.clj

scheduler/src/cook/kubernetes/controller.clj
dposada: Sounds good, will do

scheduler/src/cook/config.clj
```diff
@@ -444,12 +444,23 @@
...
+                         kubernetes
+                         lock-objects
+                         (repeatedly
+                           controller-lock-num-shards
+                           #(Object.))]
+                     (log/info "Doing" controller-lock-num-shards "way lock-sharding in k8s")
+                     (merge {:controller-lock-objects (with-meta
+                                                        lock-objects
+                                                        {:json-value (str lock-objects)})
```
nsinkov, line 464: What is the metadata for?

dposada: It's used here:

https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/rest/api.clj#L2678

when producing the JSON output for `/settings`

nsinkov: nitpick: change to `(str controller-lock-num-shards " lock-objects")`

dposada: We're already showing the # of shards in `/settings` with a different key, `controller-lock-num-shards`.

PR #1528, opened 2020-04-28
Uses add-as-decimals for synthetic pods

dposada:

- using `add-as-decimals` when calculating resource requirements for synthetic pods

To mitigate floating point rounding error introduced when adding in the sidecar resource requirements.


dposada:
This is now moot, it was resolved in a better way in #1523

PR #1539, opened 2020-05-07
Don't checkpoint when max attempts exceeded

nsinkov:

- add option to not checkpoint after some number of job attempts have failed

Checkpointing itself can cause failure. This will allow a job to complete

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -605,6 +605,28 @@
            period-sec
            (assoc "COOK_CHECKPOINT_PERIOD_SEC" (str period-sec))))))))
 
+(def default-checkpoint-failure-reasons
+  "Default set of failure reasons that should be counted against checkpointing attempts"
+  #{:mesos-unknown
+    :mesos-command-executor-failed
+    :mesos-container-limitation-memory
+    :mesos-container-launch-failed})
+
+(defn calculate-effective-checkpointing-config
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -605,6 +605,31 @@
...
+    :mesos-container-launch-failed})
+
+(defn calculate-effective-checkpointing-config
+  "Given the job's checkpointing config, calculate the effective config. Making any adjustments such as defaults,
+  overrides, or other behavior modifications."
+  [{:keys [job/checkpoint job/instance] :as job} task-id]
+  (when checkpoint
+    (let [{:keys [default-checkpoint-config]} (config/kubernetes)
+          checkpoint (merge default-checkpoint-config (util/job-ent->checkpoint job))]
+      (if-let [max-checkpoint-attempts (:max-checkpoint-attempts checkpoint)]
```
dposada, line 622: Can we destructure this in the `let` above?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -605,6 +605,31 @@
...
+        (let [checkpoint-failure-reasons (or (:checkpoint-failure-reasons checkpoint) default-checkpoint-failure-reasons)
```
dposada, line 623: Destructure?

scheduler/src/cook/kubernetes/api.clj

scheduler/src/cook/kubernetes/api.clj

dposada:
Will merge when green

PR #1540, opened 2020-05-08
Check for preemption using preemption pod label

nsinkov:

- check preemption pod label instead of assuming running to missing transition implies preemption

We can tell the difference between a node just disappearing or getting killed vs being preempted, which is expected on preemptable machines

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -923,69 +923,69 @@
           job-status (first (filter (fn [c] (= cook-container-name-for-job (.getName c)))
                                     container-statuses))]
       (if (some-> pod .getMetadata .getDeletionTimestamp)
+        ; If a pod has been ordered deleted, treat it as if it was gone, It's being async removed.
+        ; Note that we distinguish between this explicit :missing, and not being there at all when processing
+        ; (:cook-expected-state/killed, :missing) in cook.kubernetes.controller/process
+        {:state :missing
+         :reason "Pod was explicitly deleted"
+         :pod-deleted? true}
         (if (some-> pod .getMetadata .getLabels (get "node-preempted"))
           {:state :missing
            :reason "Node preempted"
            :pod-deleted? true
            :pod-preempted? true}
-          ; If a pod has been ordered deleted, treat it as if it was gone, It's being async removed.
-          ; Note that we distinguish between this explicit :missing, and not being there at all when processing
-          ; (:cook-expected-state/killed, :missing) in cook.kubernetes.controller/process
-          {:state :missing
-           :reason "Pod was explicitly deleted"
-           :pod-deleted? true})
-        ; If pod isn't being async removed, then look at the containers inside it.
-        (if job-status
-          (let [^V1ContainerState state (.getState job-status)]
-            (cond
-              (.getWaiting state)
-              (if (pod-containers-not-initialized? (V1Pod->name pod) pod-status)
-                ; If the containers are not getting initialized,
-                ; then we should consider the pod failed. This
-                ; state can occur, for example, when volume
-                ; mounts fail.
-                {:state :pod/failed
-                 :reason "ContainersNotInitialized"}
-                {:state :pod/waiting
-                 :reason (-> state .getWaiting .getReason)})
-              (.getRunning state)
-              {:state :pod/running
-               :reason "Running"}
-              (.getTerminated state)
-              (let [exit-code (-> state .getTerminated .getExitCode)]
-                (if (= 0 exit-code)
-                  {:state :pod/succeeded
-                   :exit exit-code
-                   :reason (-> state .getTerminated .getReason)}
+          ; If pod isn't being async removed, then look at the containers inside it.
+          (if job-status
+            (let [^V1ContainerState state (.getState job-status)]
+              (cond
+                (.getWaiting state)
+                (if (pod-containers-not-initialized? (V1Pod->name pod) pod-status)
+                  ; If the containers are not getting initialized,
+                  ; then we should consider the pod failed. This
+                  ; state can occur, for example, when volume
+                  ; mounts fail.
                   {:state :pod/failed
-                   :exit exit-code
-                   :reason (-> state .getTerminated .getReason)}))
-              :default
-              {:state :pod/unknown
-               :reason "Unknown"}))
-          (cond
-            ; https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
-            ; Failed means:
-            ; All Containers in the Pod have terminated, and at least
-            ; one Container has terminated in failure. That is, the
-            ; Container either exited with non-zero status or was
-            ; terminated by the system.
-            (= (.getPhase pod-status) "Failed") {:state :pod/failed
-                                                 :reason (.getReason pod-status)}
+                   :reason "ContainersNotInitialized"}
+                  {:state :pod/waiting
+                   :reason (-> state .getWaiting .getReason)})
+                (.getRunning state)
+                {:state :pod/running
+                 :reason "Running"}
+                (.getTerminated state)
+                (let [exit-code (-> state .getTerminated .getExitCode)]
+                  (if (= 0 exit-code)
+                    {:state :pod/succeeded
+                     :exit exit-code
+                     :reason (-> state .getTerminated .getReason)}
+                    {:state :pod/failed
+                     :exit exit-code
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -922,67 +922,70 @@
...
-              (let [exit-code (-> state .getTerminated .getExitCode)]
-                (if (= 0 exit-code)
-                  {:state :pod/succeeded
-                   :exit exit-code
-                   :reason (-> state .getTerminated .getReason)}
+        (if (some-> pod .getMetadata .getLabels (get (or (-> (config/kubernetes) :node-preempted-label) "node-preempted")))
+          {:state :missing
+           :reason "Node preempted"
+           :pod-deleted? true
+           :pod-preempted? true}
```
dposada, line 936: Maybe we should use `cond` instead of nesting more `if`s? If we do, I think we could merge with the `cond` below into one big `cond`.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -471,16 +471,12 @@
...
                                           ; This indicates that something deleted it behind our back
-                                          :missing (do
-                                                     (log/info "In compute cluster" name ", something deleted"
-                                                               pod-name "behind our back")
-                                                     ; A :cook-expected-state/running job suddenly disappearing in k8s is
-                                                     ; a sign of a preemption, so treat this case as a missed preemption.
-                                                     ; TODO: When we have a better story for preemption, we should switch this to
-                                                     ; go through the handle-pod-completed stack. That needs to be guarded with a null
-                                                     ; check because handle-pod-completed cannot handle null pods. So, something like:
-                                                     ; (if pod (handle-pod-completed ...) (....write a unknown reason to the pod....))
```
dposada, line 482: Do we not need the `null`-check described here?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -471,16 +471,12 @@
...
-                                                     (handle-pod-preemption compute-cluster pod-name))
+                                          :missing (if (some-> synthesized-state :pod-preempted?)
+                                                     (handle-pod-preemption compute-cluster pod-name)
+                                                     (do
+                                                       (log/info "In compute cluster" name ", something deleted"
+                                                                 pod-name "behind our back")
+                                                       (handle-pod-completed compute-cluster pod-name k8s-actual-state-dict)))
```
dposada, line 479: Does this entail a change to the state transition diagram (https://github.com/twosigma/Cook/blob/master/scheduler/docs/kubernetes-state.dot)? I'm not sure, just checking so that it stays up to date.

dposada, line 479: Can we add a test for this change?

nsinkov: @scrosby I wasn't sure how to test this. would you know how?

scheduler/src/cook/kubernetes/api.clj
nsinkov: The code below is "if job-status"
how would you want it to look?

scheduler/src/cook/kubernetes/controller.clj
nsinkov: @scrosby we agreed that we don't need it. can you confirm?

scrosby: We need the null check. When we know its preempted, we don't need the null check; we couldn't have been able to set that flag with a null Pod --- But that's the first branch of the 'if' that doesn't need a null check. We still do here.

scheduler/src/cook/kubernetes/controller.clj
nsinkov: Looks good

scheduler/src/cook/kubernetes/controller.clj
nsinkov: ok, so what does "write a unknown reason to the pod" look like

scheduler/src/cook/kubernetes/api.clj
dposada: I see what you mean. I'm not sure. But seems like this function is getting out of hand. We don't have to fix it in this PR. Maybe extract functions or something?

scheduler/src/cook/kubernetes/controller.clj
scrosby: Roughly the same code as handle-pod-preemption, but with a different reason code.

scheduler/src/cook/kubernetes/controller.clj
dposada: `test-process` is where we should test it:

https://github.com/twosigma/Cook/blob/master/scheduler/test/cook/test/kubernetes/controller.clj#L30-L106

scheduler/src/cook/kubernetes/api.clj
nsinkov: yeah

scheduler/src/cook/kubernetes/controller.clj
nsinkov: updated

scheduler/src/cook/kubernetes/controller.clj
nsinkov: updated

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -155,6 +155,19 @@
...
   {:cook-expected-state :cook-expected-state/completed})
 
+(defn handle-pod-externally-deleted
+  "Marks the corresponding job instance as failed in the database and
+  returns the `completed` cook expected state."
+  [{:keys [name] :as compute-cluster} pod-name]
+  (log/info "In compute cluster" name ", pod" pod-name "was externally deleted")
+  (when-not (api/synthetic-pod? pod-name)
+    (let [instance-id pod-name
+          status {:reason :reason-executor-terminated
```
dposada, line 165: The reason string for this reason is "Mesos executor terminated". Let's use something less Mesos-y for this.

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -103,7 +108,21 @@
...
+    (is (= :reason-slave-removed @reason))
+    (is (nil? (do-process :cook-expected-state/running :missing :custom-test-state {:state :missing
+                                                                                    :reason "Pod was explicitly deleted"
+                                                                                    :pod-deleted? true}
+                          :force-nil-pod? true)))
+    (is (= :reason-executor-terminated @reason))
+    (is (nil? (do-process :cook-expected-state/running :missing :custom-test-state {:state :missing
+                                                                                    :reason "Pod was explicitly deleted"
+                                                                                    :pod-deleted? true})))
+    (is (= :reason-executor-terminated @reason))))
```
dposada, line 125: Can we put these cases up next to the existing `[running, missing]` case?

https://github.com/twosigma/Cook/pull/1540/files/67fca4eb67ae79cf7f5e234b90a9ace700fecaef..f78ac65eb64e65b431babf309653daac0c26b111#diff-e60eb787ca7fc5a6b343e52606bd47f5R80

scheduler/src/cook/kubernetes/controller.clj
nsinkov: I didn't find anything better without creating something

dposada: Discussed offline, ended up creating a new reason

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -471,16 +484,12 @@
...
-                                                     ; TODO: When we have a better story for preemption, we should switch this to
-                                                     ; go through the handle-pod-completed stack. That needs to be guarded with a null
-                                                     ; check because handle-pod-completed cannot handle null pods. So, something like:
-                                                     ; (if pod (handle-pod-completed ...) (....write a unknown reason to the pod....))
-                                                     (handle-pod-preemption compute-cluster pod-name))
+                                          :missing (if (some-> synthesized-state :pod-preempted?)
+                                                     (handle-pod-preemption compute-cluster pod-name)
+                                                     (do
+                                                       (log/info "In compute cluster" name ", something deleted"
+                                                                 pod-name "behind our back")
```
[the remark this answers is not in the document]
nsinkov: updated

dposada:
@nsinkov: are we waiting on @scrosby to approve this, or should I go ahead and merge?

PR #1557, opened 2020-05-21
Adds reason string to Instance in jobclient

dposada:

- adding instance reason string to the `Instance` class

So that clients can access the reason string along with the other instance data.


nsinkov:
let's use autovalue :P

PR #1561, opened 2020-05-25
Do not use :missing state for preempted pod.

nsinkov:

- preempted pod label will not change pod state

Do not use :missing state for preempted pod. Preempted pod is still alive until a new instance is created.


scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -83,7 +83,7 @@
                                                                                     :reason "Node preempted"
                                                                                     :pod-deleted? true
                                                                                     :pod-preempted? true})))
```
dposada, line 85: Aren't we removing `:pod-preempted?` as a thing with this PR?

nsinkov: for now, yes

dposada:
Will merge when green

PR #1567, opened 2020-05-28
Logs start and finish of each matching round

dposada:

- 
- 
-


dposada:
@scrosby is going to open a different PR to handle this additional logging, plus adding some needed metrics

PR #1569, opened 2020-06-01
Python Client API

hdrodz97:

- Create an official Python client API for Cook
- Move the Java client API to the `java/` subdirectory of the `jobclient/` directory


jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        :type priority: int, optional
+        :param application: Application information to assign to the job,
+            defaults to None.
+        :type application: Application, optional
+        Output
+        ------
+        :return: The UUID of the newly-created job.
+        :rtype: UUID
+        """
+        uuid = str(uuid if uuid is not None else util.make_temporal_uuid())
```
dposada, line 131: 👍

dposada: Could this be simplified to:

```suggestion
        uuid = str(uuid or util.make_temporal_uuid())
```

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        _LOG.debug("Payload:")
+        _LOG.debug(json.dumps(payload, indent=4))
+        resp = requests.post(url, json=payload)
+        if not resp.ok:
+            _LOG.error(f"Could not submit job: {resp.status_code} {resp.text}")
+            resp.raise_for_status()
+
+        return UUID(uuid)
+
+    def query(self, uuid: UUID) -> Job:
```
dposada, line 163: I think it would be nice to also allow just a string for the UUID.

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        url = urlunparse(('http', self.__netloc, self.__job_endpoint, '',
+                          query, ''))
+        _LOG.debug(f'Sending GET to {url}')
+        resp = requests.get(url)
+        if not resp.ok:
+            _LOG.error(f"Could not query job: {resp.status_code} {resp.text}")
+            resp.raise_for_status()
+        return Job.from_dict(resp.json()[0])
+
+    def kill(self, uuid: UUID):
```
dposada, line 185: Same here, I think allowing string would be nice.

jobclient/python/constraints.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def parse_from(constraint: list) -> Constraint:
+    """Parse a sequence of constraints from a list."""
+    op, attr, val = constraint
+    op = Operator.from_string(op)
+    if op == Operator.EQUALS:
+        return OneToOneConstraint(op, attr, val)
+    else:
+        raise NotImplementedError(f"Operator {op} is not supported.")
```
dposada, line 94: My opinion is that we leave constraint support out of this PR to keep it small. Thoughts?

jobclient/python/groups.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def to_dict(self) -> dict:
+        return {
+            'uuid': self.uuid,
+            'status': self.status,
+            'name': self.name,
+            'host_placement': self.host_placement,
+            'straggler_handling': self.straggler_handling,
+            'jobs': self.jobs
+        }
```
dposada, line 175: My opinion is that we leave group support out of this PR as well to keep it small. What do you think?

jobclient/python/instance.py
```diff
@@ -0,0 +1,208 @@
...
+
+    @staticmethod
+    def from_string(name: str) -> 'Executor':
+        """Create an executor from a case-insensitive string representation."""
+        return _EXECUTOR_LOOKUP[name.lower()]
+
+
+_EXECUTOR_LOOKUP = {
+    'cook': Executor.COOK,
+    'executor': Executor.EXECUTOR,
```
dposada, line 67: I don't think the `executor` will ever be a value for this field.

jobclient/python/instance.py
```diff
@@ -0,0 +1,208 @@
...
+    output_url: Optional[str]
+    executor: Optional[Executor]
+    reason_mea_culpa: Optional[bool]
+
+    def __init__(self, *,
+                 # Required arguments
+                 task_id: UUID,
+                 slave_id: str,
+                 executor_id: str,
+                 start_time: int,
```
dposada, line 94: Can we convert this to a Python date?

dposada: (Same for the other timestamps below)

jobclient/python/instance.py
```diff
@@ -0,0 +1,208 @@
...
+        self.reason_mea_culpa = reason_mea_culpa
+
+    def __hash__(self):
+        PRIME = 31
+        result = 1
+        result = (PRIME * result +
+                  hash(self.status) if self.status is not None else 0)
+        result = (PRIME * result +
+                  hash(self.task_id) if self.task_id is not None else 0)
+        return result
```
dposada, line 160: What is this used for?

jobclient/python/instance.py
```diff
@@ -0,0 +1,208 @@
...
+    def __eq__(self, other):
+        if self is other:
+            return True
+        if not isinstance(other, self.__class__):
+            return False
+        if self.status != other.status:
+            return False
+        if self.task_id != other.task_id:
+            return False
+        return True
```
dposada, line 171: What is this used for?

jobclient/python/util.py
```diff
@@ -0,0 +1,117 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+import uuid
+
+from uuid import UUID
+
+
+class FetchableUri:
```
dposada, line 21: What is this used for?

jobclient/python/straggler_handling.py
```diff
@@ -0,0 +1,86 @@
...
+    def to_dict(self) -> dict:
+        """Get the `dict` representation of this object."""
+        return {
+            'type': self.handling_type,
+            'parameters': self.parameters
+        }
+
+    def __index__(self, key: str) -> Any:
+        """Get the value associated with a parameter."""
+        return self.parameter(key)
```
dposada, line 86: I think we should also leave straggler-handling support out of this PR. Thoughts?

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+from instance import Executor, Instance
+from util import FetchableUri
+
+
+class Status(Enum):
+    """
+    The curent status of a job.
+
+    A job in Cook has only three statuses: WAITING, RUNNING, COMPLETED. The
+    INITIALIZED status is for a `JobListener` to receive status updates.
```
dposada, line 35: I think we should not do the `JobListener` stuff in this PR.

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+class Job:
+    """A job object returned from Cook."""
+    command: str
+    mem: float
+    cpus: float
+
+    uuid: UUID
+    name: str
+    max_retries: int
+    max_runtime: int
```
dposada, line 119: Can we convert this to a Python timedelta?

dposada: (Same for the other time-related fields below)

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+                 cpus: float,
+                 # Optional args with defaults
+                 uuid: Optional[UUID] = None,
+                 name: str = 'cookjob',
+                 max_retries: int = 5,
+                 max_runtime: int = sys.maxsize,
+                 state: State = State.WAITING,
+                 status: Status = Status.INITIALIZED,
+                 priority: int = 50,
+                 disable_mea_culpa_retries: bool = False,
```
dposada, line 159: I don't think we should default these client-side. This is only for the query path, correct? I think we should just use what we get from the REST API.

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+        """Initializes a job object.
+
+        Normally, this function wouldn't be invoked directly. It is instead
+        used by `JobClient` when getting the data from `query`.
+
+        Required Parameters
+        -------------------
+        :param command: The command-line string that this job will execute.
+        :type command: str
+        :param mem: The amount of memory, in GB, to request from the scheduler.
```
dposada, line 190: MB?

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+        :type ports: Optional[int], optional
+        :param submit_time: The UNIX timestamp when this job was submitted.
+        :type submit_time: Optional[int], optional
+        :param retries_remaining: The number of retries remaining for this job.
+        :type retries_remaining: Optional[int], optional
+        """
+        self.command = command
+        self.mem = mem
+        self.cpus = cpus
+        self.uuid = uuid if uuid is not None else util.make_temporal_uuid()
```
dposada, line 262: This is only for the query path, correct?

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+        result = (PRIME * result
+                  + hash(self.uuid) if self.uuid is not None else 0)
+        return result
+
+    def __eq__(self, other):
+        if self is other:
+            return True
+        if not isinstance(other, self.__class__):
+            return False
+        return self.uuid == other.uuid and self.status == other.status
```
dposada, line 305: What are these used for?

dposada:
@hdrodz97 One thing I forgot to mention: can we add an integration test or two that demonstrate using the python client library?

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+    executor: Optional[Executor]
+    expected_runtime: Optional[int]
+    pool: Optional[str]
+    instances: Optional[List[Instance]]
+    env: Optional[Dict[str, str]]
+    uris: Optional[List[FetchableUri]]
+    container: Optional[dict]
+    labels: Optional[Dict[str, str]]
+    constraints: Optional[Set[Constraint]]
+    groups: Optional[List[UUID]]
```
rmanyari, line 134: can a job belong to multiple groups @dposada ?

jobclient/python/jobs.py
```diff
@@ -0,0 +1,381 @@
...
+    application: Optional[Application]
+    progress_output_file: Optional[str]
+    progress_regex_string: Optional[str]
+    user: Optional[str]
+    datasets: Optional[list]
+    gpus: Optional[float]
```
rmanyari, line 140: Optional[int]

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+    __job_endpoint: str
+    __delete_endpoint: str
+    __group_endpoint: str
+
+    __status_update_interval: int = _DEFAULT_STATUS_UPDATE_INTERVAL_SECONDS
+    __submit_retry_interval: int = _DEFAULT_SUBMIT_RETRY_INTERVAL_SECONDS
+    __batch_request_size: int = _DEFAULT_BATCH_REQUEST_SIZE
+    __request_timeout_seconds: int = _DEFAULT_REQUEST_TIMEOUT_SECONDS
+
+    def __init__(self, host: str, port: int, *,
```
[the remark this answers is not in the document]
hdrodz97: :+1:

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+               application: Optional[Application] = None) -> UUID:
+        """Submit a single job to Cook.
+
+        Required Parameters
+        -------------------
+        :param command: The command to run on Cook.
+        :type command: str
+        :param cpus: The number of CPUs to request from Cook.
+        :type cpus: float
+        :param mem: The amount of memory, in GB, to request from Cook.
```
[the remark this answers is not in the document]
hdrodz97: Whoops, not sure where I got the idea that this was in GB from. I'll fix that up.

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        :type mem: float
+        :param max_retries: The number of times this job should be retried
```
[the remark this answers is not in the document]
hdrodz97: Sounds good; I'll leave the name as `max_retries` (to keep continuity with the rest of Cook) but leave a comment on the name/function discrepancy

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+            before failing.
+        :type max_retries: int
+        Optional Parameters
+        -------------------
+        :param uuid: The UUID of the job to submit. If this value is not
+            provided, then a random UUID will be generated.
```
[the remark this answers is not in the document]
hdrodz97: It does do temporal random, it's just not mentioned in the docs. I'll add a note on that.

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        :type uuid: UUID, optional
+        :param env: Environment variables to set within the job's context,
+            defaults to None.
+        :type env: Dict[str, str], optional
+        :param labels: Labels to assign to the job, defaults to None.
+        :type labels: Dict[str, str], optional
+        :param max_runtime: The maximum number seconds this job should be
+            allowed to run, defaults to None.
```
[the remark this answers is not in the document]
hdrodz97: :+1:

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        :type max_runtime: int, optional
+        :param name: A name to assign to the job, defaults to None.
```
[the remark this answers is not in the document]
hdrodz97: :+1:

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+        :type name: str, optional
+        :param priority: A priority to assign to the job, defaults to None.
+        :type priority: int, optional
+        :param application: Application information to assign to the job,
+            defaults to None.
```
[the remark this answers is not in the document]
hdrodz97: :+1:

jobclient/python/__init__.py
hdrodz97: :+1:

jobclient/python/__init__.py
hdrodz97: :+1:

jobclient/python/__init__.py
hdrodz97: :+1:

jobclient/python/groups.py
hdrodz97: Makes sense, the `Group` class isn't instantiated anywhere at the moment so I can get rid of this

jobclient/python/instance.py
hdrodz97: :+1:

jobclient/python/instance.py
hdrodz97: It was in the original Java client code, but from a quick glance of it it's not actively used anywhere.

jobclient/python/instance.py
hdrodz97: (Same as above)

jobclient/python/util.py
hdrodz97: It's part of the `Job` object, documentation seems to hint at this being a Mesos-centric thing. Despite that, as it is part of the `Job` schema I feel like it'd be better to have a concrete definition for this instead of hiding it behind a more abstract `dict`. Thoughts?

jobclient/python/straggler_handling.py
hdrodz97: Sounds good, as we're removing groups from the PR and this is only used within the context of a group removing this makes sense.

jobclient/python/jobs.py
hdrodz97: :+1:

jobclient/python/jobs.py
hdrodz97: :+1:

jobclient/python/jobs.py
hdrodz97: :+1:

jobclient/python/jobs.py
hdrodz97: :+1:

jobclient/python/jobs.py
hdrodz97: Yup, will change!

jobclient/python/jobs.py
hdrodz97: They were implemented in the Java client, from the looks of it only for the `JobListener` stuff (as `Job`s are used as keys). I'll get rid of it

jobclient/python/jobs.py
dposada: If I remember correctly, there's room for that in the API but it's not actually currently possible to accomplish.

jobclient/python/__init__.py
```diff
@@ -0,0 +1,206 @@
...
+               mem: float,
+               max_retries: int,
+
+               uuid: Optional[UUID] = None,
+               env: Optional[Dict[str, str]] = None,
+               labels: Optional[Dict[str, str]] = None,
+               max_runtime: Optional[int] = None,
+               name: Optional[str] = None,
+               priority: Optional[int] = None,
+               application: Optional[Application] = None) -> UUID:
```
adcramer, line 92: Can this just be a string? Is `application.version` required?

hdrodz97: [This](https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/rest/api.clj#L266-L269) seems to imply that the version is required to be present and non-empty. One alternative would be to allow omission of the version string and fill in with some default value, e.g. `versionless`.

dposada: That's correct, the API requires a version if you provide an application name. You can, however, not specify either.

jobclient/python/cook/scheduler/client/__init__.py
```diff
@@ -12,25 +12,28 @@
...
+from . import util
+from .jobs import Application, Job
 
 _LOG = logging.getLogger(__name__)
 _LOG.addHandler(logging.StreamHandler())
 _LOG.setLevel(logging.DEBUG)
 
 _COOK_IMPERSONATE_HEADER = 'X-Cook-Impersonate'
 
+_CLIENT_APP = Application('cook-python-client', '0.1')
```
dposada, line 35: Let's use semantic versioning (https://semver.org/)

jobclient/python/cook/scheduler/client/__init__.py
```diff
@@ -67,29 +68,24 @@ def __init__(self, host: str, port: int, *,
...
-        self.__delete_endpoint = delete_endpoint
+        self.__netloc = url
+        self.__job_endpoint = _DEFAULT_JOB_ENDPOINT
+        self.__delete_endpoint = _DEFAULT_DELETE_ENDPOINT
 
     def submit(self, *,
                command: str,
                cpus: float,
                mem: float,
                max_retries: int,
```
dposada: Maybe we should make these have defaults too? 

The CLI defaults these so that you can so something as simple as:

```
cs submit env
```

jobclient/python/cook/scheduler/client/__init__.py
```diff
@@ -160,19 +157,20 @@ def submit(self, *,
...
-        :type uuid: UUID
+        :type uuid: Union[str, UUID]
         :return: A Job object containing the job's information.
         :rtype: Job
         """
+        uuid = str(uuid)
         if self.__job_endpoint == '/jobs':
             param_name = 'uuid'
         else:
             param_name = 'job'
```
dposada: I think we should get rid of this complexity and only allow using `/jobs`, since the other endpoint is deprecated.

jobclient/python/cook/scheduler/client/constraints.py
```diff
@@ -53,18 +53,6 @@ def __init__(self, operator: Operator, attribute: str, value: str):
...
-            return True
-        if not isinstance(other, self.__class__):
-            return False
-        return (self.operator == other.operator and
-                self.attribute == other.attribute and
-                self.value == other.value)
-
     def to_list(self) -> list:
         """Generate this constraint's `list` representation."""
         return [
```
dposada, line 58: Do we need any of the constraint stuff in this PR? Seems like maybe we can just rip it all out for now?

jobclient/python/cook/scheduler/client/instance.py
```diff
@@ -150,39 +153,19 @@ def __init__(self, *,
...
-            return True
-        if not isinstance(other, self.__class__):
-            return False
-        if self.status != other.status:
-            return False
-        if self.task_id != other.task_id:
-            return False
-        return True
-
     def to_dict(self) -> dict:
```
dposada, line 156: Where is `Instance.to_dict` used?

jobclient/python/cook/scheduler/client/instance.py
```diff
@@ -201,8 +184,11 @@ def to_dict(self) -> dict:
     def from_dict(cls, d: dict) -> 'Instance':
         """Create an instance from its `dict` representation."""
         d = deepcopy(d)
-        d['uuid'] = UUID(d['uuid'])
+        d['task_id'] = UUID(d['task_id'])
+        d['start_time'] = datetime.fromtimestamp(d['start_time'] / 1000)
         d['status'] = Status.from_string(d['status'])
+        if 'end_time' in d:
+            d['end_time'] = datetime.fromtimestamp(d['end_time'] / 1000)
```
dposada, line 191: Since we do this in a couple of spots, should we have a util function `milliseconds_to_datetime` or similar?

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -112,197 +104,181 @@ class Job:
...
+    user: str
     executor: Optional[Executor]
-    expected_runtime: Optional[int]
+    container: Optional[dict]
+    disable_mea_culpa_retries: Optional[bool]
+    expected_runtime: Optional[timedelta]
     pool: Optional[str]
     instances: Optional[List[Instance]]
     env: Optional[Dict[str, str]]
     uris: Optional[List[FetchableUri]]
```
dposada, line 125: Maybe just `List[str]` for this?

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -112,197 +104,181 @@ class Job:
...
-    container: Optional[dict]
     labels: Optional[Dict[str, str]]
     constraints: Optional[Set[Constraint]]
```
dposada, line 127: Ah, this is where the `Constraint` class is used. I guess the question is: maybe we should wait to add this until we support specifying constraints on the job submission spec?

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -112,197 +104,181 @@ class Job:
...
+    group: Optional[UUID]
     groups: Optional[List[UUID]]
     application: Optional[Application]
     progress_output_file: Optional[str]
     progress_regex_string: Optional[str]
-    user: Optional[str]
-    datasets: Optional[list]
-    gpus: Optional[float]
-    framework_id: Optional[str]
+    datasets: Optional[List[Dataset]]
```
dposada, line 133: Support for data locality never finished getting implemented. I would say we leave this out for now.

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -112,197 +104,181 @@ class Job:
...
-        return result
-
-    def __eq__(self, other):
-        if self is other:
-            return True
-        if not isinstance(other, self.__class__):
-            return False
-        return self.uuid == other.uuid and self.status == other.status
 
     def to_dict(self) -> dict:
```
dposada, line 283: Where is `Job.to_dict` used?

jobclient/python/cook/scheduler/client/util.py
```diff
@@ -96,6 +77,38 @@ def from_dict(cls, d: dict) -> 'FetchableUri':
...
+        d = {'dataset': self.dataset}
+        if self.partitions is not None:
+            d['partitions'] = self.partitions
+        return d
+
+    @classmethod
+    def from_dict(cls, d: dict) -> 'Dataset':
+        """Parse a Dataset from its dict representation."""
+        d = copy.deepcopy(d)
+        return cls(**d)
```
dposada, line 109: Since data locality is not yet supported, I think we should leave this out for now.

jobclient/python/tests/integration/test_client.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertAlmostEqual(job.max_retries, 5)
+
+    def test_kill(self):
+        uuid = self.client.submit(command='ls',
+                                  cpus=0.5,
+                                  mem=1.0,
+                                  max_retries=5)
+        self.client.kill(uuid)
+        job = self.client.query(uuid)
+        self.assertEqual(job.status, JobStatus.COMPLETED)
```
dposada, line 53: I think we should put this file under the top-level `integration` folder so that it's side-by-side with all of the other integration tests.

jobclient/python/tests/integration/test_client.py
```diff
@@ -0,0 +1,53 @@
...
+# limitations under the License.
+
+import os
+
+from unittest import TestCase
+
+from cook.scheduler.client import JobClient
+from cook.scheduler.client.jobs import Status as JobStatus
+
+CLIENT_HOST = os.environ.get('TEST_CLIENT_HOST', 'localhost:12321')
```
dposada, line 22: The pattern we use for retrieving the URL under test is:

```
    @classmethod
    def setUpClass(cls):
        cls.cook_url = util.retrieve_cook_url()
        util.init_cook_session(cls.cook_url)

    def setUp(self):
        self.cook_url = type(self).cook_url
        ...
```

(e.g. https://github.com/twosigma/Cook/blob/master/integration/tests/cook/test_basic.py#L28-L34)

jobclient/python/tests/integration/test_client.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_query(self):
+        uuid = self.client.submit(command='ls',
+                                  cpus=0.5,
+                                  mem=1.0,
+                                  max_retries=5)
+        job = self.client.query(uuid)
+        self.assertEqual(job.command, 'ls')
+        self.assertAlmostEqual(job.cpus, 0.5)
+        self.assertAlmostEqual(job.max_retries, 5)
```
dposada, line 44: Can't we use `assertEqual` for this one?

jobclient/python/tests/unit/test_instance.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_dict_parse_optional(self):
+        instdict = INSTANCE_DICT_WITH_OPTIONALS
+        inst = Instance.from_dict(instdict)
+        self._check_optional_fields(inst, instdict)
+
+    def test_dict_output(self):
+        inst = INSTANCE_EXAMPLE
+        instdict = inst.to_dict()
+        self._check_required_fields(inst, instdict)
+        self._check_optional_fields(inst, instdict)
```
dposada, line 98: Should we set these unit tests up to run in Travis, probably here?

https://github.com/twosigma/Cook/blob/master/.travis.yml#L33-L40

jobclient/python/cook/scheduler/client/constraints.py
hdrodz97: We probably could and just leave them as raw lists

jobclient/python/cook/scheduler/client/jobs.py
hdrodz97: Yeah, we could probably hold off on this and either leave it in raw list form or just ignore the parameter entirely

jobclient/python/tests/integration/test_client.py
hdrodz97: You're right, looks like my head was in-between looking at `mem` (which should use `assertAlmostEqual`) and `max_retries` (which doesn't need to)

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -112,197 +104,181 @@ class Job:
...
+    expected_runtime: Optional[timedelta]
     pool: Optional[str]
     instances: Optional[List[Instance]]
     env: Optional[Dict[str, str]]
     uris: Optional[List[FetchableUri]]
-    container: Optional[dict]
     labels: Optional[Dict[str, str]]
     constraints: Optional[Set[Constraint]]
+    group: Optional[UUID]
     groups: Optional[List[UUID]]
```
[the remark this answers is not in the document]
hdrodz97: The API seems to always return a `groups` key when fetching a job. However, as mentioned in an earlier conversation, it's not currently possible to set multiple groups on a job?

In that case, how is the single-job case handled? Is the `groups` key just set to a single-element list, e.g. `"groups": ["$SOME_UUID"]`, or is the `group` key set to the UUID? (seeing as how both [`group`](https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/rest/api.clj#L335) and [`groups`](https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/rest/api.clj#L389) are defined in api.clj)

jobclient/python/cook/scheduler/client/__init__.py
hdrodz97: What version should we start with? `0.1.0`?

jobclient/python/cook/scheduler/client/jobs.py
hdrodz97: It was mostly a convenience function for showing the entire state of the object at once (something in the vein of `print(json.dumps(job.to_dict(), indent=4))`), although if this were to stay it'd probably be called within an implementation for `__str__`

jobclient/python/cook/scheduler/client/__init__.py
dposada: Yeah, `0.1.0` sounds good

jobclient/python/cook/scheduler/client/jobs.py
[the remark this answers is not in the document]
dposada: Yeah, currently you get back a single-element list, e.g.:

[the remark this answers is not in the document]
dposada: I would say we just have `group` for now, and have it be the 0th group in the list returned by the API. There are no near-term plans to support a job belonging to multiple groups.

integration/tests/cook/test_client.py
```diff
@@ -41,13 +41,17 @@ def test_query(self):
         job = self.client.query(uuid)
         self.assertEqual(job.command, 'ls')
         self.assertAlmostEqual(job.cpus, 0.5)
-        self.assertAlmostEqual(job.max_retries, 5)
+        self.assertAlmostEqual(job.mem, 1.0)
+        self.assertEqual(job.max_retries, 5)
 
     def test_kill(self):
-        uuid = self.client.submit(command='ls',
+        uuid = self.client.submit(command='sleep 10',
```
dposada, line 48: Just to be on the safe side, let's do:

```suggestion
        uuid = self.client.submit(command=f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}',
```

jobclient/python/cook/scheduler/client/__init__.py
```diff
@@ -32,7 +32,7 @@
 
 _COOK_IMPERSONATE_HEADER = 'X-Cook-Impersonate'
 
-_CLIENT_APP = Application('cook-python-client', '0.1')
+_CLIENT_APP = Application('cook-python-client', '0.1.0')
 
 _DEFAULT_STATUS_UPDATE_INTERVAL_SECONDS = 10
 _DEFAULT_BATCH_REQUEST_SIZE = 32
```
dposada: Are these used currently?

jobclient/python/cook/scheduler/client/jobs.py
```diff
@@ -359,15 +361,20 @@ def from_dict(cls, d: dict) -> 'Job':
...
-            d['uris'] = list(map(FetchableUri.from_dict, d['uris']))
-        if 'constraints' in d:
-            d['constraints'] = set(map(constraints.parse_from, d['constraints']))  # noqa E501
+            d['uris'] = [uri['value'] for uri in d['uris']]
+        # Assigning the Job's group works in the following manner:
+        # * If the key 'group' is defined, use that UUID.
+        # * Otherwise, if the key 'groups' is defined and has more than one
+        #   item, use the first element of that list.
         if 'group' in d:
             d['group'] = UUID(d['group'])
```
dposada: We should be able to get rid of this check. If the job is in a group, the API will always return `groups` (the list) with a single group UUID in it.

rmanyari:
:shipit:

jobclient/python/setup.py
```diff
@@ -0,0 +1,28 @@
...
+from cook.scheduler.client import CLIENT_VERSION
+
+with open('README.md') as fd:
+    readme = fd.read()
+
+requirements = [
+    'requests'
+]
+
+setup(name='cook-scheduler-client',
```
DaoWen, line 14: I like the fact that we didn't call this "jobclient" because it gives us flexibility to include things other than just job management in this API. We also avoid people calling the library "the job client", which might be confused with other things that manage "jobs" in other contexts. I think "cook (scheduler) client" is both clear and concise. 👍

jobclient/python/docs/source/conf.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'Cook Python Client API'
+copyright = '2020, Two Sigma'
+author = 'Two Sigma'
+
+# The full version, including alpha/beta/rc tags
+release = '0.1.0'
```
dposada, line 25: Is it possible to only have this defined once, instead of both here and in the client itself?

PR #1575, opened 2020-06-09
Avoids autoscaling beyond what users have quota to run

dposada:

- filtering pending jobs based on quota before calling `trigger-autoscaling!`

So that we don't trigger autoscaling beyond what users have quota to actually run.


dposada:
Closing this - moving to PR #1586

PR #1576, opened 2020-06-12
Rename package from cookclient to cook-client-api

hdrodz97:

- Rename the Python client API from `cookclient` to `cook-client-api`
- The actual module name is still `cookclient`

Match the convention of Cook's existing pip projects


dposada:
Will merge when green

PR #1577, opened 2020-06-12
Dask on Cook Design Proposal

hdrodz97:

- Write a design proposal for the design of Dask on Cook.


dask/docs/design.md
```diff
@@ -0,0 +1,190 @@
...
+
+```python
+from dask_kubernetes import KubeCluster
+cluster = KubeCluster.from_yaml('worker-template.yaml')
+cluster.scale(20)  # add 20 workers
+cluster.adapt()    # or create and destroy workers dynamically based on workload
+
+from dask.distributed import Client
+client = Client(cluster)
+```
```
hjoo, line 30: I'd suggest putting the CookCluster example here at the top, side by side with the KubeCluster example you have. I think it's okay that it's repeated at the bottom of the doc. Putting it up top gives readers a clear picture for what you want to build and they can compare the KubeCluster and proposed CookCluster APIs side by side.

PR #1579, opened 2020-06-17
Update Python client to 0.2.0

hdrodz97:

- Adds node pool and container support
- Fixes Instance parsing issue

### DIFF
--- a/integration/tests/cook/test_client.py
+++ b/integration/tests/cook/test_client.py
@@ -68,7 +68,8 @@ class ClientTest(util.CookTest):
         job = self.client.query(uuid)
 
         remote_container = job.container
-        self.assertEqual(remote_container['type'], 'DOCKER')
+        self.assertEqual(remote_container['type'].lower(),
+                         'DOCKER'.lower())
         self.assertEqual(remote_container['docker']['image'], container.image)
 
         self.client.kill(uuid)
--- a/integration/tests/cook/test_client.py
+++ b/integration/tests/cook/test_client.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import unittest
+
 from cookclient import JobClient
 from cookclient.containers import DockerContainer
 from cookclient.jobs import Job, Status as JobStatus
--- a/integration/tests/cook/test_client.py
+++ b/integration/tests/cook/test_client.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 from cookclient import JobClient
-from cookclient.jobs import Status as JobStatus
+from cookclient.containers import DockerContainer
+from cookclient.jobs import Job, Status as JobStatus
 
 from tests.cook import util
 
@@ -33,6 +34,7 @@ class ClientTest(util.CookTest):
                                   mem=1.0,
                                   max_retries=5)
         self.assertTrue(uuid is not None)
+        self.client.kill(uuid)
 
     def test_query(self):
         uuid = self.client.submit(command='ls',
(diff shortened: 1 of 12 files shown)


jobclient/python/cookclient/instance.py
```diff
@@ -98,6 +98,8 @@ class Instance:
     :type status: cookclient.instance.Status
     :param preempted: If true, then this instance was preempted.
     :type preempted: bool
+    :param backfilled: If true, than this instance was backfilled.
+    :type backfilled: bool
```
dposada, line 102: The `backfilled` field is deprecated -

https://github.com/twosigma/Cook/blob/5f44d7818f1a2b9132bf6cd11b7cf0586b2adb07/scheduler/src/cook/rest/api.clj#L1106

I would say we just leave it out of the python client.

jobclient/python/docs/source/api.rst
```diff
@@ -42,6 +42,33 @@ this may be overridden in the ``JobClient.submit`` method.
 .. autoclass:: Application
    :members:
 
+Containers
+^^^^^^^^^^
+
+By default, jobs submitted to Cook will use the ``python:3.6`` container, which
```
dposada, line 48: This actually depends on how Cook Scheduler is configured, so I don't think we should say this as a blanket statement.

integration/tests/cook/test_client.py
```diff
@@ -44,6 +46,7 @@ def test_query(self):
         self.assertAlmostEqual(job.cpus, 0.5)
         self.assertAlmostEqual(job.mem, 1.0)
         self.assertEqual(job.max_retries, 5)
+        self.client.kill(uuid)
```
dposada, line 51: If the assertion fails, this won't get called. Should we do this in the `finally` block of a `try`/`finally`?

integration/tests/cook/test_client.py
```diff
@@ -56,3 +59,37 @@ def test_kill(self):
         self.client.kill(uuid)
         job = self.client.query(uuid)
         self.assertEqual(job.status, JobStatus.COMPLETED)
+
+    def test_container_submit(self):
```
dposada, line 67: We probably want:

```suggestion
    @unittest.skipUnless(util.docker_tests_enabled(), "Requires a test docker image")
    def test_container_submit(self):
```

integration/tests/cook/test_client.py
```diff
@@ -56,3 +59,37 @@ def test_kill(self):
...
+        self.assertIsNotNone(container.image)
+        uuid = self.client.submit(command='ls',
+                                  container=DockerContainer(image))
+        job = self.client.query(uuid)
+
+        remote_container = job.container
+        self.assertEqual(remote_container['type'], 'DOCKER')
+        self.assertEqual(remote_container['docker']['image'], container.image)
+
+        self.client.kill(uuid)
```
dposada, line 77: `try`/`finally` here too?

integration/tests/cook/test_client.py
```diff
@@ -56,3 +59,37 @@ def test_kill(self):
...
+                                  max_retries=5)
+
+        def is_instance_available(job: Job):
+            any_instances = len(job.instances) > 0
+            if not any_instances:
+                util.logger.info(f"Job {job.uuid} has no instances yet.")
+            return any_instances
+            
+        util.wait_until(lambda: self.client.query(uuid),
+                        is_instance_available)
```
dposada, line 93: This can simplify to just:

```suggestion
        util.wait_for_instance(type(self).cook_url, uuid)
```

integration/tests/cook/test_client.py
hdrodz97: I'm a little confused here, cause I did put this decorator (otherwise the integration tests would fail) and it doesn't seem to be showing up on the Github UI here. But if you look [here](https://github.com/hdrodz97/Cook/blob/client-fixes/integration/tests/cook/test_client.py#L65-L77), the decorator is in fact showing up. Hopefully this is a snag on the Github UI's end and it'll clear out after the next round of commits.

hdrodz97: Okay it's showing up now, that was weird.

dposada:
Are we waiting on @rmanyari or should I go ahead and merge?

PR #1583, opened 2020-06-24
Fixes whitespace bug in .travis.yml

dposada:

- replacing tab characters with spaces

Travis builds are not working because they can't parse the yaml:

image

(https://travis-ci.org/github/twosigma/Cook/requests)


dposada:
Seems to be working!

image

PR #1585, opened 2020-06-25
Make separate matching chime per pool

nsinkov:

- split signal from match-trigger-chan per pool

Every pool will get its own chime to do matching instead of randomly getting to match by winning the race to listen on a single chime

### DIFF
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -899,10 +899,7 @@ class MultiUserCookTest(util.CookTest):
                            "EQUALS",
                            "won't get scheduled"]]
         group = {'uuid': str(util.make_temporal_uuid())}
-        job_spec = {'group': group['uuid'],
-                    'command': 'sleep 30',
-                    'cpus': util.max_cpus()}
-        uuids, resp = util.submit_jobs(self.cook_url, job_spec, clones=1, groups=[group], constraints=bad_constraints)
+        uuid, resp = util.submit_job(self.cook_url, command='sleep 30', group=group['uuid'], constraints=bad_constraints)
         self.assertEqual(201, resp.status_code, resp.content)
         try:
             default_pool = util.default_submit_pool() or util.default_pool(self.cook_url)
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_multi_user.py
```diff
@@ -895,11 +895,11 @@ def test_user_limits_change(self):
                         self.assertEqual(resp.status_code, 200, resp.text)
 
     def test_queue_endpoint(self):
+        bad_constraints = [["HOSTNAME",
```
scrosby, line 898: This change seems unrelated?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1554,6 +1555,7 @@
   (persist-mea-culpa-failure-limit! conn mea-culpa-failure-limit)
 
   (let [{:keys [match-trigger-chan rank-trigger-chan]} trigger-chans
+        match-trigger-chan-mult (async/mult match-trigger-chan)
```
scrosby, line 1558: Looks good. One question. Did you do a before-after comparison? 

Before, this log line (in match-jobs-event): should print roughly one second apart across each pool:
        (log/info "In" pool-name "pool, starting offer matching")

With the change, you should see them all start at the same time, within tens of milliseconds. This should be true even if some of them are really slow; they'll just skip the chime and get the next one. (E.g., during testing, make this pause a variable time. e.g., (+ 2 (% (hash pool) 10)) seconds and make sure that nothing seems to pause an excessive amount.

integration/tests/cook/test_multi_user.py
nsinkov: this test was flaky and the change seems to have exposed the flakiness

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: the logs looked as expected in terms of chiming at the same time

not sure how to test the very long scheduling time. open to suggestions

scrosby: Within the body of the function, put in a Thread/sleep with a delay calculated based on the pool name.

integration/tests/cook/test_multi_user.py
scrosby: Cook. Thanks for fixing.

PR #1586, opened 2020-06-25
Avoids autoscaling beyond what users have quota to run

dposada:

- filtering pending jobs based on quota before calling `trigger-autoscaling!`

So that we don't trigger autoscaling beyond what users have quota to actually run.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -901,11 +901,16 @@
                                @pool-name->pending-jobs-atom)
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
-                    matched-considerable-jobs-head?))]
+                    matched-considerable-jobs-head?))
+                ;; We need to filter pending jobs based on quota so that we don't
+                ;; trigger autoscaling beyond what users have quota to actually run
+                autoscalable-jobs (->> pool-name
+                                       (get @pool-name->pending-jobs-atom)
+                                       (tools/filter-based-on-quota user->quota user->usage))]
             ;; This call needs to happen *after* launch-matched-tasks!
             ;; in order to avoid autoscaling tasks taking up available
             ;; capacity that was already matched for real Cook tasks.
-            (trigger-autoscaling! (get @pool-name->pending-jobs-atom pool-name) pool-name compute-clusters)
+            (trigger-autoscaling! autoscalable-jobs pool-name compute-clusters)
             matched-head-or-no-matches?))
         (catch Throwable t
           (meters/mark! handle-resource-offer!-errors)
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -901,11 +901,16 @@
                                @pool-name->pending-jobs-atom)
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
-                    matched-considerable-jobs-head?))]
+                    matched-considerable-jobs-head?))
+                ;; We need to filter pending jobs based on quota so that we don't
+                ;; trigger autoscaling beyond what users have quota to actually run
+                autoscalable-jobs (->> pool-name
+                                       (get @pool-name->pending-jobs-atom)
+                                       (tools/filter-based-on-quota user->quota user->usage))]
```
scrosby, line 909: I think we should use considerable jobs, or refactor this code and that code to use the same filtering logic. That way both fenzo and autoscaler see rate limits, user quotas and pool quotas, etc. the same way. Oherwise, the two could become inconsistent. considerable queue indicates the true demand we have. 

Also, related to this, I think there should be a limit here of, say, 1000 jobs. Otherwise this code can be O(# pending jobs) or vastly more expensive in the match cycle. What if the queue gets to be 500k jobs?

dposada: This code matches what we're doing with `/queue`, which is how we trigger autoscaling in the Mesos pools:

https://github.com/twosigma/Cook/blob/da5f3038e43d3cf9737b16e49b6ac3acf2ee7251/scheduler/src/cook/rest/api.clj#L2132-L2139

I vote that for now, we make them consistent, and then consider improving both as a separate task if we think we need to.

As for the limit, it's already being limited inside each KCC via `max-outstanding-synthetic-pods`.

dposada: @scrosby Thoughts?

dposada: @scrosby Ping

scrosby: Sorry. Didn't see your response. 

I'd like to refactor the code so that there's a common helper function called by both pending-jobs->considerable-jobs the /queue processing and syntthetic pod generation that does quota filtering for queuing purposes. Now we've got three locations and no hints that they should be kept in sync.

Maybe '(defn filter-pending-jobs-for-autoscaling ...' that gets re-used across all three locations.

dposada: Thanks, will take a look.

scrosby: Your thoughts?

dposada: @scrosby I created `util/filter-pending-jobs-for-autoscaling` and switched `/queue` and the `trigger-autoscaling!` argument to use it. I don't think it makes sense to use it in `pending-jobs->considerable-jobs`, because that function is not specific to autoscaling, it's for deciding what jobs to send to Fenzo.

PR #1590, opened 2020-06-30
Refactor out common method for matching regexps against a pool name.

scrosby:

- Refactor some redundant code for matching pools against a list of regexps.

Remove redundant code that will be used a third time.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -893,6 +893,15 @@
     (cond-> {:count 1 :cpus cpus :mem mem}
       gpus (assoc :gpus gpus))))
 
+(defn match-based-on-pool-name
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -894,6 +894,15 @@
     (cond-> {:count 1 :cpus cpus :mem mem}
       gpus (assoc :gpus gpus))))
 
+(defn match-based-on-pool-name
```
nsinkov, line 897: add a unit test?

scrosby: There is one via (get-default-container-for-pool)

nsinkov: nitpick: I would make it separate still

scrosby: Ok. I added one. :)

PR #1591, opened 2020-07-01
Adds port mappings to k8s pod spec

dposada:

- adds ports from `port-mapping` to the k8s pod spec

To allow k8s-bound jobs to specify ports to open.


dposada:
This got merged into PR #1594.

PR #1592, opened 2020-07-01
Support config-based global quotas for a pool

scrosby:

- Support config-based global quotas for a pool

With cloud autoscaling, cook has no limits on autoscaling within cook, only in the underlying k8s cluster. To add more refined control, allow Cook to have specific autoscaling limits.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2629,10 +2629,45 @@ class CookTest(util.CookTest):
 
     @unittest.skipUnless(util.pool_quota_test_pool() is not None, 'Test requires a test pool.')
     def test_pool_quota(self):
-        job_count = 10
-        self.logger.info(f'Submitting {job_count} jobs')
+        settings = util.settings(self.cook_url)
+        quotas_map = settings.get("pools", {}).get("quotas", [])
+        pool_name = util.pool_quota_test_pool()
+        match_quota = [ii["quota"] for ii in quotas_map if
+                                   re.match(ii["pool-regex"], pool_name)]
+        # If there are no supported GPU models for pool, assert submission gets rejected
+        logging.info("Quota: " + repr(match_quota))
+        if(len(match_quota) == 0):
+            self.skipTest(f"Pool {pool_name} lacks quota assignment.")
+        quota = match_quota[0]
+        logging.info(f"Pool quota: {quota}")
+        if (quota["count"] >= 10):
+            self.skipTest(f"Job count quota too large for test")
+        quota_count = quota["count"]
+        job_count = quota_count + 2
+        job_mem = 16
+        job_cpus = .01
+
+        if quota["mem"] < job_mem * (job_count + 3):
+            self.skipTest("Quota memory too small for test")
+        if quota["cpus"] < job_cpus * (job_count + 3):
+            self.skipTest("Quota cpus to small for test")
+
+        # Now lookup the user quota and make sure it fits.
+        pools, resp = util.all_pools(self.cook_url)
+
+        user = self.determine_user()
+        resp = util.get_limit(self.cook_url, 'quota', user)
+        user_quota = resp.json()['pools'][pool_name]
+        logging.info(f"User quota {user_quota}")
+        if user_quota["count"] < job_count + 3:
+            self.skipTest("User quota count too small for test")
+        if user_quota["mem"] < job_mem * (job_count + 3):
+            self.skipTest("User quota memory too small for test")
+        if user_quota["cpus"] < job_cpus * (job_count + 3):
+            self.skipTest("User Quota cpus to small for test")
+
         sleep_command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
-        job_resources = {'cpus': 0.01, 'mem': 16}
+        job_resources = {'cpus': job_cpus, 'mem': job_mem}
         job_specs = util.minimal_jobs(job_count, command=sleep_command, **job_resources)
         job_uuids, resp = util.submit_jobs(self.cook_url, job_specs, pool=util.pool_quota_test_pool())
         self.assertEqual(resp.status_code, 201, resp.content)
@@ -2642,7 +2677,8 @@ class CookTest(util.CookTest):
 
             def predicate(resp):
                 jobs = resp.json()
-                logging.info("Job statuses", str([(job['uuid'], job['status']) for job in jobs]))
+                str([(job['uuid'], job['status']) for job in jobs])
+                logging.info("Job statuses: " + str([(job['uuid'], job['status']) for job in jobs]))
                 return len([job for job in jobs if job['status'] == 'running']) >= 2
 
             # Wait until at least 2 are running.
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2646,9 +2646,12 @@ class CookTest(util.CookTest):
         job_mem = 16
         job_cpus = .01
 
-        if quota["mem"] < job_mem * (job_count + 3):
+        total_cpus_requested = job_count * job_cpus
+        total_mem_requested = job_count * job_mem
+
+        if quota["mem"] < total_mem_requested:
             self.fail("Quota memory too small for test")
-        if quota["cpus"] < job_cpus * (job_count + 3):
+        if quota["cpus"] < total_cpus_requested:
             self.fail("Quota cpus to small for test")
 
         # Now lookup the user quota and make sure it fits and fail if otherwise.
@@ -2656,11 +2659,11 @@ class CookTest(util.CookTest):
         resp = util.get_limit(self.cook_url, 'quota', user)
         user_quota = resp.json()['pools'][pool_name]
         logging.info(f"User quota {user_quota}")
-        if user_quota["count"] < job_count + 3:
+        if user_quota["count"] < job_count:
             self.fail("User quota count too small for test")
-        if user_quota["mem"] < job_mem * (job_count + 3):
+        if user_quota["mem"] < total_mem_requested:
             self.fail("User quota memory too small for test")
-        if user_quota["cpus"] < job_cpus * (job_count + 3):
+        if user_quota["cpus"] < total_cpus_requested:
             self.fail("User Quota cpus to small for test")
 
         sleep_command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2626,45 +2626,42 @@ class CookTest(util.CookTest):
                 job_uuid, resp = util.submit_job(self.cook_url, pool=pool_name, mem=mem_over_quota)
                 self.assertEqual(422, resp.status_code, msg=resp.content)
 
-
     @unittest.skipUnless(util.pool_quota_test_pool() is not None, 'Test requires a test pool.')
     def test_pool_quota(self):
         settings = util.settings(self.cook_url)
         quotas_map = settings.get("pools", {}).get("quotas", [])
         pool_name = util.pool_quota_test_pool()
         match_quota = [ii["quota"] for ii in quotas_map if
-                                   re.match(ii["pool-regex"], pool_name)]
-        # If there are no supported GPU models for pool, assert submission gets rejected
+                       re.match(ii["pool-regex"], pool_name)]
+        # If the pool doesn't have the right quota constraints, fail the test.
         logging.info("Quota: " + repr(match_quota))
-        if(len(match_quota) == 0):
-            self.skipTest(f"Pool {pool_name} lacks quota assignment.")
+        if (len(match_quota) == 0):
+            self.fail(f"Pool {pool_name} lacks quota assignment.")
         quota = match_quota[0]
         logging.info(f"Pool quota: {quota}")
         if (quota["count"] >= 10):
-            self.skipTest(f"Job count quota too large for test")
+            self.fail(f"Job count quota too large for test")
         quota_count = quota["count"]
         job_count = quota_count + 2
         job_mem = 16
         job_cpus = .01
 
         if quota["mem"] < job_mem * (job_count + 3):
-            self.skipTest("Quota memory too small for test")
+            self.fail("Quota memory too small for test")
         if quota["cpus"] < job_cpus * (job_count + 3):
-            self.skipTest("Quota cpus to small for test")
+            self.fail("Quota cpus to small for test")
 
-        # Now lookup the user quota and make sure it fits.
-        pools, resp = util.all_pools(self.cook_url)
-
+        # Now lookup the user quota and make sure it fits and fail if otherwise.
         user = self.determine_user()
         resp = util.get_limit(self.cook_url, 'quota', user)
         user_quota = resp.json()['pools'][pool_name]
         logging.info(f"User quota {user_quota}")
         if user_quota["count"] < job_count + 3:
-            self.skipTest("User quota count too small for test")
+            self.fail("User quota count too small for test")
         if user_quota["mem"] < job_mem * (job_count + 3):
-            self.skipTest("User quota memory too small for test")
+            self.fail("User quota memory too small for test")
         if user_quota["cpus"] < job_cpus * (job_count + 3):
-            self.skipTest("User Quota cpus to small for test")
+            self.fail("User Quota cpus to small for test")
 
         sleep_command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
         job_resources = {'cpus': job_cpus, 'mem': job_mem}
@@ -2677,7 +2674,6 @@ class CookTest(util.CookTest):
 
             def predicate(resp):
                 jobs = resp.json()
-                str([(job['uuid'], job['status']) for job in jobs])
                 logging.info("Job statuses: " + str([(job['uuid'], job['status']) for job in jobs]))
                 return len([job for job in jobs if job['status'] == 'running']) >= 2
 
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2680,7 +2680,7 @@ class CookTest(util.CookTest):
             # Wait until at least 2 are running.
             util.wait_until(query, predicate)
             # Wait an extra 60 seconds to see if anything else starts.
-            time.sleep(60.0)
+            time.sleep(20.0)
             jobs = util.query_jobs(self.cook_url, True, uuid=job_uuids).json()
             running = [job for job in jobs if job['status'] == 'running']
             self.assertEqual(2, len(running), jobs)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2626,6 +2626,36 @@ class CookTest(util.CookTest):
                 job_uuid, resp = util.submit_job(self.cook_url, pool=pool_name, mem=mem_over_quota)
                 self.assertEqual(422, resp.status_code, msg=resp.content)
 
+
+    @unittest.skipUnless(util.pool_quota_test_pool() is not None, 'Test requires a test pool.')
+    def test_pool_quota(self):
+        job_count = 10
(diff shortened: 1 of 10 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -902,20 +902,39 @@
...
    max-jobs, max-cpus or max-mem"
   [user->quota user->usage queue]
   (letfn [(filter-with-quota [user->usage job]
             (let [user (:job/user job)
                   job-usage (job->usage job)
                   user->usage' (update-in user->usage [user] #(merge-with + job-usage %))]
-              (log/debug "Quota check" {:user user
+              (log/debug "User quota check" {:user user
                                         :usage (get user->usage' user)
                                         :quota (user->quota user)})
```
dposada, line 920: Indentation seems off here

scrosby: Thanks. Fixed.

scrosby:
> A couple of suggestions:
> 
>     1. Can we add the pool-level usage to the `/usage` endpoint so it's easy for us to answer the question "what's the total pool usage for pool X?"
> 
>     2. Can we add an integration test for this?

1: We already have metrics that export task count, memory and CPU's per pool which are currently dashboarded. Per pool usage is different enough from per-user usage that we should write a new endpoint for it. (e.g. permissions, should you show all pools or some pools, etc). 

2: I thought of that, but think its a bad idea. By design, this quota isn't dynamic, we can't set a small quota, run some tests, and resize to another quota. Then, we want our integration tests to be able to scale high to finish quickly. Any artificial quota just makes the tests run slower or take longer by more scale up.

dposada:
Sounds good on #1. We can always add the data to an endpoint later if we're missing it.

On #2, how about if we allow for specifying a specific pool to run the test jobs on, e.g. `COOK_TEST_QUOTA_POOL=test-pool-name`, and then when we run the tests, we can use a separate pool from where all of the other tests run?

integration/tests/cook/util.py
```diff
@@ -1773,6 +1773,8 @@ def has_one_agent():
 def supports_exit_code():
     return using_kubernetes() or is_cook_executor_in_use()
 
+def pool_quota_test_pool():
+    return os.getenv("COOK_TEST_POOL_QUOTA_TEST_POOL", None)
```
dposada, line 1777: Instead of requiring the environment variable, how about if we just test in any active pools where the global quota is set to something that's below a certain threshold and also less than our user quota?

integration/tests/cook/test_basic.py
```diff
@@ -2610,6 +2610,35 @@ def test_pool_specific_quota_check_on_submit(self):
...
+                jobs = resp.json()
+                return len([job for job in jobs if job['status'] == 'running']) > 2
+
+            # Wait until at least 2 are running.
+            util.wait_until(query, predicate)
+            # Wait an extra 60 seconds to see if anything else starts.
+            time.sleep(60.0)
+            jobs = util.query_jobs(self.cook_url, True, uuid=job_uuids).json()
+            running = [job for job in jobs if job['status'] == 'running']
+            self.assertLessEqual(2, len(running), jobs)
```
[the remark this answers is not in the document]
scrosby: I'll make the change, but I can't imagine a user quota that's only a couple of jobs.

integration/tests/cook/util.py
scrosby: I can do that. Tradeoff is that it'll be much easier to accidentally disable this test. Tweak a setting and the test silently starts skipping itself.

dposada: Maybe fail if the number of candidate pools is 0?

scrosby: This whole idea of getting the pool isn't robust. I think we want to exclude the quota test pool from the all_pools so that other dozen tests that use all_pools won't run on it (and potentially break). This is a special pool that we should keep other tests away from, or do we want to have a 'all_pools_and_i_really_mean_all_pools()' function and use that? (IMO: No).

dposada: Ah, valid point, sounds good

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
                 self.assertEqual(422, resp.status_code, msg=resp.content)
 
+
+    @unittest.skipUnless(util.pool_quota_test_pool() is not None, 'Test requires a test pool.')
+    def test_pool_quota(self):
+        settings = util.settings(self.cook_url)
+        quotas_map = settings.get("pools", {}).get("quotas", [])
+        pool_name = util.pool_quota_test_pool()
+        match_quota = [ii["quota"] for ii in quotas_map if
+                                   re.match(ii["pool-regex"], pool_name)]
```
dposada, line 2636: Is indentation off on this line?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        # If there are no supported GPU models for pool, assert submission gets rejected
```
dposada, line 2637: Comment seems wrong

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        logging.info("Quota: " + repr(match_quota))
+        if(len(match_quota) == 0):
```
dposada, line 2639: Did you run through python code formatter?

```suggestion
        if (len(match_quota) == 0):
```

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+            self.skipTest(f"Pool {pool_name} lacks quota assignment.")
```
dposada, line 2640: Should we fail instead of skip here?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        quota = match_quota[0]
+        logging.info(f"Pool quota: {quota}")
+        if (quota["count"] >= 10):
+            self.skipTest(f"Job count quota too large for test")
```
dposada, line 2644: Should we fail instead of skip here?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        quota_count = quota["count"]
+        job_count = quota_count + 2
+        job_mem = 16
+        job_cpus = .01
+
+        if quota["mem"] < job_mem * (job_count + 3):
+            self.skipTest("Quota memory too small for test")
+        if quota["cpus"] < job_cpus * (job_count + 3):
+            self.skipTest("Quota cpus to small for test")
```
dposada, line 2653: Should we fail instead of skip here?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        user = self.determine_user()
+        resp = util.get_limit(self.cook_url, 'quota', user)
+        user_quota = resp.json()['pools'][pool_name]
+        logging.info(f"User quota {user_quota}")
+        if user_quota["count"] < job_count + 3:
+            self.skipTest("User quota count too small for test")
+        if user_quota["mem"] < job_mem * (job_count + 3):
+            self.skipTest("User quota memory too small for test")
+        if user_quota["cpus"] < job_cpus * (job_count + 3):
+            self.skipTest("User Quota cpus to small for test")
```
dposada, line 2667: Should we fail instead of skip here?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        job_specs = util.minimal_jobs(job_count, command=sleep_command, **job_resources)
+        job_uuids, resp = util.submit_jobs(self.cook_url, job_specs, pool=util.pool_quota_test_pool())
+        self.assertEqual(resp.status_code, 201, resp.content)
+        try:
+            def query():
+                return util.query_jobs(self.cook_url, True, uuid=job_uuids)
+
+            def predicate(resp):
+                jobs = resp.json()
+                str([(job['uuid'], job['status']) for job in jobs])
```
dposada, line 2680: What does this do?

scheduler/src/cook/tools.clj
```diff
@@ -916,27 +916,50 @@
...
 (defn filter-pending-jobs-for-autoscaling
   "Lazily filters jobs to those that should be considered
   for autoscaling purposes. Note that this is used in two
   places: the /queue endpoint (Mesos only) and as an
   argument to scheduler/trigger-autoscaling! (k8s only)."
-  [user->quota user->usage queue]
-  (filter-based-on-quota user->quota user->usage queue))
+  [user->quota user->usage pool-quota queue]
+  (let [pool-usage (reduce (partial merge-with +) (vals user->usage))]
+  (->> queue
```
dposada, line 960: Indentation seems off here

scheduler/src/cook/tools.clj
```diff
@@ -916,27 +916,50 @@
...
+  (log/debug "Pool quota and usage:" {:quota quota :usage usage})
+  (if (nil? quota)
+    queue
+    (letfn [(filter-with-quota [usage job]
+              (let [job-usage (job->usage job)
+                    usage' (merge-with + job-usage usage)]
+                (log/debug "Pool quota check" {:usage usage'
+                                               :quota quota})
+                [usage' (below-quota? quota usage')]))]
+      (filter-sequential filter-with-quota usage queue))))
```
dposada, line 951: It's hard to tell what the shapes of `quota` and `usage` are. Is it `pool -> {:cpus cpus :mem mem}`? Can we make the names more descriptive (and maybe describe the shapes in the docstring)?

scheduler/src/cook/tools.clj
```diff
@@ -916,27 +916,50 @@
...
+
 (defn filter-pending-jobs-for-autoscaling
   "Lazily filters jobs to those that should be considered
   for autoscaling purposes. Note that this is used in two
   places: the /queue endpoint (Mesos only) and as an
   argument to scheduler/trigger-autoscaling! (k8s only)."
-  [user->quota user->usage queue]
-  (filter-based-on-quota user->quota user->usage queue))
+  [user->quota user->usage pool-quota queue]
+  (let [pool-usage (reduce (partial merge-with +) (vals user->usage))]
```
dposada, line 959: Same comment as above about it being hard to tell what the shape of `pool-quota` and `pool-usage` are.

scheduler/src/cook/tools.clj
```diff
@@ -916,27 +916,50 @@
        first
        field))
 
-(defn filter-based-on-quota
+(defn get-quota-for-pool
```
dposada, line 919: Maybe `global-pool-quota`?

scheduler/src/cook/tools.clj
```diff
@@ -916,27 +916,50 @@
...
+  "Given a pool name, determine the quota for that pool."
+  [quotas effective-pool-name]
+  (match-based-on-pool-name quotas effective-pool-name :quota))
```
dposada, line 922: Why `effective-pool-name` instead of just `pool-name`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -653,7 +654,8 @@
             (not (and is-rate-limited? enforcing-job-launch-rate-limit?))))
         considerable-jobs
         (->> pending-jobs
-             (tools/filter-based-on-quota user->quota user->usage)
+             ;(tools/filter-based-on-pool-quota (tools/get-quota-for-pool (config/pool-quotas) pool-name))
```
dposada, line 657: Delete?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -653,7 +654,8 @@
...
+             (tools/filter-pending-jobs-for-autoscaling user->quota user->usage (tools/get-quota-for-pool (config/pool-quotas) pool-name))
```
dposada, line 658: It seems odd to me to use a function named `for-autoscaling` here, when this has nothing to do with autoscaling. Maybe the function this calls should just be named `filter-based-on-quota`, and that would do both the user-level and the global filtering?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,72 @@ def test_pool_specific_quota_check_on_submit(self):
...
+                logging.info("Job statuses: " + str([(job['uuid'], job['status']) for job in jobs]))
+                return len([job for job in jobs if job['status'] == 'running']) >= 2
+
+            # Wait until at least 2 are running.
+            util.wait_until(query, predicate)
+            # Wait an extra 60 seconds to see if anything else starts.
+            time.sleep(60.0)
```
dposada, line 2687: Maybe instead of waiting 60 seconds, we could submit a job to the default submit pool and wait for it to get scheduled? That way, we know another matching round has happened and the over-global-quota job has been filtered out.

scheduler/src/cook/tools.clj
```diff
@@ -906,7 +906,7 @@
   (let [{:keys [cpus gpus mem]} (job-ent->resources job-ent)]
     (cond-> {:count 1 :cpus cpus :mem mem}
       gpus (assoc :gpus gpus))))
-
+
```
dposada, line 909: ?

integration/tests/cook/test_basic.py
scrosby: Yes. Fixed.

integration/tests/cook/test_basic.py
scrosby: That and another. Fixed.

integration/tests/cook/test_basic.py
scrosby: No, but I will and have.

integration/tests/cook/test_basic.py
scrosby: Ok.

integration/tests/cook/test_basic.py
scrosby: Ok.

integration/tests/cook/test_basic.py
scrosby: Ok.

integration/tests/cook/test_basic.py
scrosby: Yeah. Ok.

integration/tests/cook/test_basic.py
scrosby: Leftover debug line from when str() wasn't working right. (You reviewed this before I had my usual cleanup to get it ready for review. It works externally, but isn't working internally. I'm actually running that test as we speak.)

scheduler/src/cook/tools.clj
scrosby: Yes. Fixed.

scheduler/src/cook/tools.clj
scrosby: Docstring updated. Not sure what would be more descriptive names. I'm happy to rename if you have a proposal.

scheduler/src/cook/tools.clj
scrosby: Done.

scheduler/src/cook/tools.clj
scrosby: Ok.

scheduler/src/cook/tools.clj
scrosby: Because it includes the mapping from nil to "no-pool".

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Already in my branch but not pushed.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Works for me.

integration/tests/cook/test_basic.py
scrosby: There's no guarantee of schedule order. Pools will schedule concurrently especially with the new chime.

scheduler/src/cook/tools.clj
scrosby: Not sure where it came from. Deleted.

integration/tests/cook/test_basic.py
dposada: Makes sense, 20 seconds seems like a good idea.

scheduler/src/cook/tools.clj
```diff
@@ -950,16 +953,21 @@
                 [usage' (below-quota? quota usage')]))]
       (filter-sequential filter-with-quota usage queue))))
 
-(defn filter-pending-jobs-for-autoscaling
+(defn filter-pending-jobs-for-quota
   "Lazily filters jobs to those that should be considered
   for autoscaling purposes. Note that this is used in two
   places: the /queue endpoint (Mesos only) and as an
-  argument to scheduler/trigger-autoscaling! (k8s only)."
+  argument to scheduler/trigger-autoscaling! (k8s only).
```
dposada: This is used in more than 2 places now, right?

integration/tests/cook/test_basic.py
```diff
@@ -2626,6 +2626,68 @@ def test_pool_specific_quota_check_on_submit(self):
...
+        # Now lookup the user quota and make sure it fits and fail if otherwise.
+        user = self.determine_user()
+        resp = util.get_limit(self.cook_url, 'quota', user)
+        user_quota = resp.json()['pools'][pool_name]
+        logging.info(f"User quota {user_quota}")
+        if user_quota["count"] < job_count + 3:
+            self.fail("User quota count too small for test")
+        if user_quota["mem"] < job_mem * (job_count + 3):
+            self.fail("User quota memory too small for test")
+        if user_quota["cpus"] < job_cpus * (job_count + 3):
```
dposada, line 2663: Why `job_count + 3`?

scheduler/src/cook/tools.clj
dposada: Isn't that true for almost every variable named `pool-name` in the codebase?

scheduler/src/cook/tools.clj
dposada: Docstring is helpful, thanks

integration/tests/cook/test_basic.py
scrosby: Ok.

scheduler/src/cook/tools.clj
scrosby: No. Some of them are the actual pool name based on the user. Some are with nil mapped to "no-pool" (and I hate that inconsistency).

integration/tests/cook/test_basic.py
scrosby: These checks were put in because you wanted the test to be self-protective against a future hypothetical environments where there was a very constrained user quota.

Not knowing exactly what hypothetical environment you were thinking of, I put  in a fudge factor of to handle roundoff or other accounting irregularities, but I'd be happy to put in any other number you think is appropriate.

scheduler/src/cook/tools.clj
scrosby: Comment fixed.

integration/tests/cook/test_basic.py
dposada: Why wouldn't it just be `job_count`? 

Maybe define a couple of variables:

```
total_cpus_requested = job_count * job_cpus
total_mem_requested = job_count * mem
```

And then make sure that the `user_quota` values are >= those? 

The point is to make sure that the test is passing specifically because of the global quota and not because of the user's quota. If we don't have these checks, this test could pass without any of your code changes.

scrosby: Making these changes.

PR #1596, opened 2020-07-08
add frozen version of requirements.txt that will be more stable

nsinkov:

- add frozen version of requirements.txt

frozen version of requirements.txt that will be more stable when building. A change to a package in pypi can cause an install to fail


integration/requirements.txt
```diff
@@ -1,15 +1,24 @@
+# NOTE: make sure to update the requirements-frozen.txt file when making a change here
+# Sort this file lexicographically
+# Generate the requirements-frozen.txt file as follows:
+#  pip install -r requirements.txt
+#  pip freeze > requirements-frozen.txt
```
dposada, line 5: Does this need to be in a "clean" environment, like a fresh virtualenv where I haven't already pip installed other stuff not needed for the integration tests?

integration/requirements-frozen.txt
```diff
@@ -0,0 +1,110 @@
...
+ufw==0.36
+urllib3==1.22
+virtualenv==20.0.25
+wcwidth==0.2.5
+webencodings==0.5.1
+widgetsnbextension==3.5.1
+zipp==3.1.0
+zope.interface==4.3.2
+file:../cli#egg=cook_client
+pip==9.0.1; python_version >= '3.6'
```
dposada, line 110: Should we use this file in places where we currently use `requirements.txt`? e.g. the Vagrantfile, Dockerfiles, etc.?

integration/requirements.txt
nsinkov: it doesn't really matter. if you get a green run it's good.

nsinkov: The key thing is that every dependency is frozen

integration/requirements-frozen.txt
nsinkov: updated

integration/requirements.txt
dposada: If I have pip installed a bunch of extra garbage that the tests don't need, they will get included in this file, right? That seems wrong.

nsinkov: I don't think it matters much if the tests run, but I'll say to do it from Vagrant

nsinkov: updated the instructions. you were right - I was able to pull things in that at least Travis couldn't install. that's a problem

PR #1602, opened 2020-07-11
match pools in round-robin order

nsinkov:

- match pools in round-robin order
- add configurable match interval

right now pools match all at the same time on every matching tick. this causes a thundering herd problem

### DIFF
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -483,7 +483,11 @@
                              :pod-condition-unschedulable-seconds 60
                              :reconnect-delay-ms 60000
                              :set-container-cpu-limit? true}
-                            kubernetes)))}))
+                            kubernetes)))
+     :matching-settings (fnk [[:config {matching-settings {}}]]
+                          (merge {:per-pool-match-interval-millis 3000
+                                  :global-min-match-interval-millis 100}
+                                 matching-settings))}))
 
 (defn read-config
   "Given a config file path, reads the config and returns the map"
@@ -614,3 +618,7 @@
 (defn task-constraints
   []
   (get-in config [:settings :task-constraints]))
+
+(defn matching-settings
+  []
+  (-> config :settings :matching-settings))
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -485,8 +485,8 @@
                              :set-container-cpu-limit? true}
                             kubernetes)))
      :offer-matching (fnk [[:config {offer-matching {}}]]
-                          (merge {:per-pool-match-interval-millis 3000
-                                  :global-min-match-interval-millis 100}
+                          (merge {:global-min-match-interval-millis 100
+                                  :target-per-pool-match-interval-millis 3000}
                                  offer-matching))}))
 
 (defn read-config
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -484,10 +484,10 @@
                              :reconnect-delay-ms 60000
                              :set-container-cpu-limit? true}
                             kubernetes)))
-     :matching-settings (fnk [[:config {matching-settings {}}]]
+     :offer-matching (fnk [[:config {offer-matching {}}]]
                           (merge {:per-pool-match-interval-millis 3000
                                   :global-min-match-interval-millis 100}
-                                 matching-settings))}))
+                                 offer-matching))}))
 
 (defn read-config
   "Given a config file path, reads the config and returns the map"
@@ -619,6 +619,6 @@
   []
   (get-in config [:settings :task-constraints]))
 
-(defn matching-settings
+(defn offer-matching
   []
-  (-> config :settings :matching-settings))
+  (-> config :settings :offer-matching))
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1568,6 +1568,10 @@
                                   (/ (count pools)) 
                                   int 
                                   (max global-min-match-interval-millis))]
+    (when (= match-interval-millis global-min-match-interval-millis)
+      (log/warn "match-trigger-chan is set to the global minimum interval of " global-min-match-interval-millis " ms. "
+                "This is a sign that we have more pools, " (count pools) " than we expect to have and we will "
+                "schedule each pool less often than the desired setting of every " per-pool-match-interval-millis " ms."))
     (async/pipe (chime-ch (tools/time-seq (time/now) (time/millis match-interval-millis))) match-trigger-chan)))
 
 (defn create-datomic-scheduler
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1583,34 +1583,23 @@
                  [{:pool/name "no-pool"}])
         pool-name->fenzo (pool-map pools' (fn [_] (make-fenzo-scheduler offer-incubate-time-ms
                                                                         fenzo-fitness-calculator good-enough-fitness)))
-        pool->match-trigger-chan (reduce
-                                   (fn [m pool-ent]
-                                     (let [pool-name (:pool/name pool-ent)]
-                                       (assoc m pool-name (async/chan (async/sliding-buffer 1)))))
-                                   {}
-                                   pools')
+        pool->match-trigger-chan (pool-map pools' (fn [_] (async/chan (async/sliding-buffer 1))))
         pool-names-array-list (ArrayList. (map :pool/name pools'))
         pool-name-iterator-atom (atom (.iterator pool-names-array-list))
-        pool->resources-atom
-        (reduce (fn [m pool-ent]
-                  (let [pool-name (:pool/name pool-ent)
-                        fenzo (pool-name->fenzo pool-name)
-                        resources-atom
-                        (make-offer-handler
-                          conn fenzo pool-name->pending-jobs-atom agent-attributes-cache fenzo-max-jobs-considered
-                          fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset
-                          (get pool->match-trigger-chan pool-name)
-                          rebalancer-reservation-atom mesos-run-as-user pool-name compute-clusters)]
-                    (assoc m pool-name resources-atom)))
-                {}
-                pools')]
+        pool->resources-atom (pool-map
+                               pools'
+                               (fn [{:keys [pool/name]}]
+                                 (make-offer-handler
+                                   conn (pool-name->fenzo name) pool-name->pending-jobs-atom agent-attributes-cache
+                                   fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn
+                                   fenzo-floor-iterations-before-reset (get pool->match-trigger-chan name)
+                                   rebalancer-reservation-atom mesos-run-as-user name compute-clusters)))]
     (prepare-match-trigger-chan match-trigger-chan pools')
-    (tools/chime-at-ch
-      match-trigger-chan
-      (fn process-match-chime []
+    (async/go-loop []
+      (when-let [x (async/<! match-trigger-chan)]
         (let [iterator (swap! pool-name-iterator-atom #(if (.hasNext %) % (.iterator pool-names-array-list)))]
-          (async/offer! (get pool->match-trigger-chan (.next iterator)) :trigger)))
-      {:error-handler (fn [ex] (log/error ex "Error forwarding match-trigger-chan to pool"))})
+          (async/offer! (get pool->match-trigger-chan (.next iterator)) x))
+        (recur)))
     (log/info "Pool name to fenzo scheduler map:" pool-name->fenzo)
     (start-jobs-prioritizer! conn pool-name->pending-jobs-atom task-constraints rank-trigger-chan)
     {:pool-name->fenzo pool-name->fenzo
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -54,7 +54,8 @@
   (:import (com.netflix.fenzo
              TaskAssignmentResult TaskRequest TaskScheduler TaskScheduler$Builder VirtualMachineCurrentState
              VirtualMachineLease VirtualMachineLease$Range)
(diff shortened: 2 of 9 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1569,27 +1570,39 @@
...
                  [{:pool/name "no-pool"}])
         pool-name->fenzo (pool-map pools' (fn [_] (make-fenzo-scheduler offer-incubate-time-ms
                                                                         fenzo-fitness-calculator good-enough-fitness)))
-        {:keys [pool->resources-atom]}
+        pool->match-trigger-chan (reduce
+                                   (fn [m pool-ent]
+                                     (let [pool-name (:pool/name pool-ent)]
+                                       (assoc m pool-name (async/chan (async/sliding-buffer 1)))))
+                                   {}
+                                   pools')
```
dposada, line 1584: Can this use the `pool-map` function?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1569,27 +1570,39 @@
...
                           conn fenzo pool-name->pending-jobs-atom agent-attributes-cache fenzo-max-jobs-considered
                           fenzo-scaleback fenzo-floor-iterations-before-warn fenzo-floor-iterations-before-reset
-                          (async/tap match-trigger-chan-mult (async/chan (async/sliding-buffer 1)))
+                          (get pool->match-trigger-chan pool-name)
                           rebalancer-reservation-atom mesos-run-as-user pool-name compute-clusters)]
-                    (-> m
-                        (assoc-in [:pool->resources-atom pool-name] resources-atom))))
+                    (assoc m pool-name resources-atom)))
                 {}
                 pools')]
```
dposada, line 1599: Can this use the `pool-map` function?

nsinkov: yes

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: yes

scheduler/src/cook/config.clj
```diff
@@ -483,7 +483,11 @@
                              :pod-condition-unschedulable-seconds 60
                              :reconnect-delay-ms 60000
                              :set-container-cpu-limit? true}
-                            kubernetes)))}))
+                            kubernetes)))
+     :matching-settings (fnk [[:config {matching-settings {}}]]
```
dposada, line 487: Maybe just `:matching` since we're already under `:settings`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1560,6 +1560,13 @@
                (counters/dec! in-order-queue-counter)
                (body-fn))))))
 
+(defn prepare-match-trigger-chan
+  "Calculate the required interval for match-trigger-chan and start the chimes"
+  [match-trigger-chan pools]
+  (let [{:keys [per-pool-match-interval-millis global-min-match-interval-millis]} (config/matching-settings)
+        match-interval-millis (max (int (/ per-pool-match-interval-millis (count pools))) global-min-match-interval-millis)]
```
dposada, line 1567: Thoughts on using `->` here?

```suggestion
        match-interval-millis (-> per-pool-match-interval-millis 
                                  (/ (count pools)) 
                                  int 
                                  (max global-min-match-interval-millis))]
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1576,33 +1583,23 @@
...
+                                   fenzo-floor-iterations-before-reset (get pool->match-trigger-chan name)
+                                   rebalancer-reservation-atom mesos-run-as-user name compute-clusters)))]
+    (prepare-match-trigger-chan match-trigger-chan pools')
+    (async/go-loop []
+      (when-let [x (async/<! match-trigger-chan)]
         (let [iterator (swap! pool-name-iterator-atom #(if (.hasNext %) % (.iterator pool-names-array-list)))]
-          (async/offer! (get pool->match-trigger-chan (.next iterator)) :trigger)))
-      {:error-handler (fn [ex] (log/error ex "Error forwarding match-trigger-chan to pool"))})
+          (async/offer! (get pool->match-trigger-chan (.next iterator)) x))
+        (recur)))
```
dposada: Should we have a `try`/`catch` somewhere in here?

scheduler/src/cook/scheduler/scheduler.clj
dposada: Seems like we should log a warning-level message if `(= match-interval-millis global-min-match-interval-millis)` -- basically "heads up, you're being throttled".

scheduler/src/cook/test/testutil.clj
```diff
@@ -392,7 +392,8 @@
      (loop []
        (when-not (pred)
          (if (< (.getTime (java.util.Date.)) (+ start-ms max-wait-ms))
-           (recur)
+           (do (java.lang.Thread/sleep interval-ms)
+               (recur))
```
dposada, line 396: Interesting, this has just been broken for a long time?

scheduler/test/cook/test/zz_simulator.clj
```diff
@@ -87,7 +88,7 @@
                                :retry-limit 5})
 
 (defmacro with-cook-scheduler
-  [conn make-mesos-driver-fn scheduler-config & body]
+  [conn make-mesos-driver-fn scheduler-config trigger-matching & body]
```
dposada, line 91: Maybe `trigger-matching?`

scheduler/src/cook/config.clj
nsinkov: seems a bit too short. but I don't have a strong preference

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: I would have expected it to bubble up. is it not the case for async go loop?

scheduler/src/cook/test/testutil.clj
nsinkov: apparently. my code caused it to fail

scheduler/test/cook/test/zz_simulator.clj
nsinkov: ok

scheduler/src/cook/config.clj
```diff
@@ -613,4 +617,8 @@
 
 (defn task-constraints
   []
-  (get-in config [:settings :task-constraints]))
\ No newline at end of file
+  (get-in config [:settings :task-constraints]))
+
+(defn matching-settings
```
scrosby, line 622: Does this fit in this file not at the end? (If everyone puts new config entries at the end, you guarantee merge conflicts.) E.g., after estimated-completion-config?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1559,6 +1560,20 @@
...
 
+(defn prepare-match-trigger-chan
+  "Calculate the required interval for match-trigger-chan and start the chimes"
+  [match-trigger-chan pools]
+  (let [{:keys [per-pool-match-interval-millis global-min-match-interval-millis]} (config/matching-settings)
+        match-interval-millis (-> per-pool-match-interval-millis 
+                                  (/ (count pools)) 
+                                  int 
+                                  (max global-min-match-interval-millis))]
+    (when (= match-interval-millis global-min-match-interval-millis)
```
scrosby, line 1571: I would remove the (max global-min-match-interval-millis). The target is having each pool visited at least as often as per-pool-match-interval-millis. The impact of match-interva-millis being too small (i.e., less than global-min-match-intetrval-millis) is increase of CPU (and the fallout of such). The cost of match-interval-millis * (count pools) too high is we have stale offers. The (when ...) then warns when match-interval-millis is less than **global-min-match-interval-millis-warn-threshold**. (I renamed it to indicate its purpose.)

(Also, related, if you don't agree with the logic above. We cannnot simutaneously have global-min-match-interval-millis and per-pool-match-interval-millis. We should call it per-pool-match-interval-millis-target.)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1569,27 +1584,29 @@
...
+                               (fn [{:keys [pool/name]}]
+                                 (make-offer-handler
+                                   conn (pool-name->fenzo name) pool-name->pending-jobs-atom agent-attributes-cache
+                                   fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn
+                                   fenzo-floor-iterations-before-reset (get pool->match-trigger-chan name)
+                                   rebalancer-reservation-atom mesos-run-as-user name compute-clusters)))]
+    (prepare-match-trigger-chan match-trigger-chan pools')
+    (async/go-loop []
+      (when-let [x (async/<! match-trigger-chan)]
+        (let [iterator (swap! pool-name-iterator-atom #(if (.hasNext %) % (.iterator pool-names-array-list)))]
```
scrosby, line 1607: Simpler than this, why not just have a java.util.Queue (implemented by LinkedList?) Make a helper function that remove the head and immediately add() it on to the end and returns the element. Just call that function to get the element. No (atom) or (swap!).

scheduler/src/cook/config.clj
nsinkov: don't we just need a newline at the end of the file?

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: I'm ok calling it per-pool-match-interval-millis-target but leaving the max

The max warns, but also stops CPU usage from going up if for some reason we didn't or couldn't act on the warning right away. Protects against adding a large number of pools at once.

scheduler/src/cook/config.clj
dposada: What's the downside to short?

scheduler/src/cook/scheduler/scheduler.clj
dposada: Not sure, but if it bubbles up, then what happens?

scheduler/src/cook/config.clj
scrosby: THe idea is to reduce merge conflicts. If I add to the end and you add to the end, there's often going to be a merge conflict.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: With the tradeoff of breaking mesos if too many pools get added at once. Which is more important: Saving CPU or not breaking mesos?

scheduler/src/cook/config.clj
nsinkov: how about offer-matching?

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: not sure. I think we'd want everything to die if this fails. I'll add a try catch log and rethrow

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: If we don't save CPU we might break everything, including mesos. Mesos loses either way

scheduler/src/cook/scheduler/scheduler.clj
nsinkov: thanks

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1604,8 +1603,13 @@
...
     (async/go-loop []
       (when-let [x (async/<! match-trigger-chan)]
-        (let [iterator (swap! pool-name-iterator-atom #(if (.hasNext %) % (.iterator pool-names-array-list)))]
-          (async/offer! (get pool->match-trigger-chan (.next iterator)) x))
+        (try
+          (let [pool-name (.removeFirst pool-names-linked-list)
+                _ (.addLast pool-names-linked-list pool-name)]
+            (async/offer! (get pool->match-trigger-chan pool-name) x))
+          (catch Exception e
+            (log/error e "Exception in match-trigger-chan chime handler ")
```
dposada, line 1611:
```suggestion
            (log/error e "Exception in match-trigger-chan chime handler")
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1604,8 +1603,13 @@
     (prepare-match-trigger-chan match-trigger-chan pools')
     (async/go-loop []
       (when-let [x (async/<! match-trigger-chan)]
-        (let [iterator (swap! pool-name-iterator-atom #(if (.hasNext %) % (.iterator pool-names-array-list)))]
-          (async/offer! (get pool->match-trigger-chan (.next iterator)) x))
+        (try
+          (let [pool-name (.removeFirst pool-names-linked-list)
+                _ (.addLast pool-names-linked-list pool-name)]
```
dposada, line 1608: Why not move this line to after the bindings?

nsinkov: updated

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Fair. Only probably a concern if we get to 100+ pools and we're still using mesos.

But can you please rename it to per-pool-match-interval-millis-target  because its a target. We may miss it.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1559,6 +1560,20 @@
...
+(defn prepare-match-trigger-chan
+  "Calculate the required interval for match-trigger-chan and start the chimes"
+  [match-trigger-chan pools]
+  (let [{:keys [per-pool-match-interval-millis global-min-match-interval-millis]} (config/offer-matching)
+        match-interval-millis (-> per-pool-match-interval-millis 
+                                  (/ (count pools)) 
+                                  int 
+                                  (max global-min-match-interval-millis))]
+    (when (= match-interval-millis global-min-match-interval-millis)
+      (log/warn "match-trigger-chan is set to the global minimum interval of " global-min-match-interval-millis " ms. "
```
scrosby, line 1572: Like this!

nsinkov: hmm, I thought I renamed already. renaming.

PR #1603, opened 2020-07-11
Retries failed pod submissions

dposada:

Running `process` for jobs that Cook wants to start and that don't yet have a corresponding pod in k8s.

Without this, jobs that have failed pod submissions for a transient reason (e.g. a timeout when talking to a mutating admission controller) will get a job instance in the Unknown state forever, and will not continue trying to do the pod submission.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -583,22 +583,31 @@
 (defn synthesize-state-and-process-pod-if-changed
   "Synthesizes the k8s-actual-state for the given
   pod and calls process if the state changed."
-  [{:keys [k8s-actual-state-map name] :as compute-cluster} pod-name ^V1Pod pod]
-  (let [new-state {:pod pod
-                   :synthesized-state (api/pod->synthesized-pod-state pod)
+  [{:keys [cook-expected-state-map k8s-actual-state-map name] :as compute-cluster} pod-name ^V1Pod pod]
+  (let [synthesized-pod-state (api/pod->synthesized-pod-state pod)
+        new-state {:pod pod
+                   :synthesized-state synthesized-pod-state
                    :sandbox-file-server-container-state (api/pod->sandbox-file-server-container-state pod)}
-        old-state (get @k8s-actual-state-map pod-name)]
+        old-state (get @k8s-actual-state-map pod-name)
+        pod-state-changed? (not (k8s-actual-state-equivalent? old-state new-state))
+        unsubmitted-pod? (and (nil? synthesized-pod-state)
+                              (= (get-in @cook-expected-state-map [pod-name :cook-expected-state])
+                                 :cook-expected-state/starting))
+        should-process? (or pod-state-changed? unsubmitted-pod?)]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -583,22 +583,31 @@
...
-                   :synthesized-state (api/pod->synthesized-pod-state pod)
+  [{:keys [cook-expected-state-map k8s-actual-state-map name] :as compute-cluster} pod-name ^V1Pod pod]
+  (let [synthesized-pod-state (api/pod->synthesized-pod-state pod)
+        new-state {:pod pod
+                   :synthesized-state synthesized-pod-state
                    :sandbox-file-server-container-state (api/pod->sandbox-file-server-container-state pod)}
-        old-state (get @k8s-actual-state-map pod-name)]
+        old-state (get @k8s-actual-state-map pod-name)
+        pod-state-changed? (not (k8s-actual-state-equivalent? old-state new-state))
+        unsubmitted-pod? (and (nil? synthesized-pod-state)
```
scrosby, line 593: This seems off to me. If the expected state is missing, the task is in the state (:starting :missing). That's the same as an unlaunched task whose launch hasn't even been attempted. The state scanner (controller/scan-tasks) will retry on the next scan.

dposada: The expected state is `:starting`, not missing.

The state scanner is not retrying, that's what this PR is fixing (`controller/scan-tasks` ends up calling this function).

scrosby: Whats the existing state when this bug strikes?

dposada: The Cook expected state is `:cook-expected-state :cook-expected-state/starting` and the k8s actual state is `:synthesized-state {:state :missing}` (`nil`).

dposada: @scrosby Ping

scrosby: The retry should happen automatically without this patch because the state scanner should visit every pair of states. 

Retry and initial submission are the same codepath: 
   https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/kubernetes/controller.clj#L534

The state scanner scans every state every few minutes:
   https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/kubernetes/compute_cluster.clj#L138

Ah. The bug is here where we disabled the state scanner: https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/kubernetes/controller.clj#L670 this effectively disables retries for everything. We should visit every state pair unconditionally every 15 minutes to catch these bugs. 

(I believe we're currently configured to 2 minutes, but with larger scale, we should increase this to 10 until we implement rate limits.)

scrosby: I expect this bug to have other symptoms, e.g., a failed delete pod won't be retried.

dposada:
Closing this in favor of PR #1614

PR #1605, opened 2020-07-14
Fix NPE when we ask a k8s pool for offers when it has none

scrosby:

- Fix NPE when we ask a k8s pool for offers when it has none

Avoid spam in the log.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -247,10 +247,15 @@
                     :pod-namespace pod-namespace
                     :task-metadata task-metadata})))))
 
+(defn get-name->node-for-pool
+  "Given a pool->node-name->node and pool, return a node-name->pool, for any pool"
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -335,7 +335,7 @@
     (let [timer (timers/start (metrics/timer "cc-pending-offers-compute" name))
           pods (add-starting-pods this @all-pods-atom)
           nodes @current-nodes-atom
-          offers-this-pool (generate-offers this (@pool->node-name->node pool-name)
+          offers-this-pool (generate-offers this (get @pool->node-name->node pool-name {})
```
dposada, line 338: Would it make sense to extract a function for this since we're doing it in two places?

scrosby: I can do that.

scrosby:
Test added.

PR #1606, opened 2020-07-15
Fix "unable to get node from node name XXXX" errors.

scrosby:

- Fix bug where we try to calculate consumption for pods that don't have a node.
- Do some more filtering to keep wrong-pool nodes from being injected into consumption processing.

Avoid thousands of "unable to get node from node name" ERROR logged.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -56,7 +56,11 @@
   [compute-cluster node-name->node node-name->pods]
   (let [compute-cluster-name (cc/compute-cluster-name compute-cluster)
         node-name->capacity (api/get-capacity node-name->node)
-        node-name->consumed (api/get-consumption node-name->pods)
+        ; Note that node-name->pods may include extra pods that are from other pools. So, we filter out only
+        ; nodes that have a capacity.
+        node-name->consumed (->> (api/get-consumption node-name->pods)
+                                 (filter #(node-name->capacity (first %)))
+                                 (into {}))
         node-name->available (tools/deep-merge-with - node-name->capacity node-name->consumed)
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -56,14 +56,19 @@
   [compute-cluster node-name->node node-name->pods]
   (let [compute-cluster-name (cc/compute-cluster-name compute-cluster)
         node-name->capacity (api/get-capacity node-name->node)
-        node-name->consumed (api/get-consumption node-name->pods)
+        ; Note that node-name->pods may include extra pods that are from other pools. So, we filter out only
+        ; nodes that have a capacity.
+        node-name->consumed (->> (api/get-consumption node-name->pods)
+                                 (filter #(node-name->capacity (first %)))
+                                 (into {}))
```
dposada, line 63: Why does filtering to only nodes that have capacity fix the pods from other pools problem? I don't understand the connection. Is there a way to make the comment more clear? Also, should the comment say "filter in" instead of "filter out"?

dposada: Can we write a unit test for this?

scrosby: Done. there should be no functionality change with this. The filter path is now tested with the nodeD change --- we have an off-node pod, which shouldn't (and doesn't) alter the offer generation.

dposada: Did you see the first comment in this thread?

> Why does filtering to only nodes that have capacity fix the pods from other pools problem? I don't understand the connection. Is there a way to make the comment more clear? Also, should the comment say "filter in" instead of "filter out"?

scrosby: Comment elaborated.

PR #1607, opened 2020-07-15
Adds support for EQUALS constraints in k8s

dposada:

- adding two new node pools (e2-small) in `help-make-cluster`
- adding support for default job constraints by pool (regex)
- translating EQUALS constraints to node selectors
- mapping all k8s node labels to offer attributes so they can be referenced by constraints

We want to allow users to experiment with different machine types in k8s, and this change allows users to use the existing job constraint mechanism to specify (for example) the machine type they want to use. Note that there's nothing specific here to machine types -- this allows any k8s node label to be selected.

### DIFF
--- a/scheduler/bin/help-make-cluster
+++ b/scheduler/bin/help-make-cluster
@@ -18,9 +18,25 @@ CLUSTERNAME=$3
 COOK_KUBECONFIG=$4
 GKE_CLUSTER_OWNER=${GKE_CLUSTER_OWNER:-$USER}
 
-VERSION=1.15.11-gke.12
+VERSION=1.15.11-gke.15
 
 gcloud="gcloud --project $PROJECT"
+
+function create_node_pool {
+  local cook_pool=$1
+  local machine_type=$2
+  echo "---- Creating $machine_type node pool for Cook pool $cook_pool"
+  $gcloud container node-pools create "cook-pool-$cook_pool-$machine_type" \
+         --zone "$ZONE" \
+         --cluster="$CLUSTERNAME" \
+         --disk-size=20gb \
+         --machine-type="$machine_type" \
+         --node-taints="cook-pool=$cook_pool:NoSchedule" \
+         --enable-autoscaling \
+         --min-nodes=0 \
+         --max-nodes=6 \
+         --node-labels="cook-pool=$cook_pool,test-label=true,node-type=$machine_type"
+}
 
 echo "---- Building kubernetes cluster for project=$PROJECT zone=$ZONE clustername=$CLUSTERNAME kubeconfig=$COOK_KUBECONFIG"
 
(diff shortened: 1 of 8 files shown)


scheduler/config-k8s.edn
```diff
@@ -91,8 +91,15 @@
...
                                                      :force-pull-image false
                                                      :parameters []}}}]
-        :valid-gpu-models
-        [{:pool-regex "k8s-alpha" :valid-models #{"nvidia-tesla-k80" "nvidia-tesla-p100"} :default-model "nvidia-tesla-p100"}]}
+         :valid-gpu-models
+         [{:pool-regex "k8s-alpha"
+           :valid-models #{"nvidia-tesla-k80" "nvidia-tesla-p100"}
+           :default-model "nvidia-tesla-p100"}]
+         :default-job-constraints
+         [{:pool-regex "k8s-.*"
```
scrosby, line 99: Should be "^k8s-.*"

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -929,6 +929,22 @@
       ; synthetic pods (which don't specify a node name), but it
       ; doesn't hurt to add it for all pods we submit.
       (add-node-selector pod-spec cook-pool-label pool-name))
+
+    (when pod-constraints
+      (doseq [{:keys [constraint/attribute
+                      constraint/operator
+                      constraint/pattern]}
+              pod-constraints]
+        (condp = operator
```
scrosby, line 938: Shouldn't we be rejecting bad constraints at submit time instead of here, where they'll be invisibly ignored? Also, do we need/want any nil protection here?

scheduler/bin/help-make-cluster
```diff
@@ -48,37 +64,11 @@ KUBECONFIG=${COOK_KUBECONFIG} kubectl create -f "${DIR}/priority-class-synthetic
...
-       --cluster="$CLUSTERNAME" \
-       --disk-size=20gb \
-       --machine-type=g1-small \
-       --node-taints=cook-pool=k8s-quota:NoSchedule \
-       --enable-autoscaling \
-       --min-nodes=0 \
-       --max-nodes=6 \
-       --node-labels=cook-pool=k8s-quota,test-label=true
-
+create_node_pool k8s-alpha g1-small
```
scrosby, line 67: Like this!

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -177,11 +177,57 @@
...
+  [job]
+  (util/match-based-on-pool-name
+    (config/default-job-constraints)
+    (util/job->pool-name job)
+    :default-constraints))
+
+(def machine-type-constraint-attributes
+  #{"cpu-architecture" "node-family" "node-type"})
+
+(defn job->constraints
```
scrosby, line 191: I think we should de-lazy this output; dump into a vector or list. Fenzo can call from multiple threads.

scheduler/config-k8s.edn
dposada: Good call, will do, thanks

scheduler/src/cook/kubernetes/api.clj
dposada: Yep, will take a look

scheduler/src/cook/scheduler/constraints.clj
dposada: Will do, thanks

scheduler/config-k8s.edn
dposada: Done

scheduler/src/cook/kubernetes/api.clj
dposada: The bad constraint operators are already rejected at submission time:

```
$ echo '{"command": "true", "constraints": [["foo", "BOGUS", "bar"]]}' | cs submit --raw
Enter the raw job(s) JSON (press Ctrl+D on a blank line to submit)
Attempting to submit on dev0 cluster...
Job submission failed on dev0:
{"jobs": [{"constraints": [[null, "(named (not (constraint-operator-exists? \"BOGUS\")) \"operator\")", null]]}]}

Job submission failed on all of your configured clusters.
```

dposada: Changed the log line to a `throw` so a bad constraint operator (which is already blocked at submission time) would be more visible -- the pod would not get submitted.

dposada: `null` attributes and patterns are also blocked at submission time:

```
$ echo '{"command": "true", "constraints": [[null, "EQUALS", null]]}' | cs submit --raw                                                                                                                                                          
Enter the raw job(s) JSON (press Ctrl+D on a blank line to submit)
Attempting to submit on dev0 cluster...
Job submission failed on dev0:
{"jobs": [{"constraints": [["(named (not (instance? java.lang.String nil)) \"attribute\")", null, "(named (not (instance? java.lang.String nil)) \"pattern\")"]]}]}

Job submission failed on all of your configured clusters.
```

scheduler/src/cook/scheduler/constraints.clj
dposada: Done

scheduler/src/cook/kubernetes/api.clj
scrosby: Looks good to me, but I'd put in the throw string (or an adjacent comment) that these invalid situations should have been rejected at submit time.

dposada: Good call, added the comment below

```
          ; A bad constraint operator should have
          ; been blocked at job submission time
```

PR #1611, opened 2020-07-21
One GPU job per GPU node

kathryn-zhou:

- Change the GPU constraint to only allow one GPU job on each GPU node.
- Require that the number of GPUs requested by the job is equal to the number of GPUs available on the node.

- Each GPU job requires a SSD, and Google currently only allows one SSD per node.


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -788,7 +788,8 @@
     (when (pos? gpus)
       (.putLimitsItem resources "nvidia.com/gpu" (double->quantity gpus))
       (.putRequestsItem resources "nvidia.com/gpu" (double->quantity gpus))
-      (add-node-selector pod-spec "cloud.google.com/gke-accelerator" gpu-model-requested))
+      (add-node-selector pod-spec "cloud.google.com/gke-accelerator" gpu-model-requested)
+      (add-node-selector pod-spec "gpu-count" gpus))
```
dposada, line 792: Might be nice to have an explanatory comment above this line

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -108,14 +108,14 @@
   (job-constraint-evaluate
     [{:keys [job]} _ vm-attributes _]
     (let [k8s-vm? (= (get vm-attributes "compute-cluster-type") "kubernetes")
-          job-gpu-count-requested (-> job util/job-ent->resources :gpus (or 0))]
+          job-gpu-count-requested (-> job util/job-ent->resources :gpus (or 0) int)]
```
dposada, line 111: Might also be nice to explain why we need the `int` here

scheduler/src/cook/kubernetes/api.clj
kathryn-zhou: added

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: add

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -108,14 +108,16 @@
...
           (if k8s-vm?
             (let [job-gpu-model-requested (job->gpu-model-requested
                                             job-gpu-count-requested job (util/job->pool-name job))
                   vm-gpu-model->count-available (get vm-attributes "gpus")
                   vm-satisfies-constraint? (if (pos? job-gpu-count-requested)
                                              ; If job requests GPUs, require that the VM has enough gpus available in the same model as the job requested.
-                                             (>= (get vm-gpu-model->count-available job-gpu-model-requested 0) job-gpu-count-requested)
+                                             ; cast to int because the API allows both float and int gpu values
+                                             (= (int (get vm-gpu-model->count-available job-gpu-model-requested 0)) job-gpu-count-requested)
```
nsinkov, line 120: is it better style to use (== ...) ?

nsinkov:
Will you ever get a job that wants a number of GPUs that doesn't match the number of GPUs on any node, or is that not possible?

scheduler/src/cook/kubernetes/api.clj
dposada: I don't see the comment

kathryn-zhou:
@nsinkov We are going to add in some validation for that at the job submission level. So users can only request 1, 2, 4, and 8 GPUs and if they do otherwise then the job submission will fail.

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: yes thank you! didn't know that existed but that is definitely cleaner and I modified this

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -115,7 +115,7 @@
                   vm-gpu-model->count-available (get vm-attributes "gpus")
                   vm-satisfies-constraint? (if (pos? job-gpu-count-requested)
                                              ; If job requests GPUs, require that the VM has enough gpus available in the same model as the job requested.
```
dposada, line 117:
```suggestion
                                             ; If job requests GPUs, require that the VM has the same number of gpus available in the same model as the job requested.
```

dposada:
Will merge when green

PR #1614, opened 2020-07-22
Forces processing when state scanning in k8s

dposada:

- adding overload to `synthesize-state-and-process-pod-if-changed` that takes in a `force-process?` arg
- when `force-process?` is true, calling `process` regardless of changes
- calling it with `force-process?` set to true from the state scanner

Otherwise, we're not retrying certain states like failed pod submissions.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -582,23 +582,27 @@
 
 (defn synthesize-state-and-process-pod-if-changed
   "Synthesizes the k8s-actual-state for the given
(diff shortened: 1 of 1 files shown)


scrosby:
If nothing has changed then most of the code in synthesize-state-and-process-pod-if-changed is a noop. Why not call process directly (within a try-catch), unless datomic fails and we want to retry inserting the sandbox state.

dposada:
@scrosby We need to synthesize state and store the updated state in the state scanning path (because there are certain transitions that are time-dependent, like container initialization timing out), so I don't think most of the code is a no-op.

PR #1618, opened 2020-07-24
One GPU task per VM

kathryn-zhou:

- Modify GPU host constraint to only allow one GPU task per VM
- Added unit test for this special case

- Brian Bao noticed that when he submitted multiple GPU jobs rapidly in succession, they sometimes got matched to the same VM because the GPU constraint was not accounting for tasks that were already assigned to that VM.

### DIFF
--- a/scheduler/test/cook/test/scheduler/constraints.clj
+++ b/scheduler/test/cook/test/scheduler/constraints.clj
@@ -155,7 +155,7 @@
                               (getTasksCurrentlyAssigned [_] [gpu-task-assignment])
                               (getCurrAvailableResources [_] (sched/->VirtualMachineLeaseAdapter k8s-gpu-offer 0)))
                             nil)))
-          (str "GPU task on GPU host with the correct number of GPUs but already has task assigned should fail"))
+          (str "We only allow one GPU job per VM. This VM already has a GPU job assigned to it, so it should not get any additional GPU tasks."))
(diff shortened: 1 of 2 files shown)


scheduler/test/cook/test/scheduler/constraints.clj
```diff
@@ -141,6 +146,16 @@
...
+      (is (not (.isSuccessful
+                 (.evaluate (constraints/fenzoize-job-constraint (constraints/build-gpu-host-constraint gpu-job-4))
+                            (sched/make-task-request db gpu-job-4 nil)
+                            (reify com.netflix.fenzo.VirtualMachineCurrentState
+                              (getHostname [_] "test-host")
+                              (getRunningTasks [_] [])
+                              (getTasksCurrentlyAssigned [_] [gpu-task-assignment])
+                              (getCurrAvailableResources [_] (sched/->VirtualMachineLeaseAdapter k8s-gpu-offer 0)))
+                            nil)))
+          (str "GPU task on GPU host with the correct number of GPUs but already has task assigned should fail"))
```
scrosby, line 158: What about "We only allow one GPU job per VM. This VM already has a GPU job, so shouldn't get another"?

kathryn-zhou: thanks that makes the comment way less confusing, updated to something pretty similar.

PR #1621, opened 2020-07-27
Runs only one pool in test_list_jobs_by_pool

dposada:

- running the test for 1 pool instead of 2

This is our 4th longest running non-`xfail` test over the last 30 days, taking an average of ~125 seconds, and there's no reason it should take this long.


nsinkov:
so why do we care that it takes 125 seconds? is this in anticipation that this will become the slowest test?

dposada:
@nsinkov The test can be about twice as fast by only running against 1 pool. Faster tests are generally better, and here we get a faster test without losing anything.

nsinkov:
couple questions:
was there some good reason it was set to use two pools originally?
is the new runtime half or the original?

PR #1623, opened 2020-07-28
Uses 1024*1024 as the k8s memory multiplier

dposada:

- using 1024^2 (MiB) as the k8s memory multiplier instead of 1000^2 (MB)

Cook, Fenzo, and Mesos use mebibytes (MiB) for memory, so we need to convert bytes from k8s to MiB when passing to Fenzo, and MiB back to bytes when submitting to k8s.

### DIFF
--- a/integration/tests/cook/test_cli.py
+++ b/integration/tests/cook/test_cli.py
@@ -499,7 +499,8 @@ if __name__ == '__main__':
         waiting_uuid = uuids[0]
 
         # Submit a long-running job
-        cp, uuids = cli.submit('sleep 300', self.cook_url, submit_flags='--name %s' % name)
+        sleep_command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
+        cp, uuids = cli.submit(sleep_command, self.cook_url, submit_flags='--name %s' % name)
         self.assertEqual(0, cp.returncode, cp.stderr)
         running_uuid = uuids[0]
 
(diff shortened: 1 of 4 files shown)


integration/tests/cook/test_cli.py
```diff
@@ -499,7 +499,8 @@ def test_list_by_state(self):
         waiting_uuid = uuids[0]
 
         # Submit a long-running job
-        cp, uuids = cli.submit('sleep 300', self.cook_url, submit_flags='--name %s' % name)
+        sleep_command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
+        cp, uuids = cli.submit(sleep_command, self.cook_url, submit_flags='--name %s' % name)
```
dposada, line 503: This test fix is unrelated to the rest of the PR, but the test failed on this branch, so I fixed it here.

integration/tests/cook/test_client.py
```diff
@@ -115,6 +115,7 @@ def test_container_port_submit(self):
         uuid = self.client.submit(command=f'{hostname_progress_cmd} && nc -l -p {JOB_PORT} $(hostname -I)',
                                   container=container,
                                   env={progress_file_env: 'progress.txt'},
+                                  max_retries=5,
```
dposada, line 118: This test fix is unrelated to the rest of the PR, but the test failed on this branch, so I added this here.

PR #1628, opened 2020-07-31
handle preemption on pod initialization

nsinkov:

- check for preemption label when pod goes missing during initialization

If a pod is preempted during initialization we currently show it as "Command exited non-zero" but we need to show it as  "Agent removed"

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -530,7 +530,9 @@
                                                    (do
                                                      (log/info "In compute cluster" name ", pod" pod-name
                                                                "was deleted while it was expected starting")
-                                                     (handle-pod-killed compute-cluster pod-name))
+                                                     (if (some-> synthesized-state :pod-preempted-timestamp)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -530,7 +530,9 @@
                                                    (do
                                                      (log/info "In compute cluster" name ", pod" pod-name
                                                                "was deleted while it was expected starting")
-                                                     (handle-pod-killed compute-cluster pod-name))
+                                                     (if (some-> synthesized-state :pod-preempted-timestamp)
+                                                       (handle-pod-preemption compute-cluster pod-name)
+                                                       (handle-pod-externally-deleted compute-cluster pod-name)))
```
dposada, line 535: Since we have the same three lines in another part of the machine, should we extract a function, e.g. `handle-pod-missing-unexpectedly`?

nsinkov: updated

dposada:
Will merge when green

PR #1629, opened 2020-08-02
Adds compute-cluster-type and url to test metrics

dposada:

- adding `compute-cluster-type` and `url` to the test metrics entries

It's useful to distinguish Mesos from Kubernetes test runs, and runs against different URLs.

### DIFF
--- a/integration/tests/conftest.py
+++ b/integration/tests/conftest.py
@@ -79,7 +79,9 @@ if 'TEST_METRICS_URL' in os.environ:
                 'build-id': os.getenv('TEST_METRICS_BUILD_ID', None),
                 'result': result,
                 'runtime-milliseconds': (end - start) * 1000,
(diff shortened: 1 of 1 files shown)


integration/tests/conftest.py
```diff
@@ -79,7 +79,9 @@ def record_test_metric(request):
                 'build-id': os.getenv('TEST_METRICS_BUILD_ID', None),
                 'result': result,
                 'runtime-milliseconds': (end - start) * 1000,
-                'expected-to-fail': expected_to_fail
+                'expected-to-fail': expected_to_fail,
+                'compute-cluster-type': os.getenv('TEST_METRICS_COMPUTE_CLUSTER_TYPE', None),
+                'url': os.getenv('COOK_SCHEDULER_URL', None)
```
shamsimam, line 84: Nitpick: lexicographic ordering

PR #1631, opened 2020-08-04
Allow explicit `None` in JobClient.submit_all

hdrodz97:

- Fix the Python client so that submitting a job with `None` explicitly set on certain parameters no longer causes a `NoneError`

The following code snippet:

```python
client = JobClient(COOK_URL)
client.submit_all([{'command': 'ls', 'container': None}])
```

will incorrectly cause a `NoneError`. This is because since the `container` key is explicitly set in the jobspec, the client erroneously assumes it is not `None` and attempts to call `jobspec['container'].to_dict()`. This patch fixes this issue.

### DIFF
--- a/jobclient/python/cookclient/__init__.py
+++ b/jobclient/python/cookclient/__init__.py
@@ -378,7 +378,8 @@ class JobClient:
 
         This function will convert the higher-level Python types used in job
         submissions into their JSON primitive counterparts (e.g., timedelta is
-        converted into the number of milliseconds).
+        converted into the number of milliseconds). Additionally, this function
+        will also remove all keys with a value of ``None``.
 
         The provided jobspec will not be touched and a copy will be returned.
         """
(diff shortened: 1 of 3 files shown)


jobclient/python/cookclient/__init__.py
```diff
@@ -356,19 +356,19 @@ def _apply_jobspec_defaults(jobspec: dict) -> dict:
...
+        if jobspec.get('max-retries', None) is None:
             jobspec['max-retries'] = 1
-        if 'max-runtime' not in jobspec:
+        if jobspec.get('max-runtime', None) is None:
             jobspec['max-runtime'] = timedelta(days=1)
-        if 'name' not in jobspec:
+        if jobspec.get('name', None) is None:
             jobspec['name'] = f'{getpass.getuser()}-job'
-        if 'application' not in jobspec:
+        if jobspec.get('application', None) is None:
```
dposada, line 371: Maybe we should define a function, e.g. `is_field_set` or similar, that takes in a dictionary and field and returns `d.get(field, None) is not None` and use that for all of these? I think you could use it below as well.

PR #1634, opened 2020-08-07
Reverse order of quota application

scrosby:

- Reverse order of quota application

Applying quotas in the wrong order can also cause us to not launch
jobs that are within quota. A user who's already at their quota will
appear to falsely use global quota that they can't really use.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -963,8 +963,8 @@
   ; Use the already precomputed user->usage map and just aggregate by users to get pool usage.
   (let [pool-usage (reduce (partial merge-with +) (vals user->usage))]
     (->> queue
-         (filter-based-on-pool-quota pool-quota pool-usage)
-         (filter-based-on-user-quota user->quota user->usage))))
+         (filter-based-on-user-quota user->quota user->usage)
+         (filter-based-on-pool-quota pool-quota pool-usage))))
 
 (defn pool->user->usage
   "Returns a map from pool name to user name to usage for all users in all pools."
--- a/scheduler/test/cook/test/tools.clj
+++ b/scheduler/test/cook/test/tools.clj
@@ -791,6 +791,27 @@
       (is (= [(make-job 1 2 2048) (make-job 2 1 1024)]
              (util/filter-based-on-user-quota {test-user {:count 4, :cpus 20, :mem 6144}} user->usage queue))))))
 
+(deftest test-filter-pending-jobs-for-quota
+  (let [test-user-1 "john"
+        test-user-2 "bob"
+        user->usage {test-user-1 {:count 1, :cpus 1, :mem 1} test-user-2 {:count 1, :cpus 1, :mem 1}}
+        pool-quota {:count 4, :cpus 100, :mem 100}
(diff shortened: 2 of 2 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -963,8 +963,8 @@
   ; Use the already precomputed user->usage map and just aggregate by users to get pool usage.
   (let [pool-usage (reduce (partial merge-with +) (vals user->usage))]
     (->> queue
-         (filter-based-on-pool-quota pool-quota pool-usage)
-         (filter-based-on-user-quota user->quota user->usage))))
+         (filter-based-on-user-quota user->quota user->usage)
+         (filter-based-on-pool-quota pool-quota pool-usage))))
```
dposada, line 967: Could we add a unit or integration test?

scrosby: Done. Currently running through a just-in-case jenkins run.

scheduler/test/cook/test/tools.clj
```diff
@@ -791,6 +791,27 @@
...
 
+(deftest test-filter-pending-jobs-for-quota
+  (let [test-user-1 "john"
+        test-user-2 "bob"
+        user->usage {test-user-1 {:count 1, :cpus 1, :mem 1} test-user-2 {:count 1, :cpus 1, :mem 1}}
+        pool-quota {:count 4, :cpus 100, :mem 100}
+        job-user (atom 0)
+        make-job (fn [user]
+                   (swap! job-user inc)
+                   {:db/id @job-user
```
dposada, line 802: Do we actually need the `:db/id` for this test? Seems like we could get rid of it and the `job-user` `atom`.

scheduler/test/cook/test/tools.clj
```diff
@@ -791,6 +791,27 @@
...
+                    :job/user user
+                    :job/resource [{:resource/type :cpus, :resource/amount 1}
+                                   {:resource/type :mem, :resource/amount 1}]})
+        queue [(make-job test-user-1) (make-job test-user-1) (make-job test-user-1) (make-job test-user-2)]
+        user->quota {test-user-1 {:count 2, :cpus 100, :mem 100} test-user-2 {:count 2, :cpus 100, :mem 100}}
+        [job-1 job-2 job-3 job-4] queue]
+    ; If we filtered pool quota first, only the first two jobs in the
+    ; queue would be seen by the user quota and we'd only launch job-1.
```
dposada, line 810: I don't understand this comment -- each of the 4 jobs in the queue is asking for 1 cpu and 1 mem, and the pool quota is 4 jobs / 100 cpus / 100 mem. So why would the pool quota filter to only the first 2 jobs?

scrosby: There's two jobs already running (user->usage)

scheduler/test/cook/test/tools.clj
scrosby: Nope. Good find.

scheduler/test/cook/test/tools.clj
dposada: ah, thanks!

PR #1638, opened 2020-08-12
Adds API for compute clusters

dposada:

- adding API endpoint for creating and query compute clusters

This is part of a larger effort to support dynamic (k8s) compute clusters in Cook.


dposada:
This has been absorbed into #1639

PR #1641, opened 2020-08-14
Add checkpoint field to jobclient

nsinkov:

- Add checkpoint job field to jobclient


dposada:
Will merge when green

PR #1645, opened 2020-08-17
Un-lazy list of hosts a job has run on as a perf optimization

scrosby:

- Un-lazy the lost of hosts a job has already run on as a perf optimization.

When diagnosing a production problem, I noticed a lot of of threads blocked on unlazy'ing this list.

### DIFF
--- a/scheduler/src/cook/scheduler/constraints.clj
+++ b/scheduler/src/cook/scheduler/constraints.clj
@@ -81,7 +81,8 @@
   (->> (:job/instance job)
        (remove #(true? (:instance/preempted? %)))
        (mapv :instance/hostname)
-       distinct))
+       distinct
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -81,7 +81,8 @@
   (->> (:job/instance job)
        (remove #(true? (:instance/preempted? %)))
        (mapv :instance/hostname)
```
dposada, line 83: I thought `mapv` is not lazy?

sradack: `distinct` is lazy https://clojuredocs.org/clojure.core/distinct

PR #1648, opened 2020-08-18
Use set instead of distinct

sradack:

Use `set` instead of `mapv`->`distinct`

After discussion, using a `set` will reduce lock contention when reading already realized `lazy-seq`


shamsimam:
https://github.com/twosigma/Cook/blob/ee129b9b079dab1776f2bf5036520d121f83443b/scheduler/src/cook/scheduler/constraints.clj#L71-L72 can now be improved to use `contains?` if `previous-hosts` is known to be a set.

PR #1649, opened 2020-08-18
De-lazy the list of constraints to avoid locking in Fenzo.

scrosby:

- Unlazy another argument to Fenzo.

Avoid and reduce lock contention and make nicer profiles.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -526,9 +526,10 @@
                                         (constraints/make-fenzo-group-constraint
                                           db group #(guuid->considerable-cotask-ids (:group/uuid group)) running-cotask-cache))
                                       (:group/_job job)))))
+        constraints-list (into (list) constraints)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -526,9 +526,10 @@
                                         (constraints/make-fenzo-group-constraint
                                           db group #(guuid->considerable-cotask-ids (:group/uuid group)) running-cotask-cache))
                                       (:group/_job job)))))
```
sradack, line 528: Nitpick: Could have just threaded this in the threading macro.

sradack (approved):

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Didn't see any idiomatic way to do so ---- macro uses ->, not ->>, so the list would be in the wrong spot.

PR #1650, opened 2020-08-20
Adds HOST_IP environment variable for k8s

dposada:

- adding `HOST_IP` as an environment variable to k8s containers

So that jobs can access services running on the host, e.g. on a DaemonSet.


integration/tests/cook/test_cli.py
```diff
@@ -429,7 +429,7 @@ def test_show_running_job(self):
         self.assertEqual(0, cp.returncode, cp.stderr)
 
     def test_quoting(self):
-        cp, uuids = cli.submit('echo "Hello; exit 1"', self.cook_url)
+        cp, uuids = cli.submit('echo "Hello; exit 1"', self.cook_url, submit_flags=f'--max-retries 5')
```
dposada, line 432: This change is unrelated to the rest of the PR, but the test failed on this branch, so I fixed this here.

PR #1662, opened 2020-09-04
Fix errors when using dynamic clusters

nsinkov:

- fix pool-name->fenzo map
- remove sandbox url error for dynamic clusters

These errors were being logged but were not affecting jobs and were not noticed initially

A malformed pool-name->fenzo map was being passed in to write-status-to-datomic

We were attempting to get output_url from non-leader cluster which had no references to dynamic clusters. But we don't actually need a reference to a kubernetes cluster for output_url

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -419,7 +419,7 @@
                             {:template template})))
         full-cluster-config (-> (:config cluster-definition-template) (merge config) (assoc :dynamic-cluster-config? true))
         cluster (resolved full-cluster-config {:exit-code-syncer-state @exit-code-syncer-state-atom})]
-    (initialize-cluster cluster {:pool-name->fenzo @pool-name->fenzo-atom})))
+    (initialize-cluster cluster @pool-name->fenzo-atom)))
 
 (defn execute-update!
   "Attempt to execute a valid cluster configuration update.
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -146,7 +146,9 @@
   [compute-cluster-name]
   (let [result (@cluster-name->compute-cluster-atom compute-cluster-name)]
     (when-not result
-      (log/error "Was asked to lookup db-id for" compute-cluster-name "and got nil"))
+      (try (throw (ex-info "Exception to show stack trace" {}))
+           (catch Exception e
+             (log/error e "Was asked to lookup db-id for" compute-cluster-name "and got nil"))))
     result))
 
 (defn get-default-cluster-for-legacy
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -146,9 +146,7 @@
   [compute-cluster-name]
   (let [result (@cluster-name->compute-cluster-atom compute-cluster-name)]
     (when-not result
-      (try (throw (ex-info "Exception to show stack trace" {}))
-           (catch Exception e
-             (log/error e "Was asked to lookup db-id for" compute-cluster-name "and got nil"))))
+      (log/error (ex-info (str "Was asked to lookup db-id for " compute-cluster-name " and got nil") {})))
     result))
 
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -146,7 +146,7 @@
   [compute-cluster-name]
   (let [result (@cluster-name->compute-cluster-atom compute-cluster-name)]
     (when-not result
-      (log/error "Was asked to lookup db-id for" compute-cluster-name "and got nil"))
+      (log/error (ex-info (str "Was asked to lookup db-id for " compute-cluster-name " and got nil") {})))
```
dposada, line 149: Why `ex-info`?

scheduler/src/cook/compute_cluster.clj
```diff
@@ -419,7 +419,7 @@
                             {:template template})))
         full-cluster-config (-> (:config cluster-definition-template) (merge config) (assoc :dynamic-cluster-config? true))
         cluster (resolved full-cluster-config {:exit-code-syncer-state @exit-code-syncer-state-atom})]
-    (initialize-cluster cluster {:pool-name->fenzo @pool-name->fenzo-atom})))
```
dposada, line 422: How was this ever working?

scheduler/src/cook/rest/api.clj
```diff
@@ -1062,9 +1062,8 @@
   [instance-entity]
   (if-let [sandbox-url (:instance/sandbox-url instance-entity)]
     sandbox-url
-    (if-let [compute-cluster (task/task-ent->ComputeCluster instance-entity)]
-      (cc/retrieve-sandbox-url-path compute-cluster instance-entity)
-      (log/error "Unable to get sandbox URL for" instance-entity "whose compute cluster resolves to nil"))))
```
dposada, line 1067: Why are we getting rid of this error log?

scheduler/src/cook/task.clj
```diff
@@ -47,6 +47,13 @@
           task-entity->compute-cluster-name
           cc/compute-cluster-name->ComputeCluster))
 
+(defn get-ComputeCluster-for-task-ent-if-present
```
dposada, line 50: I would favor `get-compute-cluster-for-task-ent-if-present`.

dposada (commented):
Was there any functional issues? Anything we should write an integration test for?

scheduler/src/cook/compute_cluster.clj
nsinkov: I don't expect any calls to this. If we do get a call, the ex-info will show the stack and help diagnose

scheduler/src/cook/compute_cluster.clj
nsinkov: always failed to remove from fenzo when jobs completed. but jobs ran normally otherwise (this is when using dynamic clusters only)

scheduler/src/cook/rest/api.clj
nsinkov: this is no longer an error. We don't expect to find dynamic clusters on a Cook follower. The logic above, [sandbox-url (:instance/sandbox-url instance-entity)], already does what a k8s cluster would do.

scheduler/src/cook/task.clj
nsinkov: updated

scheduler/src/cook/compute_cluster.clj
dposada: Would it be better to actually `throw`?

dposada:
@nsinkov Were there any functional issues? Anything we should write an integration test for?

scheduler/src/cook/compute_cluster.clj
nsinkov: it doesn't throw now, so least disruptive would be to not throw

nsinkov:
I'm going to finish the integration tests you started our current tests catch the issues being fixed here when dynamic clusters are turned on

dposada (approved):

PR #1666, opened 2020-09-08
Deletes unschedulable synthetic pods

dposada:

- adding a new configuration field `:kubernetes` `:synthetic-pod-condition-unschedulable-seconds`, which defaults to 900 seconds (15 minutes)
- changing `pod-unschedulable?` to use the same logic for synthetic pods as for non-synthetic pods, just with a different timeout

So that unscheduled synthetic pods don't stay around forever.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -650,8 +650,7 @@ class CookTest(util.CookTest):
         command = f'{line_1} && sleep 2 && {line_2} && sleep 2 && ' \
                   f'{line_3} && sleep 2 && {line_4} && sleep 2 && ' \
                   f'{line_5} && sleep 2 && echo "Done" && exit 0'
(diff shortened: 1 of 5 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -650,8 +650,7 @@ def test_multiple_progress_updates_submit(self):
         command = f'{line_1} && sleep 2 && {line_2} && sleep 2 && ' \
                   f'{line_3} && sleep 2 && {line_4} && sleep 2 && ' \
                   f'{line_5} && sleep 2 && echo "Done" && exit 0'
-        job_uuid, resp = util.submit_job(self.cook_url, command=command, executor=job_executor_type,
-                                         max_runtime=60000, max_retries=5)
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, executor=job_executor_type, max_retries=5)
```
dposada, line 653: This change is unrelated to the rest of the PR, but the test flaked on this branch, so I'm fixing it here.

PR #1667, opened 2020-09-09
Logs offer and job percentiles

dposada:

- logging p50, p95, and p100 offer and job size by CPUs and memory when matching

To be able to better troubleshoot matching issues by understanding the "shapes" of jobs and offers from the logs.


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -531,67 +533,103 @@
...
+               (do
+                 (log/warn "Encountered unexpected resource type"
+                           {:resource resource :type type})
+                 resource-map)))
+           {}
+           resources))
+       offers))
+
+(defn jobs->resource-maps
   [jobs]
```
scrosby, line 571: docstring then args.

dposada: Good catch, fixed!

scrosby (approved):

PR #1671, opened 2020-09-14
Refactor regexp list matching code.

scrosby:

- Make the match-based-on-regexp code more generic.

We want to be able to reuse this code for matching compute clusters too.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -911,13 +911,13 @@
   "Given a list of dictionaries [{:<regexp-name> <regexp> :<field-name> <field>} {:<regexp-name> <regexp> :<field-name> <field>} ...], match-list,
    a key <field-name> and <regexp-name> name, return the first matching <field> where the <regexp> matches the key."
   [regexp-name field-name match-list key]
-  (->> match-list
-       (filter (fn [map]
-                 (let [regexp (get map regexp-name)
-                       pattern (re-pattern regexp)]
-                   (re-find pattern key))))
-       first
-       field-name))
+  (-> match-list
+      (->> (filter (fn [map]
+                     (let [regexp (get map regexp-name)
+                           pattern (re-pattern regexp)]
+                       (re-find pattern key)))))
+      first
+      (get field-name)))
 
 (defn match-based-on-pool-name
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -907,14 +907,23 @@
     (cond-> {:count 1 :cpus cpus :mem mem}
       gpus (assoc :gpus gpus))))
 
+(defn match-based-on-regexp
+  "Given a list of dictionaries [{:<regexp-name> regexp :<field-name> ...} {:<regexp-name> .. :<field-name> ...}
```
nsinkov, line 911:
```suggestion
  "Given a list of dictionaries [{:<regexp-name> <regexp> :<field-name> <field>} {:<regexp-name> <regexp> :<field-name> <field>} ...], match-list,
```

scheduler/src/cook/tools.clj
```diff
@@ -907,14 +907,23 @@
...
+   a key <field-name> and <regexp-name> name, return the first matching <field> where the <regexp> matches the key."
```
nsinkov, line 912:
```suggestion
   a key, a <field-name>, and a <regexp-name> name, return the first matching <field> where the <regexp> matches the key."
```

scheduler/src/cook/tools.clj
```diff
@@ -907,14 +907,23 @@
...
+  "Given a list of dictionaries [{:<regexp-name> <regexp> :<field-name> <field>} {:<regexp-name> <regexp> :<field-name> <field>} ...], match-list,
+   a key <field-name> and <regexp-name> name, return the first matching <field> where the <regexp> matches the key."
+  [regexp-name field-name match-list key]
+  (->> match-list
+       (filter (fn [map]
+                 (let [regexp (get map regexp-name)
+                       pattern (re-pattern regexp)]
+                   (re-find pattern key))))
+       first
+       field-name))
```
nsinkov, line 920: let's change to 
(-> match-list
       (->> (filter (fn [map]
                 (let [regexp (get map regexp-name)
                       pattern (re-pattern regexp)]
                   (re-find pattern key)))))
       first
       (get field-name))

nsinkov (approved):

PR #1674, opened 2020-09-15
Reduces excessive logging

dposada:

- making several logs much less long

To make the log size more manageable.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -173,15 +173,19 @@
         [(util/make-atom-updater all-pods-atom) ; Update the set of all pods.
          (util/make-nested-atom-updater node-name->pod-name->pod pod->node-name get-pod-namespaced-key)
          (partial cook-pod-callback-wrap cook-pod-callback compute-cluster-name)] ; Invoke the cook-pod-callback if its a cook pod.
-        old-all-pods @all-pods-atom]
+        old-all-pods @all-pods-atom
+        new-pod-names (set (keys namespaced-pod-name->pod))
+        old-pod-names (set (keys old-all-pods))]
     (log/info "In" compute-cluster-name "compute cluster, pod watch processing pods"
-              {:number-pods (count namespaced-pod-name->pod)
-               :pod-names (->> namespaced-pod-name->pod keys (str/join "\n"))})
+              {:number-pods (count namespaced-pod-name->pod)})
     ; We want to process all changes through the callback process.
     ; So compute the delta between the old and new and process those via the callbacks.
     ; Note as a side effect, the callbacks mutate all-pods-atom
-    (doseq [task (set/union (set (keys namespaced-pod-name->pod)) (set (keys old-all-pods)))]
-      (log/info "In" compute-cluster-name "compute cluster, pod watch doing (startup) callback for" task)
+    (doseq [task (set/union new-pod-names old-pod-names)]
+      (log/info "In" compute-cluster-name "compute cluster, pod watch doing callback"
+                {:new? (contains? new-pod-names task)
+                 :old? (contains? old-pod-names task)
+                 :pod-name task})
       (doseq [callback callbacks]
         (try
           (callback task (get old-all-pods task) (get namespaced-pod-name->pod task))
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -174,12 +174,14 @@
...
-    (log/info "In" compute-cluster-name "compute cluster, pod watch processing pods:" (keys namespaced-pod-name->pod))
+    (log/info "In" compute-cluster-name "compute cluster, pod watch processing pods"
+              {:first-10-pod-names (->> namespaced-pod-name->pod keys (take 10))
+               :number-pods (count namespaced-pod-name->pod)})
     ; We want to process all changes through the callback process.
     ; So compute the delta between the old and new and process those via the callbacks.
     ; Note as a side effect, the callbacks mutate all-pods-atom
     (doseq [task (set/union (set (keys namespaced-pod-name->pod)) (set (keys old-all-pods)))]
-      (log/info "In" compute-cluster-name "compute cluster, pod watch doing (startup) callback for" task)
+      (log/debug "In" compute-cluster-name "compute cluster, pod watch doing (startup) callback for" task)
```
scrosby, line 184: Keep this at info. Its only once an hour and gives us a solid ground truth of what pods are in the system, so we can determine if there is an inconsistency. (This was *extremely* helpful for me to diagnose the cluster rename issue.)

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -78,16 +78,22 @@
         gpu-models (->> node-name->capacity vals (map :gpus) (apply merge) keys set)
         ; The following variables are only being used setting counters for monitor
         gpu-model->total-capacity (total-gpu-resource node-name->capacity)
-        gpu-model->total-consumed (total-gpu-resource node-name->consumed)]
-
-    (log/info "In" compute-cluster-name "compute cluster, all node names:" (keys node-name->node))
```
scrosby, line 83: This, or something like it is also useful for ground truth.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -174,12 +174,14 @@
          (util/make-nested-atom-updater node-name->pod-name->pod pod->node-name get-pod-namespaced-key)
          (partial cook-pod-callback-wrap cook-pod-callback compute-cluster-name)] ; Invoke the cook-pod-callback if its a cook pod.
         old-all-pods @all-pods-atom]
-    (log/info "In" compute-cluster-name "compute cluster, pod watch processing pods:" (keys namespaced-pod-name->pod))
+    (log/info "In" compute-cluster-name "compute cluster, pod watch processing pods"
```
scrosby, line 177: Keep this at the full list. Its only once an hour and gives us a solid ground truth of what pods are in the system, so we can determine if there is an inconsistency. (This was extremely helpful for me to diagnose the cluster rename issue.)

Do split into one line for each pod, instead of one huge line with all pods.

scheduler/src/cook/kubernetes/compute_cluster.clj
scrosby: Can you add a list of all nodes at one node per line in the node watch callback initialization to compensate with this being removed here?

scheduler/src/cook/kubernetes/api.clj
dposada: Sounds good, done

scheduler/src/cook/kubernetes/api.clj
dposada: Sounds good, done

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Done

scheduler/src/cook/kubernetes/api.clj
scrosby: I think we want different log lines. Newlines are seen by us in the log, but doing it as lots of separate log lines avoids creating a single huge unwieldy message.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -260,7 +262,9 @@
         [(util/make-atom-updater current-nodes-atom) ; Update the set of all pods.
          (util/make-nested-atom-updater pool->node-name->node get-node-pool node->node-name)]
          old-current-nodes @current-nodes-atom]
-    (log/info "In" compute-cluster-name "compute cluster, node watch processing nodes:" (keys @current-nodes-atom))
+    (log/info "In" compute-cluster-name "compute cluster, node watch processing nodes"
```
scrosby, line 265: I think we want different log lines. Newlines are seen by us in the log, but doing it as lots of separate log lines avoids creating a single huge unwieldy message.

scheduler/src/cook/kubernetes/api.clj
dposada: Made the changes we discussed offline.

scheduler/src/cook/kubernetes/api.clj
dposada: Made the changes we discussed offline.

scrosby (approved):

PR #1677, opened 2020-09-15
fix missing cluster check

nsinkov:

- don't ignore static clusters when checking that there are no missing clusters for known running tasks

current code will incorrectly raise an error when an existing task is running on a static cluster

### DIFF
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -17,6 +17,7 @@
   (:require [chime :refer [chime-ch]]
             [clj-time.core :as time]
             [clojure.core.async :as async]
+            [clojure.data :as data]
             [clojure.tools.logging :as log]
             [cook.compute-cluster :as cc]
             [cook.config :as config]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/mesos.clj
```diff
@@ -130,7 +131,9 @@
                                                  (group-by #(-> %
                                                               :instance/compute-cluster
                                                               :compute-cluster/cluster-name)))
-            [missing-cluster-names _ _] (util/diff-map-keys cluster-name->running-task-ents saved-cluster-configurations)]
+            cluster-names-for-running-tasks (set (keys cluster-name->running-task-ents))
+            known-cluster-names (set (concat (keys @cc/cluster-name->compute-cluster-atom) (keys saved-cluster-configurations)))
+            [missing-cluster-names _ _] (data/diff cluster-names-for-running-tasks known-cluster-names)]
```
dposada, line 136: Should we have a unit test for this?

nsinkov: added

dposada (approved):

dposada:
Will merge when green

PR #1680, opened 2020-09-18
Fix dynamic cluster integration tests

nsinkov:

- fix dynamic cluster integration tests

### DIFF
--- a/integration/tests/cook/test_dynamic_clusters.py
+++ b/integration/tests/cook/test_dynamic_clusters.py
@@ -62,6 +62,12 @@ class TestDynamicClusters(util.CookTest):
             }
             data, resp = util.create_compute_cluster(self.cook_url, test_cluster)
             self.assertEqual(201, resp.status_code, resp.content)
+            # Test create cluster with duplicate name
+            data, resp = util.create_compute_cluster(self.cook_url, test_cluster)
+            self.assertEqual(422, resp.status_code, resp.content)
+            self.assertEqual(f'Compute cluster with name {test_cluster_name} already exists',
+                             data['error']['message'],
+                             resp.content)
 
         # Check that a job schedules successfully
         command = "true"
--- a/integration/tests/cook/test_dynamic_clusters.py
+++ b/integration/tests/cook/test_dynamic_clusters.py
@@ -1,0 +1,107 @@
+import logging
+import os
+import subprocess
+import time
+import unittest
+
+import pytest
+
+from tests.cook import util
+
+
+@pytest.mark.timeout(util.DEFAULT_TEST_TIMEOUT_SECS)  # individual test timeout
+class TestDynamicClusters(util.CookTest):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.cook_url = util.retrieve_cook_url()
+        util.init_cook_session(cls.cook_url)
+
+    def setUp(self):
+        self.cook_url = type(self).cook_url
+        self.logger = logging.getLogger(__name__)
+        self.user_factory = util.UserFactory(self)
+
+    @unittest.skipUnless(util.using_kubernetes(), 'Test requires kubernetes')
+    @unittest.skipUnless(os.getenv('COOK_CHANGE_LEADER_COMMAND') is not None,
+                         'Requires setting the COOK_CHANGE_LEADER_COMMAND environment variable')
+    @pytest.mark.serial
+    def test_dynamic_clusters(self):
+        """
+        Test that dynamic cluster configuration functionality is working.
+        """
+        docker_image = util.docker_image()
+        container = {'type': 'docker',
+                     'docker': {'image': docker_image}}
+        admin = self.user_factory.admin()
+        # Force all clusters to have state = deleted via the API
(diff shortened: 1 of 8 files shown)


integration/tests/cook/test_dynamic_clusters.py
```diff
@@ -0,0 +1,107 @@
...
+                cluster["state-locked?"] = True
+                self.logger.info(f'Trying to update cluster {cluster}')
+                data, resp = util.update_compute_cluster(self.cook_url, cluster)
+                self.assertEqual(201, resp.status_code, resp.content)
+            # Then state = deleted
+            for cluster in clusters:
+                cluster["state"] = "deleted"
+                util.wait_until(lambda: util.update_compute_cluster(self.cook_url, cluster),
+                                lambda x: True if 201 == x[1].status_code else x[1].content,
+                                300000, 5000)
```
dposada, line 53: Why do we need the `wait_until` here?

integration/tests/cook/test_dynamic_clusters.py
```diff
@@ -0,0 +1,107 @@
...
+                "template": clusters[0]["template"]
+            }
+            data, resp = util.create_compute_cluster(self.cook_url, test_cluster)
+            self.assertEqual(201, resp.status_code, resp.content)
+
+        # Check that a job schedules successfully
+        command = "true"
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
+        self.assertEqual(201, resp.status_code, resp.content)
+        util.wait_for_instance(self.cook_url, job_uuid, status='success')
```
dposada, line 70: Can we also assert that the instance ran on the only running compute cluster?

integration/tests/cook/test_dynamic_clusters.py
```diff
@@ -0,0 +1,107 @@
...
+                            lambda x: True if 201 == x[1].status_code else x[1].content,
+                            300000, 5000)
+            # Hard-delete the original non-test clusters
+            for cluster in clusters:
+                self.logger.info(f'Trying to delete cluster {cluster}')
+                resp = util.delete_compute_cluster(self.cook_url, cluster)
+                self.assertEqual(204, resp.status_code, resp.content)
+            # Force give up leadership
+            self.logger.info(f"COOK_CHANGE_LEADER_COMMAND {os.getenv('COOK_CHANGE_LEADER_COMMAND')}")
+            cp = subprocess.run(os.getenv('COOK_CHANGE_LEADER_COMMAND'), shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
dposada, line 90: Why can't we just make the HTTP request to `/shutdown-leader` here?

integration/tests/cook/test_dynamic_clusters.py
```diff
@@ -0,0 +1,107 @@
...
+        # Old clusters should be re-created
+        # wait for cook to come up
+        util.wait_until(lambda: [cluster for cluster in util.compute_clusters(self.cook_url)['db-configs'] if cluster["state"] == "running"],
+                        lambda x: len(x) == len(clusters),
+                        420000, 5000)
+        # Check that a job schedules successfully
+        command = "true"
+        job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
+        self.assertEqual(201, resp.status_code, resp.content)
+        util.wait_for_instance(self.cook_url, job_uuid, status='success')
```
dposada, line 102: Should we submit jobs until each of the old clusters has successfully run a job?

integration/tests/cook/util.py
```diff
@@ -1935,12 +1939,22 @@ def kubernetes_settings():
     return settings(cook_url)['kubernetes']
 
 
-def create_compute_cluster(cook_url, cluster):
-    resp = session.post(f'{cook_url}/compute-clusters', json=cluster)
-    logger.info(f'create_compute_cluster resp: {resp.content}')
+def create_compute_cluster(cook_url, cluster, headers=None):
+    leader_url = scheduler_info_uncached(cook_url)['leader-url']
+    resp = session.post(f'{leader_url}/compute-clusters', json=cluster, headers=headers)
```
dposada, line 1944: Should we just fix the redirect-to-leader so we don't have to do this?

integration/tests/cook/test_dynamic_clusters.py
dposada: Why is it fine if one of them isn't working? What's the downside to checking that both are working?

integration/tests/cook/util.py
dposada: OK, follow-up PR sounds good

integration/tests/cook/test_dynamic_clusters.py
nsinkov: there's no reason to think that one will work but another won't
But if we want to test them all, how we get a job to schedule on each one?

dposada: I think you'd have to submit jobs until one of them gets matched to each compute cluster. Let me know what you think. If you think it's not worth the complexity, I'm fine with it.

nsinkov: we won't test here, we'll do it as a follow up test

dposada (approved):

PR #1687, opened 2020-09-29
Warns when jobs are unmatched for too long

dposada:

Logging a warning when a high (configurable) percentage of considerable jobs have gone unmatched for more than a configurable number of consecutive match cycles.

To help alert us to situations where job scheduling is not working as expected.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -20,6 +20,7 @@
             [clojure.core.async :as async]
             [clojure.core.cache :as cache]
             [clojure.edn :as edn]
+            [clojure.set :as set]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [clojure.walk :as walk]
@@ -978,6 +979,8 @@
       (catch Throwable e
         (log/error e "In" pool-name "pool, encountered error while triggering autoscaling")))))
 
+(def pool-name->unmatched-job-uuid->unmatched-cycles-atom (atom {}))
+
 (defn handle-resource-offers!
   "Gets a list of offers from mesos. Decides what to do with them all--they should all
    be accepted or rejected at the end of the function."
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1038,6 +1041,48 @@
                               :number-unmatched (- (count offers) (count offers-scheduled))
                               :stats (offers->stats offers)}})
 
+          (let [unmatched-job-uuids
```
scrosby, line 1044: Could you describe the algorithm used to determine this both high level and comments within here. In particular, I'm concerned if there are any leaks, we can OOM. I'd like to have a strong confidence that (in all paths) we prevent that.

dposada: Yep, I added some comments above and throughout, let me know what you think.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1038,6 +1041,87 @@
...
+              ;            job-uuid-d count-d
+              ;            ...}
+              ; ...}
+              ;
+              ; where the counts are the numbers of consecutive
+              ; matching cycles that the job has gone unmatched
+              pool-name->unmatched-job-uuid->unmatched-cycles-atom
+              (fn [m]
+                (let [unmatched-job-uuid->unmatched-cycles
+                      (pc/map-from-keys
```
scrosby, line 1081: Would like some text in here explaining no leak.

"""This code doesn't leak jobs and grow forever. This builds a new map from scratch of size at most unmatched-job-uuids which is less than num-considerable, which goes in. That new map gets assoc'ed in replacing the existing job-uuid->unmatched-cycles submap in pool-name->unmatched-job-uuid->unmatched-cycles-atom. Which means we won't leak historic jobs."""

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Love them. They're great! Thanks.

scheduler/src/cook/scheduler/scheduler.clj
dposada: Good call, done

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1038,6 +1041,87 @@
                               :number-unmatched (- (count offers) (count offers-scheduled))
                               :stats (offers->stats offers)}})
 
+          ; We want to log warnings when jobs have gone unmatched for a long time.
+          ; In order to do this, we keep track, per pool, of the jobs that did not
+          ; get matched to an offer, along with how many matching cycles they've
+          ; gone unmatched for. The amount of data we store is relatively small;
+          ; it's O(# pools * # unmatched considerable jobs). If a job uuid does
+          ; get matched, we stop storing it. We never store job uuids that were
```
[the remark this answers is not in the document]
dposada: Changed

scrosby (approved):

PR #1688, opened 2020-09-29
Add the ability to flush a rate-limit from the cache

scrosby:

- Add the ability to flush a ratelimit from the cache so it gets reloaded.

We want to be able to update ratelimits dynamically.


dposada (approved):

dposada:
Will merge when green

PR #1689, opened 2020-09-29
Remove circular dependency by moving a query out of cook.tools

scrosby:

- Remove circular dependency by moving a first query out of cook.tools.

Cook tools depends on a lot of cook code, so that depending on it can accidently introduce circular dependencies. Much of the code there is database queries. If we move this to a new file, then we can break the circular dependencies.


dposada (approved):

dposada:
Will merge when green

PR #1690, opened 2020-09-29
Refactor to propagate pool-name into quota filters.

scrosby:

- Refactor to propagate pool-name into quota filters.

No functionality change except to add an extra field to a debug log line, but this enables future changes where we want the quota filter to be pool specific.


dposada (approved):

dposada:
Will merge when green

Issue #1370, opened 2020-01-22
Require in components.clj cause compile time error


DaoWen:
**Describe the bug**

Adding new line to the `:require` list in components.clj (which should basically be a no-op) causes a compile-time error.

**To Reproduce**

Adding this single line to the requires list at the top of components.clj file causes an opaque compile time error when running `lein uberjar`:

```clojure
[cook.progress :as progress]
```

<details>
<summary>
<code>java.lang.RuntimeException: Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay@18c1d366, compiling:(schema.clj:16:1)</code>
</summary>

<br>Full `lein uberjar` output:
```clojure
$ lein uberjar    
Compiling 2 source files to /home/nickv/projects/cook/scheduler/target/classes
warning: No SupportedSourceVersion annotation found on io.sundr.builder.internal.processor.BuildableProcessor, returning RELEASE_6.
warning: Supported source version 'RELEASE_6' from annotation processor 'io.sundr.builder.internal.processor.BuildableProcessor' less than -source '1.8'
warning: No SupportedSourceVersion annotation found on io.sundr.builder.internal.processor.ExternalBuildableProcessor, returning RELEASE_6.
warning: Supported source version 'RELEASE_6' from annotation processor 'io.sundr.builder.internal.processor.ExternalBuildableProcessor' less than -source '1.8'
warning: No SupportedSourceVersion annotation found on io.sundr.resourcecify.internal.processor.ResourcecifyProcessor, returning RELEASE_6.
warning: Supported source version 'RELEASE_6' from annotation processor 'io.sundr.resourcecify.internal.processor.ResourcecifyProcessor' less than -source '1.8'
6 warnings                                                  
Compiling cook.components                               
log4j:WARN No appenders could be found for logger (org.jboss.logging).
log4j:WARN Please initialize the log4j system properly. 
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
java.lang.RuntimeException: Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay@18c1d366, compiling:(schema.clj:16:1)
Exception in thread "main" java.lang.RuntimeException: Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay@18c1d366, compiling:(schema.clj:16:1)
        at clojure.lang.Compiler$InvokeExpr.eval(Compiler.java:3657)
[...]

PR #1691, opened 2020-09-29
Remove metatransaction.core import that causes problem with mount.

scrosby:

- Remove metatransaction.core import that causes problem with mount.

Avoids an incomprehensible error with mount. Diagnoses #1370 with a workaround.

### DIFF
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -341,7 +341,7 @@
     ;
     ; If you get an error about "Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay"
     ; The issue is that at least metatransaction.core seems to be incompatible with mount. It cannot be in the
-    ; dependency tree of anything using mount.
+    ; dependency tree of anything using mount. See also issue #1370
     (mount/start-with-args (cook.config/read-config config-file-path))
     (pool/guard-invalid-default-pool (d/db datomic/conn))
     (metrics-jvm/instrument-jvm)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/components.clj
```diff
@@ -338,6 +338,9 @@
   (try
     ; Note: If the mount/start-with-args fails to initialize a defstate S, and/or you get weird errors on startup,
     ; you need to require S's namespace with ns :require. 'ns :require' is how mount finds defstates to initialize.
+    ;
+    ; If you get an error about "Can't embed object in code, maybe print-dup not defined: clojure.lang.Delay"
+    ; The issue is about metatransaction.core seems to be incompatible with mount. It cannot be in the dependency tree.
```
DaoWen, line 343: This comment seems overly specific in blaming `metatransaction.core` for the error. What if we have another namespace that causes the same error when `require`'d into the dependency tree? How did you trace the blame back to `metatransaction.core`? The first sentence on 343 has a typo, so it needs to be reworded anyway.

Maybe just mention the error and link to issue #1370 instead. I probably should have done something like that when I opened the issue.

scrosby: I was able to create the error only adding a dependency to metatransaction.core, and was able to make it go away by deleting the reference. The issue you found adding progress was because of this dependency chain: cook.progress -> cook.tools -> metatransaction.core. I reworded it.

DaoWen (approved):

PR #1700, opened 2020-10-07
Lowers default auth-bypass request rate limit

dposada:

Lowering the default request rate limit for auth-bypass requests to 600 / minute.

To have a more conservative default.


scheduler/src/cook/config.clj
```diff
@@ -234,7 +234,7 @@
                                      {:json-value "config-impersonation"})))
      :rate-limit (fnk [[:config {rate-limit nil}]]
                    (let [{:keys [auth-bypass-limit-per-m expire-minutes job-launch job-submission user-limit-per-m]
-                          :or {auth-bypass-limit-per-m 10000
+                          :or {auth-bypass-limit-per-m 400
```
scrosby, line 237: 600 to be consistent with user-limit-per-m.

scrosby (approved):

PR #1704, opened 2020-10-13
Enable tracking of how rate limiting is affecting the queue.

scrosby:

- Enable tracking of whenever we rate limit.
- Indicate when we are updating the backing data for /unscheduled (and a sense as to what the changes are) 
- 

Track down the consequences of the (expected) race where /unscheduled is showing what we expect. That endpoint is inherently racy, but we think the race is behaving worse than we expected.


dposada:
Will merge when green

PR #1706, opened 2020-10-14
Make our cache sizes configurable from the configuration file.

scrosby:

- Make our cache sizes configurable from the configuration file, so they can be sized to an expected working set size.

- Better handling with longer queues and reduce the risk of hitting a performance cliff.


dposada (approved):

dposada:
Will merge when green

PR #1709, opened 2020-10-15
Add API for Disk Limits

kathryn-zhou:
- Added API & job validation for specifying disk size and type
- Added unit and integration tests
- Added valid-disk-types in config file

- Allow users to request disk space and disk type


scheduler/src/cook/config.clj
```diff
@@ -554,6 +554,11 @@
       nil
       pool)))
 
+(defn valid-disk-types
+  "Returns valid disk models "
```
dposada, line 558:
```suggestion
  "Returns valid disk types"
```

scheduler/src/cook/queries.clj
```diff
@@ -28,3 +28,8 @@
             [?e :resource.type/mesos-name ?ident]]
           db)
        (map first)))
+
+(defn get-quota-resource-types
+  "Return a list of resource types used for quota, :cpus :mem :gpus ..."
+  [db]
+  (remove #{:disk} (get-all-resource-types db)))
```
dposada, line 35: Instead of removing `:disk`, should we include the ones we do want to use for quota purposes?

scheduler/src/cook/rest/api.clj
```diff
@@ -296,6 +296,11 @@
   {:dataset {non-empty-max-128-characters-str non-empty-max-128-characters-str}
    (s/optional-key :partitions) #{{non-empty-max-128-characters-str non-empty-max-128-characters-str}}})
 
+(def Disk
+  "Schema for disk limit specifications"
+  {(s/optional-key :size) (s/both PosDouble (s/pred #(<= 1 % 250000) 'between-1-and-250000))
```
dposada, line 301: I think we should make the max value configurable instead of hard-coding it.

scheduler/src/cook/rest/api.clj
```diff
@@ -737,6 +749,7 @@
                          {:db/id env-var-id
                           :environment/name k
                           :environment/value v}]))
+
```
dposada, line 752: inadvertent newline?

scheduler/src/cook/rest/api.clj
```diff
@@ -771,6 +784,17 @@
...
+                                (let [disk-id (d/tempid :db.part/user)
+                                      params {:db/id disk-id
+                                              :resource/type :resource.type/disk
+                                              :resource.disk/size (:size disk)}]
+                                  [[:db/add db-id :job/resource disk-id]
+                                   (if-not (nil? (:type disk))
+                                     (assoc params :resource.disk/type (:type disk))
+                                     params)]
+                                  )
+                                )
```
dposada, line 797:
```suggestion
                                     params)]))
```

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +978,27 @@
                (not (contains? (get-gpu-models-on-pool (config/valid-gpu-models) pool-name) requested-gpu-model)))
       (throw (ex-info (str "The following GPU model is not supported: " requested-gpu-model) {})))))
 
+(defn validate-disk-job
+  "Validates that a job requesting disk is satisfying the following conditions:
+    - User can request size but not type; User can request both size and type; User cannot request only type and not size
+    - Requested type must be a valid type
+    - TODO: validate that requested size is less than a certain amount
+  "
```
dposada, line 986:
```suggestion
    - TODO: validate that requested size is less than a certain amount"
```

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +978,27 @@
                (not (contains? (get-gpu-models-on-pool (config/valid-gpu-models) pool-name) requested-gpu-model)))
       (throw (ex-info (str "The following GPU model is not supported: " requested-gpu-model) {})))))
 
+(defn validate-disk-job
+  "Validates that a job requesting disk is satisfying the following conditions:
+    - User can request size but not type; User can request both size and type; User cannot request only type and not size
+    - Requested type must be a valid type
+    - TODO: validate that requested size is less than a certain amount
```
dposada, line 985: This is done by the schema validation, right?

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +978,27 @@
                (not (contains? (get-gpu-models-on-pool (config/valid-gpu-models) pool-name) requested-gpu-model)))
       (throw (ex-info (str "The following GPU model is not supported: " requested-gpu-model) {})))))
 
+(defn validate-disk-job
```
dposada, line 981: Maybe `validate-job-disk`?

scheduler/src/cook/schema.clj
```diff
@@ -1001,6 +1001,20 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
    {:db/id (d/tempid :db.part/user)
     :db/ident :resource.type/mem
     :resource.type/mesos-name :mem}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :resource.type/disk
+    ;TODO: Check if resource.type/mesos-name is needed
```
dposada, line 1006: Is this TODO done?

scheduler/src/cook/tools.clj
```diff
@@ -71,6 +71,11 @@
           db)
        (map first)))
 
+(defn get-quota-resource-types
+  "Return a list of resource types used for quota, :cpus :mem :gpus ..."
+  [db]
+  (remove #{:disk} (get-all-resource-types db)))
```
dposada, line 77: Can this be deleted since it's defined in `queries`?

scheduler/test/cook/test/rest/api.clj
```diff
@@ -434,6 +435,91 @@
...
+                               :uri "/jobs"
+                               :headers {"Content-Type" "application/json"}
+                               :authorization/user "dgrnbrg"
+                               :body-params {"jobs" [successful-job-1] "pool" "test-pool"}}))
+                  299)))
+        (let [resp (h {:request-method     :get
+                       :scheme             :http
+                       :uri                "/rawscheduler"
+                       :authorization/user "dgrnbrg"
+                       :query-params       {"job" (str (get successful-job-1 "uuid"))}})
```
dposada, line 490: Cursive strikes again?

scheduler/test/cook/test/rest/api.clj
```diff
@@ -434,6 +435,91 @@
...
+        h (basic-handler conn)]
+    (testing "negative disk invalid"
+      (is (<= 400
+              (:status (h {:request-method :post
+                           :scheme :http
+                           :uri "/jobs"
+                           :headers {"Content-Type" "application/json"}
+                           :authorization/user "dgrnbrg"
+                           :body-params {"jobs" [(job {"size" -4})] "pool" "test-pool"}}))
+              499)))
```
dposada, line 452: Instead of ranges, can we check for specific status codes on these?

scheduler/test/cook/test/rest/api.clj
```diff
@@ -2413,6 +2500,39 @@
...
+  (testing "valid request not specifying disk size and type"
+    (with-redefs [config/valid-disk-types (constantly [{:pool-regex "test-pool"
+                                                        :valid-types #{"valid-disk-type"}
+                                                        :default-type "valid-disk-type"}])]
+      (is (nil? (api/validate-disk-job "test-pool" {})))))
+  (testing "valid request of disk size and type"
+    (with-redefs [config/valid-disk-types (constantly [{:pool-regex "test-pool"
+                                                        :valid-types #{"valid-disk-type"}
+                                                        :default-type "valid-disk-type"}])]
+      (is (nil? (api/validate-disk-job "test-pool" { :disk {:size 20 :type "valid-disk-type"}})))))
```
dposada, line 2513:
```suggestion
      (is (nil? (api/validate-disk-job "test-pool" {:disk {:size 20 :type "valid-disk-type"}})))))
```

dposada: Same thing in a few spots below

scheduler/src/cook/queries.clj
kathryn-zhou: Yep, done

scheduler/src/cook/rest/api.clj
kathryn-zhou: yep

scheduler/src/cook/rest/api.clj
kathryn-zhou: yep, will remove

scheduler/src/cook/rest/api.clj
kathryn-zhou: yes, done

scheduler/src/cook/tools.clj
kathryn-zhou: yes, removed

scheduler/test/cook/test/rest/api.clj
kathryn-zhou: yeah, is there a reason why previous tests followed the convention with ranges?

scheduler/test/cook/test/rest/api.clj
kathryn-zhou: yep, fixed

scheduler/test/cook/test/rest/api.clj
kathryn-zhou: got it, changed

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,70 @@ def test_basic_docker_job(self):
...
+                    self.assertEqual(job["disk"]["type"], expected_type)
+
+                    # Job submission with default disk type
+                    self.logger.info(f'Submitting to {pool}')
+                    job_uuid, resp = util.submit_job(
+                        self.cook_url,
+                        pool=pool_name,
+                        disk={'size': expected_size})
+                    self.assertEqual(resp.status_code, 201, resp.text)
+                    self.assertEqual(job["disk"]["size"], expected_size)
```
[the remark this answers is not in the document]
kathryn-zhou: done

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,70 @@ def test_basic_docker_job(self):
...
+        # If no pools support disks, submit a job to the default pool and assert the submission gets rejected
+        if not valid_disk_types_config_map:
+            default_pool = util.default_submit_pool()
+            job_uuid, resp = util.submit_job(
+                self.cook_url,
+                pool=default_pool,
+                disk={'size': 20000, 'type': 'pd-ssd'})
+            self.assertEqual(resp.status_code, 400)
+            self.assertTrue(f"The following disk type is not supported: pd-ssd" in resp.text,
+                msg=resp.content)
```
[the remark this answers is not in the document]
kathryn-zhou: good point, done

scheduler/src/cook/rest/api.clj
kathryn-zhou: done

scheduler/src/cook/rest/api.clj
kathryn-zhou: done

scheduler/src/cook/schema.clj
kathryn-zhou: yeah, we need it for the "get-all-resource-types" function

scheduler/test/cook/test/rest/api.clj
dposada: Hmm, not sure, but it seems better to me to be precise.

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,67 @@ def test_basic_docker_job(self):
...
+        else:
+            # Check if there are any active pools
+            active_pools, _ = util.active_pools(self.cook_url)
```
scrosby, line 1897: Not worth considering if active_pools is empty. If thats true either we're configured for no pools (which is not production) or we're completely busted, in which case not interesting.

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,67 @@ def test_basic_docker_job(self):
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
         self.assertIn('success', [i['status'] for i in job['instances']])
 
+    def test_submit_job_with_disk(self):
+        settings_dict = util.settings(self.cook_url)
+        valid_disk_types_config_map = settings_dict.get("pools", {}).get("valid-disk-types", [])
```
scrosby, line 1884: Its not a map, its a list of regexp's. Probably worth a comment here, and renaming the variable.

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,67 @@ def test_basic_docker_job(self):
...
+            # Check if there are any active pools
+            active_pools, _ = util.active_pools(self.cook_url)
+            if len(active_pools) == 0:
+                self.logger.info('There are no pools to submit jobs to')
+                self.skipTest("There are no active pools that support disk")
+            for pool in active_pools:
+                pool_name = pool['name']
+                matching_disk_types = [ii["valid-types"] for ii in valid_disk_types_config_map if
+                                       re.match(ii["pool-regex"], pool_name)]
+                # If there are no supported disk types for pool, assert submission gets rejected
```
scrosby, line 1905: You don't actually submit and make sure its rejected, but should.

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,67 @@ def test_basic_docker_job(self):
...
+                    job_uuid, resp = util.submit_job(
+                        self.cook_url,
+                        pool=pool_name,
+                        disk={'size': expected_size, 'type': expected_type})
+                    self.assertEqual(resp.status_code, 201, resp.text)
+                    job = util.load_job(self.cook_url, job_uuid)
+                    self.assertEqual(job["disk"]["size"], expected_size)
+                    self.assertEqual(job["disk"]["type"], expected_type)
+
+                    # Job submission with default disk type
```
scrosby, line 1923: Actually submitting with a size only.

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +975,25 @@
                (not (contains? (get-gpu-models-on-pool (config/valid-gpu-models) pool-name) requested-gpu-model)))
       (throw (ex-info (str "The following GPU model is not supported: " requested-gpu-model) {})))))
 
+(defn validate-job-disk
+  "Validates that a job requesting disk is satisfying the following conditions:
+    - User can request size but not type; User can request both size and type; User cannot request only type and not size
+    - Requested type must be a valid type in config"
+  [pool-name {:keys [disk]}]
+  (let [requested-disk-size (:size disk)
```
scrosby, line 983: You can destructure this out too:
 (let [{requested-disk-size :size requested-disk-type:type}] ...
)

scheduler/src/cook/queries.clj
```diff
@@ -28,3 +28,8 @@
             [?e :resource.type/mesos-name ?ident]]
           db)
        (map first)))
+
+(defn get-quota-resource-types
```
scrosby, line 32: I don't particularly like this function. Now we have 3 kinds of resources. 

1. Those that go into fenzo without being quota. (disk)
2. Those that go into quota but do not feed fenzo. (rate limits, count)
3. Those that go into quota and fenzo. (mem, cpu gpus)

We have #2 and #3. This code creates #1. 

If you don't give a :resource.type /mesos-name for the resource, the result doesn't show up with queries/get-all-resource-types  so maybe that lets you get rid of this function. 

Alternately, whats wrong with setting a quota for disk as well as everything else? Just have a default quota of infinity and we're good.  @dposada that would get my vote. We allow setting a per user per-pool disk quota. 

Regardless of the above, we should document that distinction carefully, because its getting muddy.

scheduler/src/cook/rest/api.clj
scrosby: And move the validation to `validate-job-disk` where you can report a nice error about the maximum supported disk size.

scrosby: Also, I think we should make :size be a required key. If you include a :disk section at all, you have to include a size. You can't include an empty :disk section. @dposada  thoughts?

scheduler/src/cook/rest/api.clj
kathryn-zhou: done

integration/tests/cook/test_basic.py
kathryn-zhou: done

integration/tests/cook/test_basic.py
kathryn-zhou: done

integration/tests/cook/test_basic.py
kathryn-zhou: interesting. for gpu testing, we followed a similar structure of checking if there are any active pools. is that case also useless then?

scheduler/test/cook/test/rest/api.clj
kathryn-zhou: updated

scheduler/src/cook/queries.clj
kathryn-zhou: Great point. @dposada and I discussed earlier today and came up with the following options for solutions.

1. Leave disk out of quota but add documentation in this function for the different kinds of resources. This is the most simple and we can implement quotas per user at a later time.
2. Use a scalar value for quota (agnostic to type). This is the second simplest and could probably be included in this PR.
3. Make quotas for different types of disks. This is definitely the most complicated, and we are trying to find out from Hyonjee and Sri if this is something that is valuable and worth spending time on. DPO and I agree that it probably wouldn't be included in the scope of this PR.

Let us know your thoughts!

integration/tests/cook/test_basic.py
scrosby: Also interesting. I guess its fine, just seems strange and we should explain why...
Ah hah do we need to add something like this there (and here)?:
"""
GPU's may not be enabled for COOK_TEST_DEFAULT_SUBMIT_POOL, so we iterate over all active pools to find a pool where we can run this test.
"""

scheduler/src/cook/queries.clj
scrosby: I like #1, but we need to write the comment explaining the subtlety of whats going on here and on get-all-resource-types. (explaining the 3 resource types and why they are the way they are.)

scheduler/src/cook/queries.clj
```diff
@@ -28,3 +28,8 @@
...
+  "Return a list of resource types used for quota, :cpus :mem :gpus ..."
```
scrosby, line 33: E.g., 'quota' is an overloaded term.We should update comments to indicate what is meant, 'e.g., these resource types are used by fenzo' or 'these are used to filter the queue' or 'these resource types are used for both'. (Per our phone call, this is being refactored, but when you do that, keep an eye out for analogous ambiguity.

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +987,32 @@
...
+    - Requested size must be less than the max size in config"
+  [pool-name {:keys [disk]}]
+  (let [{disk-request :request disk-limit :limit requested-disk-type :type} disk
+        max-size (or (get-max-disk-size-on-pool (config/disk) pool-name) 10000000)
+        disk-types-on-pool (get-disk-types-on-pool (config/disk) pool-name)]
+    (when disk-limit
+      (do (when (> disk-request disk-limit)
+            (throw (ex-info (str "Disk request specified is greater than disk limit specified") disk)))
+          (when (> disk-limit max-size)
+            (throw (ex-info (str "Disk limit specified is greater than max disk size on pool") disk)))))
```
scrosby, line 1003: Maybe cleaner and more obvious with:
  (when (< disk-request disk-limit max-size)  
       (.... "Disk resource settting error, We must have disk-request < disk-limit < max-size, but have "  {:disk-request disk-request :max-size ....})

scheduler/test/cook/test/config.clj
```diff
@@ -123,6 +123,40 @@
                                                              :valid-models #{"valid-gpu-model"}
                                                              :default-model "invalid-gpu-model"}])))))
 
+(deftest test-disk-config-settings
+  (testing "empty disk"
+    (is (nil? (config/guard-invalid-disk-config []))))
+  (testing "valid default type"
+    (is (nil? (config/guard-invalid-disk-config [{:pool-regex "test-pool"
```
scrosby, line 130: People imitate unit tests. Should use ^test-pool$ here and down below.

scheduler/config-k8s.edn
```diff
@@ -119,6 +119,11 @@
                                                      :network "HOST"
                                                      :force-pull-image false
                                                      :parameters []}}}]
+         :disk
+         [{:pool-regex "k8s-alpha"
```
scrosby, line 123: ^k8s-alpha$

scheduler/src/cook/schema.clj
```diff
@@ -1001,6 +1001,24 @@ for a job. E.g. {:resources {:cpus 4 :mem 3} :constraints {\"unique_host_constra
    {:db/id (d/tempid :db.part/user)
     :db/ident :resource.type/mem
     :resource.type/mesos-name :mem}
+   {:db/id (d/tempid :db.part/user)
+    :db/ident :resource.type/disk
+    :resource.type/mesos-name :disk}
```
scrosby, line 1006: As discussed, strip off /mesos-name.

scheduler/config-k8s.edn
kathryn-zhou: done

scheduler/src/cook/rest/api.clj
kathryn-zhou: thanks for suggestion. done

scheduler/src/cook/schema.clj
kathryn-zhou: done

integration/tests/cook/test_basic.py
kathryn-zhou: I think it's worth adding to the gpu test but dpo and I think it is probably overkill for the disk test

scheduler/test/cook/test/config.clj
kathryn-zhou: done

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,94 @@ def test_basic_docker_job(self):
...
+                            msg=resp.content)
+
+            # Submit a job that specifies disk request greater than max-size on-prem
+            job_uuid, resp = util.submit_job(
+                self.cook_url,
+                pool=default_pool,
+                disk={'request': 200000000.0})
+            self.assertEqual(resp.status_code, 400)
+            self.assertTrue("Disk request specified is greater than max disk size on pool" in resp.text, msg=resp.content)
+
```
dposada, line 1917: If there are no disk configuration entries, can't we just reject any and all job submissions that specify the `disk` field on the job spec?

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,94 @@ def test_basic_docker_job(self):
...
+            # Check if there are any active pools
+            active_pools, _ = util.active_pools(self.cook_url)
+            for pool in active_pools:
+                pool_name = pool['name']
+                matching_disk_types = [ii["valid-types"] for ii in disk_config_list if
+                                       re.match(ii["pool-regex"], pool_name)]
+                # If there are no supported disk types for pool, skip test
+                if len(matching_disk_types) == 0 or len(matching_disk_types[0]) == 0:
+                    self.logger.info('There are no disk types configured')
+                    self.skipTest("There are no disk types configured for any active pools")
```
dposada, line 1928: I don't think we want to skip inside the `for` loop, right? Because we would be skipping the entire test just because one of the pools didn't have any matching disk types.

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,94 @@ def test_basic_docker_job(self):
...
+                else:
+                    expected_request = 20000.0
+                    expected_limit = 30000.0
```
dposada, line 1931: Don't we need to use a request and limit that we know are <= the max size allowed for the pool-regex?

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,94 @@ def test_basic_docker_job(self):
...
+                        pool=pool_name,
+                        disk={'request': expected_request})
+                    self.assertEqual(resp.status_code, 201, resp.text)
+                    job = util.load_job(self.cook_url, job_uuid)
+                    self.assertEqual(job["disk"]["request"], expected_request)
+                    self.assertNotIn("limit", job["disk"])
+                    self.assertNotIn("type", job["disk"])
+
+                    # Valid job submission with disk request, size, and type
+                    self.logger.info(f'Submitting to {pool}')
```
dposada, line 1947: Do we need to log this again?

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +987,31 @@
                (not (contains? (get-gpu-models-on-pool (config/valid-gpu-models) pool-name) requested-gpu-model)))
       (throw (ex-info (str "The following GPU model is not supported: " requested-gpu-model) {})))))
 
+(defn validate-job-disk
+  "Validates that a job requesting disk is satisfying the following conditions:
+    - User can request size but not type; User can request both size and type; User cannot request only type and not size
+    - Requested type must be a valid type in config
+    - Requested size must be less than the max size in config"
```
dposada, line 994: This docstring needs updating to reflect request vs limit

scheduler/src/cook/rest/api.clj
```diff
@@ -954,12 +987,31 @@
...
+  [pool-name {:keys [disk]}]
+  (let [{disk-request :request disk-limit :limit requested-disk-type :type} disk
+        max-size (or (get-max-disk-size-on-pool (config/disk) pool-name) 10000000)
```
dposada, line 997: Can we just reject the job submission in this case?

kathryn-zhou: yeah, config should reject that, forgot to update this portion to reflect that

integration/tests/cook/test_basic.py
kathryn-zhou: yep, just updated since we changed from silently accepting

integration/tests/cook/test_basic.py
kathryn-zhou: good point, updated to be proportional to the max-size

integration/tests/cook/test_basic.py
kathryn-zhou: not necessary, took out

integration/tests/cook/test_basic.py
kathryn-zhou: yes, checked that submission was rejected instead of skipping

scheduler/src/cook/rest/api.clj
kathryn-zhou: done

integration/tests/cook/test_basic.py
scrosby: I meant that as a proposed comment to add here (and at the GPU test). If thats not the reason why we're iterating, than can you put in a comment explaining why?

scheduler/config-k8s.edn
```diff
@@ -119,6 +119,11 @@
                                                      :network "HOST"
                                                      :force-pull-image false
                                                      :parameters []}}}]
+         :disk
+         [{:pool-regex "^k8s-alpha$"
```
scrosby, line 123: Indenting seems off here.

scheduler/src/cook/queries.clj
```diff
@@ -21,10 +21,11 @@
 ;; place that no other cook code depends on. DO NOT REQUITE ANY COOK NAMESPACE.
 
 (defn get-all-resource-types
-  "Return a list of all supported resources types. Example, :cpus :mem :gpus ..."
+  "Return a list of resources types that are used for both binpacking and filtering the queue. Example, :cpus :mem :gpus ...
+   Note that this function name is misleading, and does not actually return ALL resource types."
```
scrosby, line 25: :+1:

scheduler/src/cook/tools.clj
```diff
@@ -249,6 +249,9 @@
                     (let [resource (keyword (name (:resource/type r)))]
                       (condp contains? resource
                         #{:cpus :mem :gpus} (assoc m resource (:resource/amount r))
+                        #{:disk} (assoc m :disk (cond-> {:request (:resource.disk/request r)}
```
scrosby, line 252: Worth a comment here: 
"" We add these additional disk resources here so they're available for fenzo binpacking. so we don't overfill a node with disk.""

scheduler/src/cook/tools.clj
```diff
@@ -249,6 +249,9 @@
...
+                                                        (:resource.disk/limit r) (assoc :limit (:resource.disk/limit r))
```
scrosby, line 253: Also, to pose a question, if a node does not have disk resources, but a job does declare a need for disk resources (or gets resources inserted by default), will putting this code here cause fenzo to fail to binpack on that node. (But, if tets pass, then obviously we're OK, but please make sure to test this on k8s and mesos and GKE.)  I think we're safe, but we don't want to break production.

integration/tests/cook/test_basic.py
kathryn-zhou: ah, misunderstood your comment. added in the explanation as a comment

scheduler/config-k8s.edn
kathryn-zhou: fixed

scheduler/src/cook/tools.clj
kathryn-zhou: This is handled in our API validation. If a job declares a need for disk resources, we check that the pool (and thus compute cluster) has nodes with disk resources available. On k8s on-prem, the request/limit doesn't actually get enforced, but the jobs will pass. If a job requests disk resources on Mesos the job validation will fail.

scrosby (approved):

scheduler/src/cook/rest/api.clj
dposada: Yep, requiring a request sounds reasonable to me as well

integration/tests/cook/test_basic.py
```diff
@@ -1879,6 +1879,82 @@ def test_basic_docker_job(self):
...
+                pool=default_pool,
+                disk={'request': 20000.0,
+                      'type': 'pd-ssd'})
+            self.assertEqual(resp.status_code, 400)
+            self.assertTrue(f"Disk specifications are not supported on pool {default_pool}" in resp.text,
+                             msg=resp.content)
+
+        else:
+            # Disk may not be configured for COOK_TEST_DEFAULT_POOL, so we iterate over all active pools to find a pool where we can run this test.
+            # Check if there are any active pools
```
dposada, line 1900: This line seems superfluous

dposada (approved):

PR #1711, opened 2020-10-15
Shortens the 'updating dynamic clusters' log

dposada:

- removing the `:ca-cert` from the cluster configs when logging them

The log can get huge (> 80kb), which can cause problems with logstash, and the ca-cert makes the log hard to read.

### DIFF
--- a/integration/tests/cook/test_cli.py
+++ b/integration/tests/cook/test_cli.py
@@ -1944,7 +1944,7 @@ if __name__ == '__main__':
                 self.assertEqual('Hello world!', cli.decode(cp.stdout))
 
     def test_usage(self):
-        command = 'sleep 300'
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_cli.py
```diff
@@ -1944,7 +1944,7 @@ def main():
                 self.assertEqual('Hello world!', cli.decode(cp.stdout))
 
     def test_usage(self):
-        command = 'sleep 300'
+        command = f'sleep {util.DEFAULT_TEST_TIMEOUT_SECS}'
```
dposada, line 1947: This change is unrelated to the rest of the PR, but the test flaked on this branch, so I'm fixing it here.

PR #1714, opened 2020-10-22
Sets supplemental groups in the pod security context

dposada:

- adding a cache from user -> group ids
- populating the cache when needed with `id -G`
- setting the supplemental groups in the pod security context

So that the pod runs with the proper group membership for the user.


scrosby (commented):
Change looks fine to me, but I've had some problems with these uid's causing docker test breakages. Can you confirm that you've not broken the docker test stack? 

Did you run the integration test suite for this open source with and without basic-auth (with cook launched via docker stack, and tests run via pytest in a venv?)
Did you also try submitting manually at command line using 'cs' without basic auth and with cook launched via docker stack, and did it work?

scrosby (approved):

PR #1715, opened 2020-10-23
Removes group name validation from Java client

dposada:

- removing job group name validation from the Java client library

The client is validating group names, but these are not currently enforced at the Cook Scheduler level, which is inconsistent and can cause problems when reading job groups created by other clients.

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Group.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Group.java
@@ -16,7 +16,6 @@
 
 package com.twosigma.cook.jobclient;
 
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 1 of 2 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Group.java
```diff
@@ -79,6 +69,7 @@ public static Status fromString(String name) {
         private HostPlacement _hostPlacement;
         private StragglerHandling _stragglerHandling;
         private List<UUID> _jobs = new ArrayList<>();
+        private boolean _shouldValidate = true;
```
nsinkov, line 72: there's no validation in cook scheduler, so let's remove validation from the client

dposada: Sounds good will do

dposada: Done

PR #1720, opened 2020-10-28
Shortens the 'Will not checkpoint' log

dposada:

- not logging the entire job

Depending on the job, this log can get very long (> 22k), which can cause problems for logstash.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -716,7 +716,7 @@
                                               (contains? checkpoint-failure-reasons (:reason/name reason)))
                                             instance)]
             (if (-> checkpoint-failures count (>= max-checkpoint-attempts))
-              (log/info "Will not checkpoint, there are too many checkpoint failures"
+              (log/info "Will not checkpoint, there are at least" max-checkpoint-attempts "checkpoint failures"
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -716,8 +716,11 @@
                                               (contains? checkpoint-failure-reasons (:reason/name reason)))
                                             instance)]
             (if (-> checkpoint-failures count (>= max-checkpoint-attempts))
-              (log/info "Will not checkpoint task-id" task-id ", there are at least" max-checkpoint-attempts "failed instances"
-                        {:job job})
+              (log/info "Will not checkpoint, there are too many checkpoint failures"
```
nsinkov, line 719:
```suggestion
              (log/info "Will not checkpoint, there are at least" max-checkpoint-attempts "checkpoint failures"
```

dposada: Why not just have `max-checkpoint-attempts` in the dictionary?

nsinkov: yeah, it's already in the dictionary. I just want this wording is to make it clear that it's greater than or equal to.

PR #1723, opened 2020-10-29
Gracefully handles unknown job resource type

dposada:

- adding `try` / `catch` to `offers->stats` and `jobs->stats`
- ignoring unknown resource types in `job-ent->resources`

If a new resource type is added to jobs in the database, as with `disk` in PR #1709, we need old code (that is unaware of the new resource type) to not fail when encountering those jobs.

- If we can't collect stats, the whole matching cycle should not error out.
- If there is an unknown resource type, `job-ent->resources` shouldn't throw.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -264,7 +264,7 @@
   "Take a job entity and return a resource map. NOTE: the keys must be same as mesos resource keys"
   [job]
   (let [job-ent->resources-miss
-        (fn [job-ent]
+        (fn [{:keys [job/uuid] :as job-ent}]
           (reduce (fn [m r]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -273,7 +273,12 @@
                                            {:cache (:resource.uri/cache? r false)
                                             :executable (:resource.uri/executable? r false)
                                             :value (:resource.uri/value r)
-                                            :extract (:resource.uri/extract? r false)}))))
+                                            :extract (:resource.uri/extract? r false)})
+                        (do
+                          (log/info "Encountered unknown job resource type"
```
dposada, line 278:
```suggestion
                          (log/warn "Encountered unknown job resource type"
```

PR #1727, opened 2020-11-03
Fix memory leak in k8s state for deleted pods

scrosby:

- Fix a leak where we we would unconditionally write back to the k8s state when the state scanner visited a pod which had been deleted. This ends up leaking.

- Avoids a memory leak.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -600,7 +600,7 @@
      ; we may leak a deleted pod back into this dictionary where it will never be recoverable.
      ;
      ; We only reprocess if something important has actually changed.
-     (when-not (and (nil? pod))
+     (when (some? pod)
        (swap! k8s-actual-state-map assoc pod-name new-state))
      (let [new-file-server-state (:sandbox-file-server-container-state new-state)
            old-file-server-state (:sandbox-file-server-container-state old-state)]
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -596,8 +596,12 @@
                     :synthesized-state (api/pod->synthesized-pod-state pod-name pod)
                     :sandbox-file-server-container-state (api/pod->sandbox-file-server-container-state pod)}
          old-state (get @k8s-actual-state-map pod-name)]
-     ; We always store the updated state, but only reprocess it if it is genuinely different.
-     (swap! k8s-actual-state-map assoc pod-name new-state)
+     ; We always store the new state (to capture any changes in the pod) if the pod's not nil. If we write here unconditionally,
+     ; we may leak a deleted pod back into this dictionary where it will never be recoverable.
+     ;
+     ; We only reprocess if something important has actually changed.
+     (when-not (and (nil? pod))
```
dposada, line 603:
```suggestion
     (when (some? pod)
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -596,8 +596,12 @@
...
+       (swap! k8s-actual-state-map assoc pod-name new-state))
```
dposada, line 604: Can we write a unit test?

scheduler/src/cook/kubernetes/controller.clj
scrosby: Thanks. Made.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Good suggestion. Done.

dposada (approved):

PR #1728, opened 2020-11-03
add metadata env vars

nsinkov:

- add metadata env vars

We need access to more metadata like cook URL to get info about past instances, compute cluster name, pool name

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -3226,6 +3226,22 @@ class CookTest(util.CookTest):
         finally:
             util.kill_jobs(self.cook_url, [job_uuid_disabled, job_uuid_enabled])
 
+    @unittest.skipUnless(util.using_kubernetes(), 'Test requires kubernetes')
+    def test_kubernetes_metadata_env_vars(self):
+        docker_image = util.docker_image()
+        container = {'type': 'docker',
+                     'docker': {'image': docker_image}}
+        default_pool = util.default_submit_pool() or util.default_pool(self.cook_url)
+        try:
+            command = 'bash -c \'if [[ "${COOK_POOL:-zzz}" == "' + default_pool + '" ]] && ' \
+                      '[[ "${COOK_SCHEDULER_URL:-zzz}" == "' + self.cook_url + '" ]]; then exit 0; else exit 1; fi\''
+            job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
+            self.assertEqual(201, resp.status_code, resp.text)
+            util.wait_for_instance(self.cook_url, job_uuid, status='success')
+        finally:
+            util.kill_jobs(self.cook_url, [job_uuid])
+
+
     @unittest.skipUnless(util.docker_tests_enabled(), 'Requires docker support')
     def test_disallowed_docker_parameters(self):
         settings = util.settings(self.cook_url)
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -3238,7 +3238,7 @@ class CookTest(util.CookTest):
             # leader url and COOK_SCHEDULER_REST_URL can have different schemes right now
             #command = 'bash -c \'if [[ "${COOK_POOL:-zzz}" == "' + default_pool + '" ]] && [[ "${COOK_SCHEDULER_REST_URL:-zzz}" == "' + info['leader-url'] + '" ]]; then exit 0; else exit 1; fi\''
             command = 'bash -c \'if [[ "${COOK_POOL:-zzz}" == "' + default_pool + '" ]] && [[ "${COOK_SCHEDULER_REST_URL:-zzz}" != "zzz" ]]; then exit 0; else exit 1; fi\''
-            job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -800,7 +800,10 @@
         params-env (build-params-env parameters)
         progress-env (task/build-executor-environment job)
         checkpoint-env (checkpoint->env checkpoint)
-        main-env-base (merge environment params-env progress-env sandbox-env checkpoint-env)
+        metadata-env (cond-> {"COOK_COMPUTE_CLUSTER_NAME" compute-cluster-name
+                              "COOK_SCHEDULER_REST_URL" (config/scheduler-rest-url)}
```
dposada, line 804: Why not just `COOK_SCHEDULER_URL`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -800,7 +800,10 @@
...
+                       pool-name (assoc "COOK_POOL" pool-name))
```
dposada, line 805: Will this be `nil` if the user doesn't specify a pool (and lands in the default pool)? I assume we want to make it resolve to the name of the default pool?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -800,7 +800,10 @@
...
+        main-env-base (merge environment params-env progress-env sandbox-env checkpoint-env metadata-env)
```
dposada, line 806: Would it make sense to write an integration test for the URL and POOL environment variables?

nsinkov: added

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -800,7 +801,10 @@
         params-env (build-params-env parameters)
         progress-env (task/build-executor-environment job)
         checkpoint-env (checkpoint->env checkpoint)
-        main-env-base (merge environment params-env progress-env sandbox-env checkpoint-env)
+        metadata-env {"COOK_COMPUTE_CLUSTER_NAME" compute-cluster-name
+                      "COOK_POOL" (pool/pool-name-or-default pool-name)
+                      "COOK_SCHEDULER_REST_URL" (config/scheduler-rest-url)}
```
nsinkov, line 806: COOK_SCHEDULER_REST_URL is already used by the sidecar. can't rename to COOK_SCHEDULER_URL right now

integration/tests/cook/test_basic.py
```diff
@@ -3226,6 +3226,25 @@ def test_kubernetes_checkpointing(self):
...
+    def test_kubernetes_metadata_env_vars(self):
+        docker_image = util.docker_image()
+        container = {'type': 'docker',
+                     'docker': {'image': docker_image}}
+        default_pool = util.default_submit_pool() or util.default_pool(self.cook_url)
+        # info = util.scheduler_info(self.cook_url)
+        try:
+            # need to figure out how to match up COOK_SCHEDULER_REST_URL
+            # leader url and COOK_SCHEDULER_REST_URL can have different schemes right now
+            #command = 'bash -c \'if [[ "${COOK_POOL:-zzz}" == "' + default_pool + '" ]] && [[ "${COOK_SCHEDULER_REST_URL:-zzz}" == "' + info['leader-url'] + '" ]]; then exit 0; else exit 1; fi\''
```
dposada, line 3239: Remove commented out lines?

nsinkov: I have them as a TODO

dposada: Should we compare everything except the scheme?

nsinkov: it would be scheme and port.

dposada: OK. I have a slight preference for removing the commented out code lines, but if you feel strongly that it's better with them there, that's fine.

integration/tests/cook/test_basic.py
```diff
@@ -3226,6 +3226,25 @@ def test_kubernetes_checkpointing(self):
...
+            command = 'bash -c \'if [[ "${COOK_POOL:-zzz}" == "' + default_pool + '" ]] && [[ "${COOK_SCHEDULER_REST_URL:-zzz}" != "zzz" ]]; then exit 0; else exit 1; fi\''
+            job_uuid, resp = util.submit_job(self.cook_url, command=command, container=container)
+            self.assertEqual(201, resp.status_code, resp.text)
+            util.wait_for_instance(self.cook_url, job_uuid, status='success')
```
dposada, line 3243: If we're going to wait for success, I think we need to add some retries to be resilient to random failures

nsinkov: so going forward for new tests we want retries on any job submission? is there an example of this on a different test? can we implement retries on a test level instead, like in waiter?

dposada: You only want the retries if you're expecting the test job to succeed (many tests don't have that expectation). Yeah, there are lots of examples, search for `max_retries=5` in the tests.

nsinkov: oh, I wasn't thinking about cook retries. updated

dposada (approved):

dposada:
Will merge when green

PR #1729, opened 2020-11-03
add support for modifying pod image when checkpointing

nsinkov:

- add support for modifying pod image when checkpointing

We want to be able to modify the image spec when checkpointing. We want to be able to translate to a spec that is based on the image ID and is fixed. We need to get the exact same image when restoring even if new builds of the image have been made in the interim.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -728,10 +728,10 @@
 (defn calculate-effective-image
   "Transform the supplied job's image as necessary. e.g. do special transformation if checkpointing is enabled
   and an image transformation function is supplied."
-  [job-submit-time image {:keys [mode calculate-effective-image-fn]} task-id]
+  [{:keys [calculate-effective-image-fn] :as kubernetes-config} job-submit-time image {:keys [mode]} task-id]
   (if (and mode calculate-effective-image-fn)
     (try
-      ((util/lazy-load-var-memo calculate-effective-image-fn) job-submit-time image)
+      ((util/lazy-load-var-memo calculate-effective-image-fn) kubernetes-config job-submit-time image)
       (catch Exception e
         (log/error e "Error calculating effective image for checkpointing"
                    {:calculate-effective-image-fn calculate-effective-image-fn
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -780,6 +795,9 @@
         {:keys [volumes volume-mounts sandbox-volume-mount-fn]} (make-volumes volumes sandbox-dir)
         {:keys [custom-shell init-container set-container-cpu-limit? sidecar]} (config/kubernetes)
         checkpoint (calculate-effective-checkpointing-config job task-id)
+        job-submit-time (when (:job/submit-time job) ; due to a bug, submit time may not exist for some jobs
```
dposada, line 798: What's the bug?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -311,26 +311,21 @@
...
+          {:keys [default-workdir sidecar]} (config/kubernetes)
           sandbox-fileserver-port (:port sidecar)
           sandbox-health-check-endpoint (:health-check-endpoint sidecar)
           sandbox-url (try
                         (when (and sandbox-fileserver-port
                                    sandbox-health-check-endpoint
                                    (not (str/blank? pod-ip)))
                           (str "http://"
-                               ((get-pod-ip->hostname-fn pod-ip->hostname-fn) pod-ip)
+                               pod-ip
```
dposada, line 328: Is this unrelated to the rest of the PR?

scheduler/src/cook/kubernetes/api.clj
nsinkov: Don't know. This comment was added https://github.com/twosigma/Cook/commit/0208e4701c32ad1980b1e266258bec5f8bfcf7de#diff-7f153e0aaafb627dc91b7170dc98673fad41bdbca07501ff2247f0f2119563dcR469

scheduler/src/cook/kubernetes/controller.clj
nsinkov: yeah, just was looking at usage of lazy load and cleaned this up. so related a tiny bit

scheduler/src/cook/kubernetes/api.clj
nsinkov: the submit time is at least null for synthetic pods

dposada: I'm not understanding what I should be looking at in that link

dposada: Synthetic pods aren't jobs

nsinkov: that link is the commit from Dec 27, 2016 when the comment "due to a bug, submit time may not exist for some jobs" was added

synthetic pods are launched and a pod is created. So this check also happens to cover that case

dposada: Got it. Sorry, GitHub was collapsing that file even though your deep link was to a line in the file.

Is the "bug" the fact that systems can create jobs directly in the Datomic database instead of going through the API?

Either way, should we extract a function for this since it's going to be in two places?

```
(when (:job/submit-time job)
  (.getTime (:job/submit-time job)))
```

dposada (approved):

dposada:
Will merge when green

PR #1730, opened 2020-11-05
Refactor to make pool taint / label and context configurable

scrosby:
…ontext configurable

- Add configuration knobs to choose which node selector cook puts on its pods.
- Add configuration knobs to choose which taint cook uses to choose its pool name.
- Add configuration knob to select one context out of a kubeconfig file.

Give us  more configurability for how we connect to k8s.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -33,9 +33,9 @@
 
 (defn schedulable-node-filter
   "Is a node schedulable?"
-  [node-name->node node-name->pods {:keys [node-blocklist-labels] :as compute-cluster} [node-name _]]
+  [compute-cluster node-name->node node-name->pods [node-name _]]
   (if-let [^V1Node node (node-name->node node-name)]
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -623,6 +623,8 @@
                                   ; don't need the functionality. But we need to set the file anyways
                                   ; to avoid a NPE.
                                   (.setFile kubeconfig (File. config-file))
+                                  (when kubeconfig-context
+                                    (.setContext kubeconfig kubeconfig-context))
```
dposada, line 627: Why is this needed?

scrosby: In cases where you have a ~/.kube/config file that has multiple contexts, it won't know which context it should use and will pick one, which might not be the one you wanted it to use. With this change, we can now explicitly select which one.

dposada (approved):

dposada:
Will merge when green

PR #1731, opened 2020-11-05
Add support for disk field in Java job client

kathryn-zhou:
- Add Disk class
- Add set and get functions for Disk request, limit, and type

- Allow Java jobclient users to specify disk

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
@@ -44,9 +44,7 @@ public class Disk {
         this._type = type;
     }
 
-    public void setRequest(Double request) {
-        this._request = request;
-    }
+    public void setRequest(Double request) { this._request = request; }
 
     public void setLimit(Double limit) {
         this._limit = limit;
@@ -60,8 +58,8 @@ public class Disk {
         setType(type.toString());
     }
 
-    public double getRequest() { return _request; }
-    public double getLimit() { return _limit; }
+    public Double getRequest() { return _request; }
+    public Double getLimit() { return _limit; }
     public String getType() { return _type; }
 
     /**
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
@@ -44,7 +44,9 @@ public class Disk {
         this._type = type;
     }
 
-    public void setRequest(Double request) { this._request = request; }
+    public void setRequest(Double request) {
+        this._request = request;
+    }
 
     public void setLimit(Double limit) {
         this._limit = limit;
@@ -71,7 +73,7 @@ public class Disk {
 
     public JSONObject toJSONObject() throws JSONException {
         final JSONObject object = new JSONObject();
-        if(this._request != null) {
+        if (this._request != null) {
             object.put("request", _request);
         }
         if (this._limit != null) {
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
@@ -1,0 +1,104 @@
+package com.twosigma.cook.jobclient;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class Disk {
+    /**
+     *   request - disk space guaranteed for the job
+     *   limit - max disk space the job can use
+     *   type - type of disk for the job
+     */
+    private Double _request;
+    private Double _limit;
+    private String _type;
+
+    public enum DiskType {
+        STANDARD ("standard"), PD_SSD ("pd-ssd");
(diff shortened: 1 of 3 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
```diff
@@ -0,0 +1,102 @@
...
+        this(null, null, null);
+    }
+
+    public Disk(Double request, Double limit, String type) {
+        this._request = request;
+        this._limit = limit;
+        this._type = type;
+    }
+
+    public void setRequest(Double request) { this._request = request; }
```
dposada, line 47: Maybe to be consistent with the other setters:

```suggestion
    public void setRequest(Double request) { 
        this._request = request; 
    }
```

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Function to determine if disk request should be included in the JSON for the Cook payload
+     */
+    public boolean shouldIncludeInJSON() {
+        return (this._request != null || this._limit != null || this._type != null);
+    }
+
+    public JSONObject toJSONObject() throws JSONException {
+        final JSONObject object = new JSONObject();
+        if(this._request != null) {
```
dposada, line 74:
```suggestion
        if (this._request != null) {
```

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -457,6 +458,58 @@ public Builder setGpus(Integer gpus) {
...
+            _disk.setType(diskType);
+            return this;
+        }
+
+        /**
+         * Set the disk type, specifying the disk type requested, to the job.
+         * @param diskType The disk type for this job
+         * @return this builder
+         */
+        //TODO: add another setter function that takes in enum
```
dposada, line 507: Delete the TODO?

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
```diff
@@ -0,0 +1,104 @@
...
+
+public class Disk {
+    /**
+     *   request - disk space guaranteed for the job
+     *   limit - max disk space the job can use
+     *   type - type of disk for the job
+     */
+    private Double _request;
+    private Double _limit;
+    private String _type;
```
nsinkov, line 14: shouldn't the type of this field be DiskType

jobclient/java/src/test/java/com/twosigma/cook/jobclient/JobTest.java
```diff
@@ -47,6 +47,7 @@
     // Constraints which could be used for any test.
     private Constraint _constraint1;
     private Constraint _constraint2;
+    private static final double DELTA = 1e-15;
```
nsinkov, line 50: usually called EPSILON (don't have to change though)

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
kathryn-zhou: changed

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
kathryn-zhou: changed

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
kathryn-zhou: done

jobclient/java/src/test/java/com/twosigma/cook/jobclient/JobTest.java
kathryn-zhou: ah yeah, changed

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
kathryn-zhou: No, we transform the enum version of DiskType to the String representation of _type because that is what we ultimately feed into the Cook payload. Also, this allows us to not have to update the enum DiskType each time we add support for a new disk type. Added a clarifying comment

dposada (approved):

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Disk.java
nsinkov: I would get rid of the DiskType enum and just use string
@dposada thoughts?

dposada: The enum seems like a nice convenience thing -- callers don't have to know or look up the valid string choices.

nsinkov (approved):

PR #1733, opened 2020-11-10
add workload fields to job client

nsinkov:

- add workload fields to application object on job

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Application.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Application.java
@@ -11,10 +11,20 @@ public class Application {
 
     final private String _name;
     final private String _version;
+    final private String _workloadClass;
+    final private String _workloadId;
+    final private String _workloadDetails;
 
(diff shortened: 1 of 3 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Application.java
```diff
@@ -25,16 +35,44 @@ public String getVersion() {
...
+        }
+        if (application._workloadDetails != null) {
+            object.put("workload-details", application._workloadDetails);
+        }
         return object;
     }
 
+    private static String tryGetString(JSONObject object, String key) throws JSONException {
+        return object.has(key) ? object.getString(key) : null;
+    }
```
dposada, line 68: Is there a more "central" place to put this than the `Application` class?

nsinkov: I'm ok not putting it somewhere more central - it's very simple. Can re-write as a lambda

dposada (approved):

PR #1739, opened 2020-11-13
Do not schedule nodes with unschedulable node-spec

scrosby:

- Do not schedule nodes with unschedulable node-spec

This allows node cordoning to work.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -25,7 +25,7 @@
              V1DeleteOptionsBuilder V1EmptyDirVolumeSource V1EnvVar V1EnvVarSource V1Event V1HostPathVolumeSource
              V1HTTPGetAction V1ObjectFieldSelector V1Node V1NodeAffinity V1NodeSelector V1NodeSelectorRequirement
              V1NodeSelectorTerm V1ObjectMeta V1ObjectReference V1Pod V1PodCondition V1PodSecurityContext V1PodSpec
-             V1PodStatus V1Probe V1ResourceRequirements V1Toleration V1Volume V1VolumeBuilder V1VolumeMount V1NodeSpec)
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -420,20 +420,26 @@
                                   (.getKey %))
                                taints-on-node)
           node-name (some-> node .getMetadata .getName)
+          ^V1NodeSpec node-unschedulable (some-> node .getSpec .getUnschedulable)
```
nsinkov, line 423: `^V1NodeSpec node-unschedulable`
this doesn't look right
that's a boolean, not a node spec

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -25,7 +25,7 @@
              V1DeleteOptionsBuilder V1EmptyDirVolumeSource V1EnvVar V1EnvVarSource V1Event V1HostPathVolumeSource
              V1HTTPGetAction V1ObjectFieldSelector V1Node V1NodeAffinity V1NodeSelector V1NodeSelectorRequirement
              V1NodeSelectorTerm V1ObjectMeta V1ObjectReference V1Pod V1PodCondition V1PodSecurityContext V1PodSpec
-             V1PodStatus V1Probe V1ResourceRequirements V1Toleration V1Volume V1VolumeBuilder V1VolumeMount)
+             V1PodStatus V1Probe V1ResourceRequirements V1Toleration V1Volume V1VolumeBuilder V1VolumeMount V1NodeSpec)
```
nsinkov, line 28: Don't need this probably. see below

scheduler/src/cook/kubernetes/api.clj
scrosby: Good catch. Thanks.

scheduler/src/cook/kubernetes/api.clj
scrosby: Yup. Gone. Thanks.

nsinkov (approved):

PR #1741, opened 2020-11-17
Increase limit for launch-task-num-threads

scrosby:

- Let us set a range up to 511 threads.

- Give is more flexibility in tuning how many threads, in particular when we have to handle slow API response times.


dposada (approved):

dposada:
LGTM, will merge when green

PR #1742, opened 2020-11-17
turn on calculate-effective-image for all jobs

nsinkov:

- 
- 
-

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -728,7 +728,7 @@
   "Transform the supplied job's image as necessary. e.g. do special transformation if checkpointing is enabled
   and an image transformation function is supplied."
   [{:keys [calculate-effective-image-fn] :as kubernetes-config} job-submit-time image {:keys [mode]} task-id]
-  (if (and mode calculate-effective-image-fn)
+  (if (and true calculate-effective-image-fn)
     (try
       ((util/lazy-load-var-memo calculate-effective-image-fn) kubernetes-config job-submit-time image)
       (catch Exception e


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -731,7 +736,7 @@
   (if (and mode calculate-effective-image-fn)
     (try
       ((util/lazy-load-var-memo calculate-effective-image-fn) kubernetes-config job-submit-time image)
-      (catch Exception e
+      (catch Throwable e
```
scrosby, line 739: This seems weird/bad. Why this? (I saw one of the commit mentioned this is to catch an assert error, which, to me, means the assert should probably be changed or moved elsewhere.)

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -795,7 +800,10 @@
         {:keys [custom-shell init-container set-container-cpu-limit? sidecar]} (config/kubernetes)
         checkpoint (calculate-effective-checkpointing-config job task-id)
         job-submit-time (tools/job->submit-time job)
-        image (calculate-effective-image (config/kubernetes) job-submit-time image checkpoint task-id)
+        pod-name (str task-id)
+        image (if (synthetic-pod? (or pod-name ""))
```
scrosby, line 804: When is pod-name empty? (Can a pod name in k8s be the empty string?!?)

nsinkov: I don't think it should ever be. But if it is, my code would be introducing an NPE
I'm fine dropping the `or`

scrosby: Lets do that, otherwise, a future reviewer can be left guessing when this might be blank. (And may see illusionary bugs when other code isn't robust to an empty pod name.)

scheduler/src/cook/kubernetes/api.clj

PR #1746, opened 2020-11-20
Scheduling and running disk jobs

kathryn-zhou:

- Enhance offers with disk information
- Implement disk constraint for bin packing
- Fit pods with disk information

- Enforce the disk request, limit, and type specifications we are allowing users to make


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -198,11 +198,14 @@
...
   (let [container-terminated-reason (some-> container-status .getState .getTerminated .getReason)
-        pod-status-reason (.getReason pod-status)]
+        pod-status-reason (.getReason pod-status)
+        pod-status-message (.getMessage pod-status)]
     (cond
       (= "OutOfMemory" pod-status-reason) :reason-container-limitation-memory
       (= "Error" container-terminated-reason) :reason-command-executor-failed
       (= "OOMKilled" container-terminated-reason) :reason-container-limitation-memory
+      (or (str/includes? pod-status-message "Pod ephemeral local storage usage exceeds the total limit of containers")
+          (str/includes? pod-status-message "The node was low on resource: ephemeral-storage")) :reason-container-limitation-disk
```
dposada, line 208: Should we make these configurable? Not sure if it's worth the trouble -- how likely are they to change? We could alternatively make the substring slightly shorter to (maybe?) reduce the likelihood of changes breaking this?

e.g.

```
ephemeral local storage usage exceeds
low on resource: ephemeral-storage
```

It would be nice to add a comment indicating which of these is for using more than your limit versus using more than your request.

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+
+(defn build-disk-host-constraint
+  "Constructs a disk-host-constraint.
+  The constraint prevents a job from running on a host that does not have correct disk type that the job requested
+  and prevents a job from running on a host that does not have enough disk space. Use a constraint for disk binpacking instead of
+  using Fenzo because disk is not considered a first class resource in Fenzo."
+  [job]
+  (let [pool-name (cached-queries/job->pool-name job)]
+    ; If the pool does not have enable-constraint set to true, return nil
+    (if (regexp-tools/match-based-on-pool-name (config/disk) pool-name :enable-constraint? :default-value false)
```
dposada, line 186:
```suggestion
    (when (regexp-tools/match-based-on-pool-name (config/disk) pool-name :enable-constraint? :default-value false)
```

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+              ; k8s VMs all support disk, so they will only support jobs whose users requested disk or consume a default disk request
+              vm-satisfies-constraint? (if job-disk-request
+                                         (>= (get vm-disk-type->space-available job-disk-type 0) job-disk-request)
+                                         ; if job does not have a disk-request, pass true to always ignore the disk-host-constraint
+                                         true)]
+          [vm-satisfies-constraint? (when-not vm-satisfies-constraint?
+                                      "VM does not have enough disk space of requested disk type")])
+        ; Mesos jobs cannot request disk. If VM is a mesos VM, constraint always passes
+        (let [vm-satisfies-constraint? true]
+          [vm-satisfies-constraint?])))))
```
dposada, line 176: Is the variable needed here? Seems superfluous

kathryn-zhou: yeah, will just return true then

scheduler/src/cook/kubernetes/controller.clj
kathryn-zhou: added comments and made the substrings shorter -- not sure if its worth making it configurable but will check with mcf

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -810,6 +839,16 @@
...
         gpu-model-requested (constraints/job->gpu-model-requested gpus job pool-name)
+        ; if disk config for pool has enable-constraint? set to true, add disk info to the job
+        enable-disk-constraint? (regexp-tools/match-based-on-pool-name (config/disk) pool-name :enable-constraint? :default-value false)
+        ; if user did not specify disk request, use default on pool
+        disk-request (when enable-disk-constraint?
+                       (or (-> resources :disk :request)
+                           (regexp-tools/match-based-on-pool-name (config/disk) pool-name :default-request)))
+        disk-limit (when enable-disk-constraint? (-> resources :disk :limit))
+        ; if user did not specify disk type, use default on pool
+        disk-type (when enable-disk-constraint? (constraints/job->disk-type-requested (-> resources :disk :type) pool-name))
```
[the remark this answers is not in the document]
kathryn-zhou: yep, done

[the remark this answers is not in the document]
kathryn-zhou: I think it would be better to leave these lines here because outside of this `let` block, some of the variables like `resources` get changed. I actually tried to move these lines down to where they were consumed but was not able to get the correct disk specs set on the job

scheduler/config-k8s.edn
```diff
@@ -95,7 +95,9 @@
                         :resource-requirements {:cpu-request 0.1
                                                 :cpu-limit 0.1
                                                 :memory-request 128.0
-                                                :memory-limit 128.0}}}
+                                                :memory-limit 128.0}}
+              :disk-type-node-label-name [{:pool-regex "^k8s-alpha$" 
+                                           :disk-node-label "cloud.google.com/gke-boot-disk"}]}
```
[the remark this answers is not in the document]
kathryn-zhou: yeah, that makes sense. I think this was first a separate section because we assumed that the node-label would be the same for different pool-regex's, but I have now added it to the disk section

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -460,21 +466,40 @@
...
+  "Given a pool name, get the disk-type-label-name that will be used as a nodeSelector label"
+  [pool-name]
+  (regexp-tools/match-based-on-pool-name
+    (config/disk)
+    pool-name
+    :disk-node-label
+    :default-value "cloud.google.com/gke-boot-disk"))
+
+(defn add-disk-type-to-resource-map
+  "Given a map from node-name->resource-type->capacity, set the disk capacity to type->amount"
```
nsinkov, line 479: I'd rename the function to force-disk-type-in-resource-map

And then change the docstring to:
"Given a map from node-name->resource-type->capacity, set the disk capacity to type->amount or remove the disk resource type from the map if disk-type is not provided"

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -488,8 +513,12 @@
                                                                           convert-resource-map))
                                                                 containers)
                                         resource-map (apply merge-with + container-requests)
-                                        gpu-model (some-> pod .getSpec .getNodeSelector (get "cloud.google.com/gke-accelerator"))]
-                                    (add-gpu-model-to-resource-map gpu-model resource-map))))
+                                        ^V1NodeSelector nodeSelector (some-> pod .getSpec .getNodeSelector)
+                                        gpu-model (some-> nodeSelector (get "cloud.google.com/gke-accelerator"))
+                                        disk-type (some-> nodeSelector (get (pool->disk-type-label-name pool-name)))
+                                        update-gpu-in-res-map (fn [res-map] (add-gpu-model-to-resource-map gpu-model res-map))
```
nsinkov, line 519: Instead of having update-gpu-in-res-map and update-disk-in-res-map you could replace:

(-> resource-map update-gpu-in-res-map update-disk-in-res-map)

with

(->> resource-map
  (add-gpu-model-to-resource-map gpu-model)
  (add-disk-type-to-resource-map disk-type))

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
nsinkov: I'm ok with these here

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -47,17 +47,17 @@
   [node-name->resource-map resource-keyword]
   (->> node-name->resource-map vals (map resource-keyword) (filter some?) (reduce +)))
 
-(defn total-gpu-resource
+(defn total-text->scalar-resource
```
nsinkov, line 50: should this be called resource-count-map ?
or at least get-text->scalar-resource-total 

seems like "total-text" is converted to "scalar-resource"

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -92,6 +92,16 @@
   (let [previous-hosts (job->previous-hosts-to-avoid job)]
     (->novel-host-constraint job previous-hosts)))
 
+(defn job->disk-type-requested
```
nsinkov, line 95: you're not passing in or using a job, right?
maybe just call the function disk-type-requested

and it doesn't seem like you need the let, just nested or:

(or (get (regexp-tools/match-based-on-pool-name (config/disk) pool-name :type-map) disk-type-requested)
        (or disk-type-from-user
                                (regexp-tools/match-based-on-pool-name (config/disk) pool-name :default-type)))

scheduler/src/cook/kubernetes/compute_cluster.clj
kathryn-zhou: hm, reasoning for this is because the parallel for scalar resources is `total-scalar-resource`, and disk/gpu resources are represented with type `text->scalar` instead. however, I do agree that this current name looks like "total-text" -> "scalar-resource". @dposada what do you think?

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
dposada: Sounds good

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: how about `total-map-resource`?

scheduler/src/cook/kubernetes/api.clj
kathryn-zhou: done, thanks for the suggestion!!

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: `disk-type-requested` is used in 2 places, so that is why I opted to use a let block 

naming convention was consistent with `job->gpu-model-requested', but I am okay with changing it to just `disk-type-requested`

scheduler/src/cook/config.clj
```diff
@@ -167,6 +167,8 @@
         (throw (ex-info (str "Valid disk types for pool-regex " pool-regex " is not defined") entry)))
       (when-not default-type
         (throw (ex-info (str "Default disk type for pool-regex " pool-regex " is not defined") entry)))
+      (when-not default-request
+        (throw (ex-info (str "Default disk request for pool-regex " pool-regex " is not defined") entry)))
```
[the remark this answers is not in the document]
kathryn-zhou: done

scheduler/src/cook/kubernetes/compute_cluster.clj
kathryn-zhou: sounds good to me

scheduler/src/cook/kubernetes/api.clj
kathryn-zhou: agree on docstring, but similar function for gpus is `add-gpu-model-to-resource-map`, would you prefer changing that function name to `force-gpu-model-in-resource-map` too?

nsinkov: interesting. add-gpu-model-to-resource-map doesn't delete the :gpus key, just points it to an empty map

I think they should both be called "force" and both delete the key. I like not having the key better than pointing to an empty map

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -92,6 +92,16 @@
   (let [previous-hosts (job->previous-hosts-to-avoid job)]
     (->novel-host-constraint job previous-hosts)))
 
+(defn disk-type-requested
+  "Get disk type requested from job or use default disk type on pool"
+  [disk-type-from-user pool-name]
+  ; If user did not specify desired disk type, use the default disk type on the pool
+  (let [disk-type-for-job (or disk-type-from-user
+                                (regexp-tools/match-based-on-pool-name (config/disk) pool-name :default-type))]
```
dposada, line 100: Indentation seems off on this line

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+  (job-constraint-name [this] (get-class-name this))
+  (job-constraint-evaluate
+    [this _ vm-attributes]
+    (job-constraint-evaluate this nil vm-attributes []))
+  (job-constraint-evaluate
+    [_ _ vm-attributes _]
+    (let [k8s-vm? (= (get vm-attributes "compute-cluster-type") "kubernetes")]
+      (if k8s-vm?
+        (let [vm-disk-type->space-available (get vm-attributes "disk")
+              ; k8s VMs all support disk, so they will only support jobs whose users requested disk or consume a default disk request
```
dposada, line 167: The word "only" is confusing me in this comment, should it be "either"?

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+  and prevents a job from running on a host that does not have enough disk space. Use a constraint for disk binpacking instead of
+  using Fenzo because disk is not considered a first class resource in Fenzo."
+  [job]
+  (let [pool-name (cached-queries/job->pool-name job)]
+    ; If the pool does not have enable-constraint set to true, return nil
+    (when (regexp-tools/match-based-on-pool-name (config/disk) pool-name :enable-constraint? :default-value false)
+      (let [; If the user did not specify a disk request, use the default request amount for the pool
+            job-disk-request (or (-> job util/job-ent->resources :disk :request)
+                                 ; if disk config does not have default-request, use default-request of 10GiB
+                                 (regexp-tools/match-based-on-pool-name (config/disk) pool-name :default-request :default-value 10240))
```
dposada, line 189: This `or` form is also present in `task-metadata->pod`, can we extract a function for it, e.g. `job-resources->disk-request`?

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+            job-disk-type (when job-disk-request
+                            (disk-type-requested (-> job util/job-ent->resources :disk :type) pool-name))]
```
dposada, line 191: Same thing here -- should we extract `job-resources->disk-type` as a function and use it both here and in `task-metadata->pod`?

scheduler/src/cook/scheduler/constraints.clj
```diff
@@ -138,6 +148,49 @@
...
+        (->disk-host-constraint job-disk-request job-disk-type)))))
```
dposada, line 192: Would it make sense to wrap this in `(when job-disk-request ...)` so that we return `nil` from this function for jobs that have no explicit or defaulted disk request, avoiding creating the constraint altogether?

scheduler/src/cook/kubernetes/api.clj
kathryn-zhou: done

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: done

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: fixed

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: I think I originally meant that k8s vms will not support jobs that don't have a disk request, but reworded to be clearer

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: this is already wrapped in the condition `(when enable-constraint?)`, and if `enable-constraint?` is true on a pool, that also means the job will have a disk-request (by default)

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: done

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: is it necessary just to extract `(-> job util/job-ent->resources :disk :type)` as a separate function?

kathryn-zhou: changed, as discussed on call

scheduler/src/cook/scheduler/constraints.clj
kathryn-zhou: updated comments, as discussed on Skype call

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -450,31 +456,49 @@
...
-  - if the amount of gpus on the node is 0, set the gpus capacity to an empty map"
+(defn force-gpu-model-in-resource-map
+  "Given a map from node-name->resource-type->capacity, set the gpus capacity to model->amount
+  or remove the gpus resource type from the map if gpu-model is not provided"
   [gpu-model {:keys [gpus] :as resource-map}]
-  (let [gpu-model->count (if (and gpu-model (pos? gpus))
-                           {gpu-model gpus}
-                           {})]
-    (assoc resource-map :gpus gpu-model->count)))
+  (if (and gpus gpu-model)
```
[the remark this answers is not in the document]
kathryn-zhou: unintentional, changed that back

dposada (approved):

nsinkov (approved):

dposada:
Will merge when green

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -487,9 +511,30 @@
...
+                                        disk-type (some-> nodeSelector (get (pool->disk-type-label-name pool-name)))]
+                                    (log/info "Resource map after adding gpu-model and/or disk-type: "
+                                              (->> resource-map
+                                                   (force-gpu-model-in-resource-map gpu-model)
+                                                   (force-disk-type-in-resource-map disk-type)))
+                                    (->> resource-map
+                                         (force-gpu-model-in-resource-map gpu-model)
+                                         (force-disk-type-in-resource-map disk-type)))))
+                           ; remove nil resource-maps from collection before deep-merge-with
+                           (remove nil?)
```
dposada, line 537: Would be nice to include in the comment the info about why resource maps can sometimes be nil

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -904,6 +958,14 @@
       (add-node-selector pod-spec "cloud.google.com/gke-accelerator" gpu-model-requested)
       ; GKE nodes with GPUs have gpu-count label, so synthetic pods need a matching node selector
       (add-node-selector pod-spec "gpu-count" (-> gpus int str)))
+    (when disk-request
+      ; do not add disk request/limit to synthetic pod yaml
```
dposada, line 962: Would be nice to explain why

dposada (approved):

dposada:
Will merge when green

PR #1747, opened 2020-11-20
Add resource request and limit to init-container in pod

scrosby:

- When we setup the init container, attach resources to it.

When K8S is run with quotas, any container without attached resources will be rejected at submission. We don't want that, so attach resources to the init container.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -629,6 +629,7 @@
 (defn- add-as-decimals
   "Takes two doubles and adds them as decimals to avoid floating point error. Kubernetes will not be able to launch a
    pod if the required cpu has too much precision. For example, adding 0.1 and 0.02 as doubles results in 0.12000000000000001"
+  ; If you get an error like IllegalArgumentException "No suffix for exponent-18" in JSON serialization. Use this function.
   [a b]
   (double (+ (bigdec a) (bigdec b))))
 
@@ -904,7 +905,7 @@
                                (or (get-in sidecar [:resource-requirements :memory-request] 0)
                                    (get-in sidecar [:resource-requirements :memory-limit] 0))
                                0))
-              sidecar-cpu (+ cpus
+              sidecar-cpu (add-as-decimals cpus
                              (if use-cook-sidecar?
                                (or (get-in sidecar [:resource-requirements :cpu-request] 0)
                                    (get-in sidecar [:resource-requirements :cpu-limit] 0))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -900,16 +900,22 @@
     (when use-cook-init?
       (when-let [{:keys [command image]} init-container]
         (let [container (V1Container.)
-              sidecar-mem (+ computed-mem
-                             (if use-cook-sidecar?
-                               (or (get-in sidecar [:resource-requirements :memory-request] 0)
-                                   (get-in sidecar [:resource-requirements :memory-limit] 0))
-                               0))
-              sidecar-cpu (add-as-decimals cpus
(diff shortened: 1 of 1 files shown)


nsinkov:
We should refactor this into a function:

nsinkov:
let's refactor 
`    (.putRequestsItem resources "memory" (double->quantity (* memory-multiplier computed-mem)))
    (.putLimitsItem resources "memory" (double->quantity (* memory-multiplier computed-mem)))
    (when set-container-cpu-limit?
      ; Some environments may need pods to run in the "Guaranteed"
      ; QoS, which requires limits for both memory and cpu
      (.putLimitsItem resources "cpu" (double->quantity cpus)))`

into a function and use it in the 3 places: main container, sidecar, init container
so have limits in init container too

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -900,6 +920,8 @@
           (.setVolumeMounts container (filterv some? (concat [(init-container-workdir-volume-mount-fn false)
                                                               (scratch-space-volume-mount-fn false)]
                                                              init-container-checkpoint-volume-mounts)))
+          (set-mem-cpu-resources resources sidecar-mem sidecar-mem sidecar-cpu sidecar-cpu)
```
nsinkov, line 923: use the total-memory-request and total-memory-limit from above

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -890,7 +899,18 @@
     ; init container
     (when use-cook-init?
       (when-let [{:keys [command image]} init-container]
-        (let [container (V1Container.)]
+        (let [container (V1Container.)
+              sidecar-mem (+ computed-mem
```
nsinkov, line 903: call this total-memory-request and total-memory-limit and add those separately

scrosby: Updated.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -762,6 +763,19 @@
            pod-labels-from-job-labels
            pod-labels-from-job-application)))
 
+(defn set-mem-cpu-resources
+  "Given a resources object and a CPU and memory request and limit, update the resources object to reflect the
+  desired requests and limits."
+  [resources memory-request memory-limit cpu-request cpu-limit]
```
dposada, line 769: Can we add a type hint for `resources`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -890,7 +899,24 @@
...
-        (let [container (V1Container.)]
+        (let [container (V1Container.)
+              total-memory-request (+ computed-mem
+                                      (if use-cook-sidecar?
+                                        (get-in sidecar [:resource-requirements :memory-request])
+                                        0))
+              total-memory-limit (+ computed-mem
+                                    (if use-cook-sidecar?
+                                      (get-in sidecar [:resource-requirements :memory-limit])
+                                      0))
```
dposada, line 910: Shouldn't we use `add-as-decimals` here too?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -890,7 +899,24 @@
...
+              total-cpu-request (add-as-decimals cpus
+                                                 (if use-cook-sidecar?
+                                                   (get-in sidecar [:resource-requirements :cpu-request])
+                                                   0))
+              total-cpu-limit (add-as-decimals cpus
+                                               (if use-cook-sidecar?
+                                                 (get-in sidecar [:resource-requirements :cpu-limit])
+                                                 0))
```
dposada, line 918: Can we define a function in the `let` for this instead of repeating the code 4 times?

dposada (approved):

PR #1752, opened 2020-11-30
Add ability to convert JSON->Builder, set fields when JSONizing

brianbao:

- Add `throws JSONException` to `parseFromJSON`
- Set `status` and `priority` when creating a `Builder` object  from a `Job`.
- Allow users to create a `Builder` object from a `JSONObject`.

Forgot to throw the exception up. Also if you create a Job from a JSON object or vice versa it's incomplete. It makes more sense from an API perspective to allow users to parse JSON -> `Builder`, then be able to make further changes to the builder object. You can't do that with the existing code since Job is effectively immutable after built.

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -171,6 +171,8 @@ final public class Job {
             setPool(job.getPool());
             setLabels(job.getLabels());
             setDatasets(job.getDatasets());
+            setStatus(job.getStatus());
+            setPriority(job.getPriority());
             setDisk(job.getDisk());
             addConstraint(job.getConstraints());
             if (job.isMeaCulpaRetriesDisabled()) {
@@ -1084,6 +1086,7 @@ final public class Job {
         if (job.getDisk().shouldIncludeInJSON()) {
             object.put("disk", job.getDisk().toJSONObject());
         }
+        object.put("status", job.getStatus().name());
         object.put("priority", job.getPriority());
         object.put("max_retries", job.getRetries());
         object.put("disable_mea_culpa_retries", job.isMeaCulpaRetriesDisabled());
@@ -1202,7 +1205,9 @@ final public class Job {
             jobBuilder.setExecutor(json.getString("executor"));
         }
         jobBuilder.setPriority(json.getInt("priority"));
-        jobBuilder.setStatus(Status.fromString(json.getString("status")));
+        if (json.has("status")){
+            jobBuilder.setStatus(Status.fromString(json.getString("status")));
+        }
         if (json.has("disable_mea_culpa_retries") && json.getBoolean("disable_mea_culpa_retries")) {
             jobBuilder.disableMeaCulpaRetries();
         } else {
@@ -1265,7 +1270,9 @@ final public class Job {
                 }
             }
         }
-        jobBuilder.addInstances(Instance.parseFromJSON(json.getJSONArray("instances"), decorator));
+        if (json.has("instances")) {
+            jobBuilder.addInstances(Instance.parseFromJSON(json.getJSONArray("instances"), decorator));
+        }
         if (json.has("application")) {
             JSONObject applicationJson = json.getJSONObject("application");
             jobBuilder.setApplication(Application.parseFromJSON(applicationJson));
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -1184,7 +1184,8 @@ final public class Job {
      * @return a single {@link Job}.
      * @throws JSONException
      */
-    public static Job parseFromJSON(JSONObject json, InstanceDecorator decorator){
+    public static Job parseFromJSON(JSONObject json, InstanceDecorator decorator)
+            throws JSONException {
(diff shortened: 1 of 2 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -1264,7 +1268,9 @@ public static Job parseFromJSON(JSONObject json, InstanceDecorator decorator){
                 }
             }
         }
-        jobBuilder.addInstances(Instance.parseFromJSON(json.getJSONArray("instances"), decorator));
+        if (json.has("instances")) {
```
brianbao, line 1271: I'm not sure why this is required as a field.

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -1201,7 +1205,9 @@ public static Job parseFromJSON(JSONObject json, InstanceDecorator decorator){
             jobBuilder.setExecutor(json.getString("executor"));
         }
         jobBuilder.setPriority(json.getInt("priority"));
-        jobBuilder.setStatus(Status.fromString(json.getString("status")));
+        if (json.has("status")){
```
brianbao, line 1208: Not sure why this was required in the JSON either, builder.build() will set a default status of INITIALIZED on the job.

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -1084,6 +1086,7 @@ public static JSONObject jsonizeJob(Job job)
         if (job.getDisk().shouldIncludeInJSON()) {
             object.put("disk", job.getDisk().toJSONObject());
         }
+        object.put("status", job.getStatus().name());
```
dposada, line 1089: This doesn't seem right -- we don't want to send status on job submissions, right?

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
dposada: I'm not following, where is it required?

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
dposada: I'm not following, where is it required?

brianbao: Oh if you have a JSON blob that doesn't populate "instances", this line will fail. In fact there's even a hacky workaround in the tests to [convert a Job to JSON](https://github.com/twosigma/Cook/blob/master/jobclient/java/src/test/java/com/twosigma/cook/jobclient/JobTest.java#L82).

brianbao: Basically if you have a JSON payload and you call `parseFromJSON`, you won't be able to get a Job object unless the JSON has defined specific fields, e.g. "uuid", "mem", "cpus", just to name a few.

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
brianbao: I'm ok with removing this but you would need to remove the requirement that status is set when parsing from JSON -> Job.

Otherwise you would have an issue where you have a `Job job`, you initialize `JSONObject json = jsonizeJob(job)`, when you call `parseFromJson(json)`, you'll run into [an issue parsing 'status' from the JSON](https://github.com/twosigma/Cook/blob/master/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java#L1204).

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
brianbao: See above.

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
brianbao: I edited this to be able to parse from JSON -> Builder instead. This API is much more intuitive for users who need to go from JSON -> Job, but may need to make edits in the middle to the Job.

In our use case we have a map of parameters, which we want to merge with another map of parameters. To accomplish that without this CR we would have to do JSON -> Job -> Builder -> Job.

jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -722,6 +724,165 @@ public Builder setDatasets(JSONArray datasets) {
...
+         * </code>
+         * </pre>
+         *
+         * @param json {@link JSONObject} specifies a JSONObject with Builder parameters.
+         * @param decorator  specifies an instance decorator expected to decorate instances parsed from JSON string.
+         *                   If it is null, it will do nothing with it.
+         * @return {@link Builder}.
+         * @throws JSONException
+         */
+        public Builder parseFromJSON(JSONObject json, InstanceDecorator decorator)
```
brianbao, line 750: Basically this function doesn't enforce that any specific field is set, if any of them are set, the Builder will do the right thing and set the right field. Fields are only enforced on `build()` (e.g. if cpus is not defined, that will be caught by `build()`).

dposada (approved):

PR #1753, opened 2020-12-02
Allow disk request/limit to be an int

kathryn-zhou:

- Accept int values for disk request and limit

- Improve user experience since people may want to just specify int values for disk request and limit

### DIFF
--- a/scheduler/test/cook/test/rest/api.clj
+++ b/scheduler/test/cook/test/rest/api.clj
@@ -470,6 +470,14 @@
                              :headers {"Content-Type" "application/json"}
                              :authorization/user "dgrnbrg"
                              :body-params {"jobs" [(job {"request" 2.5})] "pool" "test-pool"}})))))
+      (testing "Int request of disk valid"
+        (is (= 201
+               (:status (h {:request-method :post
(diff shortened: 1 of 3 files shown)


scheduler/test/cook/test/rest/api.clj
```diff
@@ -470,6 +470,14 @@
...
                              :authorization/user "dgrnbrg"
                              :body-params {"jobs" [(job {"request" 2.5})] "pool" "test-pool"}})))))
+      (testing "Int request of disk valid"
+        (is (= 201
+               (:status (h {:request-method :post
+                            :scheme :http
+                            :uri "/jobs"
+                            :headers {"Content-Type" "application/json"}
+                            :authorization/user "dgrnbrg"
+                            :body-params {"jobs" [(job {"request" 20.0})] "pool" "test-pool"}})))))
```
dposada, line 480: Shouldn't the request be an int for this test case?

kathryn-zhou: oops! a typo, just changed

dposada (approved):

PR #1754, opened 2020-12-03
Specify that disk request and limit are in MiB

kathryn-zhou:

- Add that disk request and limit are in MiB in the docstrings

- Clarification for users


dposada (approved):

dposada:
Will merge when green

PR #1765, opened 2020-12-17
Adds location to compute cluster

dposada:

- adding a `location` field to the compute cluster

To eventually support constraining job instances to specific locations, e.g. specific cloud regions.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -185,7 +185,8 @@
            compute-cluster-config/base-path
            compute-cluster-config/ca-cert
            compute-cluster-config/state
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -324,6 +338,24 @@
                false)
     false))
 
+(defn config=?
+  "Returns true if we can consider current-config and
```
nsinkov, line 342: nitpick, but mind not wrapping so hard. like have "Returns true if we can consider current-config and new-config to be equal." on one line

dposada: Done

nsinkov (approved):

PR #1766, opened 2020-12-18
Sends synthetic pod to same location as previous checkpoint

dposada:

If checkpointing is enabled, sending the synthetic pod for a new instance to the same location (e.g. region) as the checkpointed instance.

To take advantage of data locality for the checkpoint data.

## What's not in this PR?

This PR only affects synthetic pods, not the actual matching constraint to enforce where the real job pods run. That will come in a subsequent patch.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -973,16 +973,61 @@
                                        {:job-uuid->reserved-host (apply dissoc job-uuid->reserved-host matched-job-uuids)
                                         :launched-job-uuids (into matched-job-uuids launched-job-uuids)})))
 
+(defn job->preferred-compute-clusters
+  "Given a job and a collection of compute clusters, returns the
+  subset of compute clusters that the job would prefer to run
+  (and therefore, autoscale) on"
+  [{:keys [job/checkpoint job/instance]} compute-clusters]
+  (if (and checkpoint instance)
+    ; If checkpointing is enabled, we want to run the new instance in the
+    ; same location as the checkpointed instance to take advantage of data
+    ; locality for the checkpoint data
+    (let [{{:keys [compute-cluster/cluster-name]} :instance/compute-cluster}
+          (->> instance
+               (sort-by :instance/start-time)
+               last)
+          ; This is a list, but it should never
+          ; contain more than 1 compute cluster
+          previous-compute-clusters
+          (filter
+            (fn [{:keys [name]}]
+              (= name cluster-name))
+            compute-clusters)]
+      (if (seq previous-compute-clusters)
+        (let [previous-location
+              (-> previous-compute-clusters
+                  first
+                  :cluster-definition
+                  :config
+                  :location)]
+          (filter
+            (fn [{{{:keys [location]} :config} :cluster-definition}]
+              (= location previous-location))
+            compute-clusters))
+        ; If the previous instance's compute cluster name is not
+        ; present in the list of compute clusters passed, there's
+        ; not much we can do
+        compute-clusters))
+    compute-clusters))
+
 (defn distribute-jobs-to-compute-clusters
   "Given a collection of pending jobs and a collection of
   compute clusters, distributes the jobs amongst the compute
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -973,16 +973,61 @@
...
+  [{:keys [job/checkpoint job/instance]} compute-clusters]
+  (if (and checkpoint instance)
+    ; If checkpointing is enabled, we want to run the new instance in the
+    ; same location as the checkpointed instance to take advantage of data
+    ; locality for the checkpoint data
+    (let [{{:keys [compute-cluster/cluster-name]} :instance/compute-cluster}
+          (->> instance
+               (sort-by :instance/start-time)
+               last)
+          ; This is a list, but it should never
```
nsinkov, line 989: maybe delete this "previous-compute-clusters" and replace with if-let

(let [{{:keys [compute-cluster/cluster-name]} :instance/compute-cluster}
      (->> instance
           (sort-by :instance/start-time)
           last)
      compute-cluster->location #(-> %
                                   :cluster-definition
                                   :config
                                   :location)]  
  (if-let [previous-location
           (->> compute-clusters
                (filter #(= (:name %) cluster-name))
                first
                compute-cluster->location)]
    (filter
      #(= (compute-cluster->location %) previous-location)
      compute-clusters)
    ; If the previous instance's compute cluster name is not
    ; present in the list of compute clusters passed, there's
    ; not much we can do
    compute-clusters))

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -973,16 +973,61 @@
...
   [pending-jobs compute-clusters]
-  (group-by (fn [job]
-              (nth compute-clusters
-                   (-> job :job/uuid hash (mod (count compute-clusters)))))
+  (group-by (fn choose-compute-cluster-for-autoscaling
+              [{:keys [job/uuid] :as job}]
+              (let [preferred-compute-clusters
+                    (job->preferred-compute-clusters job compute-clusters)
+                    compute-clusters-to-consider
+                    (if (seq? preferred-compute-clusters)
```
nsinkov, line 1026: when will the "seq?" check ever be false?

I don't think you need an extra check here, just do "nth" on "preferred-compute-clusters"

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -973,16 +973,61 @@
...
+              (-> previous-compute-clusters
+                  first
+                  :cluster-definition
+                  :config
+                  :location)]
+          (filter
+            (fn [{{{:keys [location]} :config} :cluster-definition}]
+              (= location previous-location))
+            compute-clusters))
+        ; If the previous instance's compute cluster name is not
```
nsinkov, line 1007: if we ever don't match the location we want here, what will happen when we try to match the offer?
will we launch synthetic pods but then never match any offers?

dposada: The matching is going to have to be implemented as a "yes"/"no" constraint. So if there are no offers in the same location as the previous instance, then we won't be able to match the job to anything.

scheduler/src/cook/scheduler/scheduler.clj
dposada: Good call, done

scheduler/src/cook/scheduler/scheduler.clj
dposada: You're right, done

nsinkov:
Ok, so if we don't find a correct cluster when launching synthetic pods,
and just pick one using the old hash, then we'll scale up and ignore those
offers.

dposada:
Yes, unless the offers are matched to other jobs.

dposada:
@nsinkov This is ready to merge if you are good with it

nsinkov:
I'll look at it in a bit, but can merge now if you don't want to wait

dposada:
@nsinkov No problem, let me know when you've had a chance to look at it

dposada:
@nsinkov This is ready for another review pass

PR #1767, opened 2020-12-21
Add disk fields to cs submit

kathryn-zhou:

- Add disk request, limit, and type to cs submit

- Allow users to request disk via cs submit


integration/tests/cook/test_cli.py
```diff
@@ -1309,6 +1309,37 @@ def test_show_disk(self):
                 self.assertIn("Disk Limit", cli.stdout(cp))
                 self.assertIn("Disk Type", cli.stdout(cp))
 
+    def test_submit_with_disk(self):
```
kathryn-zhou, line 1312: added a new test for the functionality with `cs submit` but another option is just modifying the current test (above) that is used for `cs show`. what do you think?

dposada: I think it's fine to have a separate one

integration/tests/cook/test_cli.py
```diff
@@ -1309,6 +1309,37 @@ def test_show_disk(self):
...
+        settings_dict = util.settings(self.cook_url)
+        # disk_config_list is a list of regexp's with pool regex, valid disk types, default disk type, and max requestable size of disk
+        disk_config_list = settings_dict.get("pools", {}).get("disk", [])
+        if not disk_config_list:
+            self.skipTest("There are no pools that support disk")
+
+        # Disk may not be configured for COOK_TEST_DEFAULT_POOL, so we iterate over all active pools to find a pool where we can run this test.
+        active_pools, _ = util.active_pools(self.cook_url)
+        for pool in active_pools:
```
dposada, line 1321: Do we need to run this on multiple pools? Seems like running it on just one pool would be sufficient?

kathryn-zhou: changed to just using the default pool

dposada (approved):

dposada:
Will merge when green

PR #1768, opened 2020-12-21
v1.50.0

kathryn-zhou:

- updating changelog
- bumping version

To release Cook Scheduler.

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -13,7 +13,7 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 ;;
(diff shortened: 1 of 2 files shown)


scheduler/project.clj
```diff
@@ -13,7 +13,7 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 ;;
-(defproject cook "1.50.0"
+(defproject cook "1.50.0-SNAPSHOT"
```
dposada, line 16: This should be 1.50.1-SNAPSHOT

PR #1778, opened 2021-01-19
Validates job resources by node type

dposada:

Adding a new submission plugin that validates resources across node types.

So that Cook can apply different resource limits based on the node-type, node-family, and cpu-architecture constraints that users can specify.

### DIFF
--- a/scheduler/src/cook/plugins/submission.clj
+++ b/scheduler/src/cook/plugins/submission.clj
@@ -141,55 +141,57 @@
             (regexp-tools/match-based-on-pool-name
               (:non-gpu-jobs limits-config)
               pool-name
-              :node-type-lookup)
-            constraint-pattern-fn
-            (fn [attribute-of-interest]
-              (->> constraints
-                   (filter (fn [constraint]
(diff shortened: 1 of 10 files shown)


scheduler/src/cook/plugins/submission.clj
```diff
@@ -112,15 +113,129 @@
...
+
+(defrecord JopShapeValidationPlugin
+  []
+  JobSubmissionValidator
+
+  (check-job-submission-default [_]
+    {:status :accepted})
+
+  (check-job-submission [_ {:keys [constraints cpus disk gpus mem]} pool-name]
+    (if (and gpus (pos? gpus))
```
nsinkov, line 148: so why are gpu jobs separate?

dposada: Because the GPU model and number of GPUs requested become inputs to the validation in addition to node-type, node-family, and cpu-architecture.

nsinkov: yeah, that happens during node matching?
why is the node type matching ignored if you ask for GPUs?

PR #1784, opened 2021-01-27
Add memory limit job label

kathryn-zhou:

- Let users set on configurable job label if they want to allow memory usage above memory request

- We found use cases of memory limit

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -896,6 +896,33 @@ class CookTest(util.CookTest):
         command = self.memory_limit_script_command(count=2048)
         self.memory_limit_exceeded_helper(command, 'mesos', mem=32)
 
+    @unittest.skipUnless(util.using_kubernetes())
+    def test_memory_limit(self):
+        command = "python -c 'MB = 1024 * 1024 ; a = \"a\" * (25 * MB)'"
+
+        # job requests 20MB of memory, does not allow memory usage above request, and allocates 25MB of memory
+        job_uuid1, resp1 = util.submit_job(self.cook_url, mem=20, command=command)
+
+        # job requests 20MB of memory, allows memory usage above request, and allocates 25MB of memory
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -943,8 +943,8 @@
     (.setTty container true)
     (.setStdin container true)
 
-    ; add memory limit if config flag set-memory-limit? is True
-    (set-mem-cpu-resources resources computed-mem (when (:set-memory-limit? (config/kubernetes)) computed-mem) cpus cpus)
+    ; add memory limit if user sets job label to allow memory usage above request to True
+    (set-mem-cpu-resources resources computed-mem (when-not (get labels (:memory-limit-job-label-name (config/kubernetes)) false) computed-mem) cpus cpus)
```
dposada, line 947: For readability, I'd favor:

1. pulling this `(when-not (get labels (:memory-limit-job-label-name (config/kubernetes)) false) computed-mem)` out into a `let` binding, and / or
2. breaking up this line across a few lines

integration/tests/cook/test_basic.py
```diff
@@ -898,12 +893,7 @@ def test_memory_limit_exceeded_cook_script(self):
 
     @pytest.mark.memlimit
     def test_memory_limit_exceeded_mesos_script(self):
-        settings_dict = util.settings(self.cook_url)
-        set_memory_limit = settings_dict.get("kubernetes", {}).get("set-memory-limit?", True)
-        if util.using_kubernetes() and not set_memory_limit:
-            command = self.infinite_memory_script_command(count=2048)
-        else:
-            command = self.memory_limit_script_command(count=2048)
+        command = self.memory_limit_script_command(count=2048)
```
dposada, line 896: (this is not about this line, just didn't know where else to put this question)

Are we adding new integration test(s) in this PR or no?

PR #1787, opened 2021-02-08
Enable local k8s

scrosby:

- Various changes to enable local kubernetes operation.
- Support for a supplemental group ignore list for k8s.

We need these changes for integration tests to pass in non-GKE environments.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -3016,7 +3016,7 @@ class CookTest(util.CookTest):
         self.assertEqual('failed', job['instances'][0]['status'], job)
         self.assertEqual('Invalid task', job['instances'][0]['reason_string'], job)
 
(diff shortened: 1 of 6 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -3016,6 +3016,11 @@ def test_pod_submission_failed(self):
         self.assertEqual('failed', job['instances'][0]['status'], job)
         self.assertEqual('Invalid task', job['instances'][0]['reason_string'], job)
 
+    @unittest.skipUnless(util.enable_unspecified_pool(), "Test disabled in this run as default image isn't appropriate for default pool.")
```
dposada, line 3019: Can we make this a `skipIf` instead of a `skipUnless` to make it slightly harder to skip this test?

scrosby: I don't like this as it we're adding a double negative. How about I change the sense of the flag. Rename it to disable_unspecified_pool_test (and the corresponding environmental variable) and plumb through.

dposada (approved):

PR #1788, opened 2021-02-09
Supports a default k8s pool

dposada:

Adding support for the default pool being a k8s pool.

So that Cook can be configured with a k8s pool as the default.


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1062,11 +1062,6 @@
     (.setRestartPolicy pod-spec "Never")
 
     (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
-    ; TODO:
-    ; This will need to change to allow for setting the toleration
-    ; for the default pool when the pool was not specified by the
-    ; user. For the time being, this isn't a problem only if / when
-    ; the default pool is not being fed offers from Kubernetes.
     (when pool-name
```
nsinkov, line 1065: we don't need this when anymore

dposada: good call, done

PR #1790, opened 2021-02-11
extend plugin infrastructure to support new subcommands

rmanyari:

Extend CLI plugin infrastructure to support new subcommands

We'd like to be able to support additional subcommands outside of the core already provided

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -175,6 +175,14 @@ and must return a dict containing two fields: `data` and `offset`.
 
 See the integration tests for simple examples of the subcommand plugins.
 
+#### Plugin based subcommands
+
+In addition of plugins for core subcommands, we can add new subcommands through the plugin module.
+
+Simply extend SubCommandPlugin and implement register, run and name.
+
+See the integration tests for a concrete example.
+
 ### Examples
 
 Simple job creation:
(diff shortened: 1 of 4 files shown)


cli/README.md
```diff
@@ -175,6 +175,14 @@ and must return a dict containing two fields: `data` and `offset`.
 
 See the integration tests for simple examples of the subcommand plugins.
 
+#### Plugin based subcommands
+
+In addition of plugins for core subcommands, we can add new subcommands through the plugin module.
```
dposada, line 180:
```suggestion
In addition to plugins for the core subcommands, we can add new subcommands through the plugin module.
```

cli/cook/cli.py
```diff
@@ -39,7 +40,22 @@ def run(args, plugins):
...
+    # get subcommand not found.
+    for name, instance in plugins.items():
+        if isinstance(instance, SubCommandPlugin):
+            logging.debug('Adding SubCommandPlugin %s' % name)
+            instance.register(subparsers.add_parser, configuration.add_defaults)
+            logging.debug('Done adding SubCommandPlugin %s' % name)
+            name = instance.name()
+            if name in actions:
+                raise Exception('SubCommandPlugin %s clashes with an existing subcommand.' % name)
+            actions[name] = instance.run
```
dposada, line 57: Maybe we should wrap these lines in a `try`/`catch` so that one bad plugin doesn't blow everything up?

rmanyari: yeah let me do that.

dposada (approved):

PR #1793, opened 2021-02-16
Add memory limit job label

nsinkov:

- Let users set on configurable job label if they want to allow memory usage above memory request

We found use cases of memory limit

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2911,7 +2911,7 @@ class CookTest(util.CookTest):
         finally:
             util.kill_jobs(self.cook_url, [job_uuid1, job_uuid2])
 
-    @unittest.skipUnless(util.using_kubernetes(), 'Test requires kubernetes')
+    @unittest.skipUnless(util.using_kubernetes() and util.in_cloud(), 'Test requires kubernetes')
     def test_kubernetes_checkpointing(self):
         docker_image = util.docker_image()
         container = {'type': 'docker',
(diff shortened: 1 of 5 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -898,14 +893,36 @@ def test_memory_limit_exceeded_cook_script(self):
...
+            self.assertEqual('failed', job1['instances'][0]['status'])
+            self.assertEqual(2002, job1['instances'][0]['reason_code'])
+
+            self.assertEqual(resp2.status_code, 201, msg=resp2.content)
+            job2 = util.wait_for_job(self.cook_url, job_uuid2, 'completed')
+            self.assertEqual('completed', job2['status'])
+            self.assertEqual('success', job2['instances'][0]['status'])
+
+        finally:
+            util.kill_jobs(self.cook_url, [job_uuid1, job_uuid2], assert_response=False)
```
dposada, line 924: ❤️

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -943,8 +943,12 @@
     (.setTty container true)
     (.setStdin container true)
 
-    ; add memory limit if config flag set-memory-limit? is True
-    (set-mem-cpu-resources resources computed-mem (when (:set-memory-limit? (config/kubernetes)) computed-mem) cpus cpus)
+    ; add memory limit if user sets job label to allow memory usage above request to "true"
```
dposada, line 946:
```suggestion
    ; Don't add memory limit if user sets job label to allow memory usage above request to "true"
```

dposada (approved):

PR #1795, opened 2021-02-18
cook: pass decorator through to List<Job> parseFromJSON

brianbao:

- Pass `decorator down`

My bad.

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -1401,7 +1401,7 @@ final public class Job {
         List<Job> jobs = new ArrayList<Job>(jsonArray.length());
         for (int i = 0; i < jsonArray.length(); ++i) {
             JSONObject json = jsonArray.getJSONObject(i);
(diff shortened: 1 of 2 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -1401,7 +1401,7 @@ public static Job parseFromJSON(JSONObject json) throws JSONException {
         List<Job> jobs = new ArrayList<Job>(jsonArray.length());
         for (int i = 0; i < jsonArray.length(); ++i) {
             JSONObject json = jsonArray.getJSONObject(i);
-            jobs.add(parseFromJSON(json));
+            jobs.add(parseFromJSON(json, decorator));
```
dposada, line 1404: Is there a reasonable unit test we could write for this?

brianbao: I gotta come clean, I have no idea what the `decorator` even did in this code. Do you have any good test in mind?

dposada (approved):

PR #1797, opened 2021-02-22
Marks reason correctly for pod failures from preemption

dposada:

Making pods that fail and have a `:pod-preempted-timestamp` go down the `handle-pod-preemption` path so that they get the appropriate `Agent removed` failure reason.

Currently, pod failures from node preemption result in the confusing and misleading `Unknown mesos reason`.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -307,8 +307,8 @@
   [{:keys [name] :as compute-cluster} pod-name {:keys [synthesized-state] :as k8s-actual-state-dict}]
   (if (some-> synthesized-state :pod-preempted-timestamp)
     (do
-      (log/info "In compute cluster" name ", pod" pod-name
-                "failed, seemingly due to preemption")
+      (log/info "In compute cluster" name ", pod" pod-name "failed and has"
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -302,6 +302,16 @@
   ; to deleting from the map, into (missing,missing) state.
   {:cook-expected-state :cook-expected-state/completed})
 
+(defn handle-pod-failed
+  "Handles pod failure, checking to see if the failure was caused by pod preemption."
+  [{:keys [name] :as compute-cluster} pod-name {:keys [synthesized-state] :as k8s-actual-state-dict}]
+  (if (some-> synthesized-state :pod-preempted-timestamp)
+    (do
+      (log/info "In compute cluster" name ", pod" pod-name
```
scrosby, line 310: I"d reword this to be more concrete and clear. Maybe something like:

"has a preemped timestamp and is treated as preempted"

dposada: Done, let me know what you think

scrosby: lgtm

PR #1799, opened 2021-02-23
Numerous autoscaling improvements

scrosby:
* Parallel synthetic pod launching across compute clusters
* Launch synthetic pods proportional to queue length
* Don't launch synthetic pods for the same job too frequently

* Parallel synthetic pod launching across compute clusters
* Launch synthetic pods proportional to queue length
* Don't launch synthetic pods for the same job too frequently

Reduce the overheads of autoscaling.

### DIFF
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -503,8 +503,7 @@
                          (repeatedly
                            controller-lock-num-shards
                            #(ReentrantLock.))]
-                     (merge {
-                             :autoscaling-scale-factor 1.0
+                     (merge {:autoscaling-scale-factor 1.0
                              :controller-lock-num-shards controller-lock-num-shards
                              :controller-lock-objects (with-meta
                                                         lock-objects
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -503,28 +503,30 @@
                          (repeatedly
                            controller-lock-num-shards
                            #(ReentrantLock.))]
-                     (merge {:controller-lock-num-shards controller-lock-num-shards
+                     (merge {
+                             :autoscaling-scale-factor 1.0
+                             :controller-lock-num-shards controller-lock-num-shards
                              :controller-lock-objects (with-meta
                                                         lock-objects
                                                         {:json-value (str lock-objects)})
                              :default-workdir "/mnt/sandbox"
                              :max-jobs-for-autoscaling 1000
-                             :synthetic-pod-recency-size 50000
-                             :synthetic-pod-recency-seconds 240 ; Should be greater than the time to start a node and have it accept workloads.
-                             :autoscaling-scale-factor 1.0
                              :pod-condition-containers-not-initialized-seconds 120
                              :pod-condition-unschedulable-seconds 60
                              :reconnect-delay-ms 60000
                              :set-container-cpu-limit? true
+                             :set-memory-limit? true
                              :synthetic-pod-condition-unschedulable-seconds 900
-                             :set-memory-limit? true}
+                             :synthetic-pod-recency-size 50000
+                             :synthetic-pod-recency-seconds 120 ; Should be greater than the time to start a node and have it accept workloads.
+                             }
                             kubernetes)))
      :offer-matching (fnk [[:config {offer-matching {}}]]
-                       (merge {:global-min-match-interval-millis 100
+                       (merge {:considerable-job-threshold-to-collect-job-match-statistics 20
+                               :global-min-match-interval-millis 100
                                :target-per-pool-match-interval-millis 3000
                                :unmatched-cycles-warn-threshold 500
-                               :unmatched-fraction-warn-threshold 0.5
-                               :considerable-job-threshold-to-collect-job-match-statistics 20}
+                               :unmatched-fraction-warn-threshold 0.5}
                               offer-matching))
      :queue-limits (fnk [[:config {queue-limits {}}]]
                      (merge {:update-interval-seconds 180}
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -523,7 +523,8 @@
                        (merge {:global-min-match-interval-millis 100
                                :target-per-pool-match-interval-millis 3000
                                :unmatched-cycles-warn-threshold 500
-                               :unmatched-fraction-warn-threshold 0.5}
+                               :unmatched-fraction-warn-threshold 0.5
+                               :considerable-job-threshold-to-collect-job-match-statistics 20}
                               offer-matching))
      :queue-limits (fnk [[:config {queue-limits {}}]]
                      (merge {:update-interval-seconds 180}
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -509,6 +509,8 @@
                                                         {:json-value (str lock-objects)})
                              :default-workdir "/mnt/sandbox"
                              :max-jobs-for-autoscaling 1000
+                             :autoscale-blocklist-seconds 240 ; Should be greater than the time to start a pod and have it accept workloads.
+                             :autoscaling-scale-factor 1.0
                              :pod-condition-containers-not-initialized-seconds 120
                              :pod-condition-unschedulable-seconds 60
                              :reconnect-delay-ms 60000
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -509,7 +509,8 @@
                                                         {:json-value (str lock-objects)})
                              :default-workdir "/mnt/sandbox"
                              :max-jobs-for-autoscaling 1000
-                             :autoscale-blocklist-seconds 240 ; Should be greater than the time to start a pod and have it accept workloads.
+                             :synthetic-pod-recency-size 50000
+                             :synthetic-pod-recency-seconds 240 ; Should be greater than the time to start a node and have it accept workloads.
                              :autoscaling-scale-factor 1.0
                              :pod-condition-containers-not-initialized-seconds 120
                              :pod-condition-unschedulable-seconds 60
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -485,6 +485,7 @@
                          (map (fn [{:keys [job/user job/uuid job/environment] :as job}]
                                 (let [pool-specific-resources
                                       ((adjust-job-resources-for-pool-fn pool-name) job (tools/job-ent->resources job))]
+                                  (.put cook.caches/autoscale-retry-blacklist uuid uuid)
                                   {:command {:user (or user-from-synthetic-pods-config user)
                                              :value command}
                                    :container {:docker {:image image}}
@@ -530,8 +531,9 @@
                          (take max-launchable))
                     num-synthetic-pods-to-launch (count task-metadata-seq)]
                 (meters/mark! (metrics/meter "cc-synthetic-pod-submit-rate" name) num-synthetic-pods-to-launch)
-                (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
-                          "synthetic pod(s) in" synthetic-task-pool-name "pool")
+                (when (pos? num-synthetic-pods-to-launch)
+                  (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
+                            "synthetic pod(s) in" synthetic-task-pool-name "pool"))
                 (let [timer-context-launch-tasks (timers/start (metrics/timer "cc-synthetic-pod-launch-tasks" name))]
                   (cc/launch-tasks this
                                    synthetic-task-pool-name
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -531,9 +531,8 @@
                          (take max-launchable))
                     num-synthetic-pods-to-launch (count task-metadata-seq)]
                 (meters/mark! (metrics/meter "cc-synthetic-pod-submit-rate" name) num-synthetic-pods-to-launch)
-                (when (pos? num-synthetic-pods-to-launch)
-                  (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
-                            "synthetic pod(s) in" synthetic-task-pool-name "pool"))
+                (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
+                          "synthetic pod(s) in" synthetic-task-pool-name "pool")
                 (let [timer-context-launch-tasks (timers/start (metrics/timer "cc-synthetic-pod-launch-tasks" name))]
                   (cc/launch-tasks this
                                    synthetic-task-pool-name
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -485,7 +485,7 @@
                          (map (fn [{:keys [job/user job/uuid job/environment] :as job}]
                                 (let [pool-specific-resources
                                       ((adjust-job-resources-for-pool-fn pool-name) job (tools/job-ent->resources job))]
-                                  (.put cook.caches/autoscale-retry-blacklist uuid uuid)
+                                  (.put cook.caches/recent-synthetic-pod-job-uuids uuid uuid)
                                   {:command {:user (or user-from-synthetic-pods-config user)
                                              :value command}
                                    :container {:docker {:image image}}
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -482,6 +482,7 @@
                     user-from-synthetic-pods-config user
                     task-metadata-seq
                     (->> new-jobs
+                         (take max-launchable)
                          (map (fn [{:keys [job/user job/uuid job/environment] :as job}]
                                 (let [pool-specific-resources
                                       ((adjust-job-resources-for-pool-fn pool-name) job (tools/job-ent->resources job))]
@@ -527,8 +528,7 @@
                                                   :job {:job/pool {:pool/name synthetic-task-pool-name}
                                                         :job/environment environment}
                                                   ; Need to pass in resources to task-metadata->pod for gpu count
-                                                  :resources pool-specific-resources}})))
-                         (take max-launchable))
+                                                  :resources pool-specific-resources}}))))
                     num-synthetic-pods-to-launch (count task-metadata-seq)]
                 (meters/mark! (metrics/meter "cc-synthetic-pod-submit-rate" name) num-synthetic-pods-to-launch)
                 (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1242,14 +1242,15 @@
                 {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
                 ; The fraction of jobs we tried to match that didn't actually get matched.
                 fraction-unmatched-jobs (if (pos? number-considerable-jobs) (/ (float number-unmatched-jobs) number-considerable-jobs) 0)
-                ; We want to autoscale any unmatched job or
+                ; We want to autoscale any unmatched job
+                ;     OR
                 ; we want to scale our max-jobs-for-autoscaling by the fraction of the jobs we weren't able to just match.
                 ; E.g. If we didn't match 20% of the queue, then we want to autoscale to 20% of max-jobs-for-autoscaling.
                 ; We include a scale factor however, so that if we don't match 20% and scale factor is 2.5, we'll generate
                 ; pods for 50% of max-jobs-for-autoscaling.
                 ; We do this to vary our aggression for autoscaling based on how well we're matching jobs on our existing resources.
                 ; If we're matching most of the jobs in the queue then we don't need to autoscale much. If we are not matching anything
-                ; then we want to autoscale maximally aggressive.
+                ; then we want to autoscale maximally aggressively.
                 max-jobs-for-autoscaling-scaled (-> fraction-unmatched-jobs
                                                     (* autoscaling-scale-factor)
                                                     (min 1) ; Can't match more than 100% of max-jobs-for-autoscaling.
@@ -1264,17 +1265,17 @@
                                                                             user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
                                        (take max-jobs-for-autoscaling-scaled))
                 filtered-autoscalable-jobs (remove #(.getIfPresent caches/recent-synthetic-pod-job-uuids (:job/uuid %)) autoscalable-jobs)]
-            ; When we have at least 20 jobs being looked at, metric which fraction have matched. This lets us measure how well we're matching on existing resources.
-            ; We only measure when there's at least 20 jobs being considered so that our measurements are less noisy.
+            ; When we have at least a minimum jobs being looked at, metric which fraction have matched. This lets us measure how well we're matching on existing resources.
+            ; We only measure when there's a minimum jobs being considered so that our measurements are less noisy.
             (when (> number-considerable-jobs (:considerable-job-threshold-to-collect-job-match-statistics (config/offer-matching)))
               (histograms/update! (histograms/histogram (metric-title "fraction-unmatched-jobs" pool-name)) fraction-unmatched-jobs))
             (when (pos? number-considerable-jobs)
-              (log/info "In" pool-name "autoscaling variables" {:autoscalable-jobs (count autoscalable-jobs)
+              (log/info "In" pool-name "pool, autoscaling variables" {:autoscalable-jobs (count autoscalable-jobs)
                                                                 :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
                                                                 :fraction-unmatched-jobs fraction-unmatched-jobs
                                                                 :max-jobs-for-autoscaling-scaled max-jobs-for-autoscaling-scaled
-                                                                :number-total number-considerable-jobs
-                                                                :number-unmatched number-unmatched-jobs})
+                                                                :number-considerable-jobs number-considerable-jobs
+                                                                :number-unmatched-jobs number-unmatched-jobs})
               ;; This call needs to happen *after* launch-matched-tasks!
               ;; in order to avoid autoscaling tasks taking up available
               ;; capacity that was already matched for real Cook tasks.
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1260,7 +1260,7 @@
                                        (tools/filter-pending-jobs-for-quota pool-name (atom {}) (atom {})
                                                                             user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
                                        (take number-for-autoscaling))
-                filtered-autoscalable-jobs (remove #(.getIfPresent caches/autoscale-retry-blacklist (:job/uuid %)) autoscalable-jobs)]
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/recent-synthetic-pod-job-uuids (:job/uuid %)) autoscalable-jobs)]
             ; When we have at least 20 jobs being looked at, metric which fraction have matched.
             (when (> number-total 20)
               (histograms/update! (histograms/histogram (metric-title "fraction-unmatched" pool-name)) fraction-unmatched))
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1265,17 +1265,18 @@
                                                                             user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
                                        (take max-jobs-for-autoscaling-scaled))
                 filtered-autoscalable-jobs (remove #(.getIfPresent caches/recent-synthetic-pod-job-uuids (:job/uuid %)) autoscalable-jobs)]
-            ; When we have at least a minimum jobs being looked at, metric which fraction have matched. This lets us measure how well we're matching on existing resources.
-            ; We only measure when there's a minimum jobs being considered so that our measurements are less noisy.
+            ; When we have at least a minimum number of jobs being looked at, metric which fraction have matched.
+            ; This lets us measure how well we're matching on existing resources.
+            ; We only measure when there's a minimum number of jobs being considered so that our measurements are less noisy.
             (when (> number-considerable-jobs (:considerable-job-threshold-to-collect-job-match-statistics (config/offer-matching)))
               (histograms/update! (histograms/histogram (metric-title "fraction-unmatched-jobs" pool-name)) fraction-unmatched-jobs))
             (when (pos? number-considerable-jobs)
               (log/info "In" pool-name "pool, autoscaling variables" {:autoscalable-jobs (count autoscalable-jobs)
-                                                                :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
-                                                                :fraction-unmatched-jobs fraction-unmatched-jobs
-                                                                :max-jobs-for-autoscaling-scaled max-jobs-for-autoscaling-scaled
-                                                                :number-considerable-jobs number-considerable-jobs
-                                                                :number-unmatched-jobs number-unmatched-jobs})
+                                                                      :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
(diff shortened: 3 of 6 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
                     matched-considerable-jobs-head?))
+                ; Absolute maximum jobs we will consider autoscaling to.
+                {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
+                ; The fraction of jobs we tried to match that actually matched.
+                fraction-unmatched (if (pos? number-total) (/ (float number-unmatched) number-total) 0)
+                ; We want to autoscale any unmatched job, or
+                ; We want to scale our max-jobs-for-autoscaling by the fraction of the jobs we just matched.
+                ; E.g. If we matched 20% of the queue, then we want to autoscale to 20% of max-jobs-for-autoscaling)
```
dposada, line 1247: There is no `(` for the `)`

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                ; We include a scale factor however, so that if we match 20$ and scale factor is 2.5, we'll generate
```
dposada, line 1248: `$` -> `%`

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                ; pods for 50% of max-jobs-for-autoscaling.
```
dposada, line 1249: This example also needs to be fixed (20% matched with scale factor of 2.5 would mean we do 100% of `max-jobs-for-autoscaling`)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                number-for-autoscaling (-> fraction-unmatched
```
dposada, line 1250: Maybe instead of `number-for-autoscaling`, `max-jobs-for-autoscaling-scaled`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                                           (max number-unmatched)) ; Autoscale at least the pods that failed to match.
                 ;; We need to filter pending jobs based on quota so that we don't
                 ;; trigger autoscaling beyond what users have quota to actually run
                 autoscalable-jobs (->> pool-name
                                        (get @pool-name->pending-jobs-atom)
                                        (tools/filter-pending-jobs-for-quota pool-name (atom {}) (atom {})
-                                         user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name)))]
+                                                                            user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
+                                       (take number-for-autoscaling))
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/autoscale-retry-blacklist (:job/uuid %)) autoscalable-jobs)]
```
dposada, line 1263: What's the rationale for scaling first and then filtering? 

Seems to me like we'd want to filter first (lazily), then scale based on `fraction-unmatched`.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+            ; When we have at least 20 jobs being looked at, metric which fraction have matched.
+            (when (> number-total 20)
```
dposada, line 1265: Can we make this configurable?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
                     matched-considerable-jobs-head?))
+                ; Absolute maximum jobs we will consider autoscaling to.
+                {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
+                ; The fraction of jobs we tried to match that actually matched.
+                fraction-unmatched (if (pos? number-total) (/ (float number-unmatched) number-total) 0)
+                ; We want to autoscale any unmatched job, or
+                ; We want to scale our max-jobs-for-autoscaling by the fraction of the jobs we just matched.
```
dposada, line 1246: I think it would be nice to mention the "why" for this in the comment.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
                 ;; We need to filter pending jobs based on quota so that we don't
                 ;; trigger autoscaling beyond what users have quota to actually run
                 autoscalable-jobs (->> pool-name
                                        (get @pool-name->pending-jobs-atom)
                                        (tools/filter-pending-jobs-for-quota pool-name (atom {}) (atom {})
-                                         user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name)))]
+                                                                            user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
+                                       (take number-for-autoscaling))
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/autoscale-retry-blacklist (:job/uuid %)) autoscalable-jobs)]
+            ; When we have at least 20 jobs being looked at, metric which fraction have matched.
```
dposada, line 1264: I think it would be nice to mention the "why" for this in the comment.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                                                                :fraction-unmatched fraction-unmatched
+                                                                :number-for-autoscaling number-for-autoscaling
+                                                                :autoscalable-jobs (count autoscalable-jobs)
+                                                                :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
+                                                                :number-unmatched number-unmatched}))
             ;; This call needs to happen *after* launch-matched-tasks!
             ;; in order to avoid autoscaling tasks taking up available
             ;; capacity that was already matched for real Cook tasks.
-            (trigger-autoscaling! autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn)
+            (trigger-autoscaling! filtered-autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn)
```
dposada, line 1277: Should this move to inside the 

```
(when (pos? number-total)
  ...)
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
                     matched-considerable-jobs-head?))
+                ; Absolute maximum jobs we will consider autoscaling to.
+                {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
+                ; The fraction of jobs we tried to match that actually matched.
+                fraction-unmatched (if (pos? number-total) (/ (float number-unmatched) number-total) 0)
```
dposada, line 1244: Maybe `fraction-unmatched-jobs`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
                 autoscalable-jobs (->> pool-name
                                        (get @pool-name->pending-jobs-atom)
                                        (tools/filter-pending-jobs-for-quota pool-name (atom {}) (atom {})
-                                         user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name)))]
+                                                                            user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
+                                       (take number-for-autoscaling))
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/autoscale-retry-blacklist (:job/uuid %)) autoscalable-jobs)]
+            ; When we have at least 20 jobs being looked at, metric which fraction have matched.
+            (when (> number-total 20)
+              (histograms/update! (histograms/histogram (metric-title "fraction-unmatched" pool-name)) fraction-unmatched))
```
dposada, line 1266: Maybe "fraction-unmatched-jobs" for the metric name?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+            (when (pos? number-total)
+              (log/info "In" pool-name "autoscaling variables" {:number-total number-total
+                                                                :fraction-unmatched fraction-unmatched
+                                                                :number-for-autoscaling number-for-autoscaling
+                                                                :autoscalable-jobs (count autoscalable-jobs)
+                                                                :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
+                                                                :number-unmatched number-unmatched}))
```
dposada, line 1273: Should we sort these lexicographically in the map?

scheduler/src/cook/scheduler/scheduler.clj
dposada, line 1263: It feels like all the changes above here in this function warrant some unit tests.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -530,8 +531,9 @@
                          (take max-launchable))
                     num-synthetic-pods-to-launch (count task-metadata-seq)]
                 (meters/mark! (metrics/meter "cc-synthetic-pod-submit-rate" name) num-synthetic-pods-to-launch)
-                (log/info "In" name "compute cluster, launching" num-synthetic-pods-to-launch
-                          "synthetic pod(s) in" synthetic-task-pool-name "pool")
+                (when (pos? num-synthetic-pods-to-launch)
```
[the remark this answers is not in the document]
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1099,15 +1103,17 @@
                                                           frequencies)
               user->number-total-considerable-jobs (->> considerable-jobs
                                                         (map cached-queries/job-ent->user)
-                                                        frequencies)]
+                                                        frequencies)
+              number-matched (count matched-job-uuids)
+              number-total (count considerable-jobs)
+              number-unmatched (- number-total number-matched)]
```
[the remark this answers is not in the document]
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
                     matched-considerable-jobs-head?))
+                ; Absolute maximum jobs we will consider autoscaling to.
+                {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
+                ; The fraction of jobs we tried to match that actually matched.
```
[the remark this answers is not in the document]
scrosby: Thanks.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,43 @@
...
+                fraction-unmatched (if (pos? number-total) (/ (float number-unmatched) number-total) 0)
+                ; We want to autoscale any unmatched job, or
```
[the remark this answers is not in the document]
scrosby: Yes, the next line.

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
scrosby: Good find. Yes, I got that sense inverted.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Agreed. Thanks.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Fixed.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Yes.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Fixed.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Renamed.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Sure, but I don't see a huge motivation to do so. This is a set once and never again.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Ok.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Heh, good call.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -485,6 +485,7 @@
                          (map (fn [{:keys [job/user job/uuid job/environment] :as job}]
                                 (let [pool-specific-resources
                                       ((adjust-job-resources-for-pool-fn pool-name) job (tools/job-ent->resources job))]
+                                  (.put cook.caches/autoscale-retry-blacklist uuid uuid)
```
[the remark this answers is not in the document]
scrosby: I can, but I'd vote note. The task-struct we create here does not have the job UUID in it. I can add it in, and then do a doall loop afterwards, that adds to the cache, but then we'll have that odd :job-uuid field there that someone trying to trace data flow won't necessarily realize only exists for the next line of code, and could cause confusion and tighter coupling here in the future.

But good concern. We should have had the take at the top, before the map.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: It would create the wrong answer in that we could no longer tune the loop from first principals. This scaling loop was created so we could tone it from first principals. 

1.  outstanding synthetic pods is a function of the desired scaling rate which we choose, as well as the time to create a new pod (from GKE/k8s) This defines the peak rate we intend to support in terms of SLO. Note that this is just a number. (= scale rate / typ P95 synthetic pod launch time  / typ pod size)
2. We adjust this based on fraction matched. If we're matching heavily, we don't need to autoscale as aggressively. Note that this is a number, not a list. This defines our effective scaling demand. Scaling more than this means we're scaling above our desired scaling rate. Scaling under this means that we're scaling under our desired scaling rate. (fudge factor = 1.0 or 1.50) We make sure to always synthetic pod for any unmatched job, in case it has unusual machine requirements.
3. We take the list of jobs do hard filter for out of quota. don't pass any jobs that are out of quota downstream.
4. Next, we limit the jobs for synthetic pods to be our effective scaling demand. This, combined with the time to create a new synthetic pod means we're neither oversatisfying or undersatisfying our demand to autoscale.
5. We then have the filter as an optimization so that we don't pointlessly relaunch the same pod too frequently as a synthetic pod (when we're launch starved). When we're intentionally scaling, the synthetic pod launch time is the same duration as this filter block time and this filter is effectively a noop. It only comes into play when we're in a launch starved situation and, in some sense, emulates synthetic pods as having a longer time-to-complete, as seen by the rest of the scale-up system. This gives us the benefits of the optimization, without altering the scaling loop or scaling demand behavior.

We could do steps in the order 3, 1, 2, which may seem more clear, but as steps 1 and 2 only produce a number, doing them first doesn't change the result.

Doing step 5 before step 4 means that this design isn't constructed from first principals, so will have behavior that can be dysfunctional. In particular, #1, it means that we're potentially scaling over our target demand rate. #2 It can also dive 10x deeply into the queue, requesting synthetic pods for jobs that we didn't intend to have them requested this early, e.g., if they have one-off machine type requirements.

scrosby: I'll see what I can do....

scheduler/src/cook/scheduler/scheduler.clj
dposada: Not the first time we've thought that and then ended up wanting to configure differently :) I'm not convinced this should be 20

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,47 @@
                     (launch-matched-tasks! matches conn db fenzo mesos-run-as-user pool-name)
                     (update-host-reservations! rebalancer-reservation-atom matched-job-uuids)
                     matched-considerable-jobs-head?))
+                ; Absolute maximum jobs we will consider autoscaling to.
+                {:keys [max-jobs-for-autoscaling autoscaling-scale-factor]} (config/kubernetes)
+                ; The fraction of jobs we tried to match that didn't actually get matched.
+                fraction-unmatched-jobs (if (pos? number-considerable-jobs) (/ (float number-unmatched-jobs) number-considerable-jobs) 0)
+                ; We want to autoscale any unmatched job or
+                ; we want to scale our max-jobs-for-autoscaling by the fraction of the jobs we weren't able to just match.
```
dposada, line 1246: The grammar in this sentence is confusing me, how does the part before the "or" relate to the part after the "or"? Maybe this could be re-written to be more clear?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,47 @@
...
+                ; E.g. If we didn't match 20% of the queue, then we want to autoscale to 20% of max-jobs-for-autoscaling.
+                ; We include a scale factor however, so that if we don't match 20% and scale factor is 2.5, we'll generate
+                ; pods for 50% of max-jobs-for-autoscaling.
+                ; We do this to vary our aggression for autoscaling based on how well we're matching jobs on our existing resources.
+                ; If we're matching most of the jobs in the queue then we don't need to autoscale much. If we are not matching anything
+                ; then we want to autoscale maximally aggressive.
```
dposada, line 1252:
```suggestion
                ; then we want to autoscale maximally aggressively.
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,47 @@
...
-                                         user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name)))]
-            ;; This call needs to happen *after* launch-matched-tasks!
-            ;; in order to avoid autoscaling tasks taking up available
-            ;; capacity that was already matched for real Cook tasks.
-            (trigger-autoscaling! autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn)
+                                                                            user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
+                                       (take max-jobs-for-autoscaling-scaled))
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/recent-synthetic-pod-job-uuids (:job/uuid %)) autoscalable-jobs)]
+            ; When we have at least 20 jobs being looked at, metric which fraction have matched. This lets us measure how well we're matching on existing resources.
+            ; We only measure when there's at least 20 jobs being considered so that our measurements are less noisy.
```
dposada, line 1268: Since this is configured, let's not mention 20 in the comment

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,47 @@
...
+            (when (> number-considerable-jobs (:considerable-job-threshold-to-collect-job-match-statistics (config/offer-matching)))
+              (histograms/update! (histograms/histogram (metric-title "fraction-unmatched-jobs" pool-name)) fraction-unmatched-jobs))
+            (when (pos? number-considerable-jobs)
+              (log/info "In" pool-name "autoscaling variables" {:autoscalable-jobs (count autoscalable-jobs)
```
dposada, line 1272: To be consistent with other logs, can we do:

```
"In" pool-name "pool, ..."
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,47 @@
...
+                                                                :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
+                                                                :fraction-unmatched-jobs fraction-unmatched-jobs
+                                                                :max-jobs-for-autoscaling-scaled max-jobs-for-autoscaling-scaled
+                                                                :number-total number-considerable-jobs
+                                                                :number-unmatched number-unmatched-jobs})
```
dposada, line 1277: To be consistent with the log above, can we do

```
:number-considerable-jobs number-considerable-jobs
:number-unmatched-jobs number-unmatched-jobs
```

scheduler/src/cook/kubernetes/compute_cluster.clj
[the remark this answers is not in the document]
dposada: Yeah, I think the `take` before the `map` eliminates the risk 👍

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
dposada: Created a new comment about the grammar here being confusing

scheduler/src/cook/scheduler/scheduler.clj
dposada: Let's say Cook does the math for scaling based on `fraction-unmatched`, and the answer is that we should send 100 synthetic pods this cycle. And let's say the top 100 jobs in the pending queue all had a synthetic pod sent in the last, say, 1 minute. 

The current ordering of code will mean that we send 0 synthetic pods this cycle, and this situation could persist for a while (potentially minutes, depending on what we configure `recency-seconds` to), effectively halting all autoscaling during this time. This would be true even if we have thousands of pending jobs in the queue.

I don't think we want this behavior. In this example, I think we instead want to send synthetic pods for the top 100 jobs in the queue that have not recently had a synthetic pod launched for them.

dposada: Is the test enough? I thought we would want a unit test that gives us comfort that the various math, `take`s, and `filter`s are doing what we want?

scheduler/src/cook/config.clj
```diff
@@ -509,6 +509,8 @@
                                                         {:json-value (str lock-objects)})
                              :default-workdir "/mnt/sandbox"
                              :max-jobs-for-autoscaling 1000
+                             :autoscale-blocklist-seconds 240 ; Should be greater than the time to start a pod and have it accept workloads.
```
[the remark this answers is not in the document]
scrosby: Ok.

scheduler/src/cook/config.clj
```diff
@@ -520,7 +523,8 @@
                        (merge {:global-min-match-interval-millis 100
                                :target-per-pool-match-interval-millis 3000
                                :unmatched-cycles-warn-threshold 500
-                               :unmatched-fraction-warn-threshold 0.5}
+                               :unmatched-fraction-warn-threshold 0.5
+                               :considerable-job-threshold-to-collect-job-match-statistics 20}
```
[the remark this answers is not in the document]
scrosby: Ok.

scheduler/src/cook/config.clj
```diff
@@ -503,7 +503,9 @@
                          (repeatedly
                            controller-lock-num-shards
                            #(ReentrantLock.))]
-                     (merge {:controller-lock-num-shards controller-lock-num-shards
+                     (merge {
+                             :autoscaling-scale-factor 1.0
```
dposada, line 507:
```suggestion
                     (merge {:autoscaling-scale-factor 1.0
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1038,22 +1039,25 @@
...
-            (doseq [[compute-cluster jobs-for-cluster] compute-cluster->jobs]
-              (cc/autoscale! compute-cluster pool-name jobs-for-cluster adjust-job-resources-for-pool-fn))
-            (log/info "In" pool-name "pool, done autoscaling"))))
+            (->> compute-cluster->jobs
+                 (map
+                   (fn [[compute-cluster jobs-for-cluster]]
+                     (future (cc/autoscale! compute-cluster pool-name jobs-for-cluster adjust-job-resources-for-pool-fn))))
+                 doall
+                 (run! deref)))
+          (log/info "In" pool-name "pool, done autoscaling")))
```
dposada, line 1060: Indentation seems off on this line

scheduler/src/cook/scheduler/scheduler.clj
dposada: @scrosby Thoughts on this?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,48 @@
...
-                                         user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name)))]
-            ;; This call needs to happen *after* launch-matched-tasks!
-            ;; in order to avoid autoscaling tasks taking up available
-            ;; capacity that was already matched for real Cook tasks.
-            (trigger-autoscaling! autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn)
+                                                                            user->quota user->usage (tools/global-pool-quota (config/pool-quotas) pool-name))
+                                       (take max-jobs-for-autoscaling-scaled))
+                filtered-autoscalable-jobs (remove #(.getIfPresent caches/recent-synthetic-pod-job-uuids (:job/uuid %)) autoscalable-jobs)]
+            ; When we have at least a minimum jobs being looked at, metric which fraction have matched. This lets us measure how well we're matching on existing resources.
+            ; We only measure when there's a minimum jobs being considered so that our measurements are less noisy.
```
dposada, line 1269:
```suggestion
            ; When we have at least a minimum number of jobs being looked at, metric which fraction have matched. 
            ; This lets us measure how well we're matching on existing resources.
            ; We only measure when there's a minimum number of jobs being considered so that our measurements are less noisy.
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1232,16 +1238,48 @@
...
+            (when (> number-considerable-jobs (:considerable-job-threshold-to-collect-job-match-statistics (config/offer-matching)))
+              (histograms/update! (histograms/histogram (metric-title "fraction-unmatched-jobs" pool-name)) fraction-unmatched-jobs))
+            (when (pos? number-considerable-jobs)
+              (log/info "In" pool-name "pool, autoscaling variables" {:autoscalable-jobs (count autoscalable-jobs)
+                                                                :filtered-autoscalable-jobs (count filtered-autoscalable-jobs)
+                                                                :fraction-unmatched-jobs fraction-unmatched-jobs
+                                                                :max-jobs-for-autoscaling-scaled max-jobs-for-autoscaling-scaled
+                                                                :number-considerable-jobs number-considerable-jobs
+                                                                :number-unmatched-jobs number-unmatched-jobs})
```
dposada, line 1278: Indentation seems off on these lines

scheduler/src/cook/config.clj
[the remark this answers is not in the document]
dposada: Thoughts on avoiding the prep-then-throwaway problem?

scheduler/src/cook/config.clj
scrosby: Thanks. Done.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Indentation is correct (at least per cursive's default rules) in this section.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Agree with the first paragraph, but now lets break this situation down further:

1. Lets say those 100 synthetic pods are needed in order to launch the workload, (we're at capacity or they have a unique machine type)  that means that these synthetic pods will continue to be outstanding until resources are created to run them --- I.e., we are actively autoscaling to launch the workload.  In this case, even if the filter wasn't there, we wouldn't be launching more pods --- we still have these 100 pods outstanding, and the only-one-synthetic-pod-per-job would suppress autoscaling further. So, in this case the filter is a noop; we're filtering jobs that would already be filtered downstream, doing the right thing, with higher performance.
2. Now, lets say those 100 synthetic pods are for a workload we can run now. however, we didn't match them to actually launch because they were beyond the considerable queue but still in the autoscaling queue. In this case, there's no point in creating synthetic pods for them every cycle, which the k8s scheduler will place and they'll complete. In this case, because these jobs don't require autoscaling resources to fit in, they're not evidence that we need to look even *further* into the queue for jobs. Again, the filter is doing the right thing, its avoiding k8s launch bandwidth.

What swapping does do is it makes an interesting functionality change. 
If we swap the order, we're basically avoiding the performance optimization this code intended and we're choosing to autoscale more aggresively, and in a strange way. In the first scenario above, there will be no change.  However, in the second scenario above, instead of wasting launches on the same synthetic pod in the above scenario, we're using that launch bandwidth to march another 100 pods in the queue each cycle for autoscaling because the filter suppreses that queue prefix. 

To be concrete, here's another scenario that illustrates the difference very concretely:
* The first 400 jobs in the queue are stuck waiting for a special nodetype to be created, and will be stuck unmatched for the next 2 minutes.
* The next 25k jobs can launch right now, however, no more than 1000 simultaneously.
* Our optimizations successfully bring the match cycle down to 5 seconds.

A. If we do no filter, then on each match cycle, we'll match no jobs, causing us to create synthetic pods for the next 1000 jobs, they'll launch as we have resources, and we'll try repeatedly launching those next 1000 jobs 24 times over the next 120 seconds.
B. If we do the take, then filter, then on each match cycle, we'll match no jobs, causing us to want to create synthetic pods for the next 1000 jobs. The first iteration through, we'll create them, they'll launch immediately and then complete. For the next 23 iterations, the filter will suppress them.
C. If we filter, then take, then on each match cycle, we'll match no jobs, causing us to create synthetic pods for the next 1000 jobs, they'll launch and complete. The next cycle, the filter skips jobs 400-1400 in the queue, so we'll launch jobs #1400-2400 in the queue as synthetic pods. We'll repeat 23 times and launch synthetic pods for next 24000 jobs in the queue over the next 120 seconds.

The same principal applies even if the entire queue isn't stuck for 120 seconds. 

So, my thoughts on this functionality change embodied in C:
- C is slower than B, going contrary to the goal of optimizing our code.
- C launches just as many synthetic pods as A, and much more than B. 
- I expect C to be somewhat slower than A; it will visit much more of the queue and that has implications for cache hit rate and laziness.
- C will autoscale more aggressively than A, this can be mixed. We may waste resources more aggressively autoscaling, but may also decrease autoscaling latency for resources. However, personally, if we wanted to increase autoscaling aggressiveness, we could double max-jobs-for-autoscaling at a lot less cost.

Thus,. my vote for B.

scheduler/src/cook/config.clj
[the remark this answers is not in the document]
scrosby: Too awkward to do in the code. Only the CC knows if it is running there (or is starting), so it makes sense for it to do the filtering. A better fix would be to make prep cheaper, or increase the timeout. 

Assuming no races and instantaneous operations, we could have a timeout of infinity; the reason to have a finite timeout is because of those races.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Done.

scheduler/src/cook/scheduler/scheduler.clj
scrosby: Thanks, fixed.

scheduler/src/cook/scheduler/scheduler.clj
dposada: makes sense to me, thanks! 👍🏼

dposada (approved):

PR #1806, opened 2021-03-08
Ignores the space used by synthetic pods

scrosby:

- Have synthetic pods sleep for a while after starting so that the space
cannot be reused for other synthetic pods. 
- In addition, have Cook ignore
the space used by synthetic pods, so that Cook will barge in and take
those resources away.

Improve autoscaling. At present, an empty node can have synthetic pods sequentially match on it, attenuating the autoscaling signal Cook is intending to send.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -505,6 +505,7 @@
        (filter first) ; Keep those with non-nil node names.
        (pc/map-vals (fn [pods]
                       (->> pods
+                           (remove #(some-> % .getMetadata .getName synthetic-pod?))
                            (map (fn [^V1Pod pod]
                                   (let [containers (some-> pod .getSpec .getContainers)
                                         container-requests (map (fn [^V1Container c]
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -505,6 +505,7 @@
        (filter first) ; Keep those with non-nil node names.
        (pc/map-vals (fn [pods]
                       (->> pods
+                           (remove #(some-> % .getMetadata .getName synthetic-pod?))
```
dposada, line 508: Let's put this behind a feature flag so that we can toggle it on/off via configuration.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -447,7 +448,7 @@
               total-pods (-> @all-pods-atom keys count)
               total-nodes (-> @current-nodes-atom keys count)
               {:keys [image user command max-pods-outstanding max-total-pods max-total-nodes]
-               :or {command "exit 0" max-total-pods 32000 max-total-nodes 1000}} synthetic-pods-config]
+               :or {command "sleep 300" max-total-pods 32000 max-total-nodes 1000}} synthetic-pods-config]
```
dposada, line 451: The default for `clobber-synthetic-pods` appears to be `nil`, so it seems like the default for `command` should continue to be `exit 0`, so that the two defaults make sense together.

scrosby: Good catch!

dposada (approved):

dposada:
Will merge when green

PR #1813, opened 2021-03-10
Add support for node tenuring taint

scrosby:

- Add support for node tenuring taint

This lets us redirect synthetic pods toward the autoscaler instead of having them run on existing old nodes.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1079,10 +1079,11 @@
 
     (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
     (.addTolerationsItem pod-spec (toleration-for-pool cook-pool-taint-name cook-pool-taint-prefix pool-name))
-    (.addTolerationsItem pod-spec toleration-tenured-node)
     ; We need to make sure synthetic pods --- which don't have a hostname set --- have a node selector
-    ; to run only in nodes labelled with the appropriate cook pool
-    (when-not hostname (add-node-selector pod-spec cook-pool-label-name pool-name))
+    ; to run only in nodes labelled with the appropriate cook pool and only in un-tenured nodes.
+    (when-not hostname
+      (add-node-selector pod-spec cook-pool-label-name pool-name)
+      (.addTolerationsItem pod-spec toleration-tenured-node))
 
     (when pod-constraints
       (doseq [{:keys [constraint/attribute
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1068,6 +1079,7 @@
 
     (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
     (.addTolerationsItem pod-spec (toleration-for-pool cook-pool-taint-name cook-pool-taint-prefix pool-name))
+    (.addTolerationsItem pod-spec toleration-tenured-node)
```
dposada, line 1082: Won't this make synthetic pods tolerate the taint?

dposada (approved):

scheduler/src/cook/kubernetes/api.clj
scrosby: .... yes, they would. Good catch. Thanks.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1069,8 +1080,10 @@
     (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
     (.addTolerationsItem pod-spec (toleration-for-pool cook-pool-taint-name cook-pool-taint-prefix pool-name))
     ; We need to make sure synthetic pods --- which don't have a hostname set --- have a node selector
-    ; to run only in nodes labelled with the appropriate cook pool
-    (when-not hostname (add-node-selector pod-spec cook-pool-label-name pool-name))
+    ; to run only in nodes labelled with the appropriate cook pool and only in un-tenured nodes.
+    (when-not hostname
+      (add-node-selector pod-spec cook-pool-label-name pool-name)
+      (.addTolerationsItem pod-spec toleration-tenured-node))
```
dposada, line 1086: I think this is making only synthetic pods tolerate the taint, instead of only real job pods

dposada (approved):

dposada:
Will merge when green

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -46,6 +46,13 @@
 ; This pod annotation signals to the cluster autoscaler that
 ; it's safe to remove the node on which the pod is running
 (def k8s-safe-to-evict-annotation "cluster-autoscaler.kubernetes.io/safe-to-evict")
+; We want nodes older than 5 minutes to get this taint. A separate process taints those old nodes
+; with this taint. Cook ignores it for scheduling purposes, but synthetic pods are configured so they
+; don't ignore it. Cook creates synthetic pods as a scaling signal. When they run on existing nodes,
+; the signal we intend to send to the autoscaler is attenuated and we autoscale much less than intended.
+;
+; We prefix the taint with the ignore-taint string so that the autoscaler won't use this when constructing exlempar new nodes.
```
dposada, line 54:
```suggestion
; We prefix the taint with the ignore-taint string so that the autoscaler won't use this when constructing exemplar new nodes.
```

dposada (approved):

PR #1825, opened 2021-03-26
Gracefully handles nodes with no pods in get-consumption

dposada:

- filtering out nodes with empty or `nil` pods in `get-consumption`

In the wild, we occasionally see nodes at the beginning of their lifetime come through this code path with no pods associated to them. When this happens, an exception is thrown in the offer generation code and no offers are generated.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -512,7 +512,9 @@
   (->> node-name->pods
        ; Keep those with non-nil node names
        (filter first)
-       ; Keep those with non-nil and non-empty pods
+       ; Keep those with non-nil and non-empty pods (in the wild, we occasionally see nodes at the
+       ; beginning of their lifetime come through this code path with no pods associated to them, and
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -510,7 +510,10 @@
   See https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container"
   [clobber-synthetic-pods node-name->pods pool-name]
   (->> node-name->pods
-       (filter first) ; Keep those with non-nil node names.
+       ; Keep those with non-nil node names
+       (filter first)
+       ; Keep those with non-nil and non-empty pods
```
scrosby, line 515: Worth saying that we can see these early in a node's lifetime. (As in, this code is to protect from a particular landmine that occurs in a particular place)

dposada: Good call, done

scrosby (approved):

PR #1827, opened 2021-03-26
Allows namespace to be added to synthetic pod anti-affinity

dposada:

Allowing a namespace to be specified on the synthetic pod anti-affinity.

> because pods are namespaced (and therefore the labels on pods are implicitly namespaced), a label selector over pod labels must specify which namespaces the selector should apply to

(https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity)


scrosby:
Can you please add a unit test.

dposada:
@scrosby Added a unit test, thanks

scrosby (approved):

PR #1831, opened 2021-03-30
Treats prolonged ContainersNotReady condition as failure

dposada:
Similar to PR #1489, but for a different pod condition.

- adding a new failure reason "Container readiness timed out"
- fast-failing jobs using this new reason when they get into the `ContainersNotReady` condition for a prolonged (configurable) amount of time

Without this, pods can remain in the `ContainersNotReady` state forever, and their jobs stay in the `Running` state forever, despite not actually having started running the user's workload.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1204,35 +1204,61 @@
                           (t/plus (t/seconds unschedulable-seconds))
                           (t/before? (t/now)))))))))
 
-(defn pod-containers-not-initialized?
-  "Returns true if the given pod status has an Initialized condition with status
-  False and reason ContainersNotInitialized, and the last transition was more than
-  pod-condition-containers-not-initialized-seconds seconds ago"
-  [pod-name ^V1PodStatus pod-status]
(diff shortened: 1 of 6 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1204,36 +1204,62 @@
...
                           (t/before? (t/now)))))))))
 
-(defn pod-containers-not-initialized?
-  "Returns true if the given pod status has an Initialized condition with status
-  False and reason ContainersNotInitialized, and the last transition was more than
-  pod-condition-containers-not-initialized-seconds seconds ago"
-  [pod-name ^V1PodStatus pod-status]
-  (let [{:keys [pod-condition-containers-not-initialized-seconds]} (config/kubernetes)
-        ^V1PodCondition pod-condition
+(defn pod-has-terminal-condition?
```
nsinkov, line 1207: maybe "stuck" or "unready" instead of "terminal"?
terminal sounds like pod finished running

dposada: Good call, renamed to "stuck"

PR #1833, opened 2021-03-30
Do watch event processing in parallel during watch startup.

scrosby:

- Do watch event processing in parallel during watch startup.

- Faster startup and more rapid progression when we have to reconnect after a watch disconnection.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -199,7 +199,7 @@
 (declare initialize-pod-watch)
 (defn ^Callable initialize-pod-watch-helper
   "Help creating pod watch. Returns a new watch Callable"
-  [{:keys [^ApiClient api-client all-pods-atom node-name->pod-name->pod] compute-cluster-name :name :as compute-cluster} cook-pod-callback]
+  [{:keys [^ApiClient api-client all-pods-atom launch-task-executor-service node-name->pod-name->pod] compute-cluster-name :name :as compute-cluster}  cook-pod-callback]
   (let [[current-pods namespaced-pod-name->pod] (get-all-pods-in-kubernetes api-client compute-cluster-name)
         callbacks
         [(tools/make-atom-updater all-pods-atom) ; Update the set of all pods.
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -213,18 +213,25 @@
...
-          (catch Exception e
-            (log/error e "In" compute-cluster-name
-                       "compute cluster, pod watch error while processing callback for" task)))))
-
+    (let [tasks (set/union new-pod-names old-pod-names)
+          futures
+          (doall
+            (map (fn [task]
+                   (.submit
+                     launch-task-executor-service
```
nsinkov, line 221: why re-use launch-task-executor-service and not something that's just for the parallel node watch? can there be contention?

dposada: Also seems like we might want to configure a different number of threads for task-launching than for pod processing

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -199,7 +199,7 @@
 (declare initialize-pod-watch)
 (defn ^Callable initialize-pod-watch-helper
   "Help creating pod watch. Returns a new watch Callable"
-  [{:keys [^ApiClient api-client all-pods-atom node-name->pod-name->pod] compute-cluster-name :name :as compute-cluster} cook-pod-callback]
+  [{:keys [^ApiClient api-client all-pods-atom launch-task-executor-service node-name->pod-name->pod] compute-cluster-name :name :as compute-cluster}  cook-pod-callback]
```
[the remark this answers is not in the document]
scrosby: Thanks. Fixed.

scheduler/src/cook/kubernetes/api.clj
scrosby: No. The choice to use the same executor says that we have one control for maximum concurrency level though tthe state machine. We want this. as it controls the maximum outstanding requests we'll make to the k8s server. In addition, both the scan at watch startup as well as the launch rate are bulk, high-throughput operations that occur intermittently. Now we have one maximum knob to set instead of the maximum load being the sum of both of these (and hard synchronize to loadtest)

dposada: Can we give it a more appropriate name?

dposada: (given that it's now going to do more than launching tasks)

scrosby: Will do. controller-num-threads (and renamed the executor service correspondingly)

dposada (approved):

scheduler/src/cook/kubernetes/api.clj
scrosby: There can be contention in the threads, but that's planned contention. We want a knob to control max outstanding k8s requests, and the threadpool size is it.

PR #1840, opened 2021-04-05
Do fenzo unassign in batches outside of the fenzo lock

scrosby:

- Do unassignments asyncronously, adding to a list, and later unassigning just before matching.

This avoids the fenzo lock being held when processing the k8s state machine, which would otherwise cause all pod completions to get stuck waiting for that lock.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -60,7 +60,7 @@
   (:import (com.netflix.fenzo
              TaskAssignmentResult TaskRequest TaskScheduler TaskScheduler$Builder VirtualMachineCurrentState
              VirtualMachineLease VirtualMachineLease$Range)
-           (com.netflix.fenzo.functions Action1 Func1)
+           (com.netflix.fenzo.functions Action1 Action2 Func1)
            (java.util LinkedList)))
 
 (defn now
@@ -236,7 +236,7 @@
 
 (defn write-status-to-datomic
   "Takes a status update from mesos."
-  [conn pool->fenzo status]
+  [conn pool-name->fenzo-state status]
   (log/info "Instance status is:" status)
   (timers/time!
     handle-status-update-duration
@@ -275,21 +275,12 @@
                                    (.getTime (or (:instance/start-time instance-ent) current-time)))
                job-resources (tools/job-ent->resources job-ent)
                pool-name (cached-queries/job->pool-name job-ent)
-               ^TaskScheduler fenzo (get pool->fenzo pool-name)]
+               unassign-task-set (some-> (pool-name->fenzo-state pool-name) :unassign-task-set)]
            (when (#{:instance.status/success :instance.status/failed} instance-status)
-             (if fenzo
-               (try
-                 (log/debug "In" pool-name "pool, unassigning task"
-                            task-id "from" (:instance/hostname instance-ent))
-                 (locking fenzo
-                   (.. fenzo
-                       (getTaskUnAssigner)
-                       (call task-id (:instance/hostname instance-ent))))
-                 (catch Exception e
-                   (log/error e "In" pool-name "pool, failed to unassign task"
-                              task-id "from" (:instance/hostname instance-ent))))
+             (if unassign-task-set
+               (swap! unassign-task-set conj [task-id (:instance/hostname instance-ent)])
                (log/error "In" pool-name "pool, unable to unassign task" task-id "from"
-                          (:instance/hostname instance-ent) "because fenzo is nil:" pool->fenzo)))
+                          (:instance/hostname instance-ent) "because fenzo is nil:" (keys pool-name->fenzo-state))))
            (when (= instance-status :instance.status/success)
              (handle-throughput-metrics job-resources instance-runtime :succeeded pool-name)
              (handle-throughput-metrics job-resources instance-runtime :completed pool-name))
@@ -641,7 +632,7 @@
 
    Returns {:matches (list of tasks that got matched to the offer)
             :failures (list of unmatched tasks, and why they weren't matched)}"
-  [db ^TaskScheduler fenzo considerable offers rebalancer-reservation-atom pool-name]
+  [db {:keys [^TaskScheduler fenzo unassign-task-set]} considerable offers rebalancer-reservation-atom pool-name]
   (if (and (-> considerable count zero?)
            (-> offers count pos?)
            (every? :reject-after-match-attempt offers))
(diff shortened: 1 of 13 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1456,10 +1458,20 @@
                                    (when-let [job (tools/job-ent->map (:job/_instance task-ent))]
                                      (let [pool-name (cached-queries/job->pool-name job)
                                            task-request (make-task-request db job pool-name :task-id task-id)
-                                           ^TaskScheduler fenzo (pool->fenzo pool-name)]
+                                           {:keys [^TaskScheduler fenzo unassign-task-set]} (pool-name->fenzo-state pool-name)
+                                           to-unassign (util/set-atom! unassign-task-set #{})
```
dposada, line 1462: Why not use [`reset-vals!`](https://clojuredocs.org/clojure.core/reset-vals!) instead of `util/set-atom!`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -275,21 +275,12 @@
                                    (.getTime (or (:instance/start-time instance-ent) current-time)))
                job-resources (tools/job-ent->resources job-ent)
                pool-name (cached-queries/job->pool-name job-ent)
-               ^TaskScheduler fenzo (get pool->fenzo pool-name)]
+               unassign-task-set (some-> (pool-name->fenzo-state pool-name) :unassign-task-set)]
```
dposada, line 278: How about:

```suggestion
               unassign-task-set (some-> pool-name pool-name->fenzo-state :unassign-task-set)]
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -275,21 +275,12 @@
...
-                            task-id "from" (:instance/hostname instance-ent))
-                 (locking fenzo
-                   (.. fenzo
-                       (getTaskUnAssigner)
-                       (call task-id (:instance/hostname instance-ent))))
-                 (catch Exception e
-                   (log/error e "In" pool-name "pool, failed to unassign task"
-                              task-id "from" (:instance/hostname instance-ent))))
+             (if unassign-task-set
+               (swap! unassign-task-set conj [task-id (:instance/hostname instance-ent)])
```
dposada, line 281: Instead of a set of vectors, how about a set of maps, with each map being:

```
{:hostname ...
 :task-id ...}
```

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -673,7 +664,17 @@
             ;; Need to lock on fenzo when accessing scheduleOnce because scheduleOnce and
             ;; task assigner can not be called at the same time.
             ;; task assigner may be called when reconciling
+            to-unassign (util/set-atom! unassign-task-set #{})
```
dposada, line 667: Why not use [`reset-vals!`](https://clojuredocs.org/clojure.core/reset-vals!) instead of `util/set-atom!`?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -673,7 +664,17 @@
...
+            ^Action2 unassigner (.getTaskUnAssigner fenzo)
             result (locking fenzo
+                     (doseq [[task-id hostname] to-unassign]
+                       (try
+                         (log/debug "In" pool-name "pool, unassigning task"
+                                    task-id "from" hostname)
+                         (.call unassigner task-id hostname)
+                         (catch Exception e
+                           (log/error e "In" pool-name "pool, failed to unassign task"
+                                      task-id "from" hostname))))
```
dposada, line 677: Would be nice to include a comment about why Fenzo task unassignment needs to happen.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1456,10 +1458,20 @@
...
                                        ;; scheduleOnce can not be called at the same time.
                                        (locking fenzo
+                                         (doseq [[task-id hostname] to-unassign]
+                                           (try
+                                             (log/debug "In" pool-name "pool, unassigning task"
+                                                        task-id "from" hostname)
+                                             (.call unassigner task-id hostname)
+                                             (catch Exception e
+                                               (log/error e "In" pool-name "pool, failed to unassign task"
+                                                          task-id "from" hostname))))
```
dposada, line 1474: Instead of duplicating the unassignment code in `match-offer-to-schedule` and `reconcile-tasks`, can we extract a function for it?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -275,21 +275,12 @@
...
-                       (getTaskUnAssigner)
-                       (call task-id (:instance/hostname instance-ent))))
-                 (catch Exception e
-                   (log/error e "In" pool-name "pool, failed to unassign task"
-                              task-id "from" (:instance/hostname instance-ent))))
+             (if unassign-task-set
+               (swap! unassign-task-set conj [task-id (:instance/hostname instance-ent)])
                (log/error "In" pool-name "pool, unable to unassign task" task-id "from"
-                          (:instance/hostname instance-ent) "because fenzo is nil:" pool->fenzo)))
+                          (:instance/hostname instance-ent) "because fenzo is nil:" (keys pool-name->fenzo-state))))
```
dposada, line 283: `because fenzo is nil` doesn't really apply anymore

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1980,8 +1997,8 @@
             (log/error e "Exception in match-trigger-chan chime handler")
             (throw e)))
         (recur)))
-    (log/info "Pool name to fenzo scheduler map:" pool-name->fenzo)
+    (log/info "Pool name to fenzo scheduler map:" (keys pool-name->fenzo-state))
```
dposada, line 2000: The text of this log no longer matches what's being logged

scheduler/src/cook/scheduler/scheduler.clj
dposada: @scrosby Thoughts on this?

scheduler/src/cook/scheduler/scheduler.clj
dposada: @scrosby Thoughts on this?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -633,6 +624,21 @@
...
+    (try
+      ; Fenzo tracks which tasks are on which nodes with assignment. To, e.g., make
+      ; sure it can distribute tasks across hosts. This instructs fenzo that a task
+      ; is no longer on a given host.
+      (log/debug "In" pool-name "pool, unassigning task"
+                 task-id "from" hostname)
+      (.call unassigner task-id hostname)
+      (catch Exception e
+        (log/error e "In" pool-name "pool, failed to unassign task"
+                   task-id "from" hostname)))))
```
dposada, line 640: Should we wrap this whole `doseq` with a timer?

scrosby: No need. No blocking happens here, there's good logging, and the resulting metric isn't interesting, as it would depend on the batch size which is a function of Cook load and time between match cycles.

dposada: @scrosby Thoughts on this?

dposada (approved):

PR #1842, opened 2021-04-06
Make the lock vector a vector, not a sequence.

scrosby:

- Make the lock vector a vector, not a sequence.

So we can get access to locks with O(1) instead of O(n).

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -785,6 +785,7 @@
                                                     (atom state-locked?)
                                                     dynamic-cluster-config?
                                                     compute-cluster-launch-rate-limiter cook-pool-taint-name cook-pool-taint-prefix cook-pool-label-name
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -785,7 +785,7 @@
                                                     (atom state-locked?)
                                                     dynamic-cluster-config?
                                                     compute-cluster-launch-rate-limiter cook-pool-taint-name cook-pool-taint-prefix cook-pool-label-name
-                                                    (with-meta (repeatedly lock-shard-count #(ReentrantLock.))
+                                                    (with-meta (vec (repeatedly lock-shard-count #(ReentrantLock.)))
```
dposada, line 788: I think a comment explaining why the `vec` is important would be good.

scrosby: Comment added.

dposada (approved):

PR #1845, opened 2021-04-07
Fix time-until-waiting metric.

scrosby:

- We need to detach the timer so that we don't count waiting excessively for synthetic pods. We'll count each state revisit, not the first one where the pod shows as there.

Make the metric correct.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -566,7 +566,7 @@
                                           :pod/waiting (do
                                                          (when waiting-metric-timer (timers/stop waiting-metric-timer))
                                                          ; Delete the timer so we only track the first time.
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -566,7 +566,7 @@
                                           :pod/waiting (do
                                                          (when waiting-metric-timer (timers/stop waiting-metric-timer))
                                                          ; Delete the timer so we only track the first time.
-                                                         (dissoc cook-expected-state-dict waiting-metric-timer)))
+                                                         (dissoc cook-expected-state-dict :waiting-metric-timer)))
```
dposada, line 569: Unit test?

scrosby: Added one.

dposada (approved):

dposada:
Will merge when green

PR #1851, opened 2021-04-23
add override-group-immutability option to job client

nsinkov:

- add override-group-immutability option to job client

This is part of the API but not the job client and a user needs to be able to set it

### DIFF
--- a/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
+++ b/jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
@@ -106,6 +106,7 @@ final public class Job {
         private String _progressRegexString;
         private String _user;
         private JSONArray _datasets;
+        private Boolean _overrideGroupImmutability;
 
         /**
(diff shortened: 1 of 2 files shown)


jobclient/java/src/main/java/com/twosigma/cook/jobclient/Job.java
```diff
@@ -1279,6 +1309,9 @@ public static JSONObject jsonizeJob(Job job)
         if (job._datasets != null) {
             object.put("datasets", job._datasets);
         }
+        if (job._overrideGroupImmutability != null) {
+            object.put("override_group_immutability", job._overrideGroupImmutability);
```
scrosby, line 1313: This String key name seems like it should be a separate static final variable?

nsinkov: Yeah, but I'd rather keep the style consistent for the whole java client. We will probably want to refactor the whole thing later.

PR #1859, opened 2021-04-28
Optimize TaskRequestAdapter.

scrosby:

- This swaps a dictionary lookup for a simple field query in fenzo's matching logic.

More efficient matching in Fenzo. This saves mostly CPU, not scheduling latency.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -481,9 +481,11 @@
                         (offer-resource-ranges offer "ports"))))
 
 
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -481,9 +481,11 @@
                         (offer-resource-ranges offer "ports"))))
 
 
-(defrecord TaskRequestAdapter [job resources task-id assigned-resources guuid->considerable-cotask-ids constraints scalar-requests]
+
+; job and resources appear to be unused. They're not. Other code paths destructure job and resource out.
```
nsinkov, line 485:
```suggestion
; job and resources appear to be unused, but they are used. Other code paths destructure job and resource out.
```

PR #1863, opened 2021-04-29
Hard deletes pods that won't go away

dposada:

If a pod has a deletion timestamp older than a configurable amount of time (e.g. 15 minutes), Cook will hard-delete the pod (grace period = 0).

Without this, Cook will assume any pod with a deletion timestamp is already deleted, which is sometimes not the case.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1334,15 +1334,32 @@
           ; * A job may have additional containers with the name aux-*
           job-status (first (filter (fn [c] (= cook-container-name-for-job (.getName c)))
                                     container-statuses))
-          pod-preempted-timestamp (some-> pod .getMetadata .getLabels (get (or (-> (config/kubernetes) :node-preempted-label) "node-preempted")))
+          pod-metadata (some-> pod .getMetadata)
+          {:keys [node-preempted-label pod-deletion-timeout-seconds]} (config/kubernetes)
+          pod-preempted-timestamp
+          (some-> pod-metadata
+                  .getLabels
+                  (get (or node-preempted-label "node-preempted")))
+          pod-deletion-timestamp (some-> pod-metadata .getDeletionTimestamp)
           synthesized-pod-state
-          (if (some-> pod .getMetadata .getDeletionTimestamp)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -129,6 +129,12 @@
   (log-weird-state compute-cluster pod-name cook-expected-state-dict k8s-actual-state-dict)
   (kill-pod api-client name cook-expected-state-dict pod))
 
+(defn kill-pod-hard
+  "Kills the pod with a grace period of 0"
```
scrosby, line 133: This needs an explanation why we do this and when we do it and why we wouldn't do this everytime,

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1334,15 +1334,32 @@
...
-          (if (some-> pod .getMetadata .getDeletionTimestamp)
-            ; If a pod has been ordered deleted, treat it as if it was gone, It's being async removed.
-            ; Note that we distinguish between this explicit :missing, and not being there at all when processing
-            ; (:cook-expected-state/killed, :missing) in cook.kubernetes.controller/process
-            {:state :missing
-             :reason "Pod was explicitly deleted"
-             :pod-deleted? true}
+          (if pod-deletion-timestamp
+            ; If a pod has been ordered deleted, treat it as if it were gone, unless too
+            ; long (pod-deletion-timeout-seconds) has passed since its deletion timestamp
```
scrosby, line 1347: Don't like this. Basically we're 'un-:missing' a node if it was deleted too long ago, tunneling a weird state down to the :pod/unknown state, with a special flag for special behavior.  

E.g., this transition:

:pod/running -> :missing -> :pod/unknown (with :hard-delete?) -> :missing !?!

That's really violating the nature of s state machine. We should make a new state :pod/deleting, and map :deletionTimestamp being set to it. Make a state and it makes sense:

:pod/running -> :pod/deleting -> :missing

The state scanner will visit all pods, check the timestamp then and do a hard delete if old enough.

scrosby: Also, unit test.

dposada:
Replacing this with #1866

PR #1864, opened 2021-04-29
Introduces :pod/deleting state

dposada:

- Introducing a new k8s state, `:pod/deleting`, which until now has been captured by the combination of (`:missing` + `:pod-deleted? true`).
- Adding entries to the state machine for `:pod/deleting`, but (mostly) without changing behavior (in this PR).
- The one behavior change introduced here is that prior to this change, the state machine is sometimes generating two kill requests within 100ms of each other due to the fact that pods with a deletion timestamp are being treated the same as pods that are completely missing from k8s. The "opportunistic killing" code is there to avoid a race where the kill request enters the state machine before it's become aware of the pod's existence from the pod watch. This code makes sense, but only for the "actually missing from k8s" scenario, not for the "has a pod deletion timestamp" scenario. This PR avoids the "double kill" in the `:pod/deleting` case.

This is a preparatory change for another PR which will hard-delete pods if they've been in the `:pod/deleting` state for too long.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -458,6 +458,7 @@
                                         (case pod-synthesized-state-modified
                                           ; Cause this entry to be deleted by update-or-delete! called later down.
                                           :missing nil
+                                          :pod/deleting nil
                                           ; The writeback to datomic has occurred, so there's nothing to do except to delete the pod from kubernetes
                                           ; and remove it from our tracking.
                                           :pod/failed (delete-pod api-client name cook-expected-state-dict pod)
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -498,19 +499,18 @@
...
-                                                          "was killed with cook expected state"
-                                                          (prepare-cook-expected-state-dict-for-logging cook-expected-state-dict)
-                                                          "and k8s actual state"
-                                                          (prepare-k8s-actual-state-dict-for-logging k8s-actual-state-dict))
-                                                (log-weird-state compute-cluster pod-name
-                                                                 cook-expected-state-dict k8s-actual-state-dict))
+                                              (log-weird-state compute-cluster pod-name
+                                                               cook-expected-state-dict k8s-actual-state-dict)
                                               (handle-pod-killed compute-cluster pod-name)))
+                                          :pod/deleting
```
scrosby, line 505: Maybe worth a comment that this is an expected transition in the deletion path?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -498,19 +499,18 @@
...
-                                              ; deleting, that's an expected state. So, let's be selective with our logging.
-                                              (if (= (:state synthesized-state) :missing)
-                                                (log/info "In compute cluster" name ", pod" pod-name
-                                                          "was killed with cook expected state"
-                                                          (prepare-cook-expected-state-dict-for-logging cook-expected-state-dict)
-                                                          "and k8s actual state"
-                                                          (prepare-k8s-actual-state-dict-for-logging k8s-actual-state-dict))
-                                                (log-weird-state compute-cluster pod-name
-                                                                 cook-expected-state-dict k8s-actual-state-dict))
+                                              (log-weird-state compute-cluster pod-name
```
scrosby, line 502: May be worth a comment that this shouldn't occur except in the cases of extreme staleness; if you miss the deleting intermediate states.

scheduler/src/cook/kubernetes/controller.clj
dposada: Done

scheduler/src/cook/kubernetes/controller.clj
dposada: Done

PR #1866, opened 2021-05-03
Hard deletes pods that have been deleting for too long

dposada:

If a pod has a deletion timestamp older than a configurable amount of time (e.g. 15 minutes), Cook will hard-delete the pod (with grace period = 0).

Without this, Cook will assume any pod with a deletion timestamp is already deleted, which is sometimes not the case.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1313,6 +1313,19 @@
     "ContainersNotReady"
     (:pod-condition-containers-not-ready-seconds (config/kubernetes))))
 
+(defn has-old-deletion-timestamp?
+  "Returns true if the given pod has a deletion timestamp that is older
+  than the configurable timeout, meaning we should hard-kill the pod"
+  [^V1Pod pod]
+  (let [{:keys [pod-deletion-timeout-seconds]} (config/kubernetes)
+        pod-metadata (some-> pod .getMetadata)
+        pod-deletion-timestamp (some-> pod-metadata .getDeletionTimestamp)]
+    (and pod-deletion-timestamp
+         pod-deletion-timeout-seconds
+         (-> pod-deletion-timestamp
+             (t/plus (t/seconds pod-deletion-timeout-seconds))
+             (t/before? (t/now))))))
+
 (defn pod->synthesized-pod-state
   "From a V1Pod object, determine the state of the pod, waiting running, succeeded, failed or unknown.
 
@@ -1334,8 +1347,7 @@
           ; * A job may have additional containers with the name aux-*
           job-status (first (filter (fn [c] (= cook-container-name-for-job (.getName c)))
                                     container-statuses))
-          {:keys [node-preempted-label pod-deletion-timeout-seconds]}
-          (config/kubernetes)
+          {:keys [node-preempted-label]} (config/kubernetes)
           pod-metadata (some-> pod .getMetadata)
           pod-preempted-timestamp
           (some-> pod-metadata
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1334,11 +1334,25 @@
...
+          pod-preempted-timestamp
+          (some-> pod-metadata
+                  .getLabels
+                  (get (or node-preempted-label "node-preempted")))
+          pod-deletion-timestamp (some-> pod-metadata .getDeletionTimestamp)
           synthesized-pod-state
-          (if (some-> pod .getMetadata .getDeletionTimestamp)
-            {:state :pod/deleting
-             :reason "Pod was explicitly deleted"}
+          (if pod-deletion-timestamp
```
scrosby, line 1346: I feel this code is better placed in the (only) place that immediately uses it. I.e., in the state machine processing for deleting pods. Putting in a variable that's used a long distance away is not as clear as colocating creation and use.

To be more specific, move this code into a new function, something like  (should-hard-delete-pod [config pod] ...) or maybe 'get-pod-deletion-age'. that's a nice pure function of its arguments.

Have the code around line 591 in the state machine call that function to decide if hard deleting is needed or not. Give a good comment there too why we do this.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1334,11 +1334,25 @@
           ; * A job may have additional containers with the name aux-*
           job-status (first (filter (fn [c] (= cook-container-name-for-job (.getName c)))
                                     container-statuses))
-          pod-preempted-timestamp (some-> pod .getMetadata .getLabels (get (or (-> (config/kubernetes) :node-preempted-label) "node-preempted")))
+          {:keys [node-preempted-label pod-deletion-timeout-seconds]}
+          (config/kubernetes)
```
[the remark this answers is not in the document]

scheduler/src/cook/kubernetes/api.clj
dposada: Done

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1433,9 +1452,13 @@
 
 (defn delete-pod
   "Kill this kubernetes pod. This is the same as deleting it."
-  [^ApiClient api-client compute-cluster-name ^V1Pod pod]
+  [^ApiClient api-client compute-cluster-name ^V1Pod pod & {:keys [grace-period-seconds]}]
   (let [api (CoreV1Api. api-client)
-        ^V1DeleteOptions deleteOptions (-> (V1DeleteOptionsBuilder.) (.withPropagationPolicy "Background") .build)
+        delete-options-builder
+        (cond-> (V1DeleteOptionsBuilder.)
+                true (.withPropagationPolicy "Background")
```
scrosby, line 1459: May be worth a question. Do we want to delete it with background propagation policy, or should hard deletes have another policy?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -129,6 +129,15 @@
...
   (kill-pod api-client name cook-expected-state-dict pod))
 
+(defn kill-pod-hard
+  "Kills the pod with a grace period of 0 (a 'hard kill'). This is used when pods
+  are in the :pod/deleting state for an extended period of time, to force them to
+  actually go away. We don't do this for every kill because under normal
+  circumstances, we want to give processes a chance to properly handle SIGTERM."
+  [{:keys [api-client name]} pod-name {:keys [pod]}]
+  (log/info "In compute cluster" name ", pod" pod-name "needs to be hard-killed")
+  (kill-pod api-client name nil pod :grace-period-seconds 0))
```
scrosby, line 139: We need to propagate explain why its OK to set the cook expected state dict to nil here, versus propagating the existing dictionary. (In this case, the expected state dict is missing, so what we have is equivalent.) However, this is weird enough that we should document it, or propagate it.

scheduler/test/cook/test/kubernetes/controller.clj
```diff
@@ -163,7 +169,22 @@
       (is (not (nil? (:waiting-metric-timer (do-process-full-state :cook-expected-state/starting :missing))))))
 
     (is (nil? (do-process :missing :missing)))
-    (is (nil? (do-process :missing :pod/deleting)))
+    (let [hard-kill-atom (atom false)
```
scrosby, line 172: Maybe worth putting into a (testing "Testing hard delete on old pods" .....) block.

scheduler/src/cook/kubernetes/api.clj
dposada: Sounds like it doesn't matter because we're using naked pods, so there are no child resources

scheduler/src/cook/kubernetes/controller.clj

scheduler/test/cook/test/kubernetes/controller.clj
dposada: Done

PR #1873, opened 2021-05-13
Adds features to compute cluster

dposada:

Adding a `features` field to the compute cluster.

To eventually support constraining job instances to specific compute clusters based on the features those compute clusters support, e.g. specific NVIDIA driver versions.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -186,7 +186,8 @@
            compute-cluster-config/ca-cert
            compute-cluster-config/state
            compute-cluster-config/state-locked?
-           compute-cluster-config/location]}]
+           compute-cluster-config/location
+           compute-cluster-config/features]}]
   {:name name
    :template template
    :base-path base-path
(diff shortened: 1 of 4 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -287,14 +299,15 @@
      (doseq [key both-keys]
        (let [current-db-config
              (get current-db-configs key)
-             {current-in-mem-location :location :as current-in-mem-config}
+             {current-in-mem-location :location current-in-mem-features :features :as current-in-mem-config}
              (get current-in-mem-configs key)
              keys-to-keep-synced
              (cond-> [:base-path :ca-cert :state]
                      ; The location field is treated specially when comparing db and in-mem configs --
```
nsinkov, line 306: update this comment?

scheduler/src/cook/compute_cluster.clj
```diff
@@ -343,17 +356,24 @@
...
-      (= (dissoc-non-comparable-fields current-config)
-         (dissoc-non-comparable-fields new-config)))))
+  [{current-location :location current-features :features :as current-config}
+   {new-location :location new-features :features :as new-config}]
+  (cond (and (some? current-location)
+             (not= current-location new-location))
+        false
+
+        (and (seq current-features)
+             (not= current-features new-features))
```
nsinkov, line 366: do we care about order when comparing? enough to call set on them first?

scheduler/src/cook/compute_cluster.clj
dposada: Will do

scheduler/src/cook/compute_cluster.clj
dposada: Yeah, good call, will do

scheduler/src/cook/compute_cluster.clj
dposada: Done

scheduler/src/cook/compute_cluster.clj
dposada: Added a unit test and changed to call `set`

PR #1881, opened 2021-05-29
Supports pluggable job-routing

dposada:

Adding support for configurable job-routing to a Cook pool based on a "special" (configurable) pool name.

In order to support intelligent job routing within Cook, e.g. choosing the most appropriate of multiple possible pools for a job.

### DIFF
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -36,3 +36,6 @@
       ; We blocklist a given job from being autoscaled soon after a prior autoscaling.
       (.expireAfterWrite (:synthetic-pod-recency-seconds (config/kubernetes)) TimeUnit/SECONDS)
       (.build)))
+(mount/defstate ^Cache pool-name->exists?-cache :start (new-cache config/config))
+(mount/defstate ^Cache pool-name->accepts-submissions?-cache :start (new-cache config/config))
+(mount/defstate ^Cache pool-name->db-id-cache :start (new-cache config/config))
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -39,3 +39,4 @@
 (mount/defstate ^Cache pool-name->exists?-cache :start (new-cache config/config))
 (mount/defstate ^Cache pool-name->accepts-submissions?-cache :start (new-cache config/config))
 (mount/defstate ^Cache pool-name->db-id-cache :start (new-cache config/config))
+(mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -21,10 +21,11 @@
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [congestion.limits :refer [RateLimit]]
+            [cook.plugins.util :as putil]
             [cook.rest.impersonation :refer [impersonation-authorized-wrapper]]
             [cook.util :as util]
             [mount.core :as mount]
-            [plumbing.core :refer [fnk]]
+            [plumbing.core :refer [fnk] :as pc]
             [plumbing.graph :as graph])
   (:import (com.google.common.io Files)
            (com.netflix.fenzo VMTaskFitnessCalculator)
@@ -473,8 +474,7 @@
                                          :launch-wait-seconds (get fitness-calculator :launch-wait-seconds 60)
                                          :update-interval-ms (get fitness-calculator :update-interval-ms nil)}))
      :plugins (fnk [[:config {plugins {}}]]
-                (let [{:keys [job-launch-filter job-submission-validator
-                              pool-selection]} plugins]
+                (let [{:keys [job-launch-filter job-routing job-submission-validator pool-selection]} plugins]
                   (merge plugins
                          {:job-launch-filter
                           (merge
@@ -482,6 +482,17 @@
                              :age-out-first-seen-deadline-minutes 600
                              :age-out-seen-count 10}
                             job-launch-filter)
+                          :job-routing
+                          (pc/map-vals
+                            (fn [routing-config]
+                              (update
+                                routing-config
+                                :choose-pool-for-job-fn
+                                (fn [f]
+                                  (if-let [resolved-fn (putil/resolve-symbol (symbol f))]
+                                    resolved-fn
+                                    (throw (ex-info (str "Unable to resolve fn " f) {}))))))
+                            job-routing)
                           :job-submission-validator
                           (merge {:batch-timeout-seconds 40}
                                  job-submission-validator)
@@ -700,3 +711,7 @@
 (defn job-resource-limits
   []
   (-> config :settings :plugins :job-shape-validation))
+
+(defn job-routing
+  []
+  (-> config :settings :plugins :job-routing))
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -482,6 +482,12 @@
                              :age-out-first-seen-deadline-minutes 600
                              :age-out-seen-count 10}
                             job-launch-filter)
+                          ; The job-routing config has the following shape:
+                          ;
+                          ;   pool-name -> job-router-factory-fn
+                          ;
+                          ; where job-router-factory-fn must return
+                          ; something that satisfies the JobRouter protocol
                           :job-routing
                           (pc/map-vals
                             (fn [f]
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -21,6 +21,7 @@
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [congestion.limits :refer [RateLimit]]
+            [cook.plugins.definitions :refer [JobRouter]]
             [cook.plugins.util :as putil]
             [cook.rest.impersonation :refer [impersonation-authorized-wrapper]]
             [cook.util :as util]
@@ -31,7 +32,6 @@
            (com.netflix.fenzo VMTaskFitnessCalculator)
            (java.io File)
            (java.net InetAddress)
-           (java.util.concurrent.locks ReentrantLock)
            (org.apache.curator.test TestingServer)
            (org.apache.log4j DailyRollingFileAppender Logger PatternLayout)))
 
@@ -484,14 +484,13 @@
                             job-launch-filter)
                           :job-routing
                           (pc/map-vals
-                            (fn [routing-config]
-                              (update
-                                routing-config
-                                :choose-pool-for-job-fn
-                                (fn [f]
-                                  (if-let [resolved-fn (putil/resolve-symbol (symbol f))]
-                                    resolved-fn
-                                    (throw (ex-info (str "Unable to resolve fn " f) {}))))))
+                            (fn [f]
+                              (if-let [resolved-fn (putil/resolve-symbol (symbol f))]
+                                (let [job-router (resolved-fn)]
+                                  (if (and job-router (satisfies? JobRouter job-router))
+                                    job-router
+                                    (throw (ex-info (str "Calling fn " f " did not return a JobRouter") {}))))
+                                (throw (ex-info (str "Unable to resolve fn " f) {}))))
                             job-routing)
                           :job-submission-validator
                           (merge {:batch-timeout-seconds 40}
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -741,7 +741,7 @@
 
 (s/defn make-job-txn
   "Creates the necessary txn data to insert a job into the database"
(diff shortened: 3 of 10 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -2018,22 +2023,24 @@
      [false {::error \"...\"}]
 
   where \"...\" is a detailed error string describing the quota bounds exceeded."
-  [conn {:keys [::jobs ::pool] :as ctx}]
+  [conn {:keys [::jobs ::job-pool-pairs] :as ctx}]
```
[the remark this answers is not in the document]
dposada: Good call, will do

scheduler/src/cook/config.clj
```diff
@@ -473,15 +474,25 @@
...
-                              pool-selection]} plugins]
+                (let [{:keys [job-launch-filter job-routing job-submission-validator pool-selection]} plugins]
                   (merge plugins
                          {:job-launch-filter
                           (merge
                             {:age-out-last-seen-deadline-minutes 10
                              :age-out-first-seen-deadline-minutes 600
                              :age-out-seen-count 10}
                             job-launch-filter)
+                          :job-routing
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/config.clj
```diff
@@ -473,15 +474,25 @@
...
+                          (pc/map-vals
+                            (fn [routing-config]
```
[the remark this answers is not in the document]
dposada: Will do

scheduler/src/cook/rest/api.clj
```diff
@@ -2135,25 +2161,41 @@
...
                            ; Does the plugin accept the submission?
                            (if (= :accepted status)
                              [false {::groups groups
                                      ::jobs jobs
-                                     ::pool pool}]
+                                     ::job-pool-pairs
+                                     (map
+                                       (fn [{:keys [pool-name] :as m}]
+                                         (assoc
+                                           m :pool
```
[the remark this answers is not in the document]
dposada: Definitely. This will happen in a follow-up PR.

scheduler/src/cook/rest/api.clj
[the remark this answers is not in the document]
dposada: Renamed this to `job-pool-name-maps`

scheduler/src/cook/rest/api.clj
```diff
@@ -2104,10 +2129,11 @@
                          user (get-in ctx [:request :authorization/user])
                          override-group-immutability? (boolean (get params :override-group-immutability))
                          pool-name (or (get params :pool) (get headers "x-cook-pool"))
-                         pool (when pool-name (d/entity (d/db conn) [:pool/name pool-name]))
+                         pool (pool-name->pool conn pool-name)
```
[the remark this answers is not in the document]
dposada: I moved the pool look-ups to only when they're needed and cached them

scheduler/src/cook/rest/api.clj
```diff
@@ -2121,7 +2147,7 @@
                          [true {::error (str "Must supply at least one job or group to start."
                                              "Are you specifying that this is application/json?")}]
 
-                         (and pool-name (not pool))
+                         (and (not job-routing?) pool-name (not pool))
```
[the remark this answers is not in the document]
dposada: Added caching for the pool look-ups

scheduler/src/cook/config.clj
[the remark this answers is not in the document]
dposada: Good call, made this a `JobRouter` protocol

scheduler/src/cook/config.clj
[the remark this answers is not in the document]

scheduler/src/cook/rest/api.clj
```diff
@@ -866,7 +879,13 @@
                     executor (assoc :job/executor executor)
                     progress-output-file (assoc :job/progress-output-file progress-output-file)
                     progress-regex-string (assoc :job/progress-regex-string progress-regex-string)
-                    pool (assoc :job/pool (:db/id pool))
+                    pool-name-from-submission
```
scrosby, line 882: You're checking pool-name-from-submission as well as associating job-pool here. That mismatch seems weird and worth explaining why you're using submisison pool to gate this, but inserting a different pool. Please put in a comment.

scheduler/src/cook/rest/api.clj
```diff
@@ -1959,7 +1979,10 @@
           job-asserts (map (fn [j] [:entity/ensure-not-exists [:job/uuid (:uuid j)]]) jobs)
           [commit-latch-id commit-latch] (make-commit-latch)
           db (d/db conn)
-          job-txns (mapcat (partial make-job-txn pool commit-latch-id db) jobs)
+          job-txns (mapcat
+                     (fn [{:keys [job] :as job-pool-name-map}]
+                       (make-job-txn job-pool-name-map commit-latch-id db job))
```
scrosby, line 1984: This call is weird. You're extracting job from the map in the above line, then passing it as an argument. I think you can just strip the job argument out and have make-job-txn extract it.

scheduler/src/cook/rest/api.clj
```diff
@@ -2121,10 +2158,28 @@
                          [true {::error (str "Must supply at least one job or group to start."
                                              "Are you specifying that this is application/json?")}]
 
-                         (and pool-name (not pool))
+                         :let [db (db conn)
```
scrosby, line 2161: I think a comment may be worth it here. "We reject jobs unless either the pool exists or the pool is a job-routing-pool."

And we can simplify this code, I think as:

(or (not pool-name) (job-routing-pool-name? pool-name) (pool-name->exists?-cache ....)

With a comment with pool-name->exists?-cache saying that the cached value is either true or false. (Oooh! you are storing true or false here,  right? Caches cannot store nil values or keys.)

scheduler/src/cook/rest/api.clj
```diff
@@ -2121,10 +2158,28 @@
...
+                                   (lookup-cache-pool-name!
+                                     caches/pool-name->exists?-cache
+                                     db
+                                     some?
+                                     pool-name))]
+                         (not pool-exists?)
                          [true {::error (str pool-name " is not a valid pool name.")}]
 
-                         (and pool (not (pool/accepts-submissions? pool)))
+                         :let [pool-accepts-submissions?
```
scrosby, line 2175: Ditto here.

scheduler/src/cook/rest/api.clj
dposada: Adding a comment, good call.

The reason I'm creating a variable for:

```
skip-pool-name-checks?
(or (not pool-name)
    (job-routing-pool-name? pool-name))
```

is that I re-use that same `skip-pool-name-checks?` condition in the next check (the one for whether or not the pool is accepting new job submissions).

The caches are always caching `true` or `false` (the result of either `some?` or `pool/accepts-submissions?`).

scheduler/src/cook/rest/api.clj
dposada: Added a comment, good call

scheduler/src/cook/rest/api.clj
dposada: True, fixed

scheduler/src/cook/rest/api.clj
dposada: Added comment

scheduler/src/cook/rest/api.clj
dposada: Added comment

scheduler/src/cook/rest/api.clj
```diff
@@ -1942,10 +1966,11 @@
   to Datomic.
   Preconditions:  The context must already have been populated with both
   ::jobs and ::groups, which specify the jobs and job groups."
-  [conn {:keys [::groups ::jobs ::pool] :as ctx}]
+  [conn {:keys [::groups ::jobs ::job-pool-name-maps] :as ctx}]
```
scrosby, line 1969: This code is doing something weird. If  ::job-pool-name-maps is specified, it ignores ::jobs here and never makes transactions. Then later, we use jobs (unconditionally) when doing implicit-groups. 

If this is intentional, please document why. 

To me it seems like this should be refactored so that the caller puts puts the jobs into ::job-pool-name-maps, and ::jobs goes away.

Or if thats really hard to do, there should be a comment (and an assert) to point out that both can't be non-empty and/or catch when both of these are non-empty.

scheduler/src/cook/rest/api.clj
```diff
@@ -1973,22 +2001,23 @@
           group-txns (map #(make-group-txn % (get group-uuid->job-dbids
                                                   (:uuid %)
                                                   []))
-                          groups)
-          pool-name (pool/pool-name-or-default (:pool/name pool))]
+                          groups)]
 
       (let [user (get-in ctx [:request :authorization/user])]
         (rate-limit/spend! rate-limit/job-submission-rate-limiter user (count jobs))
```
scrosby, line 2007: If we have jobs empty and as well as job-pool-name-maps not empty, then this would be wrong..

scheduler/src/cook/rest/api.clj
```diff
@@ -2018,22 +2047,24 @@
      [false {::error \"...\"}]
 
   where \"...\" is a detailed error string describing the quota bounds exceeded."
-  [conn {:keys [::jobs ::pool] :as ctx}]
+  [conn {:keys [::jobs ::job-pool-name-maps] :as ctx}]
```
scrosby, line 2050: Isn't ::jobs dead here?  (And shouldn't it be dead?)

scheduler/src/cook/caches.clj
```diff
@@ -35,4 +35,7 @@
       (.maximumSize (:synthetic-pod-recency-size (config/kubernetes)))
       ; We blocklist a given job from being autoscaled soon after a prior autoscaling.
       (.expireAfterWrite (:synthetic-pod-recency-seconds (config/kubernetes)) TimeUnit/SECONDS)
-      (.build)))
\ No newline at end of file
+      (.build)))
+(mount/defstate ^Cache pool-name->exists?-cache :start (new-cache config/config))
```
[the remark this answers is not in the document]
dposada: I agree that they don't need to be a cache, but it's actually quite convenient for them to be a cache for the unit testing path. Querying all pools on startup adds complexity to unit tests, which often use `restore-fresh-database!` in the middle of a test in order to produce a new Datomic connection object.

scheduler/src/cook/rest/api.clj
dposada: Good call, no need for `::jobs` anymore, it's now removed.

scheduler/src/cook/rest/api.clj
dposada: `::jobs` is now gone, see above

scheduler/src/cook/rest/api.clj
dposada: Good call, removed

scheduler/src/cook/caches.clj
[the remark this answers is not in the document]
scrosby: Fair reason. You should make sure to add these caches to flush-caches! though to avoid unit tests depending on accidental configuration sitting in a cache.

scheduler/src/cook/rest/api.clj
```diff
@@ -2134,26 +2207,35 @@
...
+                                      :pool-name effective-pool-name
+                                      :pool-name-from-submission pool-name}))
+                                 jobs)
+                               jobs (map :job job-pool-name-maps)
+                               {:keys [status message]}
+                               (submission-plugin/plugin-jobs-submission job-pool-name-maps)]
                            ; Does the plugin accept the submission?
                            (if (= :accepted status)
                              [false {::groups groups
                                      ::jobs jobs
```
scrosby, line 2237: Is this ::jobs variable now dead?

scheduler/src/cook/caches.clj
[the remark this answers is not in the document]
dposada: Good call, added

scheduler/src/cook/rest/api.clj
dposada: There was one place still referencing `::jobs`, which I've now changed to `::job-pool-name-maps`, and I removed this `::jobs`.

scheduler/src/cook/rest/api.clj
```diff
@@ -2052,12 +2052,29 @@
   where \"...\" is a detailed error string describing the quota bounds exceeded."
   [conn {:keys [::job-pool-name-maps] :as ctx}]
   (let [db (db conn)
+        ; We cache quota by (user, pool) here because JobRouting plugins can result in
```
scrosby, line 2055: This makes me sad as I'll be adding a cache at/behind the cook.quota level when doing the postgresql migration.

PR #1890, opened 2021-06-21
Add passport logger and json passport log events

calebhar12:

- Adding a new logger that logs Cook passport events to a passport-only log file
- Adding a structured logging wrapper to log passport events as json
- Adding passport logging for the following events:
  - Job submission through raw API
  - Pod submission
  - Pod completion

- Having structured passport logging makes it easier to work with and filter logs

### DIFF
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -511,6 +511,8 @@
                           (merge {:attribute-name "cook-pool"
                                   :default-pool "no-pool"}
                                  pool-selection)})))
+     :passport (fnk [[:config {passport {}}]]
+                 passport)
      :kubernetes (fnk [[:config {kubernetes {}}]]
                    (let [{:keys [controller-lock-num-shards
                                  telemetry-tags-key-invalid-char-pattern]
@@ -692,6 +694,10 @@
 (defn compute-cluster-templates
   []
   (:compute-cluster-templates (compute-cluster-options)))
+
+(defn passport
+  []
+  (get-in config [:settings :passport]))
 
 (defn kubernetes
   []
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -16,7 +16,6 @@
 (ns cook.config
   (:require [clj-logging-config.log4j :as log4j-conf]
             [clj-time.core :as t]
-            [clojure.data.json :as json]
             [clojure.edn :as edn]
             [clojure.stacktrace :as stacktrace]
             [clojure.string :as str]
@@ -56,10 +55,6 @@
 
 (def passport-logger-ns "passport-logger")
 
-(defn log-passport-event
-  [log-data]
-  (log/log passport-logger-ns :info nil (json/write-str log-data)))
-
 (defn init-logger
   ([] (init-logger {:levels {"datomic.db" :warn
                              "datomic.peer" :warn
@@ -88,7 +83,7 @@
          {:out (DailyRollingFileAppender.
                  (PatternLayout.
                    "%d{ISO8601} %-5p %c [%t] - %m%n")
-                 "log/passport.log"
+                 "log/cook-passport.log"
                  "'.'yyyy-MM-dd")
           :level default}))
      (catch Throwable t
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -16,6 +16,7 @@
 (ns cook.config
   (:require [clj-logging-config.log4j :as log4j-conf]
             [clj-time.core :as t]
+            [clojure.data.json :as json]
             [clojure.edn :as edn]
             [clojure.stacktrace :as stacktrace]
             [clojure.string :as str]
@@ -53,6 +54,12 @@
       'config/env-bool #(Boolean/valueOf (env %))}}
     config))
 
+(def passport-logger-ns "passport-logger")
+
+(defn log-passport-event
+  [log-data]
+  (log/log passport-logger-ns :info nil (json/write-str log-data)))
+
 (defn init-logger
   ([] (init-logger {:levels {"datomic.db" :warn
                              "datomic.peer" :warn
@@ -75,7 +82,15 @@
                       file
                       "'.'yyyy-MM-dd")
                :level default}
-              overrides))
+              overrides)
+       (log4j-conf/set-loggers!
+         (Logger/getLogger ^String passport-logger-ns)
+         {:out (DailyRollingFileAppender.
+                 (PatternLayout.
+                   "%d{ISO8601} %-5p %c [%t] - %m%n")
+                 "log/passport.log"
+                 "'.'yyyy-MM-dd")
+          :level default}))
      (catch Throwable t
        (.println System/err (str "Failed to initialize logging! Error: " (.toString t)))
        (stacktrace/print-cause-trace t)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -74,15 +74,23 @@
   "If a pod is synthetic, return the uuid of the job it was created for"
   [pod-name]
   (when (synthetic-pod? pod-name)
-    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name))
-  )
+    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name)))
 
 (defn pod-name->instance-uuid
   "If a pod is not synthetic, return the instance id associated with its pod-name"
   [pod-name]
   (when-not (synthetic-pod? pod-name)
-    pod-name)
-  )
+    pod-name))
+
+(defn create-uuid-map
+  "Return a map with :job-uuid and/or :instance-uuid based on pod-name"
+  [pod-name]
+  (let [instance-uuid (pod-name->instance-uuid pod-name)
+        job-uuid (pod-name->job-uuid pod-name)]
+    (cond->
+      {}
+      instance-uuid (assoc :instance-uuid instance-uuid)
+      job-uuid (assoc :job-uuid job-uuid))))
 
 ; DeletionCandidateTaint is a soft taint that k8s uses to mark unneeded
 ; nodes as preferably unschedulable. This taint is added as soon as the
@@ -1656,16 +1664,13 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
-        (let [event-map {:pod-name pod-name
-                         :namespace namespace
-                         :cluster-name compute-cluster-name
-                         :event-type passport/pod-launching}
-              instance-uuid (pod-name->instance-uuid pod-name)
-              job-uuid (pod-name->job-uuid pod-name)
-              event-map (cond
-                          instance-uuid (assoc event-map :instance-uuid instance-uuid)
-                          job-uuid (assoc event-map :job-uuid job-uuid))]
-          (passport/log-passport-event event-map))
+        (let [event-map (merge
+                          {:cluster-name compute-cluster-name
+                           :event-type passport/pod-launching
+                           :namespace namespace
+                           :pod-name pod-name}
+                          (create-uuid-map pod-name))]
+          (passport/log-event event-map))
         (try
           (timers/time! (metrics/timer "launch-pod" compute-cluster-name)
                         (create-namespaced-pod api namespace pod))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -8,6 +8,7 @@
             [cook.cached-queries :as cached-queries]
             [cook.config :as config]
             [cook.kubernetes.metrics :as metrics]
+            [cook.passport :as passport]
             [cook.regexp-tools :as regexp-tools]
             [cook.rest.api :as api]
             [cook.scheduler.constraints :as constraints]
@@ -1641,10 +1642,11 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
-        (cook.config/log-passport-event {"pod-name" pod-name
-                                         "namespace" namespace
-                                         "cluster-name" compute-cluster-name
-                                         "event-type" "Launching Pod"})
+        (passport/log-passport-event {"pod-name" pod-name
+                                      "instance-uuid" pod-name
+                                      "namespace" namespace
+                                      "cluster-name" compute-cluster-name
+                                      "event-type" passport/pod-launching})
         (try
           (timers/time! (metrics/timer "launch-pod" compute-cluster-name)
                         (create-namespaced-pod api namespace pod))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1642,11 +1642,10 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
-        (passport/log-passport-event {"pod-name" pod-name
-                                      "instance-uuid" pod-name
-                                      "namespace" namespace
-                                      "cluster-name" compute-cluster-name
-                                      "event-type" passport/pod-launching})
+        (passport/log-passport-event {:pod-name pod-name
+                                      :namespace namespace
+                                      :cluster-name compute-cluster-name
+                                      :event-type passport/pod-launching})
         (try
           (timers/time! (metrics/timer "launch-pod" compute-cluster-name)
                         (create-namespaced-pod api namespace pod))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -70,6 +70,20 @@
   [pod-name]
   (str/starts-with? pod-name cook-synthetic-pod-name-prefix))
 
+(defn pod-name->job-uuid
+  "If a pod is synthetic, return the uuid of the job it was created for"
+  [pod-name]
+  (when (synthetic-pod? pod-name)
+    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name))
+  )
+
+(defn pod-name->instance-uuid
+  "If a pod is not synthetic, return the instance id associated with its pod-name"
+  [pod-name]
+  (when-not (synthetic-pod? pod-name)
+    pod-name)
+  )
+
 ; DeletionCandidateTaint is a soft taint that k8s uses to mark unneeded
 ; nodes as preferably unschedulable. This taint is added as soon as the
 ; autoscaler detects that nodes are under-utilized and all pods could be
@@ -1642,10 +1656,16 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
-        (passport/log-passport-event {:pod-name pod-name
-                                      :namespace namespace
-                                      :cluster-name compute-cluster-name
-                                      :event-type passport/pod-launching})
+        (let [event-map {:pod-name pod-name
+                         :namespace namespace
+                         :cluster-name compute-cluster-name
+                         :event-type passport/pod-launching}
+              instance-uuid (pod-name->instance-uuid pod-name)
+              job-uuid (pod-name->job-uuid pod-name)
+              event-map (cond
+                          instance-uuid (assoc event-map :instance-uuid instance-uuid)
+                          job-uuid (assoc event-map :job-uuid job-uuid))]
+          (passport/log-passport-event event-map))
         (try
           (timers/time! (metrics/timer "launch-pod" compute-cluster-name)
                         (create-namespaced-pod api namespace pod))
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1641,6 +1641,10 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
+        (cook.config/log-passport-event {"pod-name" pod-name
+                                         "namespace" namespace
+                                         "cluster-name" compute-cluster-name
+                                         "event-type" "Launching Pod"})
         (try
           (timers/time! (metrics/timer "launch-pod" compute-cluster-name)
                         (create-namespaced-pod api namespace pod))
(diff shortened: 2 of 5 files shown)


scheduler/src/cook/config.clj
```diff
@@ -52,6 +53,12 @@
       'config/env-bool #(Boolean/valueOf (env %))}}
     config))
 
+(def passport-logger-ns "passport-logger")
+
+(defn log-passport-event
+  [log-data]
+  (log/log passport-logger-ns :info nil (json/write-str log-data)))
```
dposada, line 60: Is `cook.config` the right namespace for this function? Maybe we should have a `cook.passport` namespace?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1592,6 +1592,10 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
+        (cook.config/log-passport-event {"pod-name" pod-name
+                                         "namespace" namespace
+                                         "cluster-name" compute-cluster-name
+                                         "event-type" "Launching Pod"})
```
dposada, line 1598: Any thoughts on keys being :symbols instead of "strings"?

dposada: Maybe we should define all of the event-type values in a single place? Also, maybe something like `pod-launching` instead of `Launching Pod`?

dposada: Would it make sense to add `instance-uuid` as a key for this event? I know it's the same as the pod name, but we probably want to be able to have a consistent key to tie things back to a job instance.

scheduler/src/cook/config.clj
```diff
@@ -74,7 +81,15 @@
...
                       "'.'yyyy-MM-dd")
                :level default}
-              overrides))
+              overrides)
+       (log4j-conf/set-loggers!
+         (Logger/getLogger ^String passport-logger-ns)
+         {:out (DailyRollingFileAppender.
+                 (PatternLayout.
+                   "%d{ISO8601} %-5p %c [%t] - %m%n")
+                 "log/passport.log"
```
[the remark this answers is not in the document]
calebhar12: Agreed

scheduler/src/cook/config.clj
calebhar12: Yeah, having a `cook.passport` namespace probably makes more sense. Added it in `cook.config` originally b/c that's where we create the logger but I agree that the `log-passport-event` function should have a different namespace

scheduler/src/cook/kubernetes/api.clj
calebhar12: > Any thoughts on keys being :symbols instead of "strings"?

Yeah, wasn't sure what the convention would be in this case. I've generally been using :symbols as the keys for all maps, but in this case since the map is created then immediately converted to a JSON string I figured strings would be appropriate. Open to discussing more though.

> Maybe we should define all of the event-type values in a single place? Also, maybe something like `pod-launching` instead of `Launching Pod`?

I like both of these changes-- the definitions will also have a good place to 'live' once I create the `cook.passport` namespace for the logging function.

> Would it make sense to add `instance-uuid` as a key for this event? I know it's the same as the pod name, but we probably want to be able to have a consistent key to tie things back to a job instance.

Yeah, I can add it. We'll also be iterating on what fields the messages have as we go along so any suggestions for info to include are welcome.

calebhar12: > Would it make sense to add `instance-uuid` as a key for this event? I know it's the same as the pod name, but we probably want to be able to have a consistent key to tie things back to a job instance.

Are you saying to remove `pod-name` and just replace it with `instance-uuid`, or are you saying to keep both (even though they're the same)?

dposada: > Are you saying to remove pod-name and just replace it with instance-uuid, or are you saying to keep both (even though they're the same)?

I don't feel too strongly, but keeping both seems reasonable to me.

nsinkov: we don't have instance-uuid on a synthetic pod launch event. I would just keep pod name only. There will be other events that will need to be mapped to each other in a different layer.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -152,6 +153,18 @@
                                      @(:pool-name->fenzo-state-atom compute-cluster)
                                      mesos-status))
 
+(defn write-status-to-cook-passport
```
dposada, line 156: Maybe just `write-status-to-passport` (since Cook is implied)?

scheduler/src/cook/passport.clj
```diff
@@ -0,0 +1,15 @@
+(ns cook.passport
+  (:require [clojure.data.json :as json]
+            [clojure.tools.logging :as log]
+            [congestion.limits :refer [RateLimit]]
+            [cook.config :as config]
+            [cook.rest.impersonation :refer [impersonation-authorized-wrapper]]
+            [plumbing.core :refer [fnk]]))
```
dposada, line 7: Some of these are unused, aren't they?

scheduler/src/cook/passport.clj
```diff
@@ -0,0 +1,15 @@
...
+
+(defn log-passport-event
```
dposada, line 9: How about `stamp` to go along with the passport analogy? Callers would then be calling it with `(passport/stamp {...})`, which seems both clear and concise.

scheduler/src/cook/passport.clj
```diff
@@ -0,0 +1,15 @@
...
+  [log-data]
+  (log/log config/passport-logger-ns :info nil (json/write-str log-data)))
+
+(def api-job-submission "raw-api-job-submission")
```
dposada, line 13: Nitpick: is "raw-" needed? Maybe it's better without it?

scheduler/src/cook/rest/api.clj
```diff
@@ -1945,6 +1946,10 @@
   [conn {:keys [::groups ::jobs ::pool] :as ctx}]
   (try
     (log/info "Submitting jobs through raw api:" (map #(dissoc % :command) jobs))
+    (doseq [job jobs]
+      (passport/log-passport-event {"job-uuid" (str (:uuid job))
+                                    "user" (:user job)
```
dposada, line 1951:
```suggestion
    (doseq [{:keys [user uuid]} jobs]
      (passport/log-passport-event {"job-uuid" (str uuid)
                                    "user" user
```

dposada: I think destructuring as above would work

scheduler/src/cook/kubernetes/api.clj
dposada: On a synthetic pod launch event, we don't have instance uuid, but we do have job uuid. 

I'd favor including either a `job-uuid` or `instance-uuid` key on every passport event.

What's the downside?

nsinkov: The code would need to be refactored to pass in job-uuid and instance-uuid and it doesn't seen like there's a quick way to do it now. We also have a plan to have a passport query service that would have logic to join events. I want to spend some time designing that first and then finalizing the fields on all the events.

dposada: Which code would need to be refactored?

calebhar12: I don't see `job_uuid` or `instance_uuid` being passed into the `launch-pod` function, and it's not obviously seen in the `cook-expected-state-dict` map that gets passed into the function. I think the refactor @nsinkov mentioned would be refactoring to pass in the `job_uuid` and `instance_uuid` into the function.

calebhar12: Ah good point, sorry I should have been more exact with my comment. I think that `job_uuid` not being passed in is the real issue. 

> we don't have instance-uuid on a synthetic pod launch event. I would just keep pod name only. There will be other events that will need to be mapped to each other in a different layer.

It relates back to the comment from @nsinkov I quoted, but for synthetic pods we can't use `pod-name` as the `instance_uuid`. So, if we want to include `job_uuid` or `instance_uuid` in every log, we would have to refactor `launch-pod` function to pass `job_uuid` into it in order to log `job_uuid` when synthetic pods are launched.

calebhar12: Nikolai & I also talked a bit more about the passport query service a bit more after our end-user meeting yesterday and the plan is definitely for the passport query service to handle most/all of the logic for linking events. So at the moment I'm leaning towards just keeping `pod-name` as the field in this event.

scheduler/src/cook/kubernetes/controller.clj
calebhar12: Sounds good to me

scheduler/src/cook/passport.clj
calebhar12: Ah yeah, good catch I meant to remove those before pushing but forgot to :/ my bad

scheduler/src/cook/passport.clj
calebhar12: Yeah, I like that

scheduler/src/cook/passport.clj
calebhar12: I took `raw-api-job-submission` from the existing cook log in that spot-- I wasn't sure if `raw` served a purpose or not so I just kept it in, if it doesn't I'll remove it though

scheduler/src/cook/kubernetes/api.clj
nsinkov: @dposada "The pod-name argument passed to launch-pod is the instance uuid" not for synthetic pods

scheduler/src/cook/rest/api.clj
calebhar12: ah very cool I'll do this

scheduler/src/cook/kubernetes/api.clj
dposada: Yep, you're right, for synthetic pods, there is no relevant instance uuid, but we should be able to add a job-uuid field for synthetic pods.

scheduler/src/cook/passport.clj
dposada: Don't think it serves a purpose

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -69,6 +70,20 @@
   [pod-name]
   (str/starts-with? pod-name cook-synthetic-pod-name-prefix))
 
+(defn pod-name->job-uuid
+  "If a pod is synthetic, return the uuid of the job it was created for"
+  [pod-name]
+  (when (synthetic-pod? pod-name)
+    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name))
+  )
```
dposada, line 78:
```suggestion
    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name)))
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -69,6 +70,20 @@
...
+
+(defn pod-name->instance-uuid
+  "If a pod is not synthetic, return the instance id associated with its pod-name"
+  [pod-name]
+  (when-not (synthetic-pod? pod-name)
+    pod-name)
+  )
```
dposada, line 85:
```suggestion
    pod-name))
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1641,6 +1656,16 @@
...
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
+        (let [event-map {:pod-name pod-name
+                         :namespace namespace
+                         :cluster-name compute-cluster-name
+                         :event-type passport/pod-launching}
+              instance-uuid (pod-name->instance-uuid pod-name)
+              job-uuid (pod-name->job-uuid pod-name)
+              event-map (cond
+                          instance-uuid (assoc event-map :instance-uuid instance-uuid)
+                          job-uuid (assoc event-map :job-uuid job-uuid))]
```
dposada, line 1667: Thoughts on using `cond->` (something like below)?

Nitpick: We usually sort map literal keys lexicographically.

```suggestion
        (let [instance-uuid (pod-name->instance-uuid pod-name)
              job-uuid (pod-name->job-uuid pod-name)
              event-map (cond-> 
                          {:cluster-name compute-cluster-name
                           :event-type passport/pod-launching
                           :namespace namespace
                           :pod-name pod-name}
                          instance-uuid (assoc :instance-uuid instance-uuid)
                          job-uuid (assoc :job-uuid job-uuid))]
```

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -152,6 +153,23 @@
...
+        event-map {:pod-name pod-name
+                   :cluster-name name
+                   :state state
+                   :reason reason
+                   :event-type passport/pod-completed}
+        instance-uuid (api/pod-name->instance-uuid pod-name)
+        job-uuid (api/pod-name->job-uuid pod-name)
+        event-map (cond
+                    instance-uuid (assoc event-map :instance-uuid instance-uuid)
+                    job-uuid (assoc event-map :job-uuid job-uuid))]
```
dposada, line 170: Should we make this a helper function since it's being done in a couple places already?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -152,6 +153,23 @@
...
+(defn write-status-to-passport
+  "Helper function for logging completed job status to cook passport"
+  [{:keys [name]}
+   {:keys [task-id state reason]}]
+  (let [pod-name (task-id :value)
+        event-map {:pod-name pod-name
+                   :cluster-name name
+                   :state state
+                   :reason reason
+                   :event-type passport/pod-completed}
```
dposada, line 165: Nitpick: lexicographic map keys

scheduler/src/cook/passport.clj
dposada: Still planning to do this?

scheduler/src/cook/passport.clj
```diff
@@ -0,0 +1,12 @@
...
+            [clojure.tools.logging :as log]
+            [cook.config :as config]))
+
+(defn log-passport-event
+  [log-data]
+  (log/log config/passport-logger-ns :info nil (json/write-str log-data)))
+
+(def api-job-submission "api-job-submission")
+(def pod-launching "pod-launching")
+(def pod-completed "pod-completed")
```
dposada, line 12: Thoughts on these values being :symbols instead of "strings"?

scheduler/src/cook/rest/api.clj
```diff
@@ -1968,10 +1969,14 @@
...
-                     (update job-pool-name-map :job #(dissoc % :command)))
-                   job-pool-name-maps))
+    (let [jobs (map (fn [job-pool-name-map]
+                      (update job-pool-name-map :job #(dissoc % :command)))
+                    job-pool-name-maps)]
+      (log/info "Submitting jobs through raw api:" jobs)
+      (doseq [{{:keys [uuid, user]} :job} jobs]
+        (passport/log-passport-event {:job-uuid (str uuid)
+                                      :user user
+                                      :event-type passport/api-job-submission})))
```
dposada, line 1979: Would it make more sense to log this event write after the `d/transact` call returns lower in this function? That's when we're doing similar things like incrementing the queue length and marking the `jobs-created` metric.

scheduler/src/cook/passport.clj
nsinkov: Hmmm, I don't know if passport/stamp is clear enough

nsinkov: my vote is for `passport/log-event`

dposada: I'm fine with `passport/log-event`

scheduler/src/cook/kubernetes/api.clj
calebhar12: Good catch, fixing this.

scheduler/src/cook/kubernetes/api.clj
calebhar12: Fixing this.

scheduler/src/cook/kubernetes/api.clj
calebhar12: Yeah, I'll try using `cond->` like your change. Like this change also because `cond` short-circuits and `cond->` does not.

scheduler/src/cook/kubernetes/controller.clj
calebhar12: Will fix this

scheduler/src/cook/kubernetes/controller.clj
calebhar12: Yeah good call, it's the exact same in both spots. I might add it as a helper function in the `passport` namespace.

scheduler/src/cook/passport.clj
calebhar12: Sorry, I should have included a note when I posted the PR. Ultimately decided not to change it b/c I think that `passport/stamp` isn't quite clear enough for people who aren't already very familiar with the passport analogy. I'm good with changing it to `passport/log-event` though to shorten it up and avoid the repetition of `passport`.

scheduler/src/cook/passport.clj
calebhar12: I can change them to :symbols (just tested and the json log code formats them correctly).  These fields are only used as the values of maps and not as keys though-- is it still our convention to have them be :symbols?

calebhar12: I'm fine with either solution, I just didn't know that having :symbols as values was preferred.

scheduler/src/cook/rest/api.clj
calebhar12: I view "job submission" as the point at which we have the jobs and are beginning to create them, not after we create them (i.e. after the `d/transact` call returns), so to me I think that the beginning of the function is the most logical point to log the event. 

Also, if the `d/transact` function begins to slow down for some reason, we want the passport events to show that behavior. Specifically, we would want the passport metrics aggregator to show that the time between `api-job-submission` and, for example, `pod-launching` has increased. If we log `api-job-submission` after the `d/transact` call, that difference would not be captured.

scheduler/src/cook/kubernetes/controller.clj
calebhar12: actually going to keep it in the `kubernetes.api` namespace for consistency.

scheduler/src/cook/passport.clj
dposada: I think it's slightly more idiomatic Clojure to have them as symbols

scheduler/src/cook/rest/api.clj
dposada: SGTM.

In that case, is there a way we can avoid the extra `let` and the new `jobs` binding (there is already a `jobs` binding below)? Maybe leave the "Submitting jobs through raw api" code as it was and move the passport event logging into the existing `let`, right below the existing `jobs` binding?

Nitpick: lexicographic map keys for the passport event map

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -69,6 +70,28 @@
...
+    (re-find #"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}" pod-name)))
+
+(defn pod-name->instance-uuid
+  "If a pod is not synthetic, return the instance id associated with its pod-name"
+  [pod-name]
+  (when-not (synthetic-pod? pod-name)
+    pod-name))
+
+(defn create-uuid-map
+  "Return a map with :job-uuid and/or :instance-uuid based on pod-name"
```
dposada, line 86: It will currently never `assoc` both, right?

That being said, we probably eventually want to have the job uuid everywhere that we have the instance uuid.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1641,6 +1664,13 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
+        (let [event-map (merge
+                          {:cluster-name compute-cluster-name
+                           :event-type passport/pod-launching
+                           :namespace namespace
+                           :pod-name pod-name}
+                          (create-uuid-map pod-name))]
```
dposada, line 1672: Maybe to avoid the `merge`, you could call the helper function something like `assoc-uuid` and have it take in the map to `assoc` onto as an argument?

nsinkov (approved):

scheduler/src/cook/rest/api.clj
```diff
@@ -1968,10 +1969,14 @@
...
-                     (update job-pool-name-map :job #(dissoc % :command)))
-                   job-pool-name-maps))
+    (let [jobs (map (fn [job-pool-name-map]
+                      (update job-pool-name-map :job #(dissoc % :command)))
+                    job-pool-name-maps)]
+      (log/info "Submitting jobs through raw api:" jobs)
+      (doseq [{{:keys [uuid, user]} :job} jobs]
+        (passport/log-event {:job-uuid (str uuid)
+                                      :user user
+                                      :event-type passport/api-job-submission})))
```
dposada, line 1979: Indentation seems off on these two lines

scheduler/src/cook/rest/api.clj
calebhar12: If we moved it into the second let, the passport logging would have to go below the bindings, right? Similar to my other comment, I'd worry that if a function call like (d/db conn) becomes slow, we wouldn't capture that in the Passport event logs. I also sort of think that organizationally I prefer seeing the two logging statements within the same let function & occurring immediately in the `create-jobs!` function. 

But it definitely isn't strictly necessary to have it in the first `let` statement, and if you feel that it's most beneficial to avoid the extra `let` I can do that in the next revision. Let me know!

scheduler/src/cook/kubernetes/api.clj
calebhar12: Yeah, currently it will never assoc both, I just said "and/or" because at some point we might want both. I can just change it to "or" to be more exact if that's preferable.

And yeah, eventually we do want to have job uuid (and will have it). But it looks like that will involve a more involved refactor and we think it's out of scope for this PR right now.

scheduler/src/cook/rest/api.clj
nsinkov: Are you sure? What do you think is wrong?

scheduler/src/cook/kubernetes/api.clj
calebhar12: That makes sense, I can use `cond->` as before too.

scheduler/src/cook/rest/api.clj
calebhar12: Ah got this way after changing `log-passport-event` to `log-event`, I'll fix it.

scheduler/src/cook/rest/api.clj
```diff
@@ -1968,10 +1969,14 @@
...
-                     (update job-pool-name-map :job #(dissoc % :command)))
-                   job-pool-name-maps))
+    (let [jobs (map (fn [job-pool-name-map]
+                      (update job-pool-name-map :job #(dissoc % :command)))
+                    job-pool-name-maps)]
+      (log/info "Submitting jobs through raw api:" jobs)
+      (doseq [{{:keys [uuid, user]} :job} jobs]
+        (passport/log-event {:event-type passport/api-job-submission
+                             :job-uuid (str uuid)
+                             :user user})))
```
dposada, line 1979: How about this, to avoid the extra `let`?

```suggestion
    (log/info "Submitting jobs through raw api:"
              (map (fn [job-pool-name-map]
                     (update job-pool-name-map :job #(dissoc % :command)))
                   job-pool-name-maps))
    (doseq [{{:keys [uuid, user]} :job} job-pool-name-maps]
      (passport/log-event {:event-type passport/api-job-submission
                           :job-uuid (str uuid)
                           :user user}))
```

scheduler/src/cook/rest/api.clj
dposada: Sounds reasonable, I made a suggestion below that keeps them next to each other but removes the need for an additional `let`

scheduler/src/cook/rest/api.clj
calebhar12: Sure, this works for me. I'll add & test it then post another revision

dposada (approved):

dposada (approved):

PR #1898, opened 2021-07-08
[DRAFT] Add instance-uuid->job-uuid cache and update kubernetes/controller tests

calebhar12:

- Add a cache mapping `instance-uuid` to `job-uuid`
- Update `assoc-uuids` function to get `job-uuid`s from cache when needed and append them to passport event maps
- Add `with-redefs` logic to spoof cache hits in `cook.test.kubernetes.controller` tests

- Adding this cache gives us the ability to log `job-uuid` for all Passport events in the state machine where `job-uuid` is not immediately visible
- These changes will give us a way to associate `instance-uuid` with `job-uuid` without querying datomic each time we want to access it

### DIFF
--- a/scheduler/src/cook/cached_queries.clj
+++ b/scheduler/src/cook/cached_queries.clj
@@ -20,7 +20,7 @@
   [job-ent]
   (caches/lookup-cache-datomic-entity! caches/job-ent->user-cache :job/user job-ent))
 
-(defn instance-uuid->job-uuid
+(defn instance-uuid->job-uuid-datomic-query
   "Queries for the job uuid from an instance uuid.
    Returns nil if the instance uuid doesn't correspond
    to a job"
@@ -31,7 +31,7 @@
 
 (let [miss-fn
       (fn [instance-uuid]
-        (instance-uuid->job-uuid (d/db datomic/conn) instance-uuid))]
+        (instance-uuid->job-uuid-datomic-query (d/db datomic/conn) instance-uuid))]
   (defn instance-uuid->job-uuid-cache-lookup
     "Get value from cache if it is present, else search datomic for it"
     [instance-uuid]
--- a/scheduler/src/cook/cached_queries.clj
+++ b/scheduler/src/cook/cached_queries.clj
@@ -1,6 +1,9 @@
 (ns cook.cached-queries
-  (:require [cook.caches :as caches]
-            [cook.config :as config]))
+  (:require [cook.cache :as ccache]
+            [cook.caches :as caches]
+            [cook.config :as config]
+            [cook.datomic :as datomic]
+            [datomic.api :as d :refer [q]]))
 
 (let [miss-fn
       (fn [{:keys [job/pool]}]
@@ -16,3 +19,20 @@
   "Given a job entity, return the user the job runs as."
   [job-ent]
   (caches/lookup-cache-datomic-entity! caches/job-ent->user-cache :job/user job-ent))
+
+(defn instance-uuid->job-uuid
+  "Queries for the job uuid from an instance uuid.
+   Returns nil if the instance uuid doesn't correspond
+   to a job"
+  [db instance-uuid]
+  (->> (d/entity db [:instance/task-id (str instance-uuid)])
+       :job/_instance
+       :job/uuid))
+
+(let [miss-fn
+      (fn [instance-uuid]
+        (instance-uuid->job-uuid (d/db datomic/conn) instance-uuid))]
+  (defn instance-uuid->job-uuid-cache-lookup
+    "Get value from cache if it is present, else search datomic for it"
+    [instance-uuid]
+    (ccache/lookup-cache! caches/instance-uuid->job-uuid identity miss-fn instance-uuid)))
--- a/scheduler/src/cook/cached_queries.clj
+++ b/scheduler/src/cook/cached_queries.clj
@@ -31,7 +31,7 @@
 
 (let [miss-fn
       (fn [instance-uuid]
-        (instance-uuid->job-uuid-datomic-query (d/db datomic/conn) instance-uuid))]
+        (str (instance-uuid->job-uuid-datomic-query (d/db datomic/conn) instance-uuid)))]
   (defn instance-uuid->job-uuid-cache-lookup
     "Get value from cache if it is present, else search datomic for it"
     [instance-uuid]
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -40,3 +40,8 @@
 (mount/defstate ^Cache pool-name->accepts-submissions?-cache :start (new-cache config/config))
 (mount/defstate ^Cache pool-name->db-id-cache :start (new-cache config/config))
 (mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
+(mount/defstate ^Cache instance-uuid->job-uuid :start
+  (-> (CacheBuilder/newBuilder)
+    (.maximumSize (get-in config/config [:settings :cache-working-set-size])) ;1million
+    (.expireAfterAccess (get-in config/config [:settings :passport :instance-uuid->job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+    (.build)))
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -42,6 +42,6 @@
 (mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
 (mount/defstate ^Cache instance-uuid->job-uuid :start
   (-> (CacheBuilder/newBuilder)
-    (.maximumSize (get-in config/config [:settings :cache-working-set-size])) ;1million
+    (.maximumSize (get-in config/config [:settings :cache-working-set-size]))
     (.expireAfterAccess (get-in config/config [:settings :passport :instance-uuid->job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
     (.build)))
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -42,6 +42,6 @@
 (mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
 (mount/defstate ^Cache instance-uuid->job-uuid :start
   (-> (CacheBuilder/newBuilder)
-    (.maximumSize (get-in config/config [:settings :cache-working-set-size]))
-    (.expireAfterAccess (get-in config/config [:settings :passport :instance-uuid->job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+    (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
+    (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
     (.build)))
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -512,8 +512,8 @@
                                   :default-pool "no-pool"}
                                  pool-selection)})))
      :passport (fnk [[:config {passport {}}]]
-                 (merge passport
-                        {:instance-uuid->job-uuid-cache-expiry-time-hours 24}))
+                 (merge {:instance-uuid->job-uuid-cache-expiry-time-hours 24}
+                        passport))
      :kubernetes (fnk [[:config {kubernetes {}}]]
                    (let [{:keys [controller-lock-num-shards
                                  telemetry-tags-key-invalid-char-pattern]
--- a/scheduler/src/cook/config.clj
+++ b/scheduler/src/cook/config.clj
@@ -512,7 +512,8 @@
                                   :default-pool "no-pool"}
                                  pool-selection)})))
      :passport (fnk [[:config {passport {}}]]
-                 passport)
+                 (merge passport
+                        {:instance-uuid->job-uuid-cache-expiry-time-hours 24}))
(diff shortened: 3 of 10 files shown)


scheduler/src/cook/caches.clj
```diff
@@ -40,3 +40,8 @@
 (mount/defstate ^Cache pool-name->accepts-submissions?-cache :start (new-cache config/config))
 (mount/defstate ^Cache pool-name->db-id-cache :start (new-cache config/config))
 (mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
+(mount/defstate ^Cache instance-uuid->job-uuid :start
+  (-> (CacheBuilder/newBuilder)
+    (.maximumSize (get-in config/config [:settings :cache-working-set-size])) ;1million
```
dposada, line 45:
```suggestion
    (.maximumSize (get-in config/config [:settings :cache-working-set-size]))
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -82,11 +82,14 @@
...
-  "Return a map with :job-uuid or :instance-uuid based on pod-name"
+(defn assoc-uuids
+  "Returns event-map with added fields :job-uuid (based on pod-name) and :instance-uuid (if pod is not synthetic)"
   [event-map pod-name]
   (let [instance-uuid (pod-name->instance-uuid pod-name)
-        job-uuid (pod-name->job-uuid pod-name)]
+        synthetic-pod-job-uuid (synthetic-pod-name->job-uuid pod-name)
+        job-uuid (if synthetic-pod-job-uuid
+                   synthetic-pod-job-uuid
+                   (tools/instance-uuid->job-uuid-cache-lookup instance-uuid))]
```
dposada, line 92: Can be simplified to:

```suggestion
        job-uuid (or synthetic-pod-job-uuid
                   (tools/instance-uuid->job-uuid-cache-lookup instance-uuid))]
```

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -290,6 +293,8 @@
         ^V1Pod pod (api/task-metadata->pod pod-namespace compute-cluster task-metadata)
         new-cook-expected-state-dict {:cook-expected-state :cook-expected-state/starting
                                       :launch-pod {:pod pod}}]
+    (when (api/synthetic-pod? pod-name)
+      (cache/put-cache! caches/instance-uuid->job-uuid identity pod-name (get-in task-metadata [:task-request :job :job/uuid])))
```
dposada, line 297: I think it'd be nice to include a comment explaining why this is needed / giving some context for it.

scheduler/src/cook/rest/api.clj
```diff
@@ -1438,7 +1429,7 @@
   (let [jobs (wrap-seq (get-in ctx [:request :query-params :job]))
         instances (wrap-seq (get-in ctx [:request :query-params :instance]))
         allow-partial-results (get-in ctx [:request :query-params :partial])
-        instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
+        instance-uuid->job-uuid #(util/instance-uuid->job-uuid (d/db conn) %)
```
dposada, line 1432: Would it make sense to use the cached call here?

scheduler/src/cook/rest/api.clj
```diff
@@ -1490,7 +1481,7 @@
   cluster, and get back the data for those that do match."
   (let [uuids (wrap-seq (::instances ctx))
         allow-partial-results? (::allow-partial-results? ctx)
-        instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
+        instance-uuid->job-uuid #(util/instance-uuid->job-uuid (d/db conn) %)
```
dposada, line 1484: Would it make sense to use the cached call here?

scheduler/src/cook/caches.clj
calebhar12: Good catch, will fix

scheduler/src/cook/config.clj
```diff
@@ -512,7 +512,8 @@
                                   :default-pool "no-pool"}
                                  pool-selection)})))
      :passport (fnk [[:config {passport {}}]]
-                 passport)
+                 (merge passport
+                        {:instance-uuid->job-uuid-cache-expiry-time-hours 24}))
```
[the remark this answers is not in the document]
calebhar12: Will do.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -70,7 +70,7 @@
   [pod-name]
   (str/starts-with? pod-name cook-synthetic-pod-name-prefix))
 
-(defn pod-name->job-uuid
+(defn synthetic-pod-name->job-uuid
```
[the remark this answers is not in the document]
calebhar12: Yeah I see the implication, I'll change it back.

scheduler/src/cook/kubernetes/api.clj
calebhar12: Yeah I'll definitely use this. I thought Clojure would have a sleeker way to do it than my solution but I didn't know `or` worked like this, very interesting.

scheduler/src/cook/kubernetes/compute_cluster.clj
calebhar12: Do you mean specifically the `(when (api/synthetic-pod? pod-name)` part? Or do you just mean to provide context on why the cache exists.

scheduler/src/cook/rest/api.clj
calebhar12: In short, yes it can be.

My only concern is that currently we cache the `instance-uuid-->job-uuid` information right before the task is pumped into the k8s state machine, which I think is after that information is added to datomic (I'm not 100% sure). So if this `rest.api` function is called in that interim period before adding `instance-uuid` to the cache but after adding it to the DB, our cache would miss. I'm not super familiar with how often this `retrieve-jobs` api call is generally used, but I assume that it isn't used frequently enough in that interim period to give our cache a significantly increased miss rate.

calebhar12: On the subject of this time difference-- Could you point me to where the `instance-uuid-->job-uuid` information is added to datomic? I thought it might be in https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/scheduler/scheduler.clj#L773 where `:job/_instance` is in the map but I wasn't 100% sure if that's what's happening in that function (not too familiar with how datomic+clojure works). I'm interested  to see how early it occurs vs. the time when we cache it.

scheduler/src/cook/rest/api.clj
calebhar12: (see above comment, same concept)

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: I meant context on why the cache exists (but maybe the synthetic pod check warrants a little explanation too)

scheduler/src/cook/tools.clj
```diff
@@ -1037,3 +1038,22 @@
...
+
+(defn instance-uuid->job-uuid-cache-miss
+  "Wrapper function that passes in a DB to the instance-uuid->job-uuid function. Used for cache misses by the instance-uuid->job-uuid cache"
+  [instance-uuid]
+  (instance-uuid->job-uuid (d/db datomic/conn) instance-uuid))
+
+(defn instance-uuid->job-uuid-cache-lookup
+  "Gets value from cache if it is present, else search datomic for it"
+  [instance-uuid]
+  (ccache/lookup-cache! caches/instance-uuid->job-uuid identity instance-uuid->job-uuid-cache-miss instance-uuid))
```
calebhar12, line 1059: @nsinkov We discussed moving the bottom two functions, `instance-uuid->job-uuid-cache-lookup` and `instance-uuid->job-uuid-cache-miss` to the cook.cached-queries file and leaving `instance-uuid->job-uuid` in cook.tools. However, the body of `instance-uuid->job-uuid-cache-miss` calls `instance-uuid->job-uuid`, so making this change resulted in a circular dependency.

Let me know if I misunderstood what we discussed, or if there's another location that you want to move these functions to.

dposada (approved):

scheduler/src/cook/rest/api.clj
```diff
@@ -1438,7 +1429,7 @@
   (let [jobs (wrap-seq (get-in ctx [:request :query-params :job]))
         instances (wrap-seq (get-in ctx [:request :query-params :instance]))
         allow-partial-results (get-in ctx [:request :query-params :partial])
-        instance-uuid->job-uuid #(instance-uuid->job-uuid (d/db conn) %)
+        instance-uuid->job-uuid #(cached-queries/instance-uuid->job-uuid (d/db conn) %)
```
dposada, line 1432: Doesn't this namespace kind of imply that you're using a query that has caching (while this particular function doesn't actually have any caching)?

scheduler/src/cook/tools.clj
```diff
@@ -28,6 +28,7 @@
             [cook.cached-queries :as cached-queries]
             [cook.caches :as caches]
             [cook.config :as config]
+            [cook.datomic :as datomic]
```
dposada, line 31: Why was this needed?

calebhar12: Sorry that's my fault, I added the requirement for the cache functions but forgot to take it out after removing them. I'll remove it in the next revision.

scheduler/src/cook/rest/api.clj
calebhar12: Yeah, I see what you mean. Do you have ideas for what would be a better namespace for the `instance-uuid->job-uuid` function? Keeping it in `cook.rest.api` or `cook.tools` (which we originally wanted to do) resulted in circular dependency issues with `cached-queries`. And since we really want to keep `instance-uuid->job-uuid-cache-lookup` in `cached-queries`, we decided that `cached-queries` was the best namespace for `instance-uuid->job-uuid` as well since it resolves the dependency issues and makes clear the relationship between `instance-uuid->job-uuid-cache-lookup` and `instance-uuid->job-uuid`.

dposada: I'm not seeing a good namespace that already exists.

Maybe this could be solved with function naming? Like maybe the cached one should be named `instance-uuid->job-uuid` and the one with no caching should be named `instance-uuid->job-uuid-no-caching` (or something similar)?

calebhar12: That makes sense-- we could also maybe do something like `instance-uuid->job-uuid-datomic-query` for the `no-caching` call. The calls would then be:

`cached-queries/instance-uuid->job-uuid` when using the cache
`cached-queries/instance-uuid->job-uuid-no-caching` or `cached-queries/instance-uuid->job-uuid-datomic-query` when using datomic.

That makes what both calls do be pretty clear. Let me know if you have a preference, and I'll change it to that in the next revision.

dposada: Sounds great, I like the `-datomic-query` idea

scheduler/src/cook/cached_queries.clj
```diff
@@ -16,3 +19,20 @@
...
+   Returns nil if the instance uuid doesn't correspond
+   to a job"
+  [db instance-uuid]
+  (->> (d/entity db [:instance/task-id (str instance-uuid)])
+       :job/_instance
+       :job/uuid))
+
+(let [miss-fn
+      (fn [instance-uuid]
+        (instance-uuid->job-uuid-datomic-query (d/db datomic/conn) instance-uuid))]
```
dposada, line 34: Nitpick (feel free to ignore):

I think you could instead use [`partial`](https://clojuredocs.org/clojure.core/partial):

```suggestion
      (partial instance-uuid->job-uuid-datomic-query (d/db datomic/conn))]
```

scheduler/src/cook/cached_queries.clj
```diff
@@ -16,3 +19,20 @@
...
+  (defn instance-uuid->job-uuid-cache-lookup
```
dposada, line 35: Still planning on removing the `-cache-lookup` from this one?

scheduler/src/cook/cached_queries.clj
calebhar12: I think I'll stick with defining the function as-is for now because it's a bit more readable to me. We would also only ever call the partial function with exactly 1 additional argument so I don't think it's necessary to use `partial`.

(With that said I'm still new to Clojure, so if it's a highly recommended convention to use partial just let me know.)

scheduler/src/cook/cached_queries.clj
calebhar12: I decided not to because I want it to be clear that this particular function getting `job-uuid` from `instance-uuid` is looking in a cache as opposed to looking straight t datomic. And for that, I thought it made more sense to have:

`instance-uuid->job-uuid-datomic-query` & `instance-uuid->job-uuid-cache-lookup`

vs.

`instance-uuid->job-uuid-datomic-query` & `instance-uuid->job-uuid`

Imo, the former makes it more clear what each function does. Thoughts?

scheduler/src/cook/cached_queries.clj
dposada: Sounds good, I think either way is fine

scheduler/src/cook/cached_queries.clj
dposada: I don't feel too strongly one way or the other. 

I definitely see your point, but it's not consistent with the other two existing functions in `cached-queries` (`job->pool-name` and `job-ent->user`). 

I'm fine with either.

dposada (approved):

nsinkov (approved):

PR #1900, opened 2021-07-09
don't override existing job groups

nsinkov:

- don't override existing job groups

If you add a job to a group currently you will override the existing group with an implicit group

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -1994,10 +1994,10 @@
                                distinct
                                (remove #(contains? group-uuids %))
                                (map make-default-group))
-          groups (into (vec implicit-groups) groups)
+          db (d/db conn)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -1994,10 +1994,10 @@
                                distinct
                                (remove #(contains? group-uuids %))
                                (map make-default-group))
-          groups (into (vec implicit-groups) groups)
+          db (d/db conn)
+          groups (filter (fn [group] (not (group-exists? db (:uuid group)))) (into (vec implicit-groups) groups))
```
dposada, line 1998: Maybe a comment here explaining the "why?" of this would be helpful?

nsinkov: added

dposada (approved):

PR #1901, opened 2021-07-09
Adds pool source to job submission passport stamp

dposada:

- adding a new passport event type, `:api-job-creation`
- changing the existing stamp in `create-jobs!` to use `:api-job-creation`
- stamping with `:api-job-submission` from the `malformed?` check (earlier in the request flow)
- adding how the pool was sourced (request parameter vs. `x-cook-pool` header) to `:api-job-submission` 

We want to be able to distinguish between jobs that landed in a pool via the request parameter and those that landed in a pool via the `x-cook-pool` request header. This change also allows for a passport event to be logged earlier in the request flow, before all of the validation checks have run.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -1980,7 +1980,7 @@
                      (update job-pool-name-map :job #(dissoc % :command)))
                    job-pool-name-maps))
     (doseq [{{:keys [uuid, user]} :job} job-pool-name-maps]
-      (passport/log-event {:event-type passport/api-job-submission
+      (passport/log-event {:event-type passport/api-job-creation
                            :job-uuid (str uuid)
                            :user user}))
     (let [jobs (map :job job-pool-name-maps)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -2178,11 +2178,26 @@
...
                          in-debt? (not (zero? time-until-out-of-debt))]
                      (try
+                       (doseq [{:keys [uuid]} jobs]
+                         (passport/log-event
+                           {:event-type passport/api-job-submission
+                            :job-uuid (str uuid)
+                            :pool pool-name
+                            :pool-header pool-header
+                            :pool-param pool-param
+                            :pool-source
```
calebhar12, line 2195: If a user doesn't ask for a specific pool would these fields be `nil`? Or is there some default `:pool` value for when a user allows Cook to schedule their job in any pool?

dposada: If the user doesn't ask for a specific pool, then `:pool`, `:pool-header`, and `:pool-param` would all be `nil`, and `:pool-source` would be `:unspecified`.

calebhar12: Gotcha, that makes sense. Looks good to me then, giving approval now.

calebhar12 (approved):

PR #1907, opened 2021-07-20
Update passport event types with cook-scheduler source and namespace

calebhar12:

- Update names for passport event types
- Add cook-scheduler namespace on event-type strings
- Add `source` field with value `cook-scheduler`

These changes keep our `event-type` strings up to date with our design document and provide us with `source` information in every log.

### DIFF
--- a/scheduler/src/cook/passport.clj
+++ b/scheduler/src/cook/passport.clj
@@ -5,13 +5,13 @@
 
 (defn log-event
   "Log event to cook-passport log file"
-  [log-data]
+  [{:keys [event-type] :as log-data}]
   (when (:enabled? (config/passport))
     (log/log config/passport-logger-ns :info nil (json/write-str
                                                    (assoc
                                                      log-data
                                                      :source :cook-scheduler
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/passport.clj
```diff
@@ -7,9 +7,13 @@
   "Log event to cook-passport log file"
   [log-data]
   (when (:enabled? (config/passport))
-    (log/log config/passport-logger-ns :info nil (json/write-str log-data))))
+    (log/log config/passport-logger-ns :info nil (json/write-str
+                                                   (assoc
+                                                     log-data
+                                                     :source :cook-scheduler
+                                                     :event-type (str "cook-scheduler/" (name (log-data :event-type))))))))
```
dposada, line 14: Nitpick: you can destructure `:event-type` from the function argument

dposada (approved):

scheduler/src/cook/passport.clj
calebhar12: I like that, will add.

PR #1910, opened 2021-07-21
Fix Location header of redirects to include request parameters

scrosby:
**This is an API change.**

- Fixes two bugs that mostly compensated for each other. 
- When we generate Location: headers for redirect-to-leader, we should include any query parameters in the response.
- The follow redirects code in the integration tests would add those parameters (including duplicating them)

This is correct behavior.

### DIFF
--- a/scheduler/src/cook/rest/api.clj
+++ b/scheduler/src/cook/rest/api.clj
@@ -1896,8 +1896,10 @@
      [ctx]
      ;; the client is expected to cache the redirect location
      (if-let [leader-url (::leader-url ctx)]
-       (let [request-path (get-in ctx [:request :uri])]
-         [true {:location (str leader-url request-path)}])
+       (let [request-path (get-in ctx [:request :uri])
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/rest/api.clj
```diff
@@ -1896,8 +1896,10 @@
      [ctx]
      ;; the client is expected to cache the redirect location
      (if-let [leader-url (::leader-url ctx)]
-       (let [request-path (get-in ctx [:request :uri])]
-         [true {:location (str leader-url request-path)}])
+       (let [request-path (get-in ctx [:request :uri])
+             request-param (get-in ctx [:request :query-string])]
```
dposada, line 1900: Maybe `query-string` instead of `request-param`?

scrosby: Changed.

dposada (approved):

dposada:
Will merge when 🟢

PR #1911, opened 2021-07-21
Adjust test_user_pool_rate_limit to make it more reliable.

scrosby:

-  Make test_user_pool_rate_limit more reliable and more clear what its doing.

- Make the test more deterministic.

### DIFF
--- a/integration/tests/cook/test_multi_user.py
+++ b/integration/tests/cook/test_multi_user.py
@@ -114,6 +114,8 @@ class MultiUserCookTest(util.CookTest):
                 with self.user_factory.admin():
                     util.kill_jobs(self.cook_url, all_job_uuids, assert_response=False)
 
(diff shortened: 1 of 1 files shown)


integration/tests/cook/test_multi_user.py
```diff
@@ -114,6 +114,8 @@ def test_multi_user_usage(self):
                 with self.user_factory.admin():
                     util.kill_jobs(self.cook_url, all_job_uuids, assert_response=False)
 
+    # Serial so that we don't conflict with another running test that may use the same user ID.
+    @pytest.mark.serial
```
dposada, line 118: This is already guaranteed by the `UserFactory`

scrosby: Ok. Removed.

dposada (approved):

dposada:
Will merge when 🟢

PR #1914, opened 2021-07-26
[DRAFT] Add pool, user, and job-name to passport events

calebhar12:

- Change `instance-uuid->job-uuid` cache to map `pod-name->job-ent`
- Add pool, user, and job-name fields to passport events

These changes will allow us to include `pool`, `user`, and `job-name` in all passport events. This will help us better track job status and give more context to event logs.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -381,7 +381,12 @@
           (handle-watch-updates current-nodes-atom watch node->node-name
                                 callbacks) ; Update the set of all nodes.
           (catch Exception e
-            (log/warn e "Error during node watch for compute cluster" compute-cluster-name))
+            (let [cause (-> e Throwable->map :cause)]
+              (if (= cause "Socket closed")
+                ; We expect timeouts to happen on the node watch when there is
+                ; no node-related activity, so we log with INFO instead of WARN
+                (log/info e "Timeout during node watch for compute cluster" compute-cluster-name)
+                (log/warn e "Error during node watch for compute cluster" compute-cluster-name {:cause cause}))))
           (finally
             (.close watch)
             (initialize-node-watch compute-cluster)))))))


scheduler/src/cook/cached_queries.clj
```diff
@@ -20,19 +20,25 @@
   [job-ent]
   (caches/lookup-cache-datomic-entity! caches/job-ent->user-cache :job/user job-ent))
 
-(defn instance-uuid->job-uuid-datomic-query
+(defn pod-name->job-ent-datomic-query
```
dposada, line 23: Maybe this should be `instance-uuid->job-ent-datomic-query` since we're still assuming the input is a valid instance UUID?

scheduler/src/cook/cached_queries.clj
```diff
@@ -20,19 +20,25 @@
...
-       :job/_instance
-       :job/uuid))
+       :job/_instance))
+
+(defn instance-uuid->job-uuid-datomic-query
+  "Queries for the job uuid from an instance uuid.
+   Returns nil if the instance uuid doesn't correspond
+   to a job"
+  [db instance-uuid]
+  (:job/uuid (pod-name->job-ent-datomic-query db instance-uuid)))
```
dposada, line 36: Is this still needed?

scheduler/src/cook/cached_queries.clj
```diff
@@ -20,19 +20,25 @@
...
 
 (let [miss-fn
       (fn [instance-uuid]
-        (str (instance-uuid->job-uuid-datomic-query (d/db datomic/conn) instance-uuid)))]
-  (defn instance-uuid->job-uuid-cache-lookup
-    "Get value from cache if it is present, else search datomic for it"
+        (pod-name->job-ent-datomic-query (d/db datomic/conn) instance-uuid))]
+  (defn pod-name->job-ent-cache-lookup
```
dposada, line 41: Same question here as above about naming

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -82,16 +82,21 @@
...
+(defn assoc-fields
+  "Returns event-map with added fields instance-uuid (if applicable), job-name, job-uuid, pool, and user"
   [event-map pod-name]
   (let [instance-uuid (pod-name->instance-uuid pod-name)
-        job-uuid (or (pod-name->job-uuid pod-name)
-                     (cached-queries/instance-uuid->job-uuid-cache-lookup instance-uuid))]
+        job-name (str (:job/name (cached-queries/pod-name->job-ent-cache-lookup pod-name)))
+        job-uuid (str (:job/uuid (cached-queries/pod-name->job-ent-cache-lookup pod-name)))
+        pool (str (:pool/name (:job/pool (cached-queries/pod-name->job-ent-cache-lookup pod-name))))
+        user (str (:job/user (cached-queries/pod-name->job-ent-cache-lookup pod-name)))]
```
dposada, line 92: Can we extract a variable for `cached-queries/pod-name->job-ent-cache-lookup` since we call it 4 times? Then we can also potentially destructure the fields we need from the job. I think the `str` calls for job name, pool name, and job user are superfluous.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -293,9 +293,8 @@
         ^V1Pod pod (api/task-metadata->pod pod-namespace compute-cluster task-metadata)
         new-cook-expected-state-dict {:cook-expected-state :cook-expected-state/starting
                                       :launch-pod {:pod pod}}]
-    ; If a pod is not synthetic, cache a mapping of its instance-uuid to job-uuid in order to give all state machine passport events access to job-uuid
-    (when-not (api/synthetic-pod? pod-name)
-      (cache/put-cache! caches/instance-uuid->job-uuid identity pod-name (str (get-in task-metadata [:task-request :job :job/uuid]))))
+    ; Cache a mapping of pod-name to job-uuid in order to give all state machine passport events access to job attributes
+      (cache/put-cache! caches/pod-name->job-ent identity pod-name (get-in task-metadata [:task-request :job]))
```
dposada, line 297: Why did we remove the `(when-not (api/synthetic-pod? pod-name)` check? The `pod-name` for synthetic pods is not an instance UUID.

PR #1916, opened 2021-07-27
Add pool-name, job-name, and user to Passport Logs

calebhar12:

- Create a cache mapping `job-uuid` to `job-map`
- Log `pool`, `job-name`, and `user` fields in passport events
- Add `COOK_JOB_NAME` AND `COOK_JOB_USER` as pod environment variables

- These changes allow us to use those 3 fields when looking at cook metrics and to filter on them when looking at passport events.

### DIFF
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -45,3 +45,8 @@
     (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
     (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
     (.build)))
+(mount/defstate ^Cache job-uuid->job-map :start
+                (-> (CacheBuilder/newBuilder)
+                  (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
+                  (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+                  (.build)))
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -19,8 +19,8 @@
 (defn passport-cache [config]
   "Build a new passport-related cache"
   (-> (CacheBuilder/newBuilder)
-    (.maximumSize (get-in config [:settings :passport :job-uuid-cache-set-size]))
-    (.expireAfterAccess (get-in config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+    (.maximumSize (get-in config [:settings :passport :job-cache-set-size]))
+    (.expireAfterAccess (get-in config [:settings :passport :job-cache-expiry-time-hours]) TimeUnit/HOURS)
     (.build)))
 
 (defn lookup-cache-datomic-entity!
--- a/scheduler/src/cook/caches.clj
+++ b/scheduler/src/cook/caches.clj
@@ -16,6 +16,12 @@
       (.expireAfterAccess 2 TimeUnit/HOURS)
       (.build)))
 
+(defn passport-cache [config]
+  "Build a new passport-related cache"
+  (-> (CacheBuilder/newBuilder)
+    (.maximumSize (get-in config [:settings :passport :job-uuid-cache-set-size]))
+    (.expireAfterAccess (get-in config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+    (.build)))
 
 (defn lookup-cache-datomic-entity!
   "Specialized function for caching where datomic entities are the key.
@@ -40,13 +46,5 @@
 (mount/defstate ^Cache pool-name->accepts-submissions?-cache :start (new-cache config/config))
 (mount/defstate ^Cache pool-name->db-id-cache :start (new-cache config/config))
 (mount/defstate ^Cache user-and-pool-name->quota :start (new-cache config/config))
-(mount/defstate ^Cache instance-uuid->job-uuid :start
-  (-> (CacheBuilder/newBuilder)
-    (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
-    (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
-    (.build)))
-(mount/defstate ^Cache job-uuid->job-map :start
-                (-> (CacheBuilder/newBuilder)
-                  (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
-                  (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
-                  (.build)))
+(mount/defstate ^Cache instance-uuid->job-uuid :start (passport-cache config/config))
+(mount/defstate ^Cache job-uuid->job-map :start (passport-cache config/config))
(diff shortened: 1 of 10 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -82,16 +82,23 @@
...
+  "Returns event-map with added fields :instance-uuid (if pod is not synthetic), job-name, job-uuid, pool, and user."
   [event-map pod-name]
   (let [instance-uuid (pod-name->instance-uuid pod-name)
         job-uuid (or (pod-name->job-uuid pod-name)
-                     (cached-queries/instance-uuid->job-uuid-cache-lookup instance-uuid))]
+                     (cached-queries/instance-uuid->job-uuid-cache-lookup instance-uuid))
+        job-map (cached-queries/job-uuid->job-map-cache-lookup job-uuid)
+        job-name (:job/name job-map)
+        pool (:pool/name (:job/pool job-map))
+        user (:job/user job-map)]
```
dposada, line 94: I think you can destructure these from `job-map`

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -292,10 +292,14 @@
         pod-name (:task-id task-metadata)
         ^V1Pod pod (api/task-metadata->pod pod-namespace compute-cluster task-metadata)
         new-cook-expected-state-dict {:cook-expected-state :cook-expected-state/starting
-                                      :launch-pod {:pod pod}}]
+                                      :launch-pod {:pod pod}}
+        job-map (get-in task-metadata [:task-request :job])
+        job-uuid (str (:job/uuid job-map))]
```
dposada, line 297: Can we destructure `:job/uuid`?

scheduler/src/cook/caches.clj
```diff
@@ -44,4 +44,9 @@
...
     (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
     (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
-    (.build)))
\ No newline at end of file
+    (.build)))
+(mount/defstate ^Cache job-uuid->job-map :start
+                (-> (CacheBuilder/newBuilder)
+                  (.maximumSize (get-in config/config [:settings :passport :job-uuid-cache-set-size]))
+                  (.expireAfterAccess (get-in config/config [:settings :passport :job-uuid-cache-expiry-time-hours]) TimeUnit/HOURS)
+                  (.build)))
```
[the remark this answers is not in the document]
calebhar12: Yeah-- I sort of took a half-measure here and wanted to discuss it more. The reason I wrote the cache building like this for now is that I wasn't sure if we would want in the future for the `instance-uuid->job-uuid` and `job-uuid->job-map` caches to have different expiry times and sizes. So, I built each one separately but using the same parameters, so that if we did want different expiry and sizes we could easily sub them in.

However, after thinking about it I think that we will always want the two caches to have the same expiry time and size b/c they are currently always used in tandem. If that line of thinking makes sense (I might just be overthinking this) I'll make a helper function to build them.

scheduler/src/cook/kubernetes/api.clj
calebhar12: gotcha-- Does this look good: 
`{job-name :job/name user :job/user {pool-name :pool/name} :job/pool} job-map`

I also tried:
```
{:keys [job/name job/pool job/user]} jobmap
{:keys [pool/name]} pool
```
But then `name` is equal to `pool/name` and not `job/name`. And when I try to do something like `(prn job/name)`, I get a "No such namespace" exception.

If there's a better Clojure way to do this please let me know :).

scheduler/src/cook/kubernetes/compute_cluster.clj
calebhar12: So it would be created in the same line as `job-map`? I wasn't quite sure how to do that in Clojure so I just kept it like this. If there's a good way to do it I'd be interested to learn.

scheduler/src/cook/caches.clj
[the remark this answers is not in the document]
dposada: I think that line of thinking makes sense 👍🏼

scheduler/src/cook/kubernetes/compute_cluster.clj
dposada: Yeah, like I think instead of this:

```
job-map (get-in task-metadata [:task-request :job])
job-uuid (str (:job/uuid job-map))]
```

you could have this?:

```
{:keys [job/uuid] :as job-map} (get-in task-metadata [:task-request :job])
job-uuid (str uuid)]
```

scheduler/src/cook/kubernetes/api.clj
dposada: Your first proposal above looks good, except you can do it on the line where `job-map` is currently defined, and then you might not even need `job-map`, e.g.:

```
{job-name :job/name user :job/user {pool-name :pool/name} :job/pool} (cached-queries/job-uuid->job-map-cache-lookup job-uuid)
```

scheduler/src/cook/caches.clj
```diff
@@ -16,6 +16,12 @@
       (.expireAfterAccess 2 TimeUnit/HOURS)
       (.build)))
 
+(defn passport-cache [config]
+  "Build a new passport-related cache"
+  (-> (CacheBuilder/newBuilder)
+    (.maximumSize (get-in config [:settings :passport :job-uuid-cache-set-size]))
```
nsinkov, line 22: nitpick: rename job-uuid-cache-set-size to job-cache-set-size and job-uuid-cache-expiry-time-hours to job-cache-expiry-time-hours

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -82,16 +82,20 @@
   (when-not (synthetic-pod? pod-name)
     pod-name))
 
-(defn assoc-uuids
-  "Returns event-map with added fields :job-uuid (based on pod-name) and :instance-uuid (if pod is not synthetic)"
+(defn assoc-fields
```
nsinkov, line 85: I think we should rename to `pod-name->job-map` have it return a map and then call merge on the map where we use it

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1670,7 +1678,7 @@
                 (str "Pod name from pod (" pod-name-from-pod ") "
                      "does not match pod name argument (" pod-name ")"))
         (log/info "In" compute-cluster-name "compute cluster, launching pod with name" pod-name "in namespace" namespace ":" (.serialize json pod))
-        (let [event-map (assoc-uuids
+        (let [event-map (assoc-fields
```
nsinkov, line 1681: I'd probably remove the let

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -158,7 +158,7 @@
   [{:keys [name]}
    {:keys [task-id state reason]}]
   (let [pod-name (task-id :value)
```
nsinkov, line 160: I'd probably remove the let

scheduler/src/cook/kubernetes/api.clj
dposada: Works for me

dposada (approved):

scheduler/src/cook/caches.clj
calebhar12: Will change

scheduler/src/cook/kubernetes/api.clj
calebhar12: SGTM

scheduler/src/cook/kubernetes/api.clj
calebhar12: will do

scheduler/src/cook/kubernetes/controller.clj
calebhar12: will do

nsinkov (approved):

PR #1917, opened 2021-07-28
Make flag for enabling shm support on k8s.

scrosby:

- Make flag for enabling shm support on k8s.

- Some cases of python multiprocessing need SHM support to work.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -648,6 +648,7 @@
   []
   (let [^V1EmptyDirVolumeSource volume-source (V1EmptyDirVolumeSource.)
         _ (.medium volume-source "Memory")
+        ; Randomize the name to reduce the changes of name collision with any other volume name.
         ^String name (str "shmvolume-" (+ (rand-int 899999) 100000))
         ^V1VolumeBuilder builder (-> ^V1VolumeBuilder (V1VolumeBuilder.)
                                      (.withName name)
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -677,7 +677,7 @@
   "Conj the item onto the accum only if not nil"
   [accum item]
   (cond-> accum
-    (not (nil? item)) (conj item)))
+    (some? item) (conj item)))
 
 
 (defn make-volumes
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -643,6 +643,24 @@
                                      (.withEmptyDir (V1EmptyDirVolumeSource.)))]
     (.build builder)))
 
+(defn make-shm-volume
+  "Make a volume for /dev/shm and add it to the volume mount for the given image"
+  []
+  (let [^V1EmptyDirVolumeSource volume-source (V1EmptyDirVolumeSource.)
+        _ (.medium volume-source "Memory")
+        ^String name (str "shmvolume-" (+ (rand-int 899999) 100000))
+        ^V1VolumeBuilder builder (-> ^V1VolumeBuilder (V1VolumeBuilder.)
+                                     (.withName name)
+                                     (.withEmptyDir volume-source))]
+    (.build builder)))
+
+(defn- make-shm-volume-mount
+  "Make a volume mount for a shm volume mount."
+  ([^V1Volume volume]
+     (doto (V1VolumeMount.)
+       (.setName (.getName volume))
+       (.setMountPath "/dev/shm"))))
+
 (defn- make-volume-mount
   "Make a kubernetes volume mount"
   ([^V1Volume volume path read-only]
@@ -655,9 +673,16 @@
        (.setReadOnly read-only)
        (cond-> sub-path (.setSubPath sub-path))))))
 
+(defn maybe-conj
+  "Conj the item onto the accum only if not nil"
+  [accum item]
+  (cond-> accum
+    (not (nil? item)) (conj item)))
+
+
 (defn make-volumes
   "Converts a list of cook volumes to kubernetes volumes and volume mounts."
-  [cook-volumes sandbox-dir]
+  [cook-volumes sandbox-dir job-labels]
   (let [{:keys [disallowed-container-paths]} (config/kubernetes)
         allowed-cook-volumes (remove (fn [{:keys [container-path host-path mode]}]
                                        (contains? disallowed-container-paths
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -643,6 +643,24 @@
                                      (.withEmptyDir (V1EmptyDirVolumeSource.)))]
     (.build builder)))
 
+(defn make-shm-volume
+  "Make a volume for /dev/shm and add it to the volume mount for the given image"
+  []
+  (let [^V1EmptyDirVolumeSource volume-source (V1EmptyDirVolumeSource.)
+        _ (.medium volume-source "Memory")
+        ^String name (str "shmvolume-" (+ (rand-int 899999) 100000))
```
dposada, line 651: Why do we need the randomness here?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -655,9 +673,16 @@
        (.setReadOnly read-only)
        (cond-> sub-path (.setSubPath sub-path))))))
 
+(defn maybe-conj
+  "Conj the item onto the accum only if not nil"
+  [accum item]
+  (cond-> accum
+    (not (nil? item)) (conj item)))
```
dposada, line 680:
```suggestion
    (some? item) (conj item)))
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -643,6 +643,24 @@
                                      (.withEmptyDir (V1EmptyDirVolumeSource.)))]
     (.build builder)))
 
+(defn make-shm-volume
+  "Make a volume for /dev/shm and add it to the volume mount for the given image"
+  []
+  (let [^V1EmptyDirVolumeSource volume-source (V1EmptyDirVolumeSource.)
+        _ (.medium volume-source "Memory")
```
dposada, line 650: Do we need to account for this in the memory usage of the pod?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -682,10 +707,14 @@
         sandbox-volume-mount-fn #(make-volume-mount sandbox-volume sandbox-dir %)
         sandbox-volume-mount (sandbox-volume-mount-fn false)
         ; mesos-sandbox-volume-mount added for Mesos backward compatibility
-        mesos-sandbox-volume-mount (make-volume-mount sandbox-volume "/mnt/mesos/sandbox" false)]
+        mesos-sandbox-volume-mount (make-volume-mount sandbox-volume "/mnt/mesos/sandbox" false)
+        label-prefix (:add-job-label-to-pod-prefix (config/kubernetes))
+        make-shm-volume? (= "true" (get job-labels (str label-prefix "shared-memory")))
```
dposada, line 712: Can we have an integration test for this?

scheduler/src/cook/kubernetes/api.clj
scrosby: The idea is to avoid accidental name collision if someone else makes a volume with a similar name.

scheduler/src/cook/kubernetes/api.clj
scrosby: AFAIK, no you don't specify a size here. I checked and it appears that the kernel/system limit is all RAM. (I also seem to recall from discussions that this gets accounted for in cgroups. which makes sense, this shm isn't shared among pods.)

scheduler/src/cook/kubernetes/api.clj
scrosby: I don't have an opensource integration test I can run. How much do you want one?

scheduler/src/cook/kubernetes/api.clj
scrosby: ACK.

scheduler/src/cook/kubernetes/api.clj
dposada: Ah, got it. Maybe a comment explaining this would be helpful?

scheduler/src/cook/kubernetes/api.clj
dposada: So does this use the same pool of memory as what you ask for on the container?

scheduler/src/cook/kubernetes/api.clj
dposada: Maybe a simple test that submits a job with the label, and then writes to and reads from `/dev/shm`? And then a job that attempts to write to `/dev/shm` but without the label and fails?

scheduler/src/cook/kubernetes/api.clj
scrosby: https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt suggests yes as the creating cgroup container would own it and get it accounted.

scheduler/src/cook/kubernetes/api.clj
scrosby: Done.

scheduler/src/cook/kubernetes/api.clj
scrosby: Doing it that way is is not so simple of a test --- We need to create a container and send it to GKE/K8s that contains an image that has a script that fails. I'd like to not open up the quagmire of image management.

I've tried a half dozen replicators, without success. The good news is that we can try various one-liners without needing to come up with an image --- and I have tried your idea, it failed.

I tried:

`cs submit -m 20 -c .001 --pool k8s-gamma "python3.8 -c 'import multiprocessing.shared_memory; multiprocessing.shared_memory.SharedMemory(None,True,1000000).unlink()'"`

`cs submit -m 20 -c .001 --pool k8s-gamma 'ls /dev/shm/'`
or even
`cs submit -m 20 -c .001 --pool k8s-gamma 'echo hello > /dev/shm/foo && rm -f /dev/shm/foo'`

But these jobs pass even with the label not set. I tried this, which (properly) fails, but this wouldn't succeed with the label set:

`echo hello > /dev/shmx/foo && rm -f /dev/shm/foo`

Whatever it is, these tests aren't exercising the pytorch codepaths that inspired the change. (Which I have validated are fixed by this change.)

scheduler/src/cook/kubernetes/api.clj
dposada: Would it be possible to write a test where:

1. job requests 1024 MiB with the shared mem label
2. command writes 768 MiB to `/dev/shm/foo`
3. command then tries to allocate 512 MiB of memory
4. assert that the job gets OOM

scheduler/src/cook/kubernetes/api.clj
dposada: Got it, thanks for trying. Made a suggestion for a different test in the other comment thread; resolving this comment thread.

scheduler/src/cook/kubernetes/api.clj
scrosby: /dev is almost assuregly a tmpfs (in modern linux with udev), so that OOM would occur whether it was shm or tmpfs, so I would expect this to fail with or without the label.

dposada: Ah, got it.

PR #1937, opened 2021-08-30
Update Cook to run on JDK-11

scrosby:

- Update Cook to support JDK-11

- JDK8 is EOL

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -77,12 +77,12 @@
 
                  ; Bring in netty. Note that this is brought in in a lot of other places, Exclude it whenever finding another one.
                  ; Need to make sure all imports are importing the same version.
-                 [io.netty/netty-handler "4.1.63.Final"]
+                 ;[io.netty/netty-handler "4.1.63.Final"]
                  ; Netty now uses epoll with a native library. We need to tell lein our architecture so it can bring in the
                  ; right library architecture. Make sure this version smatches.
-                 [io.netty/netty-transport-native-epoll "4.1.63.Final" :classifier "linux-x86_64"]
-                 [io.netty/netty-transport-native-unix-common "4.1.63.Final" :classifier "linux-x86_64"]
-
+                 ;[io.netty/netty-transport-native-epoll "4.1.63.Final" :classifier "linux-x86_64"]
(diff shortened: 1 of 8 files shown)


scheduler/project.clj
```diff
@@ -77,6 +75,15 @@
...
                  [org.eclipse.jetty/jetty-security "9.2.6.v20141205"]
 
+                 ; TODO: Upgrade netty.
+                 ; Bring in netty. Note that this is brought in in a lot of other places, Exclude it whenever finding another one.
+                 ; Need to make sure all imports are importing the same version.
+                 ;[io.netty/netty-handler "4.1.63.Final"]
+                 ; Netty now uses epoll with a native library. We need to tell lein our architecture so it can bring in the
+                 ; right library architecture. Make sure this version matches.
+                 ;[io.netty/netty-transport-native-epoll "4.1.63.Final" :classifier "linux-x86_64"]
+                 ;[io.netty/netty-transport-native-unix-common "4.1.63.Final" :classifier "linux-x86_64"]
```
dposada, line 85: Do we really need to keep all of these comments around?

scheduler/src/cook/test/testutil.clj
```diff
@@ -97,51 +97,53 @@
...
+          (assoc :file "/dev/stdout"))
+   :mesos {:leader-path "", :master ""}
+   :metrics {}
+   :nrepl {}
+   :port 80
+   :scheduler {}
+   :kubernetes {:synthetic-pod-recency-seconds 1 :max-jobs-for-autoscaling 1000 :autoscaling-scale-factor 1000.0}
+   :unhandled-exceptions {}
+   :zookeeper {:local? true}})
+(defn setup
```
dposada, line 121:
```suggestion
   :zookeeper {:local? true}})
   
(defn setup
```

scheduler/project.clj
scrosby: No, but this is a helpful starting point for an io.netty upgrade. In particular, it explains a gotcha, and and solves a few issues. This is a battle I tactically retreated from for this PR, but expect to need to continue.

dposada (approved):

PR #1943, opened 2021-09-09
add incremental image config support for aux containers

nsinkov:

- add incremental image config support for aux containers

We want the auxiliary containers for the job pod to support incremental image configurations. This will allow us to incrementally test new functionality.

### DIFF
--- a/scheduler/src/cook/config_incremental.clj
+++ b/scheduler/src/cook/config_incremental.clj
@@ -100,7 +100,14 @@
 
 (defn resolve-incremental-config
   "Resolve an incremental config to the appropriate value.
-  With an overload that takes a fallback config to use in case the incremental config cannot be resolved."
+  With an overload that takes a fallback config to use in case the incremental config cannot be resolved.
+  If the incremental config cannot be resolved and there is no fallback then nil is returned and it is up to the caller
+  to handle appropriately.
+
+  An incremental configuration can either be a collection of incremental values or a keyword.
+  If it is a keyword, then it is used as a key to look up a collection of incremental values in the database.
+  An collection of incremental values is resolved by picking one of the values using the job uuid hash. Each incremental
+  value has an associated portion and the portions must add up to 1.0"
   ([^UUID uuid incremental-config]
    (cond
      (coll? incremental-config)
--- a/scheduler/src/cook/config_incremental.clj
+++ b/scheduler/src/cook/config_incremental.clj
@@ -99,10 +99,18 @@
   (select-config-from-values uuid (read-config config-key)))
 
 (defn resolve-incremental-config
-  "Resolve an incremental config to the appropriate value."
-  [^UUID uuid incremental-config]
-  (cond
-    (coll? incremental-config)
-    (select-config-from-values uuid incremental-config)
-    (keyword? incremental-config)
-    (select-config-from-key uuid incremental-config)))
+  "Resolve an incremental config to the appropriate value.
+  With an overload that takes a fallback config to use in case the incremental config cannot be resolved."
+  ([^UUID uuid incremental-config]
+   (cond
+     (coll? incremental-config)
+     (select-config-from-values uuid incremental-config)
+     (keyword? incremental-config)
+     (select-config-from-key uuid incremental-config)))
+  ([^UUID uuid incremental-config fallback-config]
+   (let [resolved-incremental-config (resolve-incremental-config uuid incremental-config)]
+     (if resolved-incremental-config
+       [resolved-incremental-config :resolved-incremental-config]
+       ; use a fallback config in case there is a problem resolving an incremental config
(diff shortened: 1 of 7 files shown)


scheduler/src/cook/config_incremental.clj
```diff
@@ -99,10 +99,18 @@
...
-  "Resolve an incremental config to the appropriate value."
-  [^UUID uuid incremental-config]
-  (cond
-    (coll? incremental-config)
-    (select-config-from-values uuid incremental-config)
-    (keyword? incremental-config)
-    (select-config-from-key uuid incremental-config)))
\ No newline at end of file
+  "Resolve an incremental config to the appropriate value.
+  With an overload that takes a fallback config to use in case the incremental config cannot be resolved."
```
scrosby, line 103: Should there be a warning if it was unable to be resolved? (If this is done elsewhere, probably worth a comment?)

scheduler/src/cook/config_incremental.clj
```diff
@@ -99,10 +99,18 @@
...
+  ([^UUID uuid incremental-config]
+   (cond
+     (coll? incremental-config)
+     (select-config-from-values uuid incremental-config)
+     (keyword? incremental-config)
+     (select-config-from-key uuid incremental-config)))
```
scrosby, line 109: And if none of these match, should there be a useful ERROR logged and exception thrown?

scheduler/src/cook/config_incremental.clj
```diff
@@ -99,10 +99,18 @@
...
 (defn resolve-incremental-config
-  "Resolve an incremental config to the appropriate value."
-  [^UUID uuid incremental-config]
-  (cond
-    (coll? incremental-config)
-    (select-config-from-values uuid incremental-config)
-    (keyword? incremental-config)
-    (select-config-from-key uuid incremental-config)))
\ No newline at end of file
+  "Resolve an incremental config to the appropriate value.
```
scrosby, line 102: Could you comment on the shape of an incremental config? What does the data shape look like?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1342,8 +1363,13 @@
                                 :init-container-checkpoint-volume-mounts init-container-checkpoint-volume-mounts
                                 :init-container-workdir init-container-workdir
                                 :init-container-workdir-volume-mount-fn init-container-workdir-volume-mount-fn
+                                :job job
```
scrosby, line 1366: This abstraction (stuffing a lot of local variables into the container-attributes dictionary) seems undesirarable. 

* There's an unnecessary risk of typo's; mistype a key here or in the consumer and the output is nil. Accessing the variables under their original name would catch those kinds of typo errors.
* There's redundancy here; the job and job-name are put in here twice, pool-name is put here twice. the user is put here twice.)
* Its harder to follow which consumer is using which variables.

If we wan to split it up into subfunctions, why not (let [subfn (fn add-container-to-pod ....)] and let it grab the context (user, pool, etc) from the environment?

scheduler/src/cook/rest/api.clj
```diff
@@ -593,11 +593,7 @@
         image-config (:image docker)
         image (if (string? image-config)
                 image-config
-                (let [incremental-image-config (config-incremental/resolve-incremental-config uuid image-config)
-                      [resolved-config reason] (if incremental-image-config
-                                                 [incremental-image-config :resolved-incremental-config]
-                                                 ; use a fallback image in case there is a problem resolving an incremental config
-                                                 [(:image-fallback docker) :used-image-fallback])]
+                (let [[resolved-config reason] (config-incremental/resolve-incremental-config uuid image-config (:image-fallback docker))]
```
scrosby, line 596: Q: Should we resolve this at submit time or launch time? To me, it seems like we should resolve this at launch time --- otherwise, there's a lag. if we have a bad config, it bake it into a bunch of jobs that can't be fixed even if we fix the config. Versus making a config change and then all jobs submitted after the config change are good.  Thoughts?

nsinkov: Resolving at launch time seems better, but the default image is rarely changed and errors can be mitigated with incremental rollout anyway. So it's not clear if it's worth the effort to refactor. (In either case it would not be part of this PR)

scheduler/src/cook/kubernetes/api.clj
nsinkov: I agree but want to make a separate PR

scrosby: I don't think it would take much to do that. Just copy&paste the function body and look for undefined vars. I'm not a super fan of perpetuating this further......

scheduler/src/cook/rest/api.clj
scrosby: Fair.

PR #1948, opened 2021-09-21
Trims down pod metadata logging

dposada:

Instead of logging the entire pod metadata object, hand-picking the fields from it that we want to log.

To reduce excessive Cook logging.

### DIFF
--- a/scheduler/src/cook/kubernetes/controller.clj
+++ b/scheduler/src/cook/kubernetes/controller.clj
@@ -100,6 +100,49 @@
       (assoc stripped-cook-expected-state-dict :launch-pod [:elided-for-brevity])
       stripped-cook-expected-state-dict)))
 
+(defn prepare-pod-metadata-for-logging
+  "Generates a pruned version of pod metadata that is suitable for logging.
+
+  Logging the full metadata object is too much, in particular the
+  managedFields. This function generates a pruned version of metadata
+  that's not as verbose. This cuts logging output by about 2x compared to a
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -108,7 +108,17 @@
         (dissoc :pod)
         (assoc 
           :node-name (api/pod->node-name pod)
-          :pod-metadata (some-> ^V1Pod pod .getMetadata)
+          :pod-metadata
+          (when-let [^V1ObjectMeta metadata (some-> ^V1Pod pod .getMetadata)]
```
scrosby, line 112: Reviewing all of the methods in V1ObjecMetadata, we're missing a few here:
* getClusterName
* getDeletionGracePeriodSeconds
* getGenerateName
* getGeneration
* getOwnerReference

For each of the above, can you either add them to the map, include them in a comment where we list the fields we are NOT logging with an explanation for each field as to of why (don't forget to put getManagedFields in that list)

This overall block should also have a comment as to why we're logging metadata differently, maybe something like "Logging the full metadata object is very bloated, in particular with the use of managedFields. This generates a pruned version of metadata that's not as verbose. In particular, this cuts logging output by about 2x"

Also a comment: "This field list is accurate as of 202109, with io.kubernetes/client 11.0.0" as a breadcrumb that we may notice in a few years that we should review for updates.

Final request: Can you please paste into this pull request a sample logged output line?

PR #1949, opened 2021-09-22
Make test_retrieve_jobs_with_deprecated_api test use the default pool.

scrosby:

- Make test_retrieve_jobs_with_deprecated_api test use the default pool.
- Make the wait check if the job is running run less often.

Having it run on this pool makes it more reliable. We use the default pool for other tests so the resources are likely to be available immediately. This also cuts the logging verbosity by a lot if/when the test fails.


dposada:
Will merge when 🟢

PR #1954, opened 2021-09-24
Make cook add a finalizer to its k8s pods to prevent GC.

scrosby:
This gives Cook time to collect the pod completion status.

- Have Cook put finalizers on its pods

This prevents pod-gc from deleting pods before cook can collect the completion status.

### DIFF
--- a/scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
+++ b/scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
@@ -1,0 +1,53 @@
+package com.twosigma.cook.kubernetes;
+
+import io.kubernetes.client.custom.V1Patch;
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.apis.CoreV1Api;
+import io.kubernetes.client.openapi.models.V1Pod;
+import io.kubernetes.client.util.PatchUtils;
+import org.joda.time.DateTime;
+
+import java.util.List;
+
+public class RemoveFinalizerHelper {
+    /** A finalizer that is attached to a pod to ensure that it is not GC'ed by K8s before cook
+     * has had a chance to collect the completion result (success or failed) */
+    static public final String collectResultsFinalizer = "ts.cook/prevent-pod-gc";
+
+    /** Remove the collectResultsFinalizer from a pod if it exists on a pod and the pod is morked for
+     * deletion. */
+    static public void removeFinalizer(ApiClient apiClient, V1Pod pod) throws ApiException {
+        CoreV1Api api = new CoreV1Api(apiClient);
+
+        DateTime deletionTimestamp = pod.getMetadata().getDeletionTimestamp();
+        if (deletionTimestamp != null) {
+            List<String> finalizers = pod.getMetadata().getFinalizers();
+            if (finalizers != null) {
+                for (int ii = 0; ii < finalizers.size(); ii++) {
+                    if (collectResultsFinalizer.equals(finalizers.get(ii))) {
+                        String jsonPatchStr = "[{\"op\": \"remove\", \"path\": \"/metadata/finalizers/" + ii + "\"}]";
+                        String podName = pod.getMetadata().getName();
+                        String namespaceName = pod.getMetadata().getNamespace();
+                        PatchUtils.patch(
+                                V1Pod.class,
+                                () ->
+                                        api.patchNamespacedPodCall(
+                                                podName,
+                                                namespaceName,
+                                                new V1Patch(jsonPatchStr),
+                                                null,
+                                                null,
+                                                null, // field-manager is optional
+                                                null,
+                                                null),
+                                V1Patch.PATCH_FORMAT_JSON_PATCH,
+                                apiClient);
+                        return; // Early abort of we've found the finalizer.
+                    }
+                }
+            }
+        }
+    }
+}
+
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -21,7 +21,7 @@
             [metrics.timers :as timers]
             [plumbing.core :as pc])
   (:import (com.google.gson JsonSyntaxException)
-           (com.twosigma.cook.kubernetes WatchHelper)
+           (com.twosigma.cook.kubernetes WatchHelper RemoveFinalizerHelper)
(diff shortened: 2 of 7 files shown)


scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
```diff
@@ -0,0 +1,53 @@
...
+                                                namespaceName,
+                                                new V1Patch(jsonPatchStr),
+                                                null,
+                                                null,
+                                                null, // field-manager is optional
+                                                null,
+                                                null),
+                                V1Patch.PATCH_FORMAT_JSON_PATCH,
+                                apiClient);
+                        return; // Early abort of we've found the finalizer.
```
dposada, line 46:
```suggestion
                        return; // Early abort if we've found the finalizer.
```

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -21,7 +21,7 @@
             [metrics.timers :as timers]
             [plumbing.core :as pc])
   (:import (com.google.gson JsonSyntaxException)
-           (com.twosigma.cook.kubernetes WatchHelper)
+           (com.twosigma.cook.kubernetes WatchHelper RemoveFinalizerHelper)
```
dposada, line 24: Lexicographic

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1388,6 +1388,9 @@
       (.setName metadata pod-name)
       (.setNamespace metadata namespace)
       (.setLabels metadata labels)
+      ; Only include finalizers with real pods.
+      (when-not (synthetic-pod? pod-name)
+        (.setFinalizers metadata (list RemoveFinalizerHelper/collectResultsFinalizer)))
```
dposada, line 1393: Maybe we should put this behind a feature flag

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
```diff
@@ -0,0 +1,53 @@
...
+import io.kubernetes.client.openapi.ApiClient;
+import io.kubernetes.client.openapi.ApiException;
+import io.kubernetes.client.openapi.apis.CoreV1Api;
+import io.kubernetes.client.openapi.models.V1Pod;
+import io.kubernetes.client.util.PatchUtils;
+import org.joda.time.DateTime;
+
+import java.util.List;
+
+public class RemoveFinalizerHelper {
```
dposada, line 13: Maybe just `FinalizerHelper`?

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
```diff
@@ -0,0 +1,53 @@
...
+    /** A finalizer that is attached to a pod to ensure that it is not GC'ed by K8s before cook
+     * has had a chance to collect the completion result (success or failed) */
+    static public final String collectResultsFinalizer = "ts.cook/prevent-pod-gc";
```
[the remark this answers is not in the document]
scrosby: We can name it anything; I'm happy to change it to any name you suggest. it should be something cook or 'twosigma.cook/....' to prevent name collision.  I'd prefer not configurable becase if you ever change it, you've got a messy migration problem. There's also no need to change it and having something greppable is valuable.

scheduler/src/cook/kubernetes/api.clj
scrosby: I don't like those in general, if you're just going to leave it permanently enabled, (and the system will behave incorrectly with it disabled) why make it configurable?

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
scrosby: Thanks.

scheduler/src/cook/kubernetes/api.clj
scrosby: Fixed.

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
scrosby: Thanks.

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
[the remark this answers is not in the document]
dposada: Makes sense, how about something like `cook/process-pod`?

scheduler/src/cook/kubernetes/api.clj
dposada: Agreed, we'd want to follow-up with a PR to remove the feature flag once we see it working as expected for a few days in production.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -11,7 +11,8 @@
             [cook.scheduler.scheduler :as scheduler]
             [metrics.timers :as timers])
   (:import (clojure.lang IAtom)
-           (io.kubernetes.client.openapi.models V1ContainerStatus V1Pod V1PodStatus)
+           (com.twosigma.cook.kubernetes FinalizerHelper)
+           (io.kubernetes.client.openapi.models V1ContainerStatus V1Pod V1PodStatus V1ObjectMeta)
```
dposada, line 15: Lexicographic

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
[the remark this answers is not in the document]
dposada: I'm fine with the current name (`cook/prevent-pod-gc`) too.

scheduler/src/cook/kubernetes/controller.clj
scrosby: Fixed

scheduler/java/com/twosigma/cook/kubernetes/RemoveFinalizerHelper.java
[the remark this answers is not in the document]
scrosby: Lets keep the current name then.

scheduler/src/cook/kubernetes/api.clj
scrosby: Agreed

dposada (approved):

PR #1958, opened 2021-09-28
Trims pod event logs

dposada:
This is a follow-up to PR #1948.

Trimming down the pod event log lines.

To avoid excessive and unnecessary logging.


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -470,39 +470,103 @@
...
+  - finalizers
+  - labels
+  - name
+  - namespace
+  - resourceVersion
+
+  We do not log the following fields:
+
+  - clusterName
+  - creationTimestamp
```
nsinkov, line 498: we're now using prepare-object-metadata-for-logging for different types of metadata, so let's move creationTimestamp from excluded list to included

nsinkov (approved):

PR #1964, opened 2021-09-30
add support for incremental config for checkpointing volume mounts

nsinkov:

- add support for incremental config for checkpointing volume mounts

We want to transition from one set of checkpointing volume mounts to another incrementally to make sure there are no issues in production


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1323,263 +1187,367 @@
...
+                                               sidecar-workdir-volume
+                                               checkpoint-volume)))
+    (.setSecurityContext pod-spec security-context)
+    (.setPriorityClassName pod-spec pod-priority-class)
+
+    ; pod
+    (.setMetadata pod metadata)
+    (.setSpec pod pod-spec)
+
+    (when (synthetic-pod? pod-name)
```
scrosby, line 1514: And here we do do a `synthetic-pod?`... Why?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -977,14 +977,30 @@
   (when (and mode volume-name)
     (make-empty-volume volume-name)))
 
+(defn resolve-incremental-config-volume-mounts
```
scrosby, line 980: Why the choice of 'checkpoint' for the variable name?

Can you explain a bit of the why this exists. Not what it does. but why you might want to do this? (E.g., migration to a new image? Or is there something else?)

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1323,263 +1187,367 @@
...
+        ; resources
+        (set-mem-cpu-resources resources memory-request memory-limit cpu-request cpu-limit)
+        (.setResources container resources)
+
+        (.setVolumeMounts container [(sandbox-volume-mount-fn true) (sidecar-workdir-volume-mount-fn false) (scratch-space-volume-mount-fn false)])
+        (.addContainersItem pod-spec container)))
+
+    ; We're either using the hostname (in the case of users' job pods)
+    ; or pod-hostnames-to-avoid (in the case of synthetic pods), but not both.
+    (if hostname
```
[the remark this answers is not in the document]
nsinkov: I didn't add this code, but I don't see why a non-null hostname would be isomorphic to `synththetic-pod?`

But that kind of change shouldn't be part of this PR.

scheduler/src/cook/kubernetes/api.clj
nsinkov: See above.

scheduler/src/cook/kubernetes/api.clj
scrosby: Then can you add that to the comment? The code has 'volume' in it, your response is that its involved with checkpointing and incremental rollout. Can you explain that gap in a comment?

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
scrosby: Its a cleanup. It would be nice if we did it at the time, but if you don't want to, I'm fine with that.

scheduler/src/cook/kubernetes/api.clj
nsinkov: updated

PR #1976, opened 2021-10-12
Remove dead imports.

scrosby:

- I noticed a bunch of dead/old/unused imports in my clojure upgrade work. Lets get rid of them.

Code hygiene.


dposada:
Will merge when 🟢

PR #1990, opened 2021-10-29
add waiting under quota user metrics

nsinkov:

- add waiting under quota user metrics

If Cook and the public cloud were perfectly elastic then a user would not have any waiting jobs if they are under quota. This metric will show the real world tracking error at any moment from the theoretical ideal.

### DIFF
--- a/scheduler/src/cook/monitor.clj
+++ b/scheduler/src/cook/monitor.clj
@@ -161,7 +161,7 @@
   (let [running-stats (get-job-stats running-job-ents pool-name)
         waiting-stats (get-job-stats pending-job-ents pool-name)
         starved-stats (get-starved-job-stats db running-stats waiting-stats pool-name)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/monitor.clj
```diff
@@ -133,24 +161,29 @@
   (let [running-stats (get-job-stats running-job-ents pool-name)
         waiting-stats (get-job-stats pending-job-ents pool-name)
         starved-stats (get-starved-job-stats db running-stats waiting-stats pool-name)
+        waiting-under-quota-stats (get-waiting-under-quota-job-stats  db running-stats waiting-stats pool-name)
```
dposada, line 164:
```suggestion
        waiting-under-quota-stats (get-waiting-under-quota-job-stats db running-stats waiting-stats pool-name)
```

dposada (approved):

PR #1994, opened 2021-10-31
add metrics related to node and pod usage

nsinkov:

- add metrics related to node and pod usage

We have limits to the total number of nodes and pods we can have and we alert when we hit those limits. But we don't have metrics of those limits.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -185,13 +185,15 @@
 (defn handle-watch-updates
   "When a watch update occurs (for pods or nodes) update both the state atom as well as
   invoke the callbacks on the previous and new values for the key."
-  [state-atom ^Watch watch key-fn callbacks]
+  [state-atom ^Watch watch key-fn callbacks set-metric-counters-fn]
   (while (.hasNext watch)
     (let [watch-response (.next watch)
           item (.-object watch-response)
           type (.-type watch-response)]
       (if (some? item)
-        (process-watch-response! state-atom item type key-fn callbacks)
+        (do
+          (process-watch-response! state-atom item type key-fn callbacks)
+          (set-metric-counters-fn))
         (throw (ex-info "Encountered nil object on watch response"
                         {:watch-object item
                          :watch-status (.-status watch-response)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -473,6 +478,13 @@
                       {:max-synthetic-pods max-pods-outstanding
                        :synthetic-pods num-synthetic-pods}))
 
+          (set-counter-fn "total-pods" total-pods)
+          (set-counter-fn "max-total-pods" max-total-pods)
+          (set-counter-fn "total-nodes" total-nodes)
+          (set-counter-fn "max-total-nodes" max-total-nodes)
+          (set-counter-fn "num-synthetic-pods" num-synthetic-pods)
+          (set-counter-fn "max-pods-outstanding" max-pods-outstanding)
```
dposada, line 486: For consistency with the other metric names, maybe `total-synthetic-pods` and `max-total-synthetic-pods`?

nsinkov: thanks, updated

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -473,6 +478,13 @@
                       {:max-synthetic-pods max-pods-outstanding
                        :synthetic-pods num-synthetic-pods}))
 
+          (set-counter-fn "total-pods" total-pods)
+          (set-counter-fn "max-total-pods" max-total-pods)
+          (set-counter-fn "total-nodes" total-nodes)
+          (set-counter-fn "max-total-nodes" max-total-nodes)
+          (set-counter-fn "total-synthetic-pods" num-synthetic-pods)
+          (set-counter-fn "max-total-synthetic-pods" max-pods-outstanding)
```
dposada, line 486: I just realized that these metrics are being set inside `autoscale!`, which means that pools that don't autoscale won't have these metrics. That's fine for the synthetic pods metrics and for `max-total-pods` and `max-total-nodes`, but for `total-pods` and (maybe) `total-nodes`, it might be nice to have these metrics even for non-autoscaling pools. Thoughts?

nsinkov: yeah, let me see if there's a better place to put this

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -183,13 +185,15 @@
...
+  [state-atom ^Watch watch key-fn callbacks set-metric-counters-fn]
   (while (.hasNext watch)
     (let [watch-response (.next watch)
           item (.-object watch-response)
           type (.-type watch-response)]
       (if (some? item)
-        (process-watch-response! state-atom item type key-fn callbacks)
+        (do
+          (process-watch-response! state-atom item type key-fn callbacks)
+          (set-metric-counters-fn))
```
dposada, line 196: Would it be possible to just use `metrics.gauges/defgauge` instead of this approach? Seems like these are a good fit for a gauge.

dposada: Decided offline to keep this as a counter for now

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -441,7 +455,10 @@
         (try
           (log/info "In" compute-cluster-name "compute cluster, handling node watch updates")
           (handle-watch-updates current-nodes-atom watch node->node-name
-                                callbacks) ; Update the set of all nodes.
+                                callbacks              ; Update the set of all nodes.
```
dposada, line 458: Comment indentation here seems odd

dposada (approved):

PR #2000, opened 2021-11-05
Logs pod events with type Warning

dposada:

Logging pod events that do not correspond to currently tracked pods, if they both have event type `Warning` and reference `cook` in the `fieldPath`.

Pod events tied to Cook pods can arrive after Cook has deleted the pod, and we want these events to show up in the log as well.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -550,7 +550,7 @@
                               (= event-type "Warning")
                               ; The fieldPath is something like:
                               ; "spec.containers{aux-cook-sidecar-container}"
-                              (-> field-path str/lower-case (str/includes? "cook"))))
+                              (some-> field-path str/lower-case (str/includes? "cook"))))
                       (log/info "In" compute-cluster-name
                                 "compute cluster, received pod event"
                                 {:event-reason (.getReason event)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -537,10 +537,20 @@
...
+                          (get namespaced-pod-name)
+                          (is-cook-scheduler-pod compute-cluster-name))]
+                    (when (or
+                            tracked-pod?
+                            (and
+                              ; The event type is either "Normal" or "Warning"
+                              (= event-type "Warning")
+                              ; The fieldPath is something like:
+                              ; "spec.containers{aux-cook-sidecar-container}"
+                              (some-> field-path str/lower-case (str/includes? "cook"))))
```
nsinkov, line 553: is this the best we can do?

we can't identify something like

resource.labels.container_name="required-cook-job-container"

?

nsinkov: the answer seems to be yes

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -537,10 +537,20 @@
...
+                        field-path (.getFieldPath involved-object)
+                        event-type (.getType event)
+                        tracked-pod?
+                        (some-> @all-pods-atom
+                          (get namespaced-pod-name)
+                          (is-cook-scheduler-pod compute-cluster-name))]
+                    (when (or
+                            tracked-pod?
+                            (and
+                              ; The event type is either "Normal" or "Warning"
```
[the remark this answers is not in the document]
dposada: Clarified the comment

PR #2004, opened 2021-11-11
Add Postgres support to Cook.

scrosby:

- All of the logic to for Cook to communicate to a Postgres database.
- Logic to create a Cook database in Docker.

Create a development environment for Cook to talk to Postgres to enable a migration of Cook from Datomic. Does not include any changes to actually use Postgres tables.


scheduler/postgresql/README.txt
```diff
@@ -0,0 +1,86 @@
+# For cook to function, it must have a postgresql server available. There are two approaches
+# 1. Run postgresql in docker (this script)
```
dposada, line 2: I'm confused about why this is a `.txt` file if it's a script

scheduler/postgresql/README.txt
```diff
@@ -0,0 +1,86 @@
...
+# 2. Run a database outside of docker, for, e.g., development.
+
+# For #2, its pretty easy. The main thing you need to do is create the database and a user, set COOK_DB_TEST_PG_DB, PGPASSWORD and COOK_DB_TEST_PG_USER, (or put a :pg-config section in config.edn, for integration tests), and then you can set it up using
```
dposada, line 5:
```suggestion
# For #2, it's pretty easy. The main thing you need to do is create the database and a user, set COOK_DB_TEST_PG_DB, PGPASSWORD and COOK_DB_TEST_PG_USER, (or put a :pg-config section in config.edn, for integration tests), and then you can set it up using
```

scheduler/postgresql/sql/init_cook_database.sql
```diff
@@ -0,0 +1,42 @@
...
+
+-- If you get a crazy error where the two above lines pass, but 'No schema has been selected to
+-- create in' when running the first CREATE TABLE. It can be caused by there being a capital
+-- letter in cook_schema. Schema names are lowercased when created, but case-sensitive when
+-- in the search path.
+
+--- Setup the tables.
+create table resources (
+   resource_name varchar(30) PRIMARY KEY,
+   resource_description text
```
dposada, line 19: Should we make this `NOT NULL`?

scheduler/postgresql/README.txt
scrosby: Its not a script. Its a text file with some bash commands available for copy&paste.

dposada: In that case, thoughts on making this be a Markdown file with properly formatted code blocks?

scheduler/postgresql/sql/init_cook_database.sql
dposada: It wasn't clear to me from the file that this was "currently dead" -- should we add a comment in the file explaining what you explained above?

scheduler/postgresql/README.txt
scrosby: An excellent suggestion. Done.

scheduler/postgresql/sql/init_cook_database.sql
scrosby: Better idea. I'll just strip it from this review.

PR #2011, opened 2021-11-19
allow incremental configurations for default job constraints

nsinkov:

- allow incremental configurations for default job constraints

We want different default job constraints for different jobs for example to transition to different GPU driver versions

### DIFF
--- a/scheduler/src/cook/passport.clj
+++ b/scheduler/src/cook/passport.clj
@@ -30,6 +30,7 @@
 
 (def checkpoint-volume-mounts-key-selected :checkpoint-volume-mounts-key-selected)
 (def default-image-selected :default-image-selected)
+(def default-job-constraint-pattern-selected :default-job-constraint-pattern-selected)
 (def init-container-image-selected :init-container-image-selected)
 (def job-created :job-created)
 (def job-submitted :job-submitted)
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/passport.clj
```diff
@@ -30,6 +30,7 @@
 
 (def checkpoint-volume-mounts-key-selected :checkpoint-volume-mounts-key-selected)
 (def default-image-selected :default-image-selected)
+(def default-job-constraint-pattern-selected :default-job-constraint-pattern-selected)
```
dposada, line 33: Is this used anywhere?

dposada (commented):
Should we add some tests?

scheduler/src/cook/passport.clj
nsinkov: not anymore, removed

nsinkov:
> 

I'm OK not adding a test - this works when I tested by hand, and we'll go back to a fixed config after the transition.

But if you want a test, can add an integration test that's conditional on having an incremental config active. And it will either check that some possible value was set or

dposada (approved):

dposada:
Will merge when 🟢

PR #2013, opened 2021-11-19
support using default image with a user-specified container

nsinkov:

- support using default image with a user-specified container

Currently the user can only use the default image if they do not specify any container. A user is not able to specify properties of a container such as open ports and at the same time use the default image. They must specify all or nothing.
This PR lets the user set a special symbolic image in their container spec name that will be translated into the default image.

### DIFF
--- a/integration/tests/cook/test_client.py
+++ b/integration/tests/cook/test_client.py
@@ -139,6 +139,52 @@ class ClientTest(util.CookTest):
         finally:
             self.client.kill(uuid)
 
+    @unittest.skipUnless(util.docker_tests_enabled(), "Requires setting the COOK_TEST_DOCKER_IMAGE environment variable")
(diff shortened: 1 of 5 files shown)


integration/tests/cook/test_client.py
```diff
@@ -139,6 +139,52 @@ def test_container_port_submit(self):
...
+            job = self.client.query(uuid)
+            addr = job.instances[0].progress_message
+
+            self.assertIsNotNone(addr)
+
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+                sock.connect((addr, JOB_PORT))
+                message = b"hello world!"
+
+                self.assertEqual(sock.send(message), len(message))
```
dposada, line 179: Should we additionally assert that it got the default image?

nsinkov: I think the logic to calculate default image is too complicated. I'd rather leave like this

PR #2015, opened 2021-11-22
Make offer handling unit test not depend on database at compile time.

scrosby:

- Make offer handling unit test not depend on database at compile time.

Helps with the postgres migration.

### DIFF
--- a/scheduler/test/cook/test/compute_cluster.clj
+++ b/scheduler/test/cook/test/compute_cluster.clj
@@ -51,6 +51,7 @@
     (is (= config (compute-cluster-config-ent->compute-cluster-config config-ent)))))
 
 (deftest test-db-config-ents
+  (setup)
   (let [uri "datomic:mem://test-compute-cluster-config"
         conn (restore-fresh-database! uri)
         temp-db-id (d/tempid :db.part/user)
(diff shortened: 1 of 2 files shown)


scheduler/test/cook/test/compute_cluster.clj
```diff
@@ -51,6 +51,7 @@
     (is (= config (compute-cluster-config-ent->compute-cluster-config config-ent)))))
 
 (deftest test-db-config-ents
+  (setup)
```
nsinkov, line 54: how does this relate to the rest of the PR?

scheduler/test/cook/test/compute_cluster.clj
```diff
@@ -192,6 +193,7 @@
            {:a {}}))))
 
 (deftest test-get-job-instance-ids-for-cluster-name
+  (setup)
```
nsinkov, line 196: how does this relate to the rest of the PR?

scheduler/test/cook/test/compute_cluster.clj
scrosby: Unit tests need to run this to setup mount state, however, a lot omit it and piggyback on prior tests having had already run mount. I encountered this test breaking and fixed it.

scheduler/test/cook/test/compute_cluster.clj
scrosby: Unit tests need to run this to setup mount state, however, a lot omit it and piggyback on prior tests having had already run mount. I encountered this test breaking and fixed it.

PR #2017, opened 2021-12-03
Adds annotation for using all gids

dposada:

- allowing a "use all gids" pod annotation to be incrementally added to pods

In order to avoid contradictions between which group Cook thinks a user belongs to and which group Kubernetes thinks the user belongs to.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -1373,8 +1373,20 @@
     (when-not (synthetic-pod? pod-name)
       (let [[resolved-config reason] (config-incremental/resolve-incremental-config task-id :add-finalizer "false")]
         (if (= "true" resolved-config)
-          (.setFinalizers metadata (list FinalizerHelper/collectResultsFinalizer)))))(when pod-annotations
-                                                                                       (.setAnnotations metadata pod-annotations))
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1373,8 +1373,28 @@
...
+            pod-annotations
+            ; For non-synthetic (real job) pods, when configured to do so,
+            ; we add a pod annotation indicating that Kubernetes should
+            ; use all of the group IDs the user is a member of, in order
+            ; to avoid contradictions between which group Cook thinks a
+            ; user belongs to and which group Kubernetes thinks the user
+            ; belongs to.
+            (let [[resolved-config _]
+                  (config-incremental/resolve-incremental-config
+                    task-id :add-use-all-gids-annotation "false")
```
nsinkov, line 1388: you need a java.util.UUID here and you want the job uuid not instance uuid

dposada: Above for the `:add-finalizer` one we're using `task-id`

nsinkov: yeah, that's wrong too. was approved in PR 1954

PR #2020, opened 2021-12-08
fix support for incremental default image configuration

nsinkov:

- properly resolve incremental default image configuration
- fix bad test assertion

We need to fix support incremental default image configuration. If a user omits the image from their container we must still resolve the image configuration from the default container, not try to use it directly.  The image configuration from the default container might be an incremental configuration, in which case it needs to be resolved. That is what this PR fixes.

### DIFF
--- a/integration/tests/cook/test_basic.py
+++ b/integration/tests/cook/test_basic.py
@@ -2068,25 +2068,7 @@ class CookTest(util.CookTest):
             max_retries=5)
         self.assertEqual(resp.status_code, 201)
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
-        default_image_config = expected_container["docker"]["image"]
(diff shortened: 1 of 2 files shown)


integration/tests/cook/test_basic.py
```diff
@@ -2068,25 +2068,7 @@ def test_default_container_for_pool(self):
             max_retries=5)
         self.assertEqual(resp.status_code, 201)
         job = util.wait_for_job(self.cook_url, job_uuid, 'completed')
-        default_image_config = expected_container["docker"]["image"]
```
scrosby, line 2071: Can this test be repaired in a useful way instead of basically deleted?

nsinkov: I tried to think of a way but it was too close to re-implementing the business logic in a test for me. I'd rather have it like this.

PR #2022, opened 2021-12-14
Add support for postgres-based unit tests.

scrosby:

- Add in the infrastructure so that unit tests run with a live postgres database.
- This includes all opensource test environments.
- No actual use of postgresql in this code.

To setup a development environment. This code has been tested with a Proof-of-Concept quota on postgres change. to validate correctness.

### DIFF
--- a/scheduler/src/cook/postgres.clj
+++ b/scheduler/src/cook/postgres.clj
@@ -1,88 +1,10 @@
 (ns cook.postgres
   (:require [clojure.tools.logging :as log]
-            [clojure.java.jdbc :as sql]
-            [cook.config :as config])
-  (:import (java.lang Runtime)
-           (java.util Random)))
-
-(defn setup-database
-  "Setup a fresh cook schema from scratch. At present, calls out
-  to a bash script, passing it the target schema name and hopes it does the right thing."
-  [setup-script schema-name]
-  ; This environmental variable contains the path of a setup script that, given a schema name, creates a database.
-  (let [^Process p (-> (Runtime/getRuntime)
-                       (.exec (str setup-script " " schema-name)))
-        _ (log/info "Waiting for database schema production process" (.pid p) "to exit")
-        exit-code (.waitFor p)]
-    (assert (= exit-code 0) (str "Exit code for process creation is" exit-code "not zero"))))
-
-(defn make-database-connection-dictionary-for-unit-tests
-  "Given a target schema name and assuming a variety of environmental
-   variables are set, creates the metadata to connect to a database."
-  [currentSchema]
-  (merge {:dbtype "postgresql"
-          :dbname (or (System/getenv "COOK_DB_TEST_PG_DB") "cook_local")
-          ; Note that we need to save the currentSchema in the connection info to
-          :currentSchema currentSchema
-          :host (or (System/getenv "COOK_DB_TEST_PG_SERVER") "127.0.0.1")
-          :ssl false
-          :sslfactory "org.postgresql.ssl.NonValidatingFactory"}
-         ; PGPASSWORD is the default enviornmental variable used by postgres psql.
-         (when-let [passwd (or  (System/getenv "COOK_DB_TEST_PG_PASSWORD")
-                                (System/getenv "PGPASSWORD"))]
-           {:password passwd})
-         (when-let [username (System/getenv "COOK_DB_TEST_PG_USER")]
-           {:user username})))
-
+            [cook.config :as config]))
 
 ; With unit tests, we need to create a new configuration dictionary in the test fixture.
 ; That gets stored here.
 (def saved-pg-config-dictionary (atom nil))
-
-
-(let [custom-formatter (clj-time.format/formatter "yyyyMMdd")]
-  (defn make-new-schema-name
-    "Generate the schema name we want to use for unit tests"
-    []
-    (str "db_tests_" (System/getenv "USER") "_" (clj-time.format/unparse-local-date custom-formatter (clj-time.core/today)) "_" (+ 1000 (.nextInt (Random.) 9000)))))
-
-(defn configure-database-connection-for-unit-tests
-  "Configure the database connection."
-  []
-  (if-let [schema-name (System/getenv "COOK_DB_TEST_PG_SCHEMA")]
-    (do
-      (let [pg-connection-meta (make-database-connection-dictionary-for-unit-tests schema-name)]
-        (log/info "Using existing schema " schema-name " with connection " pg-connection-meta)
-        (reset! saved-pg-config-dictionary pg-connection-meta)
-        pg-connection-meta))
-    ; If I have an environment saying to autocreate schema:
-    (do
-      (let [setup-script (System/getenv "COOK_DB_TEST_AUTOCREATE_SCHEMA")
-            schema-name (make-new-schema-name)
-            pg-connection-meta (make-database-connection-dictionary-for-unit-tests schema-name)]
-        (log/info "Auto-creating schema " schema-name "with connection" pg-connection-meta)
-        (setup-database setup-script schema-name)
-        (reset! saved-pg-config-dictionary pg-connection-meta)
-        pg-connection-meta))))
-
-(defn deconfigure-database-connection-for-unit-tests
-  "Deconfigure the database connection. Used when completing a test fixture."
-  []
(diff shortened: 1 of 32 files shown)


scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+  to a bash script, passing it the target schema name and hopes it does the right thing."
+  [setup-script schema-name]
+  ; This environmental variable contains the path of a setup script that, given a schema name, creates a database.
+  (let [^Process p (-> (Runtime/getRuntime)
+                       (.exec (str setup-script " " schema-name)))
+        _ (log/info "Waiting for database schema production process" (.pid p) "to exit")
+        exit-code (.waitFor p)]
+    (assert (= exit-code 0) (str "Exit code for process creation is" exit-code "not zero"))))
+
+(defn make-database-connection-dictionary-for-unit-tests
```
dposada, line 19: Shouldn't this function be under `scheduler/test/` instead of `scheduler/src/`?

scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+         (when-let [passwd (or  (System/getenv "COOK_DB_TEST_PG_PASSWORD")
+                                (System/getenv "PGPASSWORD"))]
+           {:password passwd})
+         (when-let [username (System/getenv "COOK_DB_TEST_PG_USER")]
+           {:user username})))
+
+
+; With unit tests, we need to create a new configuration dictionary in the test fixture.
+; That gets stored here.
+(def saved-pg-config-dictionary (atom nil))
```
dposada, line 40: Shouldn't this be under `scheduler/test/` instead of `scheduler/src/`?

scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+   There are 2 main codepaths we take here:
+
+   1. If there is a config/config with a pg-config in config.edn, (I.e., we're running
+      integration tests or production) run with that config.
+
+   2. If there's a saved configuration dictionary (used for unit tests), return that."
+  []
+  (or
+    (-> config/config :settings :pg-config)
+    @saved-pg-config-dictionary))
```
dposada, line 98: Same question as above for most things in this file, I suppose: should this stuff be under `scheduler/test/` instead of `scheduler/src/`?

scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+  []
+  (when-not (System/getenv "COOK_DB_TEST_PG_SCHEMA")
+    (let [schema-name (:currentSchema @saved-pg-config-dictionary)]
+      (log/info "Auto-destroying schema " schema-name "with connection" @saved-pg-config-dictionary)
+      ; Can't use '?' here with parameter substitution.
+      (sql/execute! @saved-pg-config-dictionary [(str "DROP SCHEMA " schema-name " CASCADE;")]))
+    (reset! saved-pg-config-dictionary nil)))
+
+(defn with-pg-db [f]
+  "Test fixture that sets up a new postgres database (if needed), runs the unit tests, and demolishes it afterwards. Setup a postgres database test fixture that creates a database and destroys it afterwards"
```
dposada, line 79: Might be nice to wrap this long line

scheduler/src/cook/postgres.clj
scrosby: I'd like to localize all postgres functions into this one file. Also, this code is going to need to refactor (in semi-uncertain ways at the moment) when we add in connection pooling. I'd like to keep it here for now.

scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+          :currentSchema currentSchema
+          :host (or (System/getenv "COOK_DB_TEST_PG_SERVER") "127.0.0.1")
+          :ssl false
+          :sslfactory "org.postgresql.ssl.NonValidatingFactory"}
+         ; PGPASSWORD is the default enviornmental variable used by postgres psql.
+         (when-let [passwd (or  (System/getenv "COOK_DB_TEST_PG_PASSWORD")
+                                (System/getenv "PGPASSWORD"))]
+           {:password passwd})
+         (when-let [username (System/getenv "COOK_DB_TEST_PG_USER")]
+           {:user username})))
```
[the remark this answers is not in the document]
scrosby: I don't see how you'd do that without either adding more syntatic cruft with nested let calls, or duplicating the code? 

This seems pretty gross to me:

```
(cond->
   ...
  (or  (System/getenv "COOK_DB_TEST_PG_PASSWORD")
                                (System/getenv "PGPASSWORD"))
    (assoc :password (or  (System/getenv "COOK_DB_TEST_PG_PASSWORD")
                                (System/getenv "PGPASSWORD")))
```

If we extract the common code as a let, there's not a nice place to put it thats local to the use. Also, FWIW, this code will change when I do database connection pooling. c3p0 does not take clojure dictionaries.

scheduler/src/cook/postgres.clj
scrosby: No. (pg-db) is how Cook gets its database connection and has to be here. This is read by that function, and is how the test fixture code communicates the chosen database out.

scheduler/src/cook/postgres.clj
```diff
@@ -0,0 +1,98 @@
...
+
+
+; With unit tests, we need to create a new configuration dictionary in the test fixture.
+; That gets stored here.
+(def saved-pg-config-dictionary (atom nil))
+
+
+(let [custom-formatter (clj-time.format/formatter "yyyyMMdd")]
+(defn make-new-schema-name
```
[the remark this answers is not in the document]
scrosby: Yes. Thanks.

scheduler/src/cook/postgres.clj
scrosby: Done.

scheduler/src/cook/postgres.clj
scrosby: Nope. This is how all of the code in Cook gets a database connection. See E.g., https://github.com/scrosby/Cook/blob/wip/postgres-docker/scheduler/src/cook/quotashare.clj#L146

scheduler/src/cook/postgres.clj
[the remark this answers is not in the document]
dposada: You could use `better-cond`, which we've already used in a couple of other spots, but I think it's also fine as is.

scheduler/src/cook/postgres.clj
dposada: Why would we keep unit-test-specific stuff in `src`?

scheduler/src/cook/postgres.clj
dposada: The comment seems to imply this is only for unit tests

scheduler/src/cook/postgres.clj
dposada: Most of the things in this file seem unit-test-specific.

scheduler/src/cook/postgres.clj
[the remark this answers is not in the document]
scrosby: Lets keep it as-is then.

scheduler/src/cook/postgres.clj
scrosby: Thinking on it further, this is used in the database configuration code, so this is called indirectly by pg-db, so therefore needs to be here (in in a file this namespace depends on) in order to be called.

scheduler/src/cook/postgres.clj
scrosby: This code is called by Cook's prod code (via pg-db), so needs to be in the call chain for Cook. This codepath however is only actually invoked when run within a unit test context.

scheduler/src/cook/postgres.clj
scrosby: Yes. Because we're having Cook create DDL in test fixtures, most of the complexity in this file is related to unit tests. Had we went with the 'database is assumed to exist' this file would have `make-database-connection-dictionary-for-unit-tests` + 4 lines of code because unit tests wouldn't have needed all of these extra codepaths.  In addition, this is going to get more coupled, stateful, and need lifecycle management when we add connection pooling.

dposada (approved):

PR #2023, opened 2021-12-15
Check that job labels that are pod labels have valid format

nsinkov:

- add check to job labels on submission

Check that job labels that are pod labels have valid format on job submission


dposada:
LGTM! Should we add tests?

nsinkov:
@dposada thanks, updated

dposada (approved):

PR #2035, opened 2022-01-10
Switch Cook to using c3p0 for database pooling.

scrosby:

- Enable Cook to use c3p0 connection pooling for postgres.
- Fix a bug where we didn't set the connection metadata from the environment in seed_pools startup.

Enable c3p0 and fix a bug in Postgres support

### DIFF
--- a/scheduler/project.clj
+++ b/scheduler/project.clj
@@ -132,6 +132,7 @@
                  [org.apache.curator/curator-test "2.7.1"
                  :exclusions [io.netty/netty
                               io.netty/netty-transport-native-epoll]]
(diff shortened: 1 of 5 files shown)


scheduler/project.clj
```diff
@@ -132,6 +132,7 @@
                  [org.apache.curator/curator-test "2.7.1"
                  :exclusions [io.netty/netty
                               io.netty/netty-transport-native-epoll]]
+                 [com.mchange/c3p0 "0.9.5.2"] ; Connection pooling.
```
nsinkov, line 135: why 0.9.5.2 and not 0.9.5.5

PR #2048, opened 2022-02-07
Change share and quota to use Postgres

nsinkov:

- change share and quota to use Postgres

Share and quota are small and simple and can be the first tables in production using Postgres


scrosby:
Worked in opensource with reads and writes matching. When combined with a preexisting database, need to transfer quota from datomic into that db.

PR #2058, opened 2022-03-24
Add second init sidecar for checkpointing.

scrosby:

- Add support for a second init container that is used for setting up checkpointing.

- This improves support when the sidecar and user container have a different platform.

This diff is much smaller if you suppress whitespace. Some refactoring was done to extract common variables.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -885,6 +885,9 @@
 (def cook-container-name-for-file-server
   "aux-cook-sidecar-container")
 ; see pod->synthesized-pod-state comment for container naming conventions
+(def cook-container-name-for-checkpoint-injection
+  "aux-cook-checkpoint-injection-container")
+; see pod->synthesized-pod-state comment for container naming conventions
 (def cook-init-container-name
   "aux-cook-init-container")
 
@@ -1334,7 +1337,7 @@
         sandbox-dir (:default-workdir (config/kubernetes))
         workdir (get-workdir parameters sandbox-dir)
         {:keys [volumes volume-mounts sandbox-volume-mount-fn]} (make-volumes volumes sandbox-dir pod-labels)
-        {:keys [custom-shell init-container sidecar telemetry-pool-regex telemetry-agent-host-var-name telemetry-env-var-name
+        {:keys [custom-shell init-container checkpoint-container sidecar telemetry-pool-regex telemetry-agent-host-var-name telemetry-env-var-name
                 telemetry-env-value telemetry-service-var-name telemetry-tags-entry-separator
                 telemetry-tags-key-invalid-char-pattern telemetry-tags-key-invalid-char-replacement
                 telemetry-tags-key-value-separator telemetry-tags-var-name telemetry-version-var-name]}
@@ -1343,6 +1346,7 @@
         pod-name (str task-id)
         checkpoint-memory-overhead (:memory-overhead checkpoint)
         use-cook-init? (and init-container pod-supports-cook-init?)
+        use-checkpoint-injection? (and checkpoint-container pod-supports-cook-init?)
         use-cook-sidecar? (and sidecar pod-supports-cook-sidecar?)
         init-container-workdir "/mnt/init-container"
         init-container-workdir-volume (when use-cook-init?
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1466,143 +1470,174 @@
...
+        ; checkpoint injection container
+        ; This container is used for injecting checkpoint/restore functionality into the user's main container.
+        ;
+        ; This is run in the same image as the user's container, but run with an alternate command line. The reason is that
+        ; the init container may not use the same platform image as the user's container, so certain binaries or other
+        ; images can't be copied between those container.
+        ;
+        ; The idea is the init container injects some scripts into a shared volume used by the checkpoint container.
+        ; The checkpoint container then runs. As it runs with the same image as the user container, those scripts can
+        ; inspect the platform the user container is on and install the right checkpoint or other code.
```
laurameng, line 1587: This comment is very helpful - thanks!

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1466,143 +1470,174 @@
...
-                                  (.setHttpGet http-get-action))]
-            (.setReadinessProbe container readiness-probe)))
-
-        ; resources
-        (set-mem-cpu-resources resources memory-request memory-limit cpu-request cpu-limit)
-        (.setResources container resources)
 
-        (.setVolumeMounts container [(sandbox-volume-mount-fn true) (sidecar-workdir-volume-mount-fn false) (scratch-space-volume-mount-fn false)])
-        (.addContainersItem pod-spec container)))
+      (let [get-resource-requirements-fn (fn [fieldname] (if use-cook-sidecar?
```
laurameng, line 1550: Groktest: you're refactoring to have some common variables between the init-container and the new checkpoint-init-container

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1466,143 +1470,174 @@
...
-          gpu-model-requested (constraints/job->gpu-model-requested gpus job pool-name)
-          ;; if disk config for pool has enable-constraint? set to true, add disk info to the job
-          enable-disk-constraint? (regexp-tools/match-based-on-pool-name (config/disk) pool-name :enable-constraint? :default-value false)
-          ;; if user did not specify disk request, use default on pool
-          disk-request (when enable-disk-constraint?
-                         (constraints/job-resources->disk-request resources pool-name))
-          disk-limit (when enable-disk-constraint? (-> resources :disk :limit))
-          ;; if user did not specify disk type, use default on pool
-          disk-type (when enable-disk-constraint? (constraints/job-resources->disk-type resources pool-name))
+    (let [{:keys [image]} docker
```
laurameng, line 1473: Groktest: you needed to pull this up a level in order to use it in both the main container and the new checkpoint-init-container

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1466,143 +1470,174 @@
...
+        ;
+        ; This is run in the same image as the user's container, but run with an alternate command line. The reason is that
+        ; the init container may not use the same platform image as the user's container, so certain binaries or other
+        ; images can't be copied between those container.
+        ;
+        ; The idea is the init container injects some scripts into a shared volume used by the checkpoint container.
+        ; The checkpoint container then runs. As it runs with the same image as the user container, those scripts can
+        ; inspect the platform the user container is on and install the right checkpoint or other code.
+        (when use-checkpoint-injection?
+          (let [{:keys [command]} checkpoint-container
```
laurameng, line 1589: So, the k8s config will have a {"checkpoint-container" {"command" "<checkpoint command>"}} entry? We don't need to do something like {"<image name>" "<checkpoint command for image"}, right?

scheduler/src/cook/kubernetes/api.clj
scrosby: Yup. I could have copy&pasted the whole block, but that would have had a lot of duplicate variable definitions.

scheduler/src/cook/kubernetes/api.clj
scrosby: Yes. Just a command.

No. Remember the comment above: We're using the user's container's image, so that doesn't need to be specified.

PR #2065, opened 2022-04-08
Splitting 'updating dynamic clusters' log into separate entries

samincheva:
- Split the "Updating dynamic clusters" log line into multiple entries for each config.

The current log is too long and hard to read.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -530,11 +530,12 @@
         dissoc-ca-cert-vals
         (fn [configs]
           (map-vals dissoc-ca-cert configs))]
-    (log/info "Updating dynamic clusters."
-              {:current-configs (dissoc-ca-cert-vals current-configs)
-               :force? force?
-               :new-configs (dissoc-ca-cert-vals new-configs)
-               :updates (map #(update % :goal-config dissoc-ca-cert) updates)})
+    (doseq [[cluster-name config] (dissoc-ca-cert-vals current-configs)]
+      (log/info "Updating dynamic cluster current config for" cluster-name config))
+    (doseq [[cluster-name config] (dissoc-ca-cert-vals new-configs)]
+      (log/info "Updating dynamic cluster new config for" cluster-name config))
+    (doseq [{:keys [goal-config] :as updated-config} (map #(update % :goal-config dissoc-ca-cert) updates)]
+      (log/info "Updating dynamic cluster updated config for" (:name goal-config) (assoc updated-config :force? force?)))
     (let [updates-with-results (map
                                  #(assoc % :update-result
                                            (when (:valid? %) (execute-update! conn %)))


scheduler/src/cook/compute_cluster.clj
```diff
@@ -530,11 +530,13 @@
         dissoc-ca-cert-vals
         (fn [configs]
           (map-vals dissoc-ca-cert configs))]
-    (log/info "Updating dynamic clusters."
-              {:current-configs (dissoc-ca-cert-vals current-configs)
-               :force? force?
-               :new-configs (dissoc-ca-cert-vals new-configs)
-               :updates (map #(update % :goal-config dissoc-ca-cert) updates)})
+    (log/info "Updating dynamic clusters.")
```
laurameng, line 533: I am in favor of removing this line since the others will provide the same context

scheduler/src/cook/compute_cluster.clj
```diff
@@ -530,11 +530,13 @@
...
+    (doseq [current-config (dissoc-ca-cert-vals current-configs)]
+      (log/info "Updating dynamic cluster" (key current-config) "current config:" current-config))
```
laurameng, line 535: current config and new config look to be logging the map entry, so adding the specific key is redundant, but harmless.

laurameng: I like the format of updating cluster [cluster], shame we can't match it on the updated configs section.

scheduler/src/cook/compute_cluster.clj
samincheva: Makes sense, I kept it mostly for context while making the change, will remove it.

scheduler/src/cook/compute_cluster.clj
```diff
@@ -530,11 +530,12 @@
         dissoc-ca-cert-vals
         (fn [configs]
           (map-vals dissoc-ca-cert configs))]
-    (log/info "Updating dynamic clusters."
-              {:current-configs (dissoc-ca-cert-vals current-configs)
-               :force? force?
-               :new-configs (dissoc-ca-cert-vals new-configs)
-               :updates (map #(update % :goal-config dissoc-ca-cert) updates)})
+    (doseq [current-config (dissoc-ca-cert-vals current-configs)]
```
scrosby, line 533: This can be a destructuring:

`(doseq ([cluster-name config] (dissoc-ca-cert-vals ...)))
   ... `

See: https://clojure.org/guides/destructuring

(Thats why you find relatively few `(key ...)` or `(val ...)` or direct dictionary gets in clojure

Also I suggest reformatting these outputs to make it easier to distinguish the in elk: 

`(log/info "Updating dynamic cluster" name "current config" ... )`

vs

`(log/info "Updating dynamic cluster current config for" name "to" ...)`

I can distinguish these much easier with the renamed version `"Updating dynamic cluster current config for"` more than I can `"Updating dynamic cluster"` Also, punctuation doesn't index will in lucene.

scheduler/src/cook/compute_cluster.clj
```diff
@@ -530,11 +530,12 @@
...
-              {:current-configs (dissoc-ca-cert-vals current-configs)
-               :force? force?
-               :new-configs (dissoc-ca-cert-vals new-configs)
-               :updates (map #(update % :goal-config dissoc-ca-cert) updates)})
+    (doseq [[cluster-name config] (dissoc-ca-cert-vals current-configs)]
+      (log/info "Updating dynamic cluster current config for" cluster-name config))
+    (doseq [[cluster-name config] (dissoc-ca-cert-vals new-configs)]
+      (log/info "Updating dynamic cluster new config for" cluster-name config))
+    (doseq [updated-config (map #(update % :goal-config dissoc-ca-cert) updates)]
+      (log/info "Updating dynamic cluster updated config for" (:name (:goal-config updated-config)) :force? force? updated-config))
```
scrosby, line 538: The clojure idiom is to use map destructuring whenever feasible, of when you want more than a single key out of a map. 

(doseq [{:keys [goal-config] :as updated-config}  (map #(update % :goal-config dissoc-ca-cert) updates)
   (log/info "Updating dynamic cluster updated config for" (:name goal-config) (assoc updated-config :force? force?)))

(The other change is to put :force into the returned dictionary. So the lines have a similar structure.)

PR #2080, opened 2022-05-13
Updating more logs to use structured logging

samincheva:
- Update more messages to use structured logging with fields like pool name and compute cluster name.

This will make it easier to filter out logs and visualize metrics.

### DIFF
--- a/scheduler/src/cook/kubernetes/api.clj
+++ b/scheduler/src/cook/kubernetes/api.clj
@@ -10,6 +10,7 @@
             [cook.config :as config]
             [cook.config-incremental :as config-incremental]
             [cook.kubernetes.metrics :as metrics]
+            [cook.log-structured :as log-structured]
             [cook.monitor :as monitor]
             [cook.passport :as passport]
             [cook.regexp-tools :as regexp-tools]
@@ -203,9 +204,10 @@
         (histograms/update!
           (metrics/histogram metric-name compute-cluster-name) millis)
         (when log?
-          (log/info
-            "In" compute-cluster-name "compute cluster, marked watch gap"
-            {:metric-name metric-name
+          (log-structured/info
+            "Marked watch gap"
+            {:compute-cluster compute-cluster-name
+             :metric-name metric-name
              :watch-gap-millis millis
              :watch-last-event (-> last-watch-response-or-connect-millis tc/from-long str)}))))
     (swap!
@@ -245,9 +247,8 @@
                   :watch-object-type watch-object-type
                   :watch-status (.-status watch-response)
                   :watch-type type})))))
-  (log/info
-    "In" compute-cluster-name "compute cluster, there are no more"
-    (name watch-object-type) "watch updates"))
+  (log-structured/info (print-str "There are no more" (name watch-object-type) "watch updates")
+                       {:compute-cluster compute-cluster-name}))
 
 (defn get-pod-namespaced-key
   [^V1Pod pod]
@@ -285,8 +286,9 @@
     (loop [all-pods []
            continue-string nil
            resource-version-string nil]
-      (log/info "In" compute-cluster-name "compute cluster, listing pods for all namespaces"
-                {:continue continue-string
+      (log-structured/info "Listing pods for all namespaces"
+                {:compute-cluster compute-cluster-name
+                 :continue continue-string
                  :limit limit
                  :number-pods-so-far (count all-pods)
                  :resource-version resource-version-string})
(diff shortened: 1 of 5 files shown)


scheduler/src/cook/kubernetes/api.clj
```diff
@@ -314,12 +316,12 @@
...
-                      {:limit limit
-                       :number-pods (count new-all-pods)
-                       :number-pods-this-chunk (count pods)
-                       :resource-version resource-version})
+            (log-structured/info "Done listing pods for all namespaces"
+                                 {:compute-cluster compute-cluster-name
+                                  :limit limit
+                                  :number-pods (count new-all-pods)
+                                  :number-pods-this-chunk (count pods)
+                                  :resource-version resource-version})
```
laurameng, line 324: Does this effectively absorb the existing logged clojure dict into the metadata dict for structured logging? Are we worried about cardinality then?

samincheva: It does absorb the dictionary, but I don’t think we need to worry about the cardinality here since it’s a small number of fixed fields. 

There were a few places where the dict data would end up adding a large number of varying fields, and for those I converted it to a string explicitly so it is all logged as one json blob under one field. 

But in general a few tens of fields across the different messages should not be a problem for log field cardinality.

scheduler/src/cook/log_structured.clj
```diff
@@ -42,30 +42,38 @@
   "Logs a structured message at the specified level.
   Supports specifying a custom logger namespace as an optional parameter."
   ([level data metadata]
-   `log/log ~level (json/write-str (assoc ~metadata :data ~data)))
-  ([level data metadata logger-ns]
-   `(log/log ~logger-ns ~level nil (json/write-str (assoc ~metadata :data ~data)))))
+   `(log/log ~level (json/write-str (assoc ~metadata :data ~data))))
```
scrosby, line 45: This code here means we may unconditionally process each log message (including super verbose /debug logs. 

My suspicion is that the logging library behind the scenes will check the log level log component and not generate strings that aren't actually logged. We should verify that that assumption holds. E.g., a unit test. 

Structured log something with a function call. Make sure that the function isn't actually invoked --- if it is invoked, then I'd advocate not structuring the /debug logs.)

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -759,7 +777,9 @@
     ;; Note that node-unschedulable may be nil or false or true.
     :let [node-unschedulable (some-> node .getSpec .getUnschedulable)]
     node-unschedulable (do
-                         (log/info "Filtering out" node-name "because it is unschedulable" node-unschedulable)
+                         (log-structured/info "Filtering out node because it is unschedulable"
```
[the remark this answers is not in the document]
samincheva: I think this change is still worth making so that these messages are more readable when looking at the logs. I'll submit another MR for the aggregate counters.

scheduler/src/cook/log_structured.clj
samincheva: According to the [documentation](https://clojure.github.io/tools.logging/#clojure.tools.logging/log) `log` only evaluates and logs the message is the logger is enabled (`log*` does not check if the logger is enabled).

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -389,22 +394,25 @@
...
                                      (callback task (get old-all-pods task) (get namespaced-pod-name->pod task))
                                      (catch Exception e
-                                       (log/error e "In" compute-cluster-name
-                                                  "compute cluster, pod watch error while processing callback for" task)))))))
+                                       (log-structured/error (print-str "Pod watch error while processing callback for" task)
+                                                             {:compute-cluster compute-cluster-name}
+                                                             e)))))))
                  tasks))]
       (run! deref futures))
     (let [watch (create-pod-watch api-client resource-version)]
```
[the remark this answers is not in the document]
samincheva: The line on 415 logs that we're handling the pod watch updates, did you want a separate log line in between 412 and 415?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -171,18 +174,21 @@
...
       (try
         (controller/scan-process kcc taskid)
         (catch Exception e
-          (log/error e "In" name
-                     "compute cluster, encountered exception scanning task"
-                     taskid))))))
+          (log-structured/error "Encountered exception scanning task"
+                                {:compute-cluster name
+                                 :task-id taskid}
+                                e))))))
```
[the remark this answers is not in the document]
samincheva: Done

scheduler/src/cook/log_structured.clj
samincheva: I added a unit test to make sure that the functions aren't called at the debug level.

PR #2085, opened 2022-05-31
Speed up rank loop worst case by improving the instance->user cache.

scrosby:

- Add a unit test that checks to make sure that running and waiting jobs sort in the right (implicit) order.
- Also checks that we successfully cache the owners of these jobs in our cache.
- Make sure that we cache the user for any job or instance.

This helps to avoid some 10x rank time outliers. The unit test helps make sure that we order jobs correctly on future changes to this code.

### DIFF
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -566,16 +566,29 @@
          (when hostname {:instance/hostname hostname})
          (when slave-id {:instance/slave-id slave-id})))
 
+
 (defn task-ent->user
+  "Given a task entity, either a real task entity, or a fake one created by
+  create-task-ent above. (This is used to create task entities for waiting jobs for
+  DRU sorting purposes, DRU sort normalizes everything as a task).
+
+  We cannot cache these using the standard cache/lookup-cache-datomic-entity! because
+  the fake create-task-ent tasks do not have a :db/id, so would be uncached.
+
+  We really want to cache the user mapping for these. This makes it faster to sort
+  and group by user lazily to compute per-user DRU and prevents us from blowing out
+  the in-memory cache at the datomic layer which can cause rank time to become 10x
+  more expensive."
   [task-ent]
-  (let [task-ent->user-miss
-        (fn [task-ent]
-          (get-in task-ent [:job/_instance :job/user]))]
-    (caches/lookup-cache-datomic-entity! caches/task-ent->user-cache task-ent->user-miss task-ent)))
+  (let [task-ent->user-miss (fn [task-ent]
+                              (get-in task-ent [:job/_instance :job/user]))
+        extract-key (fn [task-ent]
+                      (or (:db/id task-ent (-> task-ent :job/_instance :db/id))))]
+    (ccache/lookup-cache! caches/task-ent->user-cache extract-key task-ent->user-miss task-ent)))
 
 (def ^:const default-job-priority 50)
 
-
+; This code is tested with test-sorted-task-scored-task-pairs-with-running
 (defn task->feature-vector
   "Vector of comparable features of a task.
    We use :instance/start-time, because this sort sees all running and waiting jobs for a user.
--- a/scheduler/src/cook/tools.clj
+++ b/scheduler/src/cook/tools.clj
@@ -569,8 +569,9 @@
 
 (defn task-ent->user
   "Given a task entity, either a real task entity, or a fake one created by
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/tools.clj
```diff
@@ -566,16 +566,29 @@
...
+  more expensive."
   [task-ent]
-  (let [task-ent->user-miss
-        (fn [task-ent]
-          (get-in task-ent [:job/_instance :job/user]))]
-    (caches/lookup-cache-datomic-entity! caches/task-ent->user-cache task-ent->user-miss task-ent)))
+  (let [task-ent->user-miss (fn [task-ent]
+                              (get-in task-ent [:job/_instance :job/user]))
+        extract-key (fn [task-ent]
+                      (or (:db/id task-ent (-> task-ent :job/_instance :db/id))))]
```
laurameng, line 586: iiuc, we are using the db/id associated with the real instance for synthetic tasks?

scheduler/test/cook/test/scheduler/dru.clj
```diff
@@ -120,6 +122,45 @@
...
+;; to actual running instances by task->feature-vector
+(deftest test-sorted-task-scored-task-pairs-with-running
+  (setup)
+  (let [datomic-uri "datomic:mem://test-sorted-task-scored-task-pairs-order"
+        conn (restore-fresh-database! datomic-uri)
+        job1 (create-dummy-job conn :user "ljin" :memory 10.0 :ncpus 10.0 :name "1")
+        job2 (create-dummy-job conn :user "ljin" :memory 20.0 :ncpus 20.0 :name "2" :job-state :job.state/running)
+        job3 (create-dummy-job conn :user "ljin" :memory 40.0 :ncpus 40.0 :name "3")
+        job4 (create-dummy-job conn :user "ljin" :memory 80.0 :ncpus 80.0 :name "4" :job-state :job.state/running)
+        job5 (create-dummy-job conn :user "ljin" :memory 160.0 :ncpus 160.0 :name "5")
```
laurameng, line 137: Should we mix the requests too? Like, should we avoid declaring them in sorted order?

scheduler/src/cook/tools.clj
scrosby: We're using the :db/id of the task entity for running tasks, or the :db/id of the job entity for waiting job, via the synthetic task.

scheduler/src/cook/tools.clj
```diff
@@ -566,16 +566,29 @@
          (when hostname {:instance/hostname hostname})
          (when slave-id {:instance/slave-id slave-id})))
 
+
 (defn task-ent->user
+  "Given a task entity, either a real task entity, or a fake one created by
+  create-task-ent above. (This is used to create task entities for waiting jobs for
+  DRU sorting purposes, DRU sort normalizes everything as a task).
```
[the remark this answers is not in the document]
scrosby: Fixed. Thanks.

scheduler/test/cook/test/scheduler/dru.clj
```diff
@@ -120,6 +122,45 @@
...
+        tasks (into (vec running-task-ents) pending-task-ents)]
+
+    (let [share {:mem 10.0 :cpus 10.0}
+          ; Queue should be job3, job5, becasue they're running, then 1 2 4.
+          ; DRU:     8.0 10.0 (not seen) 11.0 15.0 31.0
+          ordered-drus [8.0 10.0 11.0 15.0 31.0]]
+      (testing "dru order correct"
+        (is (= ordered-drus
+               (map (comp :dru second)
+                    (dru/sorted-task-scored-task-pairs
```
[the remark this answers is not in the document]
scrosby: This test checks that all 5 jobs we visit are in the cache.

scheduler/test/cook/test/scheduler/dru.clj
laurameng: Oh interesting. We can file something for the backlog there.

scheduler/test/cook/test/scheduler/dru.clj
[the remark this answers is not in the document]
scrosby: This test didn't exist until this PR.

If I ran this test without my cache change, the DRU order part would (and does) pass, because the DRU order shouldn't change based on cache.

The line below where I make sure the cache has 5 entries however would fail because it would only find 2 running jobs in the cache, not all 5. Now it passes.

PR #2086, opened 2022-06-01
Updating structured logging utility to handle invalid input better

samincheva:
- Update the log-structured utility to handle cases in which the input cannot be converted to json.

This avoids breaking functionality when there is an issue with the logging.


scheduler/src/cook/log_structured.clj
```diff
@@ -36,23 +36,33 @@
...
+    [cook.tools :as tools]))
+
+(defn safe-jsonify
+  "Tries to jsonify the given map and returns its json string representation.
+  If conversion to json fail, returns just the map as-is."
+  [log-map]
+  (try
+    (json/write-str (tools/format-map-for-structured-logging log-map))
+    (catch Exception e
+      log-map)))
```
scrosby, line 49: I would have this log `log-map` at log level error: 
`(log/error "Unable to json convert" (str log-map))`

Also, why not do the call to tools/format-map-for-structured-logging here, instead of doing it in the caller? Doing it in the caller means that you get to play whack-a-mole with callers; If you miss a caller, you get an exception and need to fix it up. 

I'd pass data and metadata as separate arguments, to avoid the assoc.

samincheva: Added the error level log, but I couldn't figure out a good way to make it work with passing the metadata + data as separate arguments, given that we sometimes need to pass the extra error and stacktrace as well.

What do you mean regarding calling the `format-map-for-structured-logging` here? It's already called inside the safe-jsonify function, should I be calling it elsewhere instead?

scheduler/src/cook/tools.clj
```diff
@@ -1044,9 +1044,9 @@
                   (str %))
                resource-map))
 
-(defn format-resource-map-for-structured-logging
-  "Given a map with resource amount values,
-   formats the amount values for structured logging"
+(defn format-map-for-structured-logging
```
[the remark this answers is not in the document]
samincheva: Moved to the util file (along with the other map format function for consistency).

PR #2087, opened 2022-06-02
Updating format-map-for-structured-logging to not flatten nested maps

samincheva:
- Updates the `format-map-for-structured-logging` function to not flatten nested maps, but instead traverse them and format their values appropriately.

This allows us to safely call the function on nested maps without flatting its values unintentionally.

### DIFF
--- a/scheduler/test/cook/test/util.clj
+++ b/scheduler/test/cook/test/util.clj
@@ -15,7 +15,9 @@
 ;;
 (ns cook.test.util
   (:require [clojure.test :refer :all]
-            [cook.util :refer :all]))
+            [cook.util :refer :all])
+  (:import (java.util UUID)))
+
 
 (deftest test-diff-map-keys
   (is (= [#{:b} #{:c} #{:a :d}]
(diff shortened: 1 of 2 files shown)


scheduler/test/cook/test/util.clj
```diff
@@ -56,4 +56,15 @@
...
     (is (= (set-atom! state {:a :b}) "a"))
-    (is (= @state {:a :b}))))
\ No newline at end of file
+    (is (= @state {:a :b}))))
+
+(deftest test-format-map-for-structured-logging
+  "Tests that the format-map-for-structured logging preserves nested maps."
+  (let [map {:integer 2 :float 1.2 :string "foo" :nested-map {:nested-string "bar" :nested-int 3}}
+        formatted-map (format-map-for-structured-logging map)]
+    (is (number? (:integer formatted-map)))
```
scrosby, line 65: Why not also do a straight:

`(is (eq {.....}  formattted-map)`

Also, include a UUID?

PR #2095, opened 2022-06-14
Fixing bug when running in api-only mode

samincheva:
- Change `api-only?` instances to still configured leader election, without actually joining the selection process.

Currently, `api-only?` instances fail client requests that require forwarding to the leader. This change fixes that by giving `api-only?` instances the ability to forward to the elected leader.

### DIFF
--- a/scheduler/src/cook/mesos.clj
+++ b/scheduler/src/cook/mesos.clj
@@ -171,11 +171,12 @@
    progress-config               -- map, config for progress publishing. See scheduler/docs/configuration.adoc
    framework-id                  -- str, the Mesos framework id from the cook settings
    fenzo-config                  -- map, config for fenzo, See scheduler/docs/configuration.adoc for more details
-   sandbox-syncer-state          -- map, representing the sandbox syncer object"
+   sandbox-syncer-state          -- map, representing the sandbox syncer object
+   api-only?                     -- bool, true if this instance should not actually join the leader selection"
   [{:keys [curator-framework fenzo-config mea-culpa-failure-limit mesos-datomic-conn mesos-datomic-mult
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/mesos.clj
```diff
@@ -319,12 +321,20 @@
...
                                  (if server-port "http" "https") \#
                                  (java.util.UUID/randomUUID)))
-    (.autoRequeue leader-selector)
-    (.start leader-selector)
-    (log/info "Started the leader selector")
+    (if api-only?
+      ; If this is an api-only instance, we still want all the leader election configuration to be set,
+      ; without actually joing the election process.
+      ; This allows the instance to correctly forward client requests to the leader as necessary.
+      (log/info "Will not join leader election, running as api-only")
```
laurameng, line 328: Should we add the hostname here?

scheduler/src/cook/mesos.clj
```diff
@@ -319,12 +321,20 @@
...
+      (do
+        (.autoRequeue leader-selector)
+        (.start leader-selector)
+        (log/info "Started the leader selector")
+        ))
```
laurameng, line 333: reformat to bring up to the previous line

scheduler/src/cook/mesos.clj
samincheva: It is already included in the logs we emit (all logs are tagged with a hostname, along with other common metadata).

scheduler/src/cook/mesos.clj
samincheva: Yikes, not sure how that happened. I fixed it.

PR #2109, opened 2022-07-08
Implement functinality for routing jobs to pools based on job constraints.

scrosby:

- Support for joint quotas.
- Enables jobs to have a nil submit-pool-name but have pool overridden by JobRouter plugin.

Prepare the way for being able to split jobs into different pools/queues by constraints to reduce head-of-line blocking.


integration/tests/cook/test_basic.py
```diff
@@ -2217,6 +2217,10 @@ def test_user_usage_basic(self):
             util.kill_jobs(self.cook_url, [job_uuid])
 
     def test_user_usage_grouped(self):
+        settings_dict = util.settings(self.cook_url)
+        # If we're grouping quota, then we're doing HoL blocking mitigation, so this test is broken.
+        if len(settings_dict["quota-grouping"]) > 0:
+            self.skipTest("No pool match rules defined")
```
ahaysx, line 2223: Prefer `@unittest.skipIf`?

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1702,7 +1713,14 @@
...
     (.addTolerationsItem pod-spec toleration-for-deletion-candidate-of-autoscaler)
-    (.addTolerationsItem pod-spec (toleration-for-pool cook-pool-taint-name cook-pool-taint-prefix pool-name))
+    ; Cook supports two kinds of tolerations on nodes.
+    ; 1. It can tolerate a node, where the value of that toleration depends on the pool name.
+    ; I.e.,  <KEY1>=<PREFIX><POOLNAME>
+    (when cook-pool-taint-name
+      (.addTolerationsItem pod-spec (toleration-for-pool cook-pool-taint-name cook-pool-taint-prefix pool-name)))
+    ; 2. Or it can statically tolerate a particular KEY2 and VAL2.
+    (when cook-pool-taint2-name
+      (.addTolerationsItem pod-spec (toleration-for-pool2 cook-pool-taint2-name cook-pool-taint2-value)))
```
laurameng, line 1723: Ah, here's where it 1 & 2 come together. 
I don't love this. Specifically, I don't love that 1 & 2 are logically different and that difference isn't reflected in their config keys. I was assuming that 1 & 2 were applied as `1 || 2` or `1 && 2` and was ready to comment about that. I didn't realize that 1 is a prefix and 2 is not.

laurameng: Oh, I stand corrected. It's there as `cook-pool-taint-prefix` vs `cook-pool-taint2-name`

integration/tests/cook/test_basic.py
scrosby: Can't easily as we want the settings_dict variable, and we need the self.cook_url to get it, which I don't believe is available as a decorator.

PR #2129, opened 2022-08-02
Adding prometheus metric for jobs launch count

samincheva:
- Adds a counter for total jobs launched per pool and compute cluster.

- This would let us visualize the job launch rate.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -38,7 +38,7 @@
             [cook.plugins.definitions :as plugins]
             [cook.plugins.launch :as launch-plugin]
             [cook.pool :as pool]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -38,7 +38,7 @@
             [cook.plugins.definitions :as plugins]
             [cook.plugins.launch :as launch-plugin]
             [cook.pool :as pool]
-            [cook.prometheus-metrics :as prometheus]
+            [cook.prometheus-metrics :as prom]
```
samincheva, line 41: My intellij setup for some reason stopped working with this and was trying to treat prometheus as the prometheus library, so I changed the ref as a quick workaround.

laurameng (approved):

PR #2131, opened 2022-08-03
Adding configurable rate limit for the /metrics endpoint

samincheva:
- Adds support for a custom rate limit to the /metrics endpoint.

The default rate limit for endpoints with no auth is not sufficient, and we want to ensure that the /metrics endpoint always works correctly.

### DIFF
--- a/scheduler/src/cook/components.clj
+++ b/scheduler/src/cook/components.clj
@@ -217,6 +217,15 @@
         (h req)
         (auth-fn req)))))
 
+(defn conditional-rate-limit
+  "Set rate limits, with support for endpoints with custom limits."
+  [h rate-limit-storage default-user-limit user-limit-for-metrics]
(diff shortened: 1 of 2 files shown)


scheduler/src/cook/components.clj
```diff
@@ -217,6 +217,15 @@
         (h req)
         (auth-fn req)))))
 
+(defn conditional-rate-limit
```
samincheva, line 220: I modeled this after `conditional-auth-bypass` (line 210 above). I debated making it more customizable, like maybe passing a dict of endpoint -> custom limit, but I decided that we can do that if we ever need to add another custom limit instead of fidgeting with it now.

laurameng: sounds good to me

laurameng (approved):

PR #2135, opened 2022-08-10
use a factory fn for creating (future) different types of pool handlers

ahaysx:
Previously, `create-datomic-scheduler` would call `make-offer-handler` for each pool, which setup an anonymous function on the trigger-chan for the pool. This anonymous function is the beginning of the scheduling cycle for a pool.

This PR inserts a new factory function between these two in order to setup for instantiating other types of pool handlers in the future (e.g. Kubernetes Scheduler pools). Currently, the factory only calls one new handler for all pools: `make-fenzo-handler`. This new function is essentially identical to the old `make-offer-handler` function, except for some variables such as `user->quota` moved one call higher as it can be imagined this would be a common variable with other types of handlers.

There are no TODOs with this change.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1429,82 +1429,87 @@
    mesos-run-as-user pool-name compute-clusters job->acceptable-compute-clusters-fn
    user->quota user->usage-future]
   (log-structured/info "Creating handler for Fenzo pool" {:pool pool-name})
-  (let [num-considerable @fenzo-num-considerable-atom
-        next-considerable
-        (try
-          (let [offers (tracing/with-span [s {:name "scheduler.offer-handler.generate-offers"
-                                              :tags {:pool pool-name :component tracing-component-tag}}]
-                         (apply concat (map (fn [compute-cluster]
-                                              (try
-                                                (cc/pending-offers compute-cluster pool-name)
-                                                (catch Throwable t
-                                                  (log-structured/error "Error getting pending offers"
-                                                                        {:pool pool-name
-                                                                         :compute-cluster (cc/compute-cluster-name compute-cluster)}
-                                                                        t)
-                                                  (list))))
-                                            compute-clusters)))
-                _ (tracing/with-span [s {:name "scheduler.offer-handler.cache-offers-for-rebalancer"
-                                         :tags {:pool pool-name :component tracing-component-tag}}]
-                    (doseq [offer offers
-                            :let [slave-id (-> offer :slave-id :value)]]
+  (timers/time!
+   (timers/timer (metric-title "match-jobs-event" pool-name))
+   (tracing/with-span [s {:name "scheduler.offer-handler.match-jobs" :tags {:pool pool-name :component tracing-component-tag}}]
+     (let [num-considerable @fenzo-num-considerable-atom
+           next-considerable
+           (try
+             (let [;; Try to clear the channel
+                   ;; Merge the pending offers from all compute clusters.
+                   offers (tracing/with-span [s {:name "scheduler.offer-handler.generate-offers"
+                                                 :tags {:pool pool-name :component tracing-component-tag}}]
+                            (apply concat (map (fn [compute-cluster]
+                                                 (try
+                                                   (cc/pending-offers compute-cluster pool-name)
+                                                   (catch Throwable t
+                                                     (log-structured/error "Error getting pending offers"
+                                                                           {:pool pool-name
+                                                                            :compute-cluster (cc/compute-cluster-name compute-cluster)}
+                                                                           t)
+                                                     (list))))
+                                               compute-clusters)))
+                   _ (tracing/with-span [s {:name "scheduler.offer-handler.cache-offers-for-rebalancer"
+                                            :tags {:pool pool-name :component tracing-component-tag}}]
+                       (doseq [offer offers
+                               :let [slave-id (-> offer :slave-id :value)]]
                       ; Cache offers for rebalancer so it can use job constraints when doing preemption decisions.
                       ; Computing get-offer-attr-map is pretty expensive because it includes calculating
                       ; currently running pods, so we have to union the set of pods k8s says are there and
                       ; the set of pods we're trying to put on the node. Even though it's not used by
                       ; rebalancer (and not needed). So it's OK if it's stale, so we do not need to refresh
                       ; and only store if it is a new node.
-                      (when-not (ccache/get-if-present agent-attributes-cache identity slave-id)
-                        (ccache/put-cache! agent-attributes-cache identity slave-id (offer/get-offer-attr-map offer)))))
-                user->usage (tracing/with-span [s {:name "scheduler.offer-handler.resolve-user-to-usage-future"
-                                                   :tags {:pool pool-name :component tracing-component-tag}}]
-                              @user->usage-future)
-                matched-head? (handle-resource-offers! conn fenzo-state pool-name->pending-jobs-atom
-                                                       mesos-run-as-user user->usage user->quota
-                                                       num-considerable offers
-                                                       rebalancer-reservation-atom pool-name compute-clusters
-                                                       job->acceptable-compute-clusters-fn)]
-            (when (seq offers)
-              (reset! resources-atom (view-incubating-offers fenzo)))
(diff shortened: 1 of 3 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -2132,6 +2009,116 @@
                 "schedule each pool less often than the desired setting of every " target-per-pool-match-interval-millis " ms."))
     (async/pipe (chime-ch (util/time-seq (time/now) (time/millis match-interval-millis))) match-trigger-chan)))
 
+(defn handle-fenzo-pool
```
ahaysx, line 2012: This should be identical to `make-offer-handler` except for some variables defined in the caller such as `user->quota`.

scrosby: Can you run the diff for what changed. (FYI: kdiff3 lets you do manual alignment.) One slight concern is changing the order of when we fetch from the DB can made race windows wider or narrower. That may be fine, but I want such changes to be intentional.

scrosby: Alternate approach. Move the implementation of handle-fenzo-pool up to where make-offer-handler used to be, that way the diff is much smaller and github will do the right thing.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-                                                             "Fenzo has seen only 1 job for" (counters/value iterations-at-fenzo-floor)
-                                                             "iterations, and still hasn't matched it.  Cook is now giving up and will "
-                                                             "now give Fenzo" max-considerable "jobs to look at.")
-                                                  {:pool pool-name
-                                                   :iterations-count (counters/value iterations-at-fenzo-floor)
-                                                   :number-max-considerable max-considerable})
-                            (meters/mark! fenzo-abandon-and-reset-meter)
-                            max-considerable)
-                          next-considerable))))))
-        (log-structured/info "Done with offer matching" {:pool pool-name}))
```
scrosby, line 1544: The analogous 'done with pool handler' is missing in the new code.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -2156,15 +2143,22 @@
...
-                                 (make-offer-handler
-                                   conn (get pool-name->fenzo-state name)
-                                   pool-name->pending-jobs-atom agent-attributes-cache
-                                   fenzo-max-jobs-considered fenzo-scaleback fenzo-floor-iterations-before-warn
-                                   fenzo-floor-iterations-before-reset (get pool->match-trigger-chan name)
-                                   rebalancer-reservation-atom mesos-run-as-user name
-                                   cluster-name->compute-cluster-atom job->acceptable-compute-clusters-fn)))]
+                              pools'
+                              (fn [{:keys [pool/name]}]
+                                (make-pool-handler
```
scrosby, line 2148: Would have liked to not have this reformat. E.g., in 'git gui' don't include the lines that are just reformats. Now its MUCH harder to determine what variables changed. It *looks* like nothing changed. True? 

Suggestion to turn off automatic reformatting.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-   mesos-run-as-user pool-name cluster-name->compute-cluster-atom job->acceptable-compute-clusters-fn]
-  (let [fenzo (:fenzo fenzo-state)
-        resources-atom (atom (view-incubating-offers fenzo))]
-    (reset! fenzo-num-considerable-atom max-considerable)
-    (tools/chime-at-ch
-      trigger-chan
-      (fn match-jobs-event []
-        (log-structured/info "Starting offer matching" {:pool pool-name})
-        (prom/with-duration
-          prom/scheduler-match-cycle-duration {:pool pool-name}
```
scrosby, line 1438: This should be resurrected in make-pool-handler.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-                            ;;       and the job reflects that change
-                            ;;  2. Once the above two items are addressed, user->usage should always correctly
-                            ;;     reflect *Cook*'s understanding of the state of the world at this point.
-                            ;;     When this happens, users should never exceed their quota
-                            user->usage-future (future (tracing/with-span [s1 {:from s :finish? false}] ; NOTE: finish? is set to false to prevent early finishing of the span
-                                                                          (generate-user-usage-map (d/db conn) pool-name)))
-                            ;; Try to clear the channel
-                            ;; Merge the pending offers from all compute clusters.
-                            compute-clusters (vals @cook.compute-cluster/cluster-name->compute-cluster-atom)
-                            offers (tracing/with-span
```
scrosby, line 1469: This should be resurrected in handle-fenzo-pool.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-    (tools/chime-at-ch
-      trigger-chan
-      (fn match-jobs-event []
-        (log-structured/info "Starting offer matching" {:pool pool-name})
-        (prom/with-duration
-          prom/scheduler-match-cycle-duration {:pool pool-name}
-          (timers/time!
-            (timers/timer (metric-title "match-jobs-event" pool-name))
-            (tracing/with-span
-              [s {:name "scheduler.offer-handler.match-jobs" :tags {:pool pool-name :component tracing-component-tag}}]
```
scrosby, line 1442: This should be resurrected in handle-fenzo-pool.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-              (let [num-considerable @fenzo-num-considerable-atom
-                    next-considerable
-                    (try
-                      (let [
-                            ;; There are implications to generating the user->usage here:
```
scrosby, line 1447: I think we want to salvage this comment to make-pool-handler

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1422,128 +1422,89 @@
...
-                            ;;       show up in the queue. However, it is important to know about
-                            ;;    b. Cook doesn't update the job state when cook hears from mesos about the
-                            ;;       state of an instance. Cook waits until it hears from datomic about the
-                            ;;       instance state change to change the state of the job. This means that it
-                            ;;       is possible to have large delays between when an instance changes status
-                            ;;       and the job reflects that change
-                            ;;  2. Once the above two items are addressed, user->usage should always correctly
-                            ;;     reflect *Cook*'s understanding of the state of the world at this point.
-                            ;;     When this happens, users should never exceed their quota
-                            user->usage-future (future (tracing/with-span [s1 {:from s :finish? false}] ; NOTE: finish? is set to false to prevent early finishing of the span
```
scrosby, line 1464: This tracing span is lost.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: I actually reformatted this myself because it is, personally, so hard to visually separate the 13 parameters to this function. But uh, I get it, the diff would have nicely highlighted the changes in line. I will change back but again, so hard to read this.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -2132,6 +2098,54 @@
...
+        resources-atom (atom (view-incubating-offers fenzo))
+        using-pools? (not (nil? (config/default-pool)))]
+    (reset! fenzo-num-considerable-atom fenzo-max-jobs-considered)
+    (tools/chime-at-ch
+     trigger-chan
+     (fn pool-schedule-event []
+       (log-structured/info "Starting pool handler" {:pool pool-name})
+       (prom/with-duration
+         prom/scheduler-match-cycle-duration {:pool pool-name}
+         (tracing/with-span [s {:name "scheduler.pool-handler" :tags {:pool pool-name :component tracing-component-tag}}]
```
ahaysx, line 2116: Spoke with Sam, they asked me to add a span here in addition to the span inside the Fenzo handler.

laurameng (approved):

PR #2149, opened 2022-08-16
Updating synthetic pods counters even if we don't autoscale in that match cycle

samincheva:
- Updates the logic to set the synthetic pods metric counters to happen even if that match cycle is not autoscaling.

This makes the metric more accurate when we're matching at 100%.

### DIFF
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -50,10 +50,7 @@
   (restore-offers [this pool-name offers]
     "Called when offers are not processed to ensure they're still available.")
 
-  (get-outstanding-synthetic-pods [this pool-name]
-    "Returns a list of outstanding synthetic pods in the pool.")
-
-  (set-synthetic-pods-counters [this pool-name num-synthetic-pods]
+  (set-synthetic-pods-counters [this pool-name]
     "Sets the counter metrics for number of and max synthetic pods.")
 
   (autoscaling? [this pool-name]
--- a/scheduler/src/cook/compute_cluster.clj
+++ b/scheduler/src/cook/compute_cluster.clj
@@ -50,6 +50,12 @@
   (restore-offers [this pool-name offers]
     "Called when offers are not processed to ensure they're still available.")
 
+  (get-outstanding-synthetic-pods [this pool-name]
+    "Returns a list of outstanding synthetic pods in the pool.")
+
+  (set-synthetic-pods-counters [this pool-name num-synthetic-pods]
+    "Sets the counter metrics for number of and max synthetic pods.")
+
   (autoscaling? [this pool-name]
     "Returns true if this compute cluster should autoscale the provided pool to satisfy pending jobs")
 
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -479,6 +479,22 @@
 
   (restore-offers [this pool-name offers])
 
+  (get-outstanding-synthetic-pods [this pool-name]
+    (->> (get-pods-in-pool this pool-name)
+         (add-starting-pods this)
+         (filter synthetic-pod->job-uuid)))
+
+  (set-synthetic-pods-counters [_ pool-name num-synthetic-pods]
+    (let [{:keys [max-pods-outstanding]} synthetic-pods-config]
+      (monitor/set-counter!
+        (counters/counter ["cook-k8s" "total-synthetic-pods" (str "compute-cluster-" name) (str "pool-" pool-name)])
+        num-synthetic-pods)
+      (monitor/set-counter!
+        (counters/counter ["cook-k8s" "max-total-synthetic-pods" (str "compute-cluster-" name) (str "pool-" pool-name)])
+        max-pods-outstanding)
+      (prom/set prom/total-synthetic-pods {:pool pool-name :compute-cluster name} num-synthetic-pods)
+      (prom/set prom/max-synthetic-pods {:pool pool-name :compute-cluster name} max-pods-outstanding)))
+
   (autoscaling? [_ pool-name]
     (and (-> synthetic-pods-config :pools (contains? pool-name))
          (= @state-atom :running)))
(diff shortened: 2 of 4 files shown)


scheduler/src/cook/compute_cluster.clj
```diff
@@ -50,6 +50,12 @@
   (restore-offers [this pool-name offers]
     "Called when offers are not processed to ensure they're still available.")
 
+  (get-outstanding-synthetic-pods [this pool-name]
+    "Returns a list of outstanding synthetic pods in the pool.")
+
+  (set-synthetic-pods-counters [this pool-name num-synthetic-pods]
```
scrosby, line 56: Is there something forcing this into the API? Can't the implementation be outside of the compute cluster; you can get the compute cluster name with `get-compute-cluster` and incorporate that.

scheduler/src/cook/compute_cluster.clj
```diff
@@ -50,6 +50,12 @@
   (restore-offers [this pool-name offers]
     "Called when offers are not processed to ensure they're still available.")
 
+  (get-outstanding-synthetic-pods [this pool-name]
```
scrosby, line 53: The only thing we do with this is count the result. So why not make this instead be `get-outstanding-synthetic-pod-count` and a comment that this exists only for metrics?

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -479,6 +479,22 @@
 
   (restore-offers [this pool-name offers])
 
+  (get-outstanding-synthetic-pods [this pool-name]
+    (->> (get-pods-in-pool this pool-name)
+         (add-starting-pods this)
+         (filter synthetic-pod->job-uuid)))
+
+  (set-synthetic-pods-counters [_ pool-name num-synthetic-pods]
```
scrosby, line 487: This would become
`(set-synthetic-pods-counters [cluster-name pool-name num-synthetic-pods]` and be outside of the compute cluster API.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
```diff
@@ -310,6 +310,10 @@
     (async/go
       (async/>! (pool->offers-chan pool-name) offers)))
 
+  (get-outstanding-synthetic-pods [_ _] nil)
```
scrosby, line 313: This placeholder is not type compatible with the other implementation, which returns a seq. In the new proposal, this returns 0.

scheduler/src/cook/compute_cluster.clj
samincheva: We don't just count the results, it is also used on line 554.

scheduler/src/cook/compute_cluster.clj
samincheva: I tried that, but because the autoscaling code in the scheduler uses the compute cluster API and doesn't actually know if the compute cluster is Kubernetes or Mesos, separating this function out would break the abstraction.

scheduler/src/cook/kubernetes/compute_cluster.clj
samincheva: Same here as previous comment, I think this would break the abstraction we maintain in scheduler.clj that is agnostic to kubernetes vs mesos.

scheduler/src/cook/mesos/mesos_compute_cluster.clj
samincheva: I looked up whether it's better to return an empty seq or nil, and I found some suggestion that nil is better. I can update it to return an empty sequence. Would `(sequence ())` be the best approach in that case?

scheduler/src/cook/compute_cluster.clj
scrosby: Can you double-check the line number?

scheduler/src/cook/compute_cluster.clj
scrosby: Aaah. Yes, I see. This code needs `max-total-synthetic-pods` which comes from `scheduler-config` which is not visible outside. (And thats worth a comment)

Then new proposal, why fetch total synthetic pods and then immediately feed it back? Why not replace both functions with something like 
`(log-synthetic-pod-metrics [this pool-name])` thats a noop for k8s?

scheduler/src/cook/mesos/mesos_compute_cluster.clj
scrosby: `(list)`  but see my second proposal.

scheduler/src/cook/compute_cluster.clj
samincheva: I did, it is used in this chunk of code:
```
                (let [using-pools? (config/default-pool)
                      synthetic-task-pool-name (when using-pools? pool-name)
                      new-jobs (remove (fn [{:keys [job/uuid]}]
                                         (some #(= (str uuid) (synthetic-pod->job-uuid %))
                                               outstanding-synthetic-pods))
```

scheduler/src/cook/compute_cluster.clj
samincheva: I did it this way because we already need to calculate the outstanding synthetic pods for the `autoscale!` function, and I didn't want to replicate the calculation in the case when we do autoscale. If you think repeating that work is fine, I can update it to be a self-contained function that does all its calculations like you suggest.

scheduler/src/cook/compute_cluster.clj
samincheva: Here's the code on default: https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/kubernetes/compute_cluster.clj#L555 (I can't seem to link to a line within the changed MR version)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1076,19 +1076,24 @@
...
+                                              job->acceptable-compute-clusters-fn)]
+                  (log-structured/info "Starting autoscaling" {:pool pool-name})
+                  (->> compute-cluster->jobs
+                       (map
+                         (fn [[compute-cluster jobs-for-cluster]]
+                           (future (cc/autoscale! compute-cluster pool-name jobs-for-cluster adjust-job-resources-for-pool-fn))))
+                       doall
+                       (run! deref)))
+                (log-structured/info "Done autoscaling" {:pool pool-name}))
+              (do
```
scrosby, line 1093: Why the `do`?

samincheva: Leftover from when we were calling the function to get outstanding pods explicitly. I removed it.

PR #2150, opened 2022-08-17
Initial implementation for submitting jobs directly to Kubernetes Scheduler

ahaysx:
## Kubernetes Bin Packing
Configure pools to submit real job pods directly to Kubernetes, which handles the scheduling and cluster autoscaling. Pools that use the Kubernetes Scheduler will not use the "match-offer" loop, Fenzo bin packing algorithm, nor synthetic pods.

This is a proof of concept that will require a few iterations to get to a production state. However, my internal tests shows this feature behaves well. The following is my current understanding of the gaps between this changeset and a version we can test with real jobs. I'm sure there are more, please let me know and I can document it here.

### Configuration
#2139 introduced configuring schedulers per pool. Here is an example of running the Kubernetes Scheduler feature in one pool and Fenzo in all others.

```
:pools {
  ...
  :schedulers [{:pool-regex "^k8s-exp$"
                        :scheduler-config {:scheduler "kubernetes"
                                           :max-jobs-considered 1000}}
                       {:pool-regex ".*"
                        :scheduler-config {:scheduler "fenzo"
                                           :good-enough-fitness 1.0
                                           :fenzo-fitness-calculator "com.netflix.fenzo.plugins.BinPackingFitnessCalculators/cpuMemBinPacker"
                                           :fenzo-max-jobs-considered 400
                                           :fenzo-scaleback 1
                                           :fenzo-floor-iterations-before-warn 10
                                           :fenzo-floor-iterations-before-reset 1000}}]
}
```

### Remaining Work
* Real job pods now trigger autoscaling. In both Fenzo and Kubernetes Scheduler pools, the instance runtime begins when the pod is submitted to Kubernetes. This is usually fine in the Fenzo case, because a pod is only launched when there is available space for it on a node. In this case, runtime is pretty close to when the pod is actually in the Running state (although it does count initialization). However, for the Kubernetes Scheduler pools, a pod could trigger autoscaling, leading to 5 to 10 minutes of inaccurate runtime. We should consider starting instance runtime in both cases when the pod is actually running on the node.
* A backpressure mechanism remains, in order to moderate the amount of pending pods in each GKE cluster.
* Proper metrics accounting in the new Kubernetes handler and code. This functionality has been a bit of a moving target because of Sam's great work on it over the past quarter. I envision there will be a round of review and edits for this specifically at a later date.
* Audit values of the instance saved to Datomic. For example, do we need to set `ports`, `slave-id`, etc?


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+     (cond->
+      {:db/id (d/tempid :db.part/user)
+       :job/_instance job-ref
+       :instance/executor executor
+       :instance/executor-id task-id ; NB command executor uses the task-id as the executor-id 
+       :instance/hostname "Unknown"
+       :instance/ports 9876
+       :instance/preempted? false
+       :instance/progress 0
+       :instance/slave-id "Unknown"
```
scrosby, line 1558: Why not leave this empty and unspecified?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+    [[:job/allowed-to-start? job-ref]
+     ; The job will remain waiting until the instance is running.
+     [:db/add job-ref :job/state :job.state/waiting]
+     (cond->
+      {:db/id (d/tempid :db.part/user)
+       :job/_instance job-ref
+       :instance/executor executor
+       :instance/executor-id task-id ; NB command executor uses the task-id as the executor-id 
+       :instance/hostname "Unknown"
+       :instance/ports 9876
```
scrosby, line 1555: Why is this here?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+  "Generate task transaction for a job."
+  [job metadata compute-cluster]
+  (let [{:keys [job/last-waiting-start-time job/uuid]} job
+        {:keys [executor task-id]} metadata
+        job-ref [:job/uuid uuid]
+        instance-start-time (now)]
+    [[:job/allowed-to-start? job-ref]
+     ; The job will remain waiting until the instance is running.
+     [:db/add job-ref :job/state :job.state/waiting]
+     (cond->
```
scrosby, line 1549: Overall, this code feels like a copy&paste from `matches->task-txns` Can we use that function, or refactor a common helper function that this and `matches->task-txns` uses?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+  "Generate sequence of task-metadata for jobs in a compute-cluster, suitable for Kubernetes Scheduler."
+  [jobs pool-name mesos-run-as-user compute-cluster]
+  (let [jobs->task-id (plumbing.core/map-from-keys (fn [_] (str (d/squuid))) jobs)]
+    (for [{:keys [job/name job/user job/uuid job/environment] :as job} jobs]
+      (let [pool-specific-resources
+            ((adjust-job-resources-for-pool-fn pool-name) job (tools/job-ent->resources job))]
+        (merge
+         (task/job->task-metadata compute-cluster mesos-run-as-user
+                                  job (jobs->task-id job))
+         {; NB: see `kubernetes/compute_cluster/autoscale!`
```
scrosby, line 1578: What does this mean?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+          ; compares to launching a synthetic pod.
+
+          ; Job constraints need to be expressed on a real job
+          ; pod so that if it triggers the cluster autoscaler,
+          ; CA will spin up nodes that will end up satisfying it. 
+          :pod-constraints (constraints/job->constraints job)
+
+          ; We need to label the pods handled completely by Kubernetes
+          ; so that we can treat them differently throughout the scheduler.
+          :pod-labels {"twosigma.com/scheduler" "kubernetes"}
```
scrosby, line 1589: I'd consider a `cook` prefix here instead of just `twosigma.com`. Also, can this be in a `def`?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -711,6 +721,10 @@
            old-file-server-state (:sandbox-file-server-container-state old-state)]
        (when (and (= new-file-server-state :running) (not= old-file-server-state :running))
          (record-sandbox-url pod-name new-state)))
+     ; Hostname will change for pods scheduled by Kubernetes.
+     (when-not (= (api/pod->node-name (:pod new-state))
```
scrosby, line 725: Should this be gated on "twosigma.com/scheduler" "kubernetes"? (FYI: This is code that's live immediately)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+
+          ; Cook has a "novel host constraint", which disallows a job from
+          ; running on the same host twice. So, we need to avoid running a
+          ; real job pod on any of the hosts that it won't be able to run on.
+          :pod-hostnames-to-avoid (constraints/job->previous-hosts-to-avoid job)
+          :task-request {:scalar-requests (walk/stringify-keys pool-specific-resources)
+                         :job {:job/pool {:pool/name pool-name}
```
scrosby, line 1596: Why rebuild this from the components? Can the `job` be used directly, instead of destructuring and rebuilding a new one? (And if not, can you explain why in a comment?)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+  (try
+    (let [user->usage (tracing/with-span [s {:name "scheduler.kubernetes-handler.resolve-user-to-usage-future"
+                                             :tags {:pool pool-name :component tracing-component-tag}}]
+                        @user->usage-future)
+          {max-considerable :max-jobs-considered} scheduler-config
+          db (db conn)
+          pending-jobs (get @pool-name->pending-jobs-atom pool-name)
+          ;; We need to filter pending jobs based on quota so that we don't
+          ;; submit beyond what users have quota to actually run.
+          jobs (prom/with-duration
```
scrosby, line 1618: Consider refactoring this code calculating `jobs` out as it's shared with existing scheduler.... Or maybe, move the timer code into `pending-jobs->considerable-jobs`

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+          job-uuids (set (map :job/uuid jobs))]
+      (log-structured/info "Considering jobs" {:pool pool-name :number-considered-jobs (count job-uuids)})
+      (if (seq jobs)
+        (do
+          (swap! pool-name->pending-jobs-atom
+                 remove-matched-jobs-from-pending-jobs
+                 job-uuids pool-name)
+          (log-structured/debug (print-str "Updated pool-name->pending-jobs-atom:" (get @pool-name->pending-jobs-atom pool-name))
+                                {:pool pool-name})
+          (let [autoscaling-compute-clusters (filter #(cc/autoscaling? % pool-name) compute-clusters)
```
scrosby, line 1633: This block of code has a high overlap with a block inside of `trigger-autoscaling!` can you refactor out the common components?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+                                       jobs pool-name autoscaling-compute-clusters
+                                       job->acceptable-compute-clusters-fn)]
+            (log-structured/info "Starting launch directly to Kubernetes" {:pool pool-name})
+            (->> compute-cluster->jobs
+                 (map
+                  (fn [[compute-cluster jobs]]
+                    (let [kill-lock-object (cc/kill-lock-object compute-cluster)
+                          task-metadata-seq (jobs->kubernetes-task-metadata jobs pool-name mesos-run-as-user compute-cluster)
+                          task-txns (map (fn [job metadata] (job->task-txn job metadata compute-cluster)) jobs task-metadata-seq)]
+                      (try
```
scrosby, line 1646: The overall shape of this is very close to `launch-tasks-for-cluster` with the token spending code inlined in from `launch-matches!`.

Can you refactor the common components out and call from here and existing place?

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -250,7 +251,7 @@
 
 (defn container-status->failure-reason
   "Maps kubernetes failure reasons to cook failure reasons"
-  [{:keys [name]} instance-id ^V1PodStatus pod-status ^V1ContainerStatus container-status]
+  [{:keys [name]} instance-id ^V1PodStatus pod-status ^V1Pod pod ^V1ContainerStatus container-status]
```
scrosby, line 254: Can strip out the `instance-id` argument and extract it from the `pod` argument

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -280,7 +281,7 @@
         (log/info "In compute cluster" name ", encountered OutOfcpu pod status reason for" instance-id)
         :reason-invalid-offers)
 
-      (api/pod-unschedulable? instance-id pod-status)
+      (api/pod-unschedulable? instance-id pod-status pod)
```
scrosby, line 284: Worth a comment that this needs to know the pod because we have different timeouts for different depending on if we binpack via fenzo or if k8s binpacks, with a TODO that we won't need this when we're out of this hybrid state.

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1815,26 +1820,27 @@
...
   "Returns true if the given pod status has a PodScheduled
   condition with status False and reason Unschedulable"
-  [pod-name ^V1PodStatus pod-status]
+  [pod-name ^V1PodStatus pod-status ^V1Pod pod]
   (let [{:keys [pod-condition-unschedulable-seconds
                 synthetic-pod-condition-unschedulable-seconds]}
         (config/kubernetes)
         unschedulable-seconds
-        (if (some-> pod-name synthetic-pod?)
+        (if (or (some-> pod-name synthetic-pod?)
```
scrosby, line 1828: I think we should split this up into two cases:
* `synthetic pod unschedulable seconds`
* `k8s-packed-pod-unschedulable-seconds`

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -420,6 +421,15 @@
...
       (when sandbox-url (scheduler/write-sandbox-url-to-datomic datomic/conn task-id sandbox-url)))))
 
+(defn record-hostname
+  "Record the pod's hostname in Datomic."
+  [pod-name {:keys [^V1Pod pod]}]
+  (when (api/kubernetes-scheduler-pod? pod) 
+    (let [task-id (-> pod .getMetadata .getName)
+          hostname (api/pod->node-name pod)]
+      (log-structured/debug "Recording hostname for pod" {:pod-name pod-name :hostname hostname})
+      (when hostname (scheduler/write-hostname-to-datomic datomic/conn task-id hostname)))))
```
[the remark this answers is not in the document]
ahaysx: It can be nil before being placed on the node and after exiting. I think it's good to gate this write to only non-nil values.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -328,6 +328,26 @@
     (catch Exception e
       (log/error e "Sandbox file server URL update error"))))
 
+(defn write-hostname-to-datomic
+  "Takes a hostname for an instance and saves it to datomic."
+  [conn task-id hostname]
+  (try
+    (let [db (db conn)
+          _ (when-not task-id
```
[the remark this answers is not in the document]
ahaysx: Take a look at `write-sandbox-url-to-datomic` for the model code. I will restructure to make task-id log error and exit?

scheduler/src/cook/kubernetes/api.clj
ahaysx: I originally had this as a separate configuration. However, once we moved scheduler config to be per pool, it became much more involved to get this setting at this level of the code. I would need to get the pool for the pod and use the pool regex matcher to extract the setting. Would you be ok with this being globally configured on `:kubernetes`?

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Added the prefix. 

For now, I don't want to make this a def because it is accessed both here and in `cook.kubernetes.api`, which cannot be imported due to a cycle. This speaks to the eventual TODO here to move this task-metadata generation into probably `cook.kubernetes.task`. This would require refactoring `task/job->task-metadata`. I may be able to do that refactoring in this PR, but I need your help understanding some of the nuance there. I'm leaning towards adding this to list of cleanup tasks for post production experiment.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Good catch. I was providing just enough to run but was missing providing `job/application` as an example.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1516,6 +1536,152 @@
...
+        instance-start-time (now)]
+    [[:job/allowed-to-start? job-ref]
+     ; The job will remain waiting until the instance is running.
+     [:db/add job-ref :job/state :job.state/waiting]
+     (cond->
+      {:db/id (d/tempid :db.part/user)
+       :job/_instance job-ref
+       :instance/executor executor
+       :instance/executor-id task-id ; NB command executor uses the task-id as the executor-id 
+       :instance/hostname "Unknown"
```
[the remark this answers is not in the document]
ahaysx: This is remnant from hacking it to just work. I've removed what you think are not required and will test internally again.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Yes, there's quite a bit of overlap here in this and other parts to get it to work. Happy to discuss refactoring with you tomorrow? I want us to consider how much is useful to do now vs after the production test.

scheduler/src/cook/kubernetes/controller.clj
ahaysx: Done, good idea. Moved the gate to this clause to make it extra unlikely to affect live code.

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
ahaysx: Yeah, there is some dependency on a few of these being non-nil

```
Response validation failed: [{:instances [{:hostname (not (instance? java.lang.String nil)), :slave_id (not (instance? java.lang.String nil)), :agent_id (not (instance? java.lang.String nil))}]}] schema.utils.ErrorContainer@d26b59ff
	at compojure.api.coerce$coerce_response_BANG_.invokeStatic(coerce.clj:40)
	at compojure.api.coerce$coerce_response_BANG_.invoke(coerce.clj:31)
	at compojure.api.resource$coerce_response.invokeStatic(resource.clj:39)
	at compojure.api.resource$coerce_response.invoke(resource.clj:38)
	at compojure.api.resource$create_handler$fn__18174.invoke(resource.clj:59)
	at compojure.api.routes.Route.invoke(routes.clj:74)
	at compojure.core$routing$fn__9841.invoke(core.clj:144)
	at clojure.core$some.invokeStatic(core.clj:2705)
	at clojure.core$some.invoke(core.clj:2696)
	at compojure.core$routing.invokeStatic(core.clj:144)
	at compojure.core$routing.doInvoke(core.clj:141)
	at clojure.lang.RestFn.applyTo(RestFn.java:139)
	at clojure.core$apply.invokeStatic(core.clj:669)
	at clojure.core$apply.invoke(core.clj:662)
	at compojure.core$routes$fn__9845.invoke(core.clj:149)
	at compojure.core$routing$fn__9841.invoke(core.clj:144)
	at clojure.core$some.invokeStatic(core.clj:2705)
	at clojure.core$some.invoke(core.clj:2696)
	at compojure.core$routing.invokeStatic(core.clj:144)
	at compojure.core$routing.doInvoke(core.clj:141)
	at clojure.lang.RestFn.invoke(RestFn.java:423)
	at cook.rest.api$main_handler$fn__20411.invoke(api.clj:3633)```

File: scheduler/src/cook/kubernetes/api.clj [hunk 17]
```diff
@@ -1815,26 +1820,27 @@
 (defn pod-unschedulable?
   "Returns true if the given pod status has a PodScheduled
   condition with status False and reason Unschedulable"
-  [pod-name ^V1PodStatus pod-status]
+  [pod-name ^V1PodStatus pod-status ^V1Pod pod]
```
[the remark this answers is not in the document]
ahaysx: Done

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -420,6 +421,15 @@
                           nil))]
       (when sandbox-url (scheduler/write-sandbox-url-to-datomic datomic/conn task-id sandbox-url)))))
 
+(defn record-hostname
+  "Record the pod's hostname in Datomic."
+  [pod-name {:keys [^V1Pod pod]}]
```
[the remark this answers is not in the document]
ahaysx: Done

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -420,6 +421,15 @@
                           nil))]
       (when sandbox-url (scheduler/write-sandbox-url-to-datomic datomic/conn task-id sandbox-url)))))
 
+(defn record-hostname
+  "Record the pod's hostname in Datomic."
```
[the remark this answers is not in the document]
ahaysx: Done, should be more descriptive now.

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: This is the one refactor I totally agree needs to happen and would be good, but is a bit complex. I will add it to the "master list of TODOs" I'm developing for you tomorrow.

scheduler/src/cook/kubernetes/controller.clj
ahaysx: Done

scheduler/src/cook/kubernetes/controller.clj
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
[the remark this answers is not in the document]
ahaysx: Added optional as discussed. Will test tomorrow on studio54.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Removed

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/rest/api.clj
```diff
@@ -1190,13 +1191,13 @@
...
                  :compute-cluster (fetch-compute-cluster-map db (:instance/compute-cluster instance))
                  :executor_id (:instance/executor-id instance)
-                 :hostname hostname
                  :ports (vec (sort (:instance/ports instance)))
                  :preempted (:instance/preempted? instance false)
-                 :agent_id (:instance/slave-id instance)
-                 :slave_id (:instance/slave-id instance)
                  :status (name (:instance/status instance))
                  :task_id task-id}
+          hostname (assoc :hostname hostname)
```
scrosby, line 1198: Actually. Why not subsitute in the fake data in here, instead of in the database?

I.e., in the cond-> above, we unconditionally include the hostname, but with the code:

> `   :hostname (or hostname "not-binpacked-hostname")`

> The cs cli requires hostname, slave-id, and agent-id  to be set. For just-launched pods on k8s where we are letting k8s to the binpacking, they won't be available. Thus we substitute fake numbers temporarily.

With a TODO: to fix the CLI. 

This way we also don't need to store garbage data into the database. (And we can delete the code that does that.)

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -328,6 +328,27 @@
...
       (log/error e "Sandbox file server URL update error"))))
 
+(defn write-hostname-to-datomic
+  "Takes a hostname for an instance and saves it to datomic."
+  [conn task-id hostname] 
+  (log-structured/debug "Writing hostname" {:task-id task-id :hostname hostname})
+  (if task-id
+    (try
+      (let [db (db conn)
+            [instance] (first (q '[:find ?i
```
scrosby, line 338: I think this can be replaced by:

`(d/entity db [:instance/task-id task-id])`

See https://docs.datomic.com/on-prem/schema/identity.html#lookup-refs
and https://docs.datomic.com/on-prem/clojure/index.html#datomic.api/entity

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1062,6 +1086,23 @@
                               :first-ten-jobs (print-str (->> jobs (take 10) (map :job/uuid) (map str)))}))
       (dissoc compute-cluster->jobs :no-acceptable-compute-cluster))))
 
+(defn distribute-and-handle-jobs
+  "Distributes jobs across compute clusters and applies a given handler for each distribution."
```
scrosby, line 1090: Can you expand 'applies a given handler for each distribution' with something less abstract:

E.g., 

> This function takes a collection of jobs, then applies various logic (driven by configuration and pool name and other logic) to determine which compute cluster should run which job. It then launches the jobs across compute clusters in parallel. 
> A function `launch-distributed-job-fn` is used to actually do the launches. This is needed to implement the different launch logic for letting k8s pack or for launching synthetic pods for autoscaling.

I'd also suggest naming this 'distribute and launch jobs'.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1062,6 +1086,23 @@
...
+  [jobs pool-name compute-clusters job->acceptable-compute-clusters-fn handle-jobs-for-cluster-fn]
+  (try
+    (when (seq jobs)
+      (log-structured/info "Preparing for job distribution" {:pool pool-name})
+      (let [compute-cluster->jobs (distribute-jobs-to-compute-clusters jobs pool-name compute-clusters
+                                                                       job->acceptable-compute-clusters-fn)]
+        (log-structured/info "Starting job distribution" {:pool pool-name})
+        (->> compute-cluster->jobs
+             (map handle-jobs-for-cluster-fn)
```
scrosby, line 1099: The `future` call should happen here, because this code is what derefs the future. (I.e., right now, this code is broken *if* `launch-distributed-job-fn` doesn't return a future.)

scheduler/src/cook/kubernetes/api.clj
```diff
@@ -1816,26 +1821,27 @@
 (defn pod-unschedulable?
   "Returns true if the given pod status has a PodScheduled
   condition with status False and reason Unschedulable"
-  [pod-name ^V1PodStatus pod-status]
+  [^V1PodStatus pod-status ^V1Pod pod]
```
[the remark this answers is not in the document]
ahaysx: Refactoring this would require refactoring testutils/pod-helper, which is a large task. I will add this as a followup TODO.

scheduler/src/cook/rest/api.clj
```diff
@@ -1190,13 +1191,13 @@
...
+          slave-id (assoc :slave_id slave-id)
+          slave-id (assoc :agent_id slave-id)
```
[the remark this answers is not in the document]
ahaysx: This isn't a typo. Agent id is set to slave-id.

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -250,9 +251,10 @@
 
 (defn container-status->failure-reason
   "Maps kubernetes failure reasons to cook failure reasons"
-  [{:keys [name]} instance-id ^V1PodStatus pod-status ^V1ContainerStatus container-status]
+  [{:keys [name]} ^V1PodStatus pod-status ^V1Pod pod ^V1ContainerStatus container-status]
```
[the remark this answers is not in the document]
ahaysx: Refactoring this would require refactoring testutils/pod-helper, which is a large task. I will add this as a followup TODO.

scheduler/src/cook/rest/api.clj
ahaysx: I will add this as a follow-up.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: I will add this is a follow-up in addition to applying it to the above function for writing the sandbox URL.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: How does this look?
```
(map (fn [launch-args] (future (launch-distributed-job-fn launch-args))))
```

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Updated, lmk

scheduler/src/cook/kubernetes/controller.clj
```diff
@@ -420,6 +425,16 @@
                           nil))]
       (when sandbox-url (scheduler/write-sandbox-url-to-datomic datomic/conn task-id sandbox-url)))))
 
+(defn record-hostname
+  "Record the hostname in Datomic for the instance. This is necessary for pods
+   submitted directly to Kubernetes without node selection. In this case, the
```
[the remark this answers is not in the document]
ahaysx: Done

scheduler/src/cook/kubernetes/api.clj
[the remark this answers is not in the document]
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1062,6 +1086,28 @@
...
+   autoscaling or for using the Kubernetes Scheduler directly."
+  [jobs pool-name compute-clusters job->acceptable-compute-clusters-fn launch-distributed-job-fn]
+  (try
+    (when (seq jobs)
+      (log-structured/info "Preparing for job distribution" {:pool pool-name})
+      (let [compute-cluster->jobs (distribute-jobs-to-compute-clusters jobs pool-name compute-clusters
+                                                                       job->acceptable-compute-clusters-fn)]
+        (log-structured/info "Starting job distribution" {:pool pool-name})
+        (->> compute-cluster->jobs
+             (map (fn [launch-args] (future (launch-distributed-job-fn launch-args))))
```
scrosby, line 1104: This can be an anonyomous function with the syntax:   `#(future (launch-distributed-jobs-fn %))`

PR #2161, opened 2022-08-24
Launch tasks similarly for both Fenzo and K8s Scheduler pools

ahaysx:

- Refactor `launch-tasks-for-cluster` to handle launching for Fenzo and Kubernetes Scheduler pools. This required some movement of metrics handling and a boolean gate for Fenzo pools.
- Refactors Kubernetes Scheduler pools to use this function for launching, by providing its incomplete match objects (just the metadata) and db transactions.

- Follow-up feedback from #2150

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -871,6 +871,9 @@
             (ratelimit/spend! quota/per-user-per-pool-launch-rate-limiter token-key 1)
             (ratelimit/spend! compute-cluster-launch-rate-limiter ratelimit/compute-cluster-launch-rate-limiter-key 1))
           (prom/inc prom/scheduler-jobs-launched {:pool pool-name :compute-cluster (cc/compute-cluster-name compute-cluster)})
+          ;; When Fenzo is specified, we need to tell it that the task was
+          ;; scheduled. Fenzo can be nil when launching tasks for a Kubernetes
+          ;; Scheduler pool.
           (when fenzo
             (locking fenzo
               (-> fenzo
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -871,10 +871,11 @@
             (ratelimit/spend! quota/per-user-per-pool-launch-rate-limiter token-key 1)
             (ratelimit/spend! compute-cluster-launch-rate-limiter ratelimit/compute-cluster-launch-rate-limiter-key 1))
           (prom/inc prom/scheduler-jobs-launched {:pool pool-name :compute-cluster (cc/compute-cluster-name compute-cluster)})
-          (locking fenzo
-            (-> fenzo
-                (.getTaskAssigner)
-                (.call task-request hostname)))))
+          (when fenzo
```
scrosby, line 874: Can you add a comment to the function above saying that `fenzo` can be null (and when).

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -996,20 +999,22 @@
...
             (prom/with-duration
               prom/scheduler-launch-all-matched-tasks-submit-duration {:pool pool-name}
               (timers/time!
                 (timers/timer (metric-title "handle-resource-offer!-mesos-submit-duration" pool-name))
                 (let [_ (log-structured/info "Launching matched tasks for compute cluster"
                                              {:pool pool-name :compute-cluster compute-cluster-name})]
-                  (doseq [match matches]
-                    (timers/stop (-> match :leases first :offer :offer-match-timer))
-                    (-> match :leases first :offer :offer-match-timer-prom-stop-fn))
+                  ;; Offers must be launched within a timeout, otherwise they expire worthless.
```
scrosby, line 1011: "**Mesos** offers must be launched .... only applicable to **Mesos tasks on Fenzo** ..... "

"When Mesos goes away, this metric ceases to be important and this can be deleted."

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Done

PR #2162, opened 2022-08-24
Use backpressure of scheduling pods to moderate launching new pods for real jobs

ahaysx:

- The Kenzo loop will now account for available scheduling capacity before "assigning" a job to an acceptable compute cluster. This is done via  `pos?` filter in the distribution function.
- Kubernetes compute clusters now report their "max launchable" stat, which is a measurement of pods without assigned nodes. This is our best approximation for pressure inside the Kubernetes Scheduler.
- The Kenzo handler will not try scheduling more jobs than total available scheduling capacity across all compute clusters. This reduces expensive work for jobs that would be filtered out at distribution stage.
- The Kenzo handler will short-circuit and then delay the next scheduling loop if there isn't sufficient scheduling capacity (configurable). This will prevent incurring expensive cycle overhead for a small number of considerable jobs.

- This is a backpressure mechanism needed to efficiently use the Kubernetes Scheduler for real job pods.

## Newly Identified Future Work
- Move synthetic pod config on cc-template to new "autoscaling" block, in addition to max-outstanding, etc.

### DIFF
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -378,6 +378,15 @@
      (prom/set prom/total-synthetic-pods {:pool pool-name :compute-cluster name} num-synthetic-pods)
      (prom/set prom/max-synthetic-pods {:pool pool-name :compute-cluster name} max-pods-outstanding))))
 
+(defn get-scheduling-pods
+  "Get the pods that are still being scheduled by Kubernetes. It consider pods
+   not yet bound to a node when filtering for this state. It does not consider 
+   pods with 'Pending' status, since this phase includes time spent pulling images
+   onto the host."
+  [compute-cluster pool-name]
+  (->> (get-pods-in-pool compute-cluster pool-name)
+       (add-starting-pods compute-cluster)
+       (remove api/pod->node-name)))
 
 (defrecord KubernetesComputeCluster [^ApiClient api-client name entity-id exit-code-syncer-state
                                      all-pods-atom current-nodes-atom pool->node-name->node
@@ -522,6 +531,39 @@
   (autoscaling? [_ pool-name]
     (and (-> synthetic-pods-config :pools (contains? pool-name))
          (= @state-atom :running)))
+  
+  (max-launchable
+   [this pool-name]
+   (let [scheduling-pods (get-scheduling-pods this pool-name)
+         num-scheduling-pods (count scheduling-pods)
+         total-pods (-> @all-pods-atom keys count)
+         total-nodes (-> @current-nodes-atom keys count)
+         ; NOTE: the synthetic pods config is co-opted here to determine the max outstanding
+         ; pods in the compute cluster. This configuration, set on the compute cluster
+         ; template, will be refactored in the future to more generally define autoscaling 
+         ; parameters once synthetic pods are decommissioned.
+         {:keys [max-pods-outstanding max-total-pods max-total-nodes]
+          :or {max-total-pods 32000 max-total-nodes 1000}} synthetic-pods-config]
+
+     (when (>= total-pods max-total-pods)
+       (log-structured/warn "Total pods are maxed out"
+                            {:compute-cluster name
+                             :number-max-total-pods max-total-pods
+                             :number-total-pods total-pods}))
+     (when (>= total-nodes max-total-nodes)
+       (log-structured/warn "Nodes are maxed out"
+                            {:compute-cluster name
+                             :number-max-total-nodes max-total-nodes
+                             :number-total-nodes total-nodes}))
+     (when (>= num-scheduling-pods max-pods-outstanding)
+       (log-structured/warn "Outstanding pods are maxed out"
+                            {:compute-cluster name
+                             :number-max-synthetic-pods max-pods-outstanding
+                             :number-outstanding-pods num-scheduling-pods}))
+
+     (min (- max-pods-outstanding num-scheduling-pods)
+          (- max-total-nodes total-nodes)
+          (- max-total-pods total-pods))))
 
   (autoscale! [this pool-name jobs adjust-job-resources-for-pool-fn]
     (tracing/with-span [s {:name "k8s.autoscale" :tags {:compute-cluster name :pool pool-name :component tracing-component-tag}}]
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -379,11 +379,20 @@
      (prom/set prom/max-synthetic-pods {:pool pool-name :compute-cluster name} max-pods-outstanding))))
 
 (defn get-scheduling-pods
-  "Get the pods that are still being scheduled by Kubernetes. We consider pods
-   not yet bound to a node when filtering for this state. Pods with 'Pending' 
-   status are not considered, since this phase includes time spent pulling images
-   onto the host. Depending on that status would not acurately reflect the pressure
-   inside the Kubernetes Scheduler."
+  "Get the pods that are still being scheduled by Kubernetes. This is 
+   done by filtering pods not yet bound to a node from the set of 
+   pods Cook thinks is running.
+
+   This is particularly useful when utilizing the Kubernetes Scheduler
+   to bin pack jobs. Pressure inside the scheduler is measured and then
+   reduced by only having so many outstanding, unscheduled pods at any 
+   given time.
+
+   Filtering for node assignments is the best approximation of this 
+   pressure. Depending on pods with 'Pending' status is problematic. 
+   It includes time spent pulling images onto nodes, after they have 
+   been assigned by the scheduler. This status does not accurately 
+   reflect the pressure inside the Kubernetes Scheduler."
   [compute-cluster pool-name]
   (->> (get-pods-in-pool compute-cluster pool-name)
        (add-starting-pods compute-cluster)
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -552,8 +552,7 @@
          ; pods in the compute cluster. This configuration, set on the compute cluster
          ; template, will be refactored in the future to more generally define autoscaling 
          ; parameters once synthetic pods are decommissioned.
-         {:keys [max-pods-outstanding max-total-pods max-total-nodes]
-          :or {max-total-pods 32000 max-total-nodes 1000}} synthetic-pods-config]
+         {:keys [max-pods-outstanding max-total-pods max-total-nodes]} synthetic-pods-config]
 
      (when (>= total-pods max-total-pods)
        (log-structured/warn "Total pods are maxed out"
@@ -858,10 +857,12 @@
   [compute-cluster-name synthetic-pods-config]
   (when synthetic-pods-config
     (when-not (and
-                (-> synthetic-pods-config :image count pos?)
-                (-> synthetic-pods-config :max-pods-outstanding pos?)
-                (-> synthetic-pods-config :pools set?)
-                (-> synthetic-pods-config :pools count pos?))
+               (-> synthetic-pods-config :image count pos?)
+               (-> synthetic-pods-config :max-pods-outstanding pos?)
+               (-> synthetic-pods-config :max-total-pods pos?)
+               (-> synthetic-pods-config :max-total-nodes pos?)
+               (-> synthetic-pods-config :pools set?)
+               (-> synthetic-pods-config :pools count pos?))
       (throw (ex-info (str "In " compute-cluster-name " compute cluster, invalid synthetic pods config")
                       synthetic-pods-config)))))
 
--- a/scheduler/src/cook/kubernetes/compute_cluster.clj
+++ b/scheduler/src/cook/kubernetes/compute_cluster.clj
@@ -379,10 +379,11 @@
      (prom/set prom/max-synthetic-pods {:pool pool-name :compute-cluster name} max-pods-outstanding))))
 
 (defn get-scheduling-pods
-  "Get the pods that are still being scheduled by Kubernetes. It consider pods
-   not yet bound to a node when filtering for this state. It does not consider 
-   pods with 'Pending' status, since this phase includes time spent pulling images
-   onto the host."
+  "Get the pods that are still being scheduled by Kubernetes. We consider pods
+   not yet bound to a node when filtering for this state. Pods with 'Pending' 
+   status are not considered, since this phase includes time spent pulling images
+   onto the host. Depending on that status would not acurately reflect the pressure
+   inside the Kubernetes Scheduler."
   [compute-cluster pool-name]
   (->> (get-pods-in-pool compute-cluster pool-name)
        (add-starting-pods compute-cluster)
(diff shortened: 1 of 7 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
                               :first-ten-jobs (print-str (->> jobs (take 10) (map :job/uuid) (map str)))}))
       (dissoc compute-cluster->jobs :no-acceptable-compute-cluster))))
 
+(defn decrement-scheduling-capacity-counter
+  "Decrement capacity for compute cluster in the counter."
+  [capacity-counter compute-cluster]
```
scrosby, line 1089: I'd rename this argument. See below.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
...
+  (update-in capacity-counter [compute-cluster] dec))
+
+(defn scheduling-capacity-constrained-job-distributor
+  "Create a job distributor function that only assigns jobs to compute clusters
+   with available scheduling capacity. Assignments are tracked and the given
+   counter is decremented."
+  [compute-cluster->scheduling-capacity autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn]
+  (tracing/with-span [s {:name "scheduler.distribute-jobs-to-compute-cluster"
+                         :tags {:pool pool-name :component tracing-component-tag}}]
+    (let [capacity-counter-atom (atom compute-cluster->scheduling-capacity)
```
scrosby, line 1099: I'd be more consistent with this atom's name. It sort of came from nowhere and doesn't have the `->` hint that its a map.

E.g., `compute-cluster->available-scheduling-capacity-atom`. Also should probably use the non-`atom` suffixed version as the argument name to `decrement-scheduling-capacity-counter` 

I'd not name it `compute-cluster->scheduling-capacity-atom` because its semantically different than that.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
...
+           (fn choose-compute-cluster-for-autoscaling
+             [{:keys [job/uuid] :as job}]
+             (let [acceptable-compute-clusters
+                   (job->acceptable-compute-clusters-fn job compute-clusters)
+                   available-compute-clusters
+                   (filter
+                    (fn [compute-cluster]
+                      (let [capacity (get @capacity-counter-atom compute-cluster)]
+                        (when (nil? capacity)
+                          (throw (ex-info "Compute cluster given to distributor is not tracking scheduling capacity"
```
scrosby, line 1111: Log an error and move on. It'll be really noisy, but not run the risk of blowing up production if a compute cluster leaks in here via `job->acceptable-compute-clusters-fn` in an unexpected manner. Say, something underinitialized during dynamic compute clusters. 

The map won't have a nil value because we'd get a NPE from `max-launchable` when computing the min if a nil leaked in there, so that above is the only risk here.

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -522,6 +532,42 @@
...
+   [this pool-name]
+   (let [scheduling-pods (get-scheduling-pods this pool-name)
+         num-scheduling-pods (count scheduling-pods)
+         total-pods (-> @all-pods-atom keys count)
+         total-nodes (-> @current-nodes-atom keys count)
+         ; NOTE: the synthetic pods config is co-opted here to determine the max outstanding
+         ; pods in the compute cluster. This configuration, set on the compute cluster
+         ; template, will be refactored in the future to more generally define autoscaling 
+         ; parameters once synthetic pods are decommissioned.
+         {:keys [max-pods-outstanding max-total-pods max-total-nodes]
```
scrosby, line 546: Why not default max-pods-outstanding here too? (The concern if that's not in synthetic-pods-config, this will blow up with an NPE with.... impressive blast radius.)

Conversely, shouldn't this all be defaulted in synthetic-pods-config, or the factory method for this defrecord? (I'd prefer that to be honest.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
...
+                                          {:pool-name pool-name :compute-cluster compute-cluster})))
+                        (pos? capacity)))
+                    acceptable-compute-clusters)]
+               (if (empty? available-compute-clusters)
+                 :no-acceptable-compute-cluster
+                 (let [assigned-compute-cluster (nth available-compute-clusters
+                                                     (-> uuid hash (mod (count available-compute-clusters))))]
```
scrosby, line 1118: Usually nice to put steps in a -> or ->> on sequential lines.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
...
+                   (filter
+                    (fn [compute-cluster]
+                      (let [capacity (get @capacity-counter-atom compute-cluster)]
+                        (when (nil? capacity)
+                          (throw (ex-info "Compute cluster given to distributor is not tracking scheduling capacity"
+                                          {:pool-name pool-name :compute-cluster compute-cluster})))
+                        (pos? capacity)))
+                    acceptable-compute-clusters)]
+               (if (empty? available-compute-clusters)
+                 :no-acceptable-compute-cluster
```
scrosby, line 1116: I was mulling over if we should pause here. If we have a job that does have CC it can go, but that place doesn't have any available capacity, do we skip and try again later, or pause. Either way we can't place the job, and if there are too many we'll clog the queue.  But that's expected. Better to let what can match match and hope for luck soon.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
...
+   with available scheduling capacity. Assignments are tracked and the given
+   counter is decremented."
+  [compute-cluster->scheduling-capacity autoscalable-jobs pool-name compute-clusters job->acceptable-compute-clusters-fn]
+  (tracing/with-span [s {:name "scheduler.distribute-jobs-to-compute-cluster"
+                         :tags {:pool pool-name :component tracing-component-tag}}]
+    (let [capacity-counter-atom (atom compute-cluster->scheduling-capacity)
+          compute-cluster->jobs
+          (group-by
+           (fn choose-compute-cluster-for-autoscaling
+             [{:keys [job/uuid] :as job}]
```
scrosby, line 1103: Job UUID or instance UUID? I'd vote for instance UUID because it should make reruns route to a new CC, for hopefully better luck there.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1624,56 +1671,95 @@
...
-                     psuedo-matches [{:task-metadata-seq task-metadata-seq}]
-                     task-txns (map (fn [job metadata] (job->task-txn job metadata compute-cluster)) jobs task-metadata-seq)]
-                 (launch-tasks-for-cluster compute-cluster psuedo-matches task-txns pool-name conn nil false)))]
+       (let [distributor-fn (partial scheduling-capacity-constrained-job-distributor 
+                                     compute-cluster->scheduling-capacity)
+             kubernetes-jobs-launcher (create-kubernetes-jobs-launcher conn pool-name mesos-run-as-user)]
          (log-structured/info "Launching jobs in Kubernetes" {:pool pool-name})
-         (distribute-and-launch-jobs considerable-jobs pool-name compute-clusters-for-pool
-                                     job->acceptable-compute-clusters-fn launch-distributed-job-fn)
-         (log-structured/info "Launched jobs in Kubernetes" {:pool pool-name :number-launched-jobs (count considerable-jobs)}))))))
```
scrosby, line 1651: Retain this long line after its all done, that we've launched everything?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1624,56 +1671,95 @@
...
+              user->usage (tracing/with-span [s {:name "scheduler.kubernetes-handler.resolve-user-to-usage-future"
+                                                 :tags {:pool pool-name :component tracing-component-tag}}]
+                            @user->usage-future)
+              considerable-jobs (pending-jobs->considerable-jobs db pending-jobs user->quota user->usage max-considerable pool-name)
+              considerable-job-uuids (set (map :job/uuid considerable-jobs))]
+          (log-structured/info "Considering jobs" {:pool pool-name :number-considered-jobs (count considerable-job-uuids)
+                                                   :number-pending-jobs (count pending-jobs)})
+          (if (seq considerable-jobs)
+            (do
+              (swap! pool-name->pending-jobs-atom
```
scrosby, line 1746: Worth a small comment: 
`remove the considerable jobs from consideration from matching, until we do a rank cycle again.`

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1084,19 +1084,64 @@
                               :first-ten-jobs (print-str (->> jobs (take 10) (map :job/uuid) (map str)))}))
       (dissoc compute-cluster->jobs :no-acceptable-compute-cluster))))
 
+(defn decrement-scheduling-capacity-counter
+  "Decrement capacity for compute cluster in the counter."
+  [capacity-counter compute-cluster]
+  (update-in capacity-counter [compute-cluster] dec))
+
+(defn scheduling-capacity-constrained-job-distributor
```
scrosby, line 1092: Can you put a unit test over this code that exercises some of the code paths?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1624,56 +1671,95 @@
...
-          considerable-job-uuids (set (map :job/uuid considerable-jobs))]
-      (log-structured/info "Considering jobs" {:pool pool-name :number-considered-jobs (count considerable-job-uuids)})
-      (if (seq considerable-jobs)
+    (let [; This is required to properly filter compute clusters to
+          ; those configured for this pool. This is done via the synthetic-pods
+          ; config, accessed via the autoscaling? method. In the future, this list
+          ; of pools should be slightly restructured to be listed one level higher,
+          ; on the compute-cluster-template config.
+          compute-clusters-for-pool (filter #(cc/autoscaling? % pool-name) compute-clusters) 
+          max-total-considerable (:max-jobs-considered scheduler-config)
```
scrosby, line 1719: These three vars can be destructured out from `scheduler=config` at the function argument stage:
`{:keys [max-totalconsiderable ...] :as scheduler-config} `

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -378,6 +378,16 @@
      (prom/set prom/total-synthetic-pods {:pool pool-name :compute-cluster name} num-synthetic-pods)
      (prom/set prom/max-synthetic-pods {:pool pool-name :compute-cluster name} max-pods-outstanding))))
 
+(defn get-scheduling-pods
```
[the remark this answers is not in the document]
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: At this point, we do not have an instance for the job. Only after this function assigns do we create that instance and then launch. But that's a really interesting point. Do you know why we shouldn't just make this random?

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: good point, done

scheduler/src/cook/kubernetes/compute_cluster.clj
ahaysx: Done, I've made max total nodes and pods required as part of the guard below. Additionally, confirmed out current prod configs define these.

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done, tested that it ignores clusters with no capacity and dec's the counter.

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1624,56 +1678,96 @@
...
+    (let [cc-name (cc/compute-cluster-name compute-cluster) 
+          task-metadata-seq (jobs->kubernetes-task-metadata jobs pool-name mesos-run-as-user compute-cluster)
+          psuedo-matches [{:task-metadata-seq task-metadata-seq}]
+          task-txns (map (fn [job metadata] (job->task-txn job metadata compute-cluster)) jobs task-metadata-seq)]
+      (launch-tasks-for-cluster compute-cluster psuedo-matches task-txns pool-name conn nil false)
+      (log-structured/info "Launched jobs in Kubernetes compute cluster" {:pool pool-name
+                                                                          :compute-cluster cc-name
+                                                                          :number-launched-jobs jobs}))))
+
 (defn distribute-jobs-to-kubernetes
```
scrosby, line 1695: Just realized that this distributes and launches. Maybe rename it to 'scheduler-jobs-on-kubernetes'?

scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1624,56 +1678,96 @@
...
-             compute-clusters-for-pool (filter #(cc/autoscaling? % pool-name) compute-clusters)
-             launch-distributed-job-fn
-             (fn [[compute-cluster jobs]]
-               (let [task-metadata-seq (jobs->kubernetes-task-metadata jobs pool-name mesos-run-as-user compute-cluster)
-                     psuedo-matches [{:task-metadata-seq task-metadata-seq}]
-                     task-txns (map (fn [job metadata] (job->task-txn job metadata compute-cluster)) jobs task-metadata-seq)]
-                 (launch-tasks-for-cluster compute-cluster psuedo-matches task-txns pool-name conn nil false)))]
+       (let [distributor-fn (partial scheduling-capacity-constrained-job-distributor 
+                                     compute-cluster->scheduling-capacity)
+             kubernetes-jobs-launcher (create-kubernetes-jobs-launcher conn pool-name mesos-run-as-user)]
```
scrosby, line 1707: Rename `kubernetes-job-launcher-fn`

scheduler/src/cook/scheduler/scheduler.clj
scrosby: I see no reason not to make it random as long as this code isn't called from the synthetic-pod codepath. (And AFACT, its not. Can you confirm?), so excellent suggestion to make this random.

scheduler/test/cook/test/scheduler/scheduler.clj
```diff
@@ -2645,7 +2645,32 @@
...
+               compute-cluster-3 1}]
+          (is (= {compute-cluster-3 [job1]}
+                 (sched/scheduling-capacity-constrained-job-distributor
+                  compute-cluster->scheduling-capacity
+                  [job1 job2]
+                  "test-pool"
+                  [compute-cluster-1
+                   compute-cluster-2
+                   compute-cluster-3]
+                  sched/job->acceptable-compute-clusters))))))))
```
scrosby, line 2673: What about submitting two jobs with capacity 1 and making sure only one gets distributed?
What about submitting jobs when there's no capacity?
What about submitting two jobs when there's a capacity of 4?
What about submitting two jobs when there's no compute clusters?

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: done

scheduler/src/cook/scheduler/scheduler.clj
ahaysx: Yup this code is just for new Kenzo side. Done. (rand-nth is nice)

scheduler/test/cook/test/scheduler/scheduler.clj
ahaysx: - What about submitting two jobs with capacity 1 and making sure only one gets distributed?
  - Covered in existing test ([job1 job2])

Added rest.

PR #2163, opened 2022-08-25
Adding prometheus metrics to remaining modules

samincheva:
- Adds prometheus metrics to match any remaining codahale metrics.

Ongoing migration to prometheus.


scheduler/src/cook/mesos/sandbox.clj
```diff
@@ -129,6 +139,7 @@
   [{:keys [pending-sync-hosts] :as pending-sync-state}]
   (counters/clear! sandbox-pending-sync-host-count)
   (when (seq pending-sync-hosts)
+    (prom/set prom/mesos-pending-sync-host-count (count pending-sync-hosts))
```
scrosby, line 142: Should this be inc?

scheduler/src/cook/progress.clj
```diff
@@ -57,13 +59,16 @@
...
+            (prom/set prom/progress-aggregator-pending-states-count 0)
             (counters/clear! progress-aggregator-pending-states-count))
+          (prom/inc prom/progress-aggregator-pending-states-count (- new-count old-count))
           (counters/inc! progress-aggregator-pending-states-count (- new-count old-count)))
         instance-id->progress-state')
       (do
         (log/warn "skipping" data "as it is missing an integer progress-sequence")
         instance-id->progress-state))
     (do
+      (prom/inc prom/progress-aggregator-drop-count)
```
scrosby, line 71: Does this replace both the rate and the count?

samincheva: Yes, we can use the counter variable with a `rate` function to calculate the rate dynamically.

scheduler/src/cook/mesos/sandbox.clj
samincheva: I was missing the call to set the counter value to 0 (like we do for the codahale counter with the call to clear!). Updated to have the call to set to 0, and then it's functionally equivalent to set or inc by that amount.

PR #2168, opened 2022-09-02
Optimize RAM by not storing all pod metadata in the internal maps.

scrosby:

- Do not save all of the pod data when we're tracking kubernetes state. 

If you don't save it, it doesn't clutter up RAM, and in particular, clutter up oldgen. The worst case for GC is objects that live a long time but eventually die. This hopefully cuts the rate of those objects, at least in the case of Pod data.


scrosby:
Doesn't save anything (we store another copy, and have to store it), and we use the spec in other spots in the code. Do not merge.

PR #2169, opened 2022-09-02
Parallel watch processing.

scrosby:

- Do watch processing on the `controller-executor-service` threadpool

Processing watches was previously done by a single thread which could impact the rate Cook could process job completions. Doing it in parallel allows us  to handle a higher job completion throughput.


laurameng (approved):
Ok,  we are defaulting to 1k max outstanding and 200 shards. This will start affecting everyone even w/o adding new config.

PR #2176, opened 2022-09-08
do not set scalar-requests to pool specific resources for Kenzo pods

ahaysx:

- When creating the task-metadata for Kenzo jobs, do not set scalar-requests to the pool-specific resources. Instead, set to the unmodified job resource values.

- Kenzo pods include sidecar and other containers in addition to the job container. It does not need to account for the overhead like synthetic pods, which only have a single container.

### DIFF
--- a/scheduler/src/cook/scheduler/scheduler.clj
+++ b/scheduler/src/cook/scheduler/scheduler.clj
@@ -1700,9 +1700,15 @@
           ; running on the same host twice. So, we need to avoid running a
           ; real job pod on any of the hosts that it won't be able to run on.
           :pod-hostnames-to-avoid (constraints/job->previous-hosts-to-avoid job)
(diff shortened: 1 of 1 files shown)


scheduler/src/cook/scheduler/scheduler.clj
```diff
@@ -1700,9 +1700,15 @@
           ; running on the same host twice. So, we need to avoid running a
           ; real job pod on any of the hosts that it won't be able to run on.
           :pod-hostnames-to-avoid (constraints/job->previous-hosts-to-avoid job)
-          :task-request {:scalar-requests (walk/stringify-keys pool-specific-resources)
+          :task-request {:scalar-requests (job->scalar-request job)
```
laurameng, line 1703: Yep, this matches ln 502 ✔️

laurameng (approved):

PR #2180, opened 2022-09-14
Fixing compute_cluster value in metric to use :name

samincheva:
-  Updating a prometheus metric that was recording the compute cluster incorrectly.


scheduler/src/cook/kubernetes/compute_cluster.clj
```diff
@@ -106,7 +106,11 @@
                                                                  %)
                                        node-name->available)
         number-nodes-schedulable (count node-name->schedulable)
-        number-nodes-total (count node-name->node)]
+        number-nodes-total (count node-name->node)
+        total-cpus-capacity (total-resource node-name->capacity :cpus)
```
ahaysx, line 110: As discussed in person, there appears to be an existing bug here in the metrics, where capacity is being tracked as total capacity in the computer cluster when it is actually per pool. The comment says as much, but can be confirmed by inspecting `api/get-capacity` and `pending-offers`, which filters nodes to be in this pool. Consumption does appear to be total per cluster.

